PR #161, opened 2016-09-15
Add a view to allow configuring hostname.

mwhudson:
This seems like a lot of typing/copypaste for a very simple view... I didn't think very hard while doing this.

See bug https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1621380

### DIFF
--- a/subiquitycore/ui/views/hostname.py
+++ b/subiquitycore/ui/views/hostname.py
@@ -1,0 +1,109 @@
+# Copyright 2016 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+import socket
+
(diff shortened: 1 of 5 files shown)


subiquitycore/ui/views/hostname.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def cancel(self, button):
+        self.signal.prev_signal()
+
+    def done(self, button):
+        if len(self.hostname.value) < 1:
+            self.error.set_text("Please enter a non-empty name.")
+            return
+        if not self.opts.dry_run:
+            cmd = ["hostname", self.hostname.value]
```
mwhudson: Ah yeah, you're right. Is editing /etc/hosts really necessary though? After running hostname foo.local on in my KVM test system I can resolve foo.local. (Maybe this is some resolved magic?)

mwhudson:
PTAL

cyphermox:
I hadn't caught on to this, but applying the changes should go in the controller, not the view. I think we can also use hostnamectl to do the full change work, so maybe this is going to work better (I'll test).

mwhudson:
Well after hacking around in the insides of subiquity a bit, here's another go. I like this one more :-)

I think /perhaps/ using hostnamectl would be better, but it's hard to be certain. Let me know what you think!

mwhudson:
Ping?

cyphermox:
There's a question as to whether this really does make sense to have for console-conf; maybe it shouldn't be a feature there. It still makes sense in the world of subiquity, but requiring people to configure the hostname of the device seems like one more step that people would have to do before ownership, whereas they could just as well do it later -- it's not required for the device to be reachable and usable.

xnox:
From the bug report

> The DHCP server might have told you to use a hostname to use when you're assigned an address. If so, default to that in the prompt to the user.

I don't think we do this today, so I guess we should probe hostnamectl for the epehemral hostname and present that in subiquity UI too.

mwhudson:

PR #162, opened 2016-09-16
Indicate if a wired interface is connected or not

mwhudson:
Depends on https://github.com/CanonicalLtd/probert/pull/26

See https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1623897

### DIFF
--- a/subiquitycore/models/network.py
+++ b/subiquitycore/models/network.py
@@ -17,7 +17,10 @@ import errno
 import ipaddress
 import logging
 import os
+
 import netifaces
+from probert.network import IFF_UP, IFF_RUNNING
+
 from subiquitycore.prober import make_network_info
 from subiquitycore.model import BaseModel
 from subiquitycore.utils import (read_sys_net,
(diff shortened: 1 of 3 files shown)


subiquitycore/models/network.py
```diff
@@ -228,6 +231,12 @@ def ip6_providers(self):
         ip_info = self._get_ip_info()
         return ip_info['ip6_providers']
 
+    def is_connected(self):
```
cyphermox: I think this properly belongs in probert rather than directly in our models for subiquity.

subiquitycore/models/network.py
```diff
@@ -360,6 +369,7 @@ def get_menu(self):
     # --- Model Methods ----
     def probe_network(self):
         log.debug('model calling prober.get_network()')
+        self.prober.probe()
```
cyphermox: Why do we need to probe again? The probing should have already happened before, at the point where we did the initial discovery of the devices.

subiquitycore/models/network.py
mwhudson: Done.

subiquitycore/models/network.py
mwhudson: In case the user has connected an interface since that happened. Ogra has filed https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1624322 since I did this, so maybe I should dust of my "live updates of networking UI" code.

PR #166, opened 2016-09-23
show some words when we are running under qemu

mwhudson:
For https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1621550

### DIFF
--- a/console_conf/ui/views/login.py
+++ b/console_conf/ui/views/login.py
@@ -107,11 +107,16 @@ class LoginView(BaseView):
                     ips.append(ip)
 
         for ip in ips:
-                ssh_iface = "    ssh %s@%s" % (user.username, ip)
-                ssh += [Padding.center_50(Text(ssh_iface))]
+            ssh_iface = "    ssh %s@%s" % (user.username, ip)
+            ssh += [Padding.center_50(Text(ssh_iface))]
+
 
         sl += [Text(login_text),
                Padding.line_break("")] + ssh
+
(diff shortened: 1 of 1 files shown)


console_conf/ui/views/login.py
```diff
@@ -107,12 +107,17 @@ def _build_model_inputs(self):
...
-                ssh += [Padding.center_50(Text(ssh_iface))]
+            ssh_iface = "    ssh %s@%s" % (user.username, ip)
+            ssh += [Padding.center_50(Text(ssh_iface))]
+
 
         sl += [Text(login_text),
                Padding.line_break("")] + ssh
 
+        if 'QEMU Virtual CPU' in open('/proc/cpuinfo').read():
+            qemu_text = "Ubuntu Core is running under QEMU/KVM. If using user mode networking, the above addresses will be wrong!"
```
cyphermox: I strongly disagree with displaying this. It seems extremely clunky: why display addresses and then mention they are wrong? Plus it's not impossible that the setup is done "on purpose", and that the users have configured their system appropriately to access the system with whatever access is provided, even with user-mode networking.

Rather than doing this, I would support either doing better detection that permits us to make sure the IP addresses would be wrong (which seems fraught with peril too), or leave it to users to know what they are doing.

An elegant solution might be to always display the ssh line for user@hostname, and expect that users will pick that first.

mwhudson:
This doesn't seem useful.

PR #168, opened 2016-09-29
identity: implement an initial, dumb "is this device owned" check.

cyphermox:
Signed-off-by: Mathieu Trudel-Lapierre <mathieu.trudel-lapierre@canonical.com>

### DIFF
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -31,11 +31,13 @@ class IdentityController(BaseIdentityController):
         self.ui.set_header(title, excerpt)
         self.ui.set_footer(footer, 40)
         self.ui.set_body(self.identity_view(self.model, self.signal, self.opts, self.loop))
-        if self.is_device_owned():
-            # FIXME: should be email rather than realname here, but extrausers only has email
-            self.signal.emit_signal('identity:done', self.model.user.realname)
+        device_owner = self.get_device_owner()
+        if device_owner is not None:
+            # FIXME: should be email rather than realname here, but extrausers
+            # only has an email in realname.
+            self.signal.emit_signal('identity:done', device_owner['realname'])
 
-    def is_device_owned(self):
+    def get_device_owner(self):
         """ Check if device is owned """
 
         # TODO: use proper snap APIs.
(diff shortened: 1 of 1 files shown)


console_conf/controllers/identity.py
```diff
@@ -31,9 +31,37 @@ def default(self):
         self.ui.set_header(title, excerpt)
         self.ui.set_footer(footer, 40)
         self.ui.set_body(self.identity_view(self.model, self.signal, self.opts, self.loop))
+        if self.is_device_owned():
+            # FIXME: should be email rather than realname here, but extrausers only has email
+            self.signal.emit_signal('identity:done', self.model.user.realname)
+
+    def is_device_owned(self):
```
mwhudson: The side-effecty nature of a method with this name is surprising. I get it now, but maybe it could be a get_device_owner() method that returns None if there is no user? And then we just run the next-screen signal, not done().

cyphermox: Indeed, that's kind of too side-effecty.

console_conf/controllers/identity.py
```diff
@@ -31,9 +31,39 @@ def default(self):
...
+        if device_owner is not None:
+            # FIXME: should be email rather than realname here, but extrausers
+            # only has an email in realname.
+            self.signal.emit_signal('identity:done', device_owner['realname'])
+
+    def get_device_owner(self):
+        """ Check if device is owned """
+
+        # TODO: use proper snap APIs.
+        with open('/var/lib/extrausers/passwd', 'r') as extrausers_fp:
```
mwhudson: Ah, this is going to fall over in a heap in dry-run mode right? Should probably be defensive here, return None if this raises FileNotFoundError

console_conf/controllers/identity.py
```diff
@@ -31,9 +31,39 @@ def default(self):
         self.ui.set_header(title, excerpt)
         self.ui.set_footer(footer, 40)
         self.ui.set_body(self.identity_view(self.model, self.signal, self.opts, self.loop))
+        device_owner = self.get_device_owner()
+        if device_owner is not None:
+            # FIXME: should be email rather than realname here, but extrausers
+            # only has an email in realname.
+            self.signal.emit_signal('identity:done', device_owner['realname'])
```
mwhudson: I still think calling self.model.add_user() here and then directly going to the 'identity:login' makes more sense but I'm not going to make a big fuss about it.

console_conf/controllers/identity.py
```diff
@@ -31,9 +31,39 @@ def default(self):
...
+        device_owner = self.get_device_owner()
+
+        if self.opts.dry_run:
+            result = {
+                'realname': email,
+                'username': email,
+                }
+            self.model.add_user(result)
+        elif device_owner is not None:
+            self.model.add_user(device_owner)
```
mwhudson: Still need to run mark_firstboot_complete() in this path!

PR #169, opened 2016-10-04
exit console-conf immediately on a managed device

mwhudson:
I don't know if this is what we want, but I wanted to play with it so posting it up...

### DIFF
--- a/bin/console-conf-wrapper
+++ b/bin/console-conf-wrapper
@@ -8,6 +8,11 @@ trap true HUP INT QUIT TSTP
 # happened and need to force it on. Yay UNIX!
 stty icrnl
 
+if [ `snap is-managed` = "true" ]; then
+    touch /var/lib/console-conf/complete
+    exit 0
+fi
+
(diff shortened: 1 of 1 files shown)


cyphermox:
Yep, makes sense if it's indeed what we want, but then from the point where the device is managed, you couldn't run through network, keyboard, etc configuration. Seems slightly wrong, but it depends on what mvo and niemeyer will decide.

mvo5:
https://github.com/snapcore/snapd/pull/2097 has landed so this is good to go.

niemeyer:
Please note we'll probably (assuming nobody complains) tweak the name to "snap managed".

niemeyer:
Also, the expectation from the code there is wrong, as far as the current PR is implemented.

niemeyer:
snapcore/snapd#2102 is up for review. With that the logic there may be:
```
if snap managed -q; then
        ...
fi

niemeyer:
Sorry, please hold fire on the above syntax. @pedronis makes a good point in the review.

PR #170, opened 2016-10-11
remove trivial uses of signals for console-conf

mwhudson:
subiquity uses urwid's signals in a lot of places where there can only be one possible receiver. In that case, just call the method on the object directly!


mwhudson:
I'd like to merge this I think. It touches a lot of places but IMO makes things a bit clearer. Will wait until tomorrow to give people a change to object.

PR #172, opened 2016-10-19
render gateway4/gateway6 in network config

mwhudson:
See https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1634540


mwhudson:
Tested and fixes a critical bug so merging.

martinpitt:
You only have one field for the gateway? what if you use both IPv4 and IPv6?

PR #173, opened 2016-10-19
do not write the netplan config so the world can see it

mwhudson:


mwhudson:
Tested and fixes a critical bug so merging.

PR #174, opened 2016-10-20
set umask in console-conf-tui to make log files private

mwhudson:
this would also affect any other files console-conf creates, but afaict
that's only the netplan config which (a) should be private too and (b)
is handled specially anyway


mwhudson:
Merging this too.

PR #175, opened 2016-10-20
add UI for entering nameservers

mwhudson:
This depends on netplan changes that pitti is frantically hacking up right now so there's a chance the two bits won't line up. But I think it should be OK.


cyphermox:
Looks reasonable, let's merge this, we can fix the missing bits once netplan has caught up too.

PR #190, opened 2017-01-25
recommend some common mount points per spec

mwhudson:
Here's me playing with the new feature https://asciinema.org/a/0pveykhykc6hncidesxyc0md5

It needs some other "common mount points", please suggest away.


cyphermox:
/, /usr, /var, /var/lib, /boot, /srv, /home are the typical mountpoints.
Do we care about putting /snap as an option too?

PR #193, opened 2017-02-03
implement proper tab cycling

mwhudson:
I know we said this wasn't a priority but I couldn't resist one more crack and actually I was really nearly there. This makes 'tab' and 'shift tab' move to next and previous widgets respectively and cycles around as needed.


markshuttle:
Ah, that will make the whole thing feel much simpler.

Thank you!

PR #197, opened 2017-02-09
updated validation for the identity view

mwhudson:


mwhudson:
Going to do this a different way.

PR #198, opened 2017-02-10
an attempt to make shift tab work in the linux console

mwhudson:


mwhudson:
@vorlonofportland @cyphermox can you guys take a look at this one? Are there any situations where this will break?

cyphermox:
I see nothing wrong with it, aside from in cases where it won't be running just yet (serial and SSH). That said, it leads me to wonder if it was worth the trouble of fixing it (shift-tab doesn't work in d-i).

PR #207, opened 2017-03-21
Add snapcraft.yaml and some fixes for working as a snap.

cyphermox:


subiquity/curtin.py
```diff
@@ -71,6 +71,11 @@
 # Reworked for flake8, but it does make it harder to read.
 POST_INSTALL_LIST = [
     ("late_commands:"),
+    ("    05_clear_subiquity: curtin in-target -- "
```
mwhudson: Wait, so curtin works by copying it's own / to the point that snaps get transferred? That's a surprise to me.

Should we also remove the core snap?

snapcraft.yaml
```diff
@@ -0,0 +1,41 @@
...
+  subiquity:
+    plugin: python
+    build-packages: [python-setuptools]
+    stage-packages: [curtin]
+    python-packages:
+      - urwid
+      - pyyaml
+      - pyudev
+      - attrs
+    source: https://github.com/CanonicalLtd/subiquity.git
```
mwhudson: Shouldn't this be 'source: .'?

snapcraft.yaml
```diff
@@ -0,0 +1,41 @@
...
+    source-branch: cyphermox/snap
+    source-type: git
+  wrappers:
+    plugin: dump
+    source: https://github.com/CanonicalLtd/subiquity.git
```
mwhudson: ditto

snapcraft.yaml
```diff
@@ -0,0 +1,41 @@
+name: subiquity
+version: "0.0.32"
+summary: Ubuntu installer
+description: The Ubuntu server installer
+confinement: classic
+
+apps:
+  subiquity:
+    command: usr/bin/subiquity
+  console-conf:
```
mwhudson: I don't think we should have an app for console-conf (probert is just about ok :-))

subiquitycore/utils.py
```diff
@@ -68,10 +68,18 @@ def is_executable(x):
         for ftype, items in checks[check_type].items():
             for i in items:
                 if not os.path.exists(i):
-                    log.error('FAIL: {} is not found on the'
-                              ' filesystem'.format(i))
-                    env_ok = False
-                    continue
+                    if 'SNAP' in os.environ:
```
mwhudson: I wonder if we can just delete the environment check now. Most other packages I use do not explicitly check their dependencies are present at runtime. But that's a change for another day I guess.

snapcraft.yaml
cyphermox: On the contrary, there is no reason not to ship it while we're building the rest anyway.

subiquity/curtin.py
cyphermox: It's customizable, but we should take the proper steps to clean up anyway. These things *must not* remain on the system.

The core snap is a different matter, and up to discussion (but for now it also wouldn't be included on the installed image).

snapcraft.yaml
cyphermox: Currently, it works better having it set to the real location of the source code for building things. When merged and integrated into LP for daily building, we'll want to revisit (as LP probably can't reach it), but it also apparently can't do snap recipes from purely git trees.

subiquitycore/utils.py
cyphermox: You still need this in the case of running subiquitycore magic outside a snap (ie. as the package). We should allow that to work properly (and AFAIK this is shared with console-conf...). The intent there is that we shouldn't crash when things like curtin are missing, but rather warn the user early that something is wrong, and that's why the dependencies are explicitly listed and checked (and other places do similar things). The issue is in the snap, you also need to take into account that some pieces may be shipping in the snap, or already be available on a classic system outside snaps (straight on the filesystem). In classic context, non-snap paths should win, and we should only rely on trying the /snap path if is otherwise fails.

subiquity/curtin.py
cyphermox: This is already appropriately layered so that the files *should not* be there normally, obviously a work in progress. However, we should still be careful and make sure the files are removed if they do happen to show up, as otherwise they would break the installed system.

snapcraft.yaml
cyphermox: console-conf comes from the same code as subiquity, we'll eventually need it some way or another as a snap, so it will already be done and available.

snapcraft.yaml
cyphermox: It is if you're not ready to take the steps, and as above, this is only while this is not set up to do daily builds. This avoids cases where changes are not pushed and forgotten about.

subiquity/curtin.py
```diff
@@ -25,7 +25,7 @@
 
 TMPDIR = '/tmp'
 CURTIN_SEARCH_PATH = ['/usr/local/curtin/bin', '/usr/bin']
-CURTIN_INSTALL_PATH = ['/media/root-ro', '/']
+CURTIN_INSTALL_PATH = ['/media/root-ro', '/rofs', '/']
```
cyphermox: Not quite. This is just alternatives from which to grab the source data, with / being the last alternative (as it's always there). /media/root-ro does not normally exist, and /rofs is always the bottom-layer squashfs mounted by casper. Further alternatives may be added in the future, say if ubuntu-image mounts things differently.

cyphermox: Furthemore, /rofs is *not* the CD but a pristine live system, exactly as it should be once installed to disk. Another squashfs layer exists for the installer bits, and any changes done while in the installer environment as simply lost (we don't keep the overlay-work directories, I in fact have no idea where they are -- we don't need them anyway... we can revisit this if we want persistence in the future).

subiquity/curtin.py
mwhudson: I'm really not a fan of belt and braces like this. I'd much rather have something that did unsquashfs -ls installer.squashfs | grep subiquity && exit 1 or so in the image build.

subiquity/curtin.py
```diff
@@ -91,7 +94,7 @@ def curtin_userinfo_to_config(userinfo):
         '    groups: admin\\n' + \
         '    lock-passwd: false\\n'
     if 'ssh_import_id' in userinfo:
-        user_template += '    ssh-import-id: [{ssh_import_id}]\\n'
+        user_template += '    ssh-import-id: [ "{ssh_import_id}" ]\\n'
```
mwhudson: BTW, I have this fixed properly in https://github.com/CanonicalLtd/subiquity/pull/211. The reason this fix doesn't help is the levels of nesting -- this is cloud-init config, which is yaml that gets embedded as a string inside the yaml of the curtin config. Drop this bit?

PR #208, opened 2017-03-22
confirm before writing data to disks

mwhudson:
I looks like this:

confirm

thoughts on wording or formatting welcomed.


mwhudson:
Well I'm going to land this, refinements to wording etc can always be made later.

PR #209, opened 2017-03-28
fix crash on startup with no ip address

mwhudson:
console-conf-write-login-details would crash if it ran before the system
got an ip address. re-jig things a bit so that it displays an nice message
when there is no ip address but still checks for a new address every 5s until
it finds one.


mwhudson:
I tested this in a VM that was talking to a dnsmasq I could SIGSTOP to prevent it responding.

PR #214, opened 2017-04-10
[WIP] apport integration

mwhudson:
This branch:

1. reorganizes the code that decides where curtin config and log files go on disk (not strictly required but I've been wanting to do this for a while)
1. adds an apport hook that collects up the subiquity config and hook files.
2. runs apport-cli -c $tmpfile subiquity when curtin fails
3. runs a web server to serve up the crash file

It has some issues:

1. the hook needs to go into /usr/share/apport/package-hooks/source_subiquity.py. Easy enough if subiquity is installed as a deb, not if it is installed as a snap. The fix here is probably to put the hook into the apport source package.
2. the web server runs on a random port, which makes for a hard to type URL and makes it impossible (I think?) to test with qemu's user mode networking. Could run on a well known port (even 80, I guess), but then the subquities running on tty1 and on the serial line need to coordinate somehow. Could just run apache in the installer image I guess? We're only serving up a static file really.


xnox:
Usually src:apport itself ships the "installer" hooks, cause one may want to run apport-collect after an install was done (badly). I am confused about `runs a web server to serve up the crash file` why is it not uploading the crash file to launchpad direct? or save it on disk and the person is expected to extract it however they like. Am I missing something?

murraybd:
"the hook needs to go into /usr/share/apport/package-hooks/source_subiquity.py. Easy enough if subiquity is installed as a deb, not if it is installed as a snap. The fix here is probably to put the hook into the apport source package."

The ubiquity and debian-installer apport package hooks exist in the apport package exactly because of this.

One thing to consider in the subiquity package hook is the permissions on the log file, are they only readable by root and will root be submitting a bug about subiquity?

mwhudson:
I can't remember now why it doesn't upload the crash file itself. Do you need to be logged in to do that?

murraybd:
If it is an actual crash, ProblemType: Crash, then I think you'd need to run 'ubuntu-bug /var/crash/my.crash' to start the data collection / upload process.  If you are just reporting an issue, ProblemType: Bug, about subiquity running 'ubuntu-bug subiquity' should start the upload but a browser is required.  The logging in bit will come after the blob is uploaded to Launchpad and you are presented with a page to report the bug.

mwhudson:
> The logging in bit will come after the blob is uploaded to Launchpad and you are presented with a page to report the bug.

Right, this was the problem that led to the (stupid, I admit) approach taken in this bug: there is a crash report on the system running subiquity (which is of course in text mode) and you want to open a URL in a browser on some other machine of the user operating subiquity. I don't see a good way to do that. I guess we could bundle w3m or something into the snap and go that way but that doesn't seem likely to be a good experience. Ideas welcome, honestly.

xnox:
I think the default with `ubuntu-dev-tools` on the serial things to display url with 10s timeout, for people to open it elsewhere, and then try default-www or some such anyway.

I guess given we will expect people to open url elsewhere, we need short urls? qr code for the url? (there are library that can print qrcodes on the console)

mwhudson:

PR #216, opened 2017-05-03
proxy support

mwhudson:
A bit of a WIP, haven't tested it on a core image yet.

https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1657254

### DIFF
--- a/console_conf/controllers/__init__.py
+++ b/console_conf/controllers/__init__.py
@@ -19,3 +19,4 @@ from .welcome import WelcomeController  # NOQA
 from subiquitycore.controllers.network import NetworkController  # NOQA
 from subiquitycore.controllers.login import LoginController  # NOQA
 from .identity import IdentityController  # NOQA
+from .proxy import ProxyController  # NOQA
--- a/console_conf/controllers/proxy.py
+++ b/console_conf/controllers/proxy.py
@@ -1,0 +1,45 @@
+# Copyright 2015 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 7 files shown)


cyphermox:
This should go more in subiquitycore, and then called from console-conf. It's a generic enough concept that we'll need it in subiquity eventually.

mwhudson:
Yes, you're probably right. subiquity will need to do different things with the supplied proxy data though, I think? (well, mostly just pass it through to curtin?)

mwhudson:
Turns out /etc/environment isn't writable in core anyway https://forum.snapcraft.io/t/set-proxy-via-snap-set-core-proxy-http/467 so this probably needs to wait until that is sorted out (I guess we could create a drop-in file for now but that's not a great answer)

smoser:
http proxy is really just more complex than people initially think.  You almost never want to set it in /etc/environment.  At first it seems like it works, but then later it just causes more problems.   See my comment at https://bugs.launchpad.net/ubuntu/+source/maas/+bug/1671320/comments/1 for more information.

xnox:
Meanwhile, over at curtin I see top-level proxy dictinary with multiple keys. Sould we not support all three like in curtin and pass it to it, somehow? http://bazaar.launchpad.net/~curtin-dev/curtin/trunk/revision/508

xnox:
proxy support was merged already no....? should this be closed.

mwhudson:
I think this aaaannccient PR is for console-conf?

xnox:
Ah, console-conf.

mwhudson:

PR #220, opened 2017-08-22
add script and modify snapcraft.yaml to enable local testing

mwhudson:
For a while there's not been a good way of testing local changes to subiquity in an image vs dry-run mode. These changes make it possible again.


mwhudson:
merging, if anyone has issues with this we can fix them later :-)

PR #222, opened 2017-08-31
[wip] Bare-bones i18n infrastructure

xnox:


subiquity/ui/mount.py
```diff
@@ -2,6 +2,9 @@
 import os
 import re
 
+import gettext
+gettext.install('subiquity')
```
mwhudson: does it make sense to do this in `subiquity/__init__.py`?

xnox: I haven't tried running subiquity yet. And based on the docs ".install()" is new world order. It potentially makes sense to do in subiquitycore / first ever import, or possibly just at the top of the entry-point scripts. Also I'm not sure if we will need translations in the console-conf scripts, and if those need their own separate translation domain. I think next steps will be for me to get a top level script running with a translation. After that boot subiquity with a translation set in the environment. And then we can look at changing translations on the welcome screen.

xnox:
Re getting this to work with a snap - i have not tried building snap. It needs the extra build-dep, and then it should ship the usr/share/locale .mo files. and everything should be fine.

xnox:
Re still building debs, well subiquity is a package in the ubuntu archive...

xnox:
So i've managed to force snapcraft to do something sensible. I've moved translation files into `/usr`.

At the same time I've moved subiquity from `/lib/python3.5/site-packages` to `/usr/lib/python3/dist-packages` as it normally should be. No idea if this will break the world. So, still need to test an image with this new snap.

mwhudson:
I've tested this in an image and it seems to work. There are some encoding issues but that's not related to the work here so let's get this merged?

xnox:
So with these changes locale data is finally discovered correctly. At the boot screen I selected russian, and tweaked the command line to specify `LC_ALL=ru_RU.UTF-8` and at the welcome screen this should result in `Выберите предпочитаемый язык` to appear at the top of the language selection list.

xnox:
I was not sure at all about logging, thus used `syslog.syslog()` for now. Ideally a logging module would be setup for subiquity to print useful / logged messages.

xnox:
This will be superseeded by a rebased branch.

PR #224, opened 2017-09-04
move Selector widget to its own file and create a class for its Options

mwhudson:
I wanted this so I can enable/disable options after creating the widget.


xnox:
+1

PR #225, opened 2017-09-04
still use source-type: git even though source: is now .

mwhudson:
This makes the contents of the snap much more reasonable for reasons I don't entirely
understand

There is still a mystery dh_python in the snap but it's less than half the size now.


xnox:
+1 whatever this does =)

PR #226, opened 2017-09-05
allow editing/deleting partitions

mwhudson:
Spent much longer on this than I probably should have done, oh well. The code is a bit tidier now at least.


xnox:
Do I understand this right, that we have no unit-tests / interactive tests to drive testing of these flows, and one has to effectively manually click through things? I wonder if we can somehow automate dry-run tests with fake devices information. Especially to test the logic and exected states of things.

mwhudson:
Yes, that's right :(

I'm not even sure how to write tests in such a way that maintaining them would be less of a burden than not having them. expect scripts? Sounds pretty awful but maybe it wouldn't be that bad.

raharper:
Each of the "inputs" or entry values put into the View (which calls back
into the Model) is a general API.

A unittest can run tests over various inputs and call the model api just
like the View code does.

I'm not sure if it's been kept up-to-date, but there is a
subiquity/tests/tests_models_filesystem.py

which does just this (and uses the existing example probert machine configs)


This will cover the model api which ultimate result is a curtin config.

For UI stuff, that'll likely require a different approach but study of the
urwid unittests may be enlightening (anything the user interacts with
widget wise can be programmed via urwid)

mwhudson:
It probably is worth writing unit tests for the model and views, but I think care is required to come up with tests that are not excessively repetitive and brittle (I don't find the urwid unit tests to be very clear at all). Maybe something to talk about at the rally?

Anyway, merging this for now.

PR #228, opened 2017-09-11
Update color palette

xnox:
Create palette for traffic light buttons: green (default), amber and
red. Color-code most buttons.

Make all progressbars orange for the completed part.

ps.
I have attempted a screen recording at - https://asciinema.org/a/xEAbwzI0oETRoi8fCL9wKpgwn
The colors are displayed mostly right.
Not sure about the confirmation dialog "No vs Continue". I've colored them "Green and Amber" because "No" does not destroy things, yet "Continue" is the one we want. Maybe it needs to be "Amber and Green". Also not sure about some buttons, it feels like we want a non-colored button for "Guided" or "Advanced". or e.g. "green and no color".


xnox:
@mwhudson if this provokes questions, responses and feedback it means it is a work of art! And I think I will twiddle the "no/continue" back to "amber/green" as it is weird to not have "continue" as green.

mwhudson:
I guess we should aim for semantic names in code, e.g. Color.cancel_button / Color.continue_button / Color.dangerous_button (for the "continue" on the filesystem screen, the partition delete button) and then we can twiddle the colours in palette.py independently.

xnox:
I got very confused with the whole Color module and buttons "collection" and wrapping of buttons in colors. I would have thought "cancel button" would be a subclass of AttrMap with w preset to Button with the preset labels. E.g. no partials, just initialize CancelButton() and set callbacks on it... Am I odd, or is this now how one codes in urwid?

mwhudson:
I find the "convenience" partials in subiquitycore/ui/utils.py and (especially) subiquitycore/ui/buttons.py and  a bit confusing too. I'd be happy to stop using them.

mwhudson:
So I guess when I say "Color.cancel_button" what I really mean is that the palette name we give to urwid should be somewhat semantic, not that we necessarily have use a helper with that name.

PR #230, opened 2017-09-13
display events from curtins reporting system

mwhudson:
This also changes the dry-run version to run a script that feeds realistic curtin events into subiquity.

Video here https://asciinema.org/a/G6oxtAXsDMpLTGd0eBOZqGSTa

It's just displaying the 'description' as reported by curtin and as such is probably fairly mysterious to the user. OTOH, it provides a nice sense to the user that install-related stuff is happening and that is most of what I want to achieve with this.


xnox:
I'm guessing that for curtin writing things over the network is the "native" way of doing things. Writing to a unix domain / file socket may have been better - if we can make curtin do that.

W.r.t. httpd implementation. Have you ever used `asyncio`? It allows doing polling/callbacks without spawning threads, as everything is simply executed by the mainloop. I wonder if using asyncio for the httpd server socket might be "nicer".

E.g. i wonder if curtin can push structured messages to the systemd-journal and us displaying journald messages filter on a particular key, or some such. Cause sometimes the fact that there is like a kernel panic, or a disk IO errors during install is helpful to know and see it happening intermixed with the installation progress.

mwhudson:
So yes, this is the "don't even think about changing curtin" approach, which is perhaps a bit short sighted. Curtin does not support writing to a unix domain socket at this time, but it would be easy to add a different reporter to curtin that would be more straightforward to use from subiquity. What I have here does work though :)

I've not used asyncio, no. There are probably quite a few things in the code that would benefit from using it, I've certainly reinvented a bunch of primitives event based frameworks provide. Maybe something I can hack on on my long, long flight to the rally?

Logging to the journal and having subiquity show curtins output intermixed with kernel messages does sound interesting. I don't know how to do structured logging to journald but I'm sure I can find out :)

xnox:
There is python-systemd modules and thus `import systemd.journal` to write directly to journal, but it is no better nor worse than syslog. Imho, we should just use standard python logging into syslog -> which does end up being the journal.

Then from subiquity we need to somehow stream the journal. I do not believe there is python api to do so, but we should be able to come up with something interesting. And for example when streaming journal to the user - do not show debug messages, but like info and up. Or some such.

mwhudson:
OK, feeding things into syslog makes sense.

For reading, I guess we just invoke journalctl -f (and whatever other options) read its output? That's more or less how the curtin log is read already.

mwhudson:
Also, am I ok to merge this, or do you think it's better to do the better thing now?

xnox:
I do wish for curtin to log to journal; and for us simply display stream from journalctl -f. That would be really nice.

If curtin cannot / doesn't log to journal now. Can we have this webserver as a standalone systemd unit / service that we launch and it will accept curtin logs and dump them into journal and that's it? Call it a "microservices" architecture? It should make the code simplier, no? Can curtin log to a remote syslog? cause we can start remote syslog journald server on the localhost.

mwhudson:
Poking at this, the problem that comes up is that there is interesting curtin log output that does not go via the event reporting system. So shoving events into the journal and then showing the user the output of journalctl -f would lose things. We could write our own curtin entry point that sets up both general logging and event reporting to journald and then runs curtin's main function, and then use https://www.freedesktop.org/software/systemd/python-systemd/journal.html#systemd.journal.Reader to display the output. This would re-use almost exactly none of this branch <wink> but I think it has a chance to be cleaner. I'll try it and see how it goes.

mwhudson:
(and we should add a --journald or something option to curtin that we could use instead of our own entrypoint)

mwhudson:
Ah it's a little more complicated than that :( curtin does this game of re-invoking itself, and _printing_ not logging the output of these subprocesses. So we'd have to call our entry point 'curtin' and put it on $PATH before /usr/bin/curtin to get the subprocesses to have the same logging setup, and even then I think we'll miss some output that is currently presented. I guess we could just shove all of curtin's output into the journal in the style of 'curtin $args | while read; logger $REPLY; done' but eh.

It might still be worth getting the curtin events into the journal as that's what I actually want to consume here and worry about having nicely interleaved curtin/kernel messages later.

mwhudson:
Done a different way in another PR

PR #232, opened 2017-09-13
update readme and makefile a bit

mwhudson:
Main changes: 1) mention console-conf at all in the README, 2) mention that bugs should be file on Launchpad


xnox:
Yeah, i'm yet to run console-conf.... I hope none of my changes are breaking it.....

PR #233, opened 2017-09-14
have button helpers style the buttons they return

mwhudson:
The main change here is so that instead of having to say Color.menu_button(menu_btn("foo")), menu_btn() and friends return an appropriately styled button already.

I also changed the names of the palette entries to be things like "cancel_button" rather than "amberbutton".

Finally, and a bit annoyingly, you can't call connect_signal on a decorated button, so I've added a helper that can be called with a decorated widget and call's urwids connect_signal with the underyling widget.


subiquity/models/welcome.py
```diff
@@ -23,7 +23,9 @@ class WelcomeModel(object):
     """ Model representing language selection
     """
 
-    supported_languages = ['English']
+    supported_languages = [
+        ('English', 'en_US'),
+        ]
     selected_language = None
```
xnox: Is this intentional? in my in-progress i18n branch I change supported_languages to have a different structure, and it's not that.

subiquity/ui/views/welcome.py
```diff
@@ -39,20 +39,14 @@ def __init__(self, model, controller):
             Text(""),
             Padding.center_79(Text("(More language choices will appear in time)"))]))
 
-    def _build_buttons(self):
-        self.buttons = [
-            Color.button(ok_btn(on_press=self.confirm)),
-        ]
-        return Pile(self.buttons)
-
```
xnox: Yes, I also noticed this is unused.

subiquity/ui/views/welcome.py
```diff
@@ -39,20 +39,14 @@ def __init__(self, model, controller):
...
     def _build_model_inputs(self):
         sl = []
-        for lang in self.model.get_languages():
-            sl.append(Color.menu_button(menu_btn(label=lang, on_press=self.confirm)))
+        for lang, code in self.model.get_languages():
+            sl.append(menu_btn(label=lang, on_press=self.confirm, user_arg=code))
```
xnox: Yeap, i see this now. I have ended up doing something similar in my i18n branch. But also hook-up actual switch languages code. Will rebase my code on top of this. So this is fine to go in now.

subiquitycore/ui/buttons.py
```diff
@@ -27,12 +27,30 @@ class MenuSelectButton(Button):
...
-reset_btn = partial(PlainButton, label="Reset", on_press=None)
-menu_btn = partial(MenuSelectButton, on_press=None)
+def plain_btn(label, color, on_press=None, user_arg=None):
+    button = PlainButton(label=label)
+    if on_press is not None:
+        connect_signal(button, 'click', on_press, user_arg)
+    return AttrWrap(button, color, color + ' focus')
+
+
+start_btn = partial(plain_btn, label="Start", color="save_button")
```
xnox: Ha, ok. this is simple enough. I thought we were going to do something crazy like:
```
class StartBtn(AttrWrap)
    def __init__(on_press=None, user_arg=None):
         w = PlainButton(label='Start', on_press=on_press)
         super(self).__init__(w=w, "save_button", "save_button focus")
```
Or some such, but your functional implementation is fine as well.

Using classes would mean that we would be able to drop `partial(...)` completely. As this partial stuff, is like fake class / obscure lambda function =)
I guess moving to classes can be done as another refactor. This at least makes all buttons semantically color coded straight away.

subiquitycore/ui/form.py
```diff
@@ -259,10 +261,10 @@ class Form(object, metaclass=MetaForm):
...
 
     def __init__(self, initial={}):
-        self.done_btn = Toggleable(done_btn(), 'button')
-        self.cancel_btn = Toggleable(cancel_btn(), 'button')
-        connect_signal(self.done_btn.base_widget, 'click', self._click_done)
-        connect_signal(self.cancel_btn.base_widget, 'click', self._click_cancel)
+        self.done_btn = Toggleable(done_btn())
+        self.cancel_btn = Toggleable(cancel_btn())
+        connect_signal(self.done_btn, 'click', self._click_done)
+        connect_signal(self.cancel_btn, 'click', self._click_cancel)
```
xnox: I have not yet poked forms. But looks sensible.

subiquitycore/ui/utils.py
```diff
@@ -180,3 +188,8 @@ class Color:
 
     """
     pass
+
+
+def connect_signal(obj, name, callback, user_arg=None, weak_args=None, user_args=None):
+    obj = getattr(obj, 'base_widget', obj)
+    return _connect_signal(obj, name, callback, user_arg, weak_args, user_args)
```
xnox: Hm.... I have an iffy feeling about this. Can we not reuse the same name? Or e.g. somehow monkey-patch urwid to do this. I fear we will make the mistake somewhere of using `urwid.connect_signal` instead of this =/ it would be ideal to not have this at all.... and just use the stock one, somehow.

subiquity/models/welcome.py
mwhudson: Oh I was just doodling, I'll change this back.

subiquity/ui/views/welcome.py
mwhudson: Oh heh, well either way...

subiquitycore/ui/buttons.py
mwhudson: Yeah, I'm not particularly happy about this but it works.

I think we need to be able to override the labels.

subiquitycore/ui/utils.py
mwhudson: Yeah I don't like this greatly either. We could just change the places where it is needed to do connect_signal(w.base_widget) instead but that seems error-prone and annoying too.

mwhudson: I just removed the wrapper in favour of using on_press everywhere.

PR #235, opened 2017-09-15
[RFC] more conjure-up like buttons for moving between screens

mwhudson:
Looking for feedback on whether this is an improvement or not. Demo at:

https://asciinema.org/a/cH1ryq05PVMCwv4d7N4HdMvQf

There is some un-essential wonkiness left (in particular, focus goes to the wrong place by default quite a lot, some views get big buttons that shouldn't) but that's just because I didn't want to spend too long on this before deciding if this even made sense.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,22 +2,10 @@
 # Makefile for subiquity
 #
 NAME=subiquity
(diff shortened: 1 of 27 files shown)


mwhudson:
Abandoning this experiment for now.

xnox:
Ack, also not sure if mpt would like these buttons or not.

PR #236, opened 2017-09-15
Barebones i18n setup.

xnox:
One can launch this in a UTF-8 encoding, e.g. LANG=C.UTF-8 switch language to Russian, cancel from the next screen and see that welcome screen has a Russian welcome message, switch to english again, cancel to return to the welcome screen and notice that welcome message is back in English.

So a general way to run subiquity for translation support is then this:
```
unset LANG LANGUAGE LC_CTYPE LC_NUMERIC LC_TIME LC_COLLATE LC_MONETARY LC_MESSAGES LC_PAPER LC_NAME LC_ADDRESS LC_TELEPHONE LC_MEASUREMENT LC_IDENTIFICATION LC_ALL
LANG=C.UTF-8 ./bin/subiquity-tui --dry-run
```
Since untranslated strings fallback to messageid which is English, I've also added a quick backdoor to spot untranslated strings / fixed width layout errors. It is called FAKE_TRANSLATE.

```
LANG=C.UTF-8 FAKE_TRANSLATE=1 ./bin/subiquity-tui --dry-run
```
With this, when selecting Russian all STRINGS will become _(STRINGS) to demote that they have been marked for translation. Anything without _() wrapper in the UI, is thus untranslatable.

Before rebase onto master, buttons did get translated.... but I guess they are somehow now pre-initialised, instead of executing _() function dynamically at runtime thus they end up having the language that was there before subiquity.i18n got imported and do not "refresh" upon `switch_language()` calls. I will work on fixing this up. Also, at least on my machine, the initial language selection correctly gets native label for russian - but not on the custom.iso will fix this up shortly.

Please review and merge this, such that we can setup translations in rosetta and start gathering translations.

### DIFF
--- a/bin/subiquity-tui
+++ b/bin/subiquity-tui
@@ -19,6 +19,7 @@ import sys
 import logging
 import signal
 
+from subiquitycore.i18n import *
 from subiquitycore.log import setup_logger
 from subiquitycore import __version__ as VERSION
 from subiquitycore.core import ApplicationError
--- a/debian/control
+++ b/debian/control
@@ -6,9 +6,11 @@ Build-Depends: bzr,
                debhelper (>= 9),
                dh-python,
                dh-systemd,
+               iso-codes,
                git,
                python3,
                python3-setuptools,
+               python3-distutils-extra,
                python3-yaml,
                python3-attr
 Standards-Version: 3.9.5
@@ -23,6 +25,7 @@ Depends: curtin,
          python3,
          python3-attr,
          subiquitycore,
+         iso-codes,
          ${misc:Depends},
          ${python3:Depends}
 Description: Ubuntu Server Installer
--- a/debian/subiquity-tools.install
+++ b/debian/subiquity-tools.install
@@ -1,1 +1,0 @@
-installer                              usr/share/subiquity/
--- a/debian/subiquitycore.install
+++ b/debian/subiquitycore.install
@@ -1,1 +1,2 @@
 usr/share/subiquity/subiquitycore
+usr/share/locale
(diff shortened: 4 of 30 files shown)


mwhudson:
This is awesome to see, thanks. Did you miss a step in your instructions though, I get "FileNotFoundError: [Errno 2] No translation file found for domain: 'subiquity'" when I try to run it.

It would perhaps be easier to review if the infrastructure was in one commit and putting _() in another but oh well. I'm just going to approve it anyway I expect :)

mwhudson:
Oh yeah and console-conf will need to jam def _(x): return x into builtins, but otherwise should work?

xnox:
`FileNotFoundError: [Errno 2] No translation file found for domain: 'subiquity'" when I try to run it.`

Please run `python3 setup.py clean build` that will produce .mo files in the build-tree, and they will become usable (there is code to detect in-tree compiled translations)

xnox:
console-conf in the top script should `import subiquitycore.i18n` or some such. Or it can do a manual call `gettext.install('subiquity')` not sure. How do I run console-conf and does it make sense to also localise?

mwhudson:
> console-conf in the top script should import subiquitycore.i18n or some such. Or it can do a manual call gettext.install('subiquity') not sure. 

Either works I guess.

> How do I run console-conf 

You run console-conf by running ./bin/console-conf-tui instead of ./bin/subiquity-tui, that's all.

> and does it make sense to also localise?

Hm. IIUC Ubuntu Core is not in general localized for space reasons. But translating console-conf might be worth it nevertheless. Ask in the forum? :)

mwhudson:
> Please run python3 setup.py clean build that will produce .mo files in the build-tree, and they will become usable (there is code to detect in-tree compiled translations)

Ah yes, thanks. Looks great now :)

PR #238, opened 2017-09-18
hook up curtin's event reporting into journald and watch these messages during install

mwhudson:
100% less threading, maybe makes for a happier @xnox


mwhudson:
Ah so I found out why this wasn't working in kvm: it was systemd version differences after all. It can be fixed by building the snap on an artful system so that the right libsystemd-dev gets used, but that's going to make local testing more awkward as I don't run artful :)

xnox:
I thought that snapcraft cleanbuild can be told to build the snap in artful. Which should make local testing easier.

mwhudson:
Ah haha nope, it hard codes xenial. I'm attempting to hack up a PR that fixes this...

PR #240, opened 2017-09-18
Fix cancel button sizing in installpath view.

xnox:
This issue is visible with FAKE_TRANSLATE or going to russian language. `[ Cancel ]` button is fixed width, yet it is now obviously variable length due to languages. I wonder if we can somehow make the button automatically "shrink" and be centered, no? cause otherwise I'll have to fix up all the buttons layouts.

### DIFF
--- a/subiquity/ui/views/installpath.py
+++ b/subiquity/ui/views/installpath.py
@@ -23,7 +23,7 @@ from urwid import BoxAdapter
 
 from subiquitycore.ui.lists import SimpleList
 from subiquitycore.ui.buttons import back_btn, menu_btn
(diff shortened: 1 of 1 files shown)


xnox:
@mwhudson do you have any opinions on this? at least in Gtk+ one was able to say "add this button to this frame, but make it natural/packed size, do not make it span the width of the whole screen" and I'm failing to replicate with urwid without manually calculating and forcing the size of the labels.

mwhudson:
As far as I can tell urwid doesn't really have a concept of 'natural width' so I think we're going to have to do something ourselves, I guess a helper for making a pile of buttons?

PR #242, opened 2017-09-20
Fix up more fixed layouts with button_pile().

xnox:


mwhudson:
Grur how did I miss those, thanks!

PR #243, opened 2017-09-21
have enter advance to the next focus item in piles and listboxes

mwhudson:
this has been requested a few times, turns out to be easy enough.


xnox:
Ack, looks good.

PR #244, opened 2017-09-22
store curtin's output in journald

mwhudson:
This changes to put curtin's output into journald and use a
systemd.journal.Reader instance to read it. It's a bit simpler than the current
code, which is nice. It also allows us to play with displaying kernel or other
messages intermixed with the curtin output, which might be interesting to
explore.

This does mean that the curtin output is not on disk anywhere, but as it's
pretty trivially recoverable with `journalctl -t curtin_output -o cat` I think
that's OK.

### DIFF
--- a/subiquity/controllers/installprogress.py
+++ b/subiquity/controllers/installprogress.py
@@ -25,8 +25,6 @@ from subiquitycore import utils
 from subiquitycore.controller import BaseController
 
 from subiquity.curtin import (CURTIN_CONFIGS,
-                              CURTIN_INSTALL_LOG,
-                              CURTIN_POSTINSTALL_LOG,
                               curtin_install_cmd,
                               curtin_write_network_config,
                               curtin_write_reporting_config)
@@ -61,9 +59,9 @@ class InstallProgressController(BaseController):
         self.progress_view = None
         self.install_state = InstallState.NOT_STARTED
         self.postinstall_written = False
-        self.tail_proc = None
         self.journald_forwarder_proc = None
         self.curtin_event_stack = []
+        self.curtin_output_buffer = []
 
     def curtin_wrote_network_config(self, path):
         curtin_write_network_config(open(path).read())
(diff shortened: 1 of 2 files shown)


xnox:
This looks good. Imho we should be rotating and storing install time journal into the installed system, like it is done by d-i / ubiquity. Either a text dump, or better the journal files themself. As in fetch `/run/log/journal/*` and `/var/log/journal/*` and store it in the installed system at `/var/log/installer/journal/` Actually this is such a good idea that ubiquity should do it too!

subiquity/controllers/installprogress.py
```diff
@@ -87,18 +85,15 @@ def curtin_error(self):
         else:
             self.default()
 
-    def run_command_logged(self, cmd, logfile_location):
-        with open(logfile_location, 'wb', buffering=0) as logfile:
+    def run_command_logged_to_journal(self, cmd):
+        with journal.stream("curtin_output") as logfile:
             log.debug("running %s", cmd)
             cp = subprocess.run(
                 cmd, stdout=logfile, stderr=subprocess.STDOUT, stdin=subprocess.DEVNULL)
```
xnox: could be also done as `systemd-run --wait cmd` which will run requested command as a separate temporary unit with logging to journal. This just an fyi, rather than actual suggestion to do it this way.

xnox:
Looks good. Needs to resolve conflicts, and i guess land once BB opens. Imho it is a good idea to use journal, given that we have it running.

vorlonofportland:
On Sat, Oct 14, 2017 at 02:49:45PM -0700, Dimitri John Ledkov wrote:

> Looks good.  Needs to resolve conflicts, and i guess land once BB opens. 
> Imho it is a good idea to use journal, given that we have it running.

"Once bb opens"?  This is published as a snap, landing is not blocked by the
Ubuntu archive freeze.

(We want to be careful about what we choose to promote to the stable snap
channel at the 11th hour, but we should not be wary of landing to git +
edge.)

xnox:
@vorlonofportland your comment is unjust, sir. since you have switched us from edge to stable channel in the time between my comment and yours ;-) thus mine was factually correct at pixel time.

mwhudson:
I just merged another branch which does this (amongst other things)

PR #246, opened 2017-09-26
change the progress bar to show x / y rather than a %age

mwhudson:
Also make its progress more automatic.


mwhudson:
mpt thought this was an improvement, good enough for now

xnox:
+1

PR #248, opened 2017-09-28
color tweaks

mwhudson:
This adds a bit of machinery to allow us to overwrite the first 8 console colors with 24 bit values we choose and make sure that the color used in a xterm are close to those.

It also ensures that our default background color is actually used and tweaks the colors a bit (probably for the worse but it'll be possible to sensibly change those now)


xnox:
Looks good, but then i need to re-fix this up for the fbterm case too. But that could be done later.

PR #249, opened 2017-09-29
misc networking fixes

mwhudson:
I have a more complicated networking change brewing which doesn't quite work and these are bugs I found while debugging that.


subiquitycore/controllers/network.py
```diff
@@ -282,6 +282,7 @@ def data_ready(self, fd):
         if code != 0:
             log.debug("waiting 0.1 to let udev event queue settle")
```
xnox: Hm.... this should really call `udevadm settle` no?

xnox:
Ok, not sure why or how. Side comment about udevadm settle usage.

PR #250, opened 2017-10-04
Drop Installpath controller, whilst it's single option.

xnox:
@vorlonofportland

### DIFF
--- a/subiquity/core.py
+++ b/subiquity/core.py
@@ -27,7 +27,6 @@ class Subiquity(Application):
     project = "subiquity"
     controllers = [
             "Welcome",
-            "Installpath",
             "Network",
             "Filesystem",
             "Identity",


mwhudson:
I guess the only question is whether there is value in showing "Welcome to Ubuntu! The world's favorite platform for clouds, clusters, and amazing internet things. This is the installer for Ubuntu on servers and internet devices." (or its translation) to the user. We could just replace the option "list" with a button saying "begin" or something. But I don't really have a strong opinion.

vorlonofportland:
Because this install option screen will /eventually/ return (because we need an 'install maas' option), I don't think it's worth bothering with that right now and instead we should focus on not requiring the user to have to click extra buttons needlessly.

mwhudson:
+1 to merge this then :-)

PR #251, opened 2017-10-05
do not filter block devices by major

mwhudson:
instead, filter out devices with devpath /devices/virtual/...

Please review skeptically, I'm not really sure what this code is trying to do. Ping @raharper too!

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -318,12 +318,11 @@ class FilesystemModel(object):
 
     def probe(self):
         storage = self.prober.get_storage()
-        VALID_MAJORS = ['8', '253']
         currently_mounted = self._get_system_mounted_disks()
         for path, data in storage.items():
             if path in currently_mounted:
                 continue
-            if data['DEVTYPE'] == 'disk' and data['MAJOR'] in VALID_MAJORS:
+            if data['DEVTYPE'] == 'disk' and not data["DEVPATH"].startswith('/devices/virtual'):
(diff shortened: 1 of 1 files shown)


vorlonofportland:
This looks sensible to me.

cyphermox:
I agree; this should work. The intent was more to avoid picking up pre-made virtual devices (raids and whatnot) before we reach the partitioner; and we should honor virtio disks. The thing is, is this likely to bite us later for multipath/RAID/etc. ?

raharper:
virtual will block things like bcache, device-mapper and raid; so I don't think that's going to work.
It was mean to skip things like loop devices and cdroms.  We may need to invert from whitelist to blacklist;

mwhudson:
At the point this runs (just after boot from an installer CD) there will not be anything like bcache, dm or raid set up surely? My understanding was that this code was just looking for the "physical" (which obviously <wink> includes virtio) devices that can be formatted or (eventually) composed into such things.

raharper:
This image is booting on machines with unknown block devices; there could
be *anything* on the disks.  The virtio disks any of the storage devices
may already have configuration on them.
We're looking for disks that can be utilized.

mwhudson:
Well yes, subiquity currently doesn't care at all about what is already on the system it runs on, that's kinda a separate issue though isn't it?

mwhudson:
I think I've argued everyone into accepting that this is at the least no worse than the status quo :)

raharper:
Yes, you're right.  The underlying devices will still show.

PR #252, opened 2017-10-05
add a trivial oneshot service we can order the subiquity.service in t…

mwhudson:
…he install after to ensure the mounts are set up

This is part of my workaround/fix for https://bugs.launchpad.net/subiquity/+bug/1721414

### DIFF
--- a/bin/started
+++ b/bin/started
@@ -1,0 +1,2 @@
+#!/bin/sh
+true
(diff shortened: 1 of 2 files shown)


mwhudson:
Other half of the workaround/fix: https://code.launchpad.net/~mwhudson/livecd-rootfs/subiquity-ordering/+merge/331844

xnox:
May or may not do what you want, but on the surface looks sane (however insane this is, lol)

PR #255, opened 2017-10-09
Vorlon/fix up tests

vorlonofportland:
Partial fixes for failing tests.

Running 'nosetests3' fails in the subiquity tree with a number of failures due to missing modules, undefined functions, etc.  This is a partial fix for these problems.

The test suite still fails with errors around the removed 'blockdev' module.  I'm not sure if those tests should be ported to something else, or if they should be dropped.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ yaml
 testtools
 mock
 attr
+systemd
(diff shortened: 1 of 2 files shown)


mwhudson:
I'm not sure how much value these tests have but fine.

PR #256, opened 2017-10-09
add a way to run console-conf or subiquity with a subset of the screens

mwhudson:
You can break things this way but it's also super helpful when hacking on the identity screen's validation to be able to jump straight there...

Mostly PRing this for visibility but reviews welcome too :)


xnox:
yes please! (ubiquity has something like that too, a plugin loader that shows just one page, which works enough for unit-testing a single page transitions / button clicks etc)

PR #258, opened 2017-10-10
do not overwrite the top 8 colors of the palette with black

mwhudson:
Switch to vt2 and running e.g. ls was very confusing without this...


xnox:
Hahahahah =) this is cute.

PR #259, opened 2017-10-10
filter out floppy and read-only block devices

mwhudson:
Possibly we should filter a lot more block devices. But the others seem mostly mighty obscure.


xnox:
yes floppy must die

PR #260, opened 2017-10-11
add user to the correct set of default groups

mwhudson:
It's not pretty but it does work!

For https://bugs.launchpad.net/subiquity/+bug/1722574

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -50,6 +50,14 @@ parts:
       'bin/started': usr/bin/started
     prime:
       - usr/bin
+  users-and-groups:
+    plugin: dump
+    build-packages:
+      - user-setup
+    prepare: |
+      echo "get passwd/user-default-groups" | \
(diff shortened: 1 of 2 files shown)


cyphermox:
Ack

xnox:
why is this not in `curtin` itself?

mwhudson:
Curtin has no support for configuring users at all, what subiquity does is
configure cloud-init to create users on boot (cloud-init has a manually
maintained copy of this list too, but I wasn't sure how to make use of
that). This is frankly a bit shit and probably once we can get curtin to
leave the family mounted we should just call chroot /target adduser etc etc.

xnox:
@smoser is there a way to specify to cloud-init to create an "admin" user with default set of groups that it means, i.e. passwd default set + sudo? or some such.

smoser:
xnox there isnt a way to reference the default set of groups for cloud-init.

PR #261, opened 2017-10-11
clunky way of indicating which network config method is active

mwhudson:
I'm waiting on design for a proper go at this but this is probably a sufficient temporary improvement.

"Video" at https://asciinema.org/a/kmUqrUOGDVDRfdo91GV5abU5J

### DIFF
--- a/subiquitycore/ui/views/network_configure_interface.py
+++ b/subiquitycore/ui/views/network_configure_interface.py
@@ -34,8 +34,10 @@ class NetworkConfigureInterfaceView(BaseView):
     def _build_widgets(self):
         self.ipv4_info = Pile(_build_gateway_ip_info_for_version(self.dev, 4))
         self.ipv4_method = Pile(self._build_ipv4_method_buttons())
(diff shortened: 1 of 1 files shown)


cyphermox:
Looks good to me.

subiquitycore/ui/views/network_configure_interface.py
```diff
@@ -59,18 +61,47 @@ def _build_body(self):
...
+            else:
+                p = "( ) "
+            b.set_label(p + b.label[4:])
+
     def _build_ipv4_method_buttons(self):
         button_padding = 70
 
         buttons = [
-            menu_btn(label=_("Use a static IPv4 configuration"),
+            menu_btn(label=_("    Use a static IPv4 configuration"),
```
vorlonofportland: This breaks existing translations; better to express this as:
menu_btn(label="    " + _("Use a static IPv4 configuration"),

xnox: yes! will fix up.

PR #264, opened 2017-10-20
reduce default padding, tighten scrollable areas

mwhudson:
For a couple of branches I have brewing I want (a) to be able to fill the whole area between the header and footer (b) to add scrollbars, which look very strange if all of the just-mentioned area is what is in the list box.

Fixing the former is easy, just remove the default padding but then of course all existing views need the padding re-adding. While I was doing that I switched most views to a standard layout of

 1 line of padding
 a stretchy listbox containing the bulk of the content
 1 line of padding
 the done/save/cancel/etc buttons
 1 line of padding


xnox:
look ok.

PR #266, opened 2017-11-15
changes following changes in probert

mwhudson:
the only really visible effect of all this is to make --dry-run
--machine-config foo show the network config from foo, not the machine
running subiquity. (The existing configs won't work, though)


mwhudson:
See https://github.com/CanonicalLtd/probert/pull/37

PR #268, opened 2017-11-20
move curtin config rendering to a method on SubiquityModel

mwhudson:
this moves to curtin's native netplan support as a side effect


mwhudson:
@tdaitx this might conflict with the work you're doing on proxies. hopefully it's easier to make the required changes to this version though!

subiquity/models/subiquity.py
```diff
@@ -28,3 +31,81 @@ def __init__(self, common):
...
+        return {
+            'users': [user_info],
+            'hostname': self.identity.hostname,
+        }
+
+    def _write_files_config(self):
+        # TODO, this should be moved to the in-target cloud-config seed so on first
+        # boot of the target, it reconfigures datasource_list to none for subsequent
+        # boots.
+        # (mwhudson does not entirely know what the above means!)
```
xnox: It means, we should not re-initialize a server instance, if one wipes `/var/lib/cloud/instance-id` state and like reset user changed passwords. Or e.g. randomly detect a usb stick with a config drive, and randomly reinitialise this server with that usb stick data.

On the other hand, storing this enables one to easily "reset" this server back to original passwords....

subiquity/models/subiquity.py
```diff
@@ -28,3 +31,81 @@ def __init__(self, common):
...
+        userdata = '#cloud-config\n' + yaml.dump(self._cloud_init_config())
+        return {
+            'postinst_metadata': {
+                'path': 'var/lib/cloud/seed/nocloud-net/meta-data',
+                'content': 'instance-id: inst-3011\n',
```
xnox: Probably should generate a UUID here.

subiquity/models/subiquity.py
mwhudson: OK, that makes sense I guess. How do we do that? Add a cloud-init runcmd that runs something like rm -rf /var/lib/cloud/seed/nocloud-net or echoes "datasource_list: None" into some file in etc/cloud? (should we anyway set datasource_list to [NoCloud, None]?)

subiquity/models/subiquity.py
mwhudson: yes, fair. is the meta-data file yaml?

PR #269, opened 2017-11-21
refactor color / palette setup

mwhudson:
Replace some incomprehensible garbage with slightly better factored incomprehensible garbage.


mwhudson:
BTW another bonus of this is that it works properly in asciinema ...

PR #271, opened 2017-11-22
implement new color scheme from design

mwhudson:
https://asciinema.org/a/A0jnRyc0g0ksNl4LSoL1YQK0I

### DIFF
--- a/subiquity/palette.py
+++ b/subiquity/palette.py
@@ -17,50 +17,44 @@
 
 
 COLORS = [
-    ("bg",        (0x00, 0x00, 0x00)),
+    ("bg",        (0x11, 0x11, 0x11)),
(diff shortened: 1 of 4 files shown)


zyga:
My only comment, after watching the nice recording, is that the terminal output scroll bar is a bit confusing.

mwhudson:
The way it bounces around in size? yes that's a bit odd. Urwid's listbox is a thing of no small complexity :(

mwhudson:
> My only comment, after watching the nice recording, is that the terminal output scroll bar is a bit confusing. (which is not really related except for maybe the new colors)

Now fixed: https://github.com/CanonicalLtd/subiquity/commit/bc4978fd049c04c1a6dd83acadd52f66338bcc47 thanks for the nag.

PR #272, opened 2017-11-27
add some simple view unit tests and travis integration

mwhudson:
not much but it's a start


xnox:
Will it blend? looks like it did pass the checks.

PR #273, opened 2017-11-29
Update setup.py

Cesar03:


setup.py
```diff
@@ -41,9 +41,9 @@
       description="Ubuntu Server Installer",
       long_description=__doc__,
       author='Canonical Engineering',
-      author_email='ubuntu-dev@lists.ubuntu.com',
-      url='https://github.com/CanonicalLtd/subiquity',
-      license="AGPLv3+",
+      author_email='ubuntu-dev@list.ubuntu.com',
+      url='https://github.com/CanonicalLtd/suiquity',
```
zyga: This is a 404

setup.py
```diff
@@ -41,9 +41,9 @@
       description="Ubuntu Server Installer",
       long_description=__doc__,
       author='Canonical Engineering',
-      author_email='ubuntu-dev@lists.ubuntu.com',
-      url='https://github.com/CanonicalLtd/subiquity',
-      license="AGPLv3+",
+      author_email='ubuntu-dev@list.ubuntu.com',
```
raharper: lists -> list is not correct

setup.py
raharper: typo in the project name

setup.py
```diff
@@ -41,9 +41,9 @@
...
+      url='https://github.com/CanonicalLtd/suiquity',
+      license="AGPLv2+",
```
raharper: And the License hasn't changed that I'm aware of.

mwhudson:
I'm going to close this, feel free to re-open but please add an explanation if you do.

PR #275, opened 2018-01-09
Re-add installpath controller, with added MAAS views.

xnox:
This re-adds back the installpath controller.

Adds views to setup MAAS with a bit too much wording, and too little validation.

The installpath model is extended to blindly override cutin's debconf-selections; and cloud-init packages stanzas.

curtin itself can install missing packages... but there is no yaml stanza for missing packages it seems. I'm not sure how I feel about installing maas packages by cloud-init on first boot; and how I feel about logging / passing the maas admin password in plaintext.

= Testing =
I did not manage to test this successfully, and to verify that
 1) debconf selections did happen by curtin
 2) cloud-init did install the right packages
 3) maas is installed and works, and one can login


mwhudson:
I've pushed some updates to simplify the view code a bit (IMO anyway) and rebased. I'm happy with the UI side of things now but I (and @vorlonofportland ) aren't really happy with installing maas via cloud-init. I'll brain dump the plan for that in trello I guess.

mwhudson:
[add option to install maas server](https://trello.com/c/DFS8XI9s/379-add-option-to-install-maas-server)

mwhudson:
I made some more small changes. I think we should merge this to get the UI in place, and then get back to making the back end side better.

PR #276, opened 2018-01-12
implement keyboard selection

mwhudson:
Add a view immediately after language selection. Please review/try out!


mwhudson:
I've pushed up the changes I made to make the layouts translatable. It's actually a bit simpler than the previous version I think.

mwhudson:
[Subiquity: keyboard / locales / etc](https://trello.com/c/XIkUwfuJ/192-subiquity-keyboard-locales-etc)

subiquity/models/keyboard.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def load_language(self, code):
+        if code == self.current_lang:
+            return
+
+        self._clear()
+
+        raw = gzip.open(self._kbnames_file)
+        try:
+            with io.TextIOWrapper(raw) as kbdnames:
```
cyphermox: Does kbdnames need to be compressed? It would be on a CD, in a squashfs that is already compressed, etc.

cyphermox:
Looks good to me, I tested the keyboard code in a subiquity image and although the install failed in the end on hardware, the keyboard interface looked great and seemed to work as intended. I also did code review and didn't see anything horrible jump out aside from the small details in inline comments.

PR #277, opened 2018-01-17
remove the game of running curtin twice in favour of new curtin features

mwhudson:
subiquity previously ran curtin twice, once to do the install and the second
time to configure the user data. All this second run did was to drop some
files into the fs and now that we can tell curtin not to unmount everything
when it is done, this can be achieved much more simply. So do that, and rip out
a bunch of the complexity having two curtin configs etc brings.

### DIFF
--- a/subiquity/controllers/installprogress.py
+++ b/subiquity/controllers/installprogress.py
@@ -37,15 +37,13 @@ from subiquity.ui.views import ProgressView
 
 log = logging.getLogger("subiquitycore.controller.installprogress")
 
+TARGET = '/target'
 
 class InstallState:
     NOT_STARTED = 0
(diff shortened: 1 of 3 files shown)


xnox:
Yes! I couldn't understand why we need to run curtin twice, felt very counter intuitive.

xnox:
With NVMe hard-drives, and fast raids, the copying time is the least amout of time spend installing the server. Typing in the passwords and making sure one has the right partitioning / ip addresses, interactively, will always take the longest amount of time. And it is actually a feature, that one can go back to the partitioning step, even from the late user password screen. I strongly hope that we can move the install all the way to the end.

mwhudson:
I merged this, not sure why github didn't notice.

PR #278, opened 2018-01-19
Add partitions configuration in answers.yaml

musicguitar:
Currently, the answers.yaml supports auto installation, but the partitioning function only supports guided mode. Here to add a "volumes" in answers.yaml for pre-partitioning configs.

The "volumes" can use either disk number(disk-[0-9]) or partition path (/dev/<device node of disk>) to identify the disk.
Each disk can be configured by the following attributes:
- size
  The partition size with human readable units : 'B', 'K', 'M', 'G', 'T', 'P'
  Leave size as empty, it would use disk free space as partition size
- flag
  The flag attribute is optional. It can be set to 'boot' or 'grub_bios' for boot partition.
- filesystem
  The filesystem type: 'ext4', 'fat32', 'btrfs', 'swap', 'xfs'
- filesystem-label
  The 'filesystem-label' is optional. Add this attribute to set label of filesystem.
- mount
  The mount attribute is optional. Leave empty for unmounted, or set mount path for partition.

Please see the example file: examples/answers-partitions.yaml

If any volumes attributes are invalid, the auto installation would fall back to manual mode.
The error can be check in log file.

### DIFF
--- a/examples/answers-partitions.yaml
+++ b/examples/answers-partitions.yaml
@@ -10,17 +10,13 @@ Filesystem:
               filesystem: fat32
               filesystem-label: boot
               mount: /boot/efi
-            - size: 
-              flag: 
-              filesystem: ext4
+            - filesystem: ext4
               filesystem-label: rootfs
               mount: /
       - /dev/sdh:
             - size: 100G
-              flag: 
               filesystem: ext4
               filesystem-label:
-              mount: /home
             - size: 0
               flag:
               filesystem: ext4
--- a/examples/answers-partitions.yaml
+++ b/examples/answers-partitions.yaml
@@ -34,6 +34,7 @@ Filesystem:
       #        filesystem: ext4
       #        filesystem-label:
       #        mount: /tmp
+      # Fail case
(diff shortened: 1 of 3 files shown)


raharper:
Why wouldn't the answer section include actual storage configuration yaml rather than this new custom format?

http://curtin.readthedocs.io/en/latest/topics/storage.html

And I think instead of attempting to have a new format to instantiate Filesystem model objects; it may be better to store the raw storage-config yaml in the model and if present use that instead of having the model render a storage-config from objects.

musicguitar:
Yes, I think your suggestion is better.
I am planing to add a subiquity-tui opts with --storage and default file is /subiquity_config/subiquity-config-storage.yaml
When storage-config yaml presents, subiquity would ignore the filesystem setting in UI.

Please let me know, if any suggestions to improve my proposal.

I'll resend a pull request when it's ready.

musicguitar:
I had resent a pull request which follow the yaml format as curtin.
Please help to review the PR #279 

Close this PR.

PR #279, opened 2018-01-26
Add storage configuration for auto install

musicguitar:
Currently, the answers.yaml supports auto installation, but the partitioning function only supports guided mode. Here to add a "--storage" argument in subiquity-tui, or the default yaml file in /subiquity_config/subiquity-config-storage.yaml

The yaml format is same as curtin storage which as:
http://curtin.readthedocs.io/en/latest/topics/storage.html

When the storage config yaml file presents, the storage partition setting in UI would be skipped.

### DIFF
--- a/bin/subiquity-tui
+++ b/bin/subiquity-tui
@@ -66,12 +66,14 @@ def parse_options(argv):
                         help='run in uefi support mode')
     parser.add_argument('--screens', action='append', dest='screens', default=[])
     parser.add_argument('--answers')
+    parser.add_argument('--storage', dest='storage_conf')
     return parser.parse_args(argv)
 
 
 LOGDIR = "/var/log/installer/"
 
 AUTO_ANSWERS_FILE = "/subiquity_config/answers.yaml"
+STORAGE_CONFIG_FILE = "/subiquity_config/subiquity-config-storage.yaml"
 
 def main():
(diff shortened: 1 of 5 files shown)


bin/subiquity-tui
```diff
@@ -66,12 +66,14 @@ def parse_options(argv):
...
     parser.add_argument('--screens', action='append', dest='screens', default=[])
     parser.add_argument('--answers')
+    parser.add_argument('--storage', dest='storage_conf')
     return parser.parse_args(argv)
 
 
 LOGDIR = "/var/log/installer/"
 
 AUTO_ANSWERS_FILE = "/subiquity_config/answers.yaml"
+STORAGE_CONFIG_FILE = "/subiquity_config/subiquity-config-storage.yaml"
```
raharper: I was originally suggesting that within the answers file you include a 'storage' configuration in the curtin format.
This way users only have a single --answers parameter;  if that input file include a 'storage' config then, subiquity can render that 'storage' section to the subiquity-config-storage.yaml itself.  This prevents users from needing to provide both --answers <my answers> and --storage <my storage config>

raharper:
I think it would be better to move the storage config back into the answers.yaml and pull the storage config from there and hand it to the storage model.

PR #280, opened 2018-02-09
have a subiquity-service be part of the snap

mwhudson:
To avoid having systemd try to start subiquity before snapd has finished seeding the snaps, let's have snapd create the subiquity service file. We can't set all the options we want to via subiquity so some of the config is still going to have to be done in livecd-rootfs (branch coming up for that)


mwhudson:
[fix race condition that sometimes prevents subiquity from starting](https://trello.com/c/5fvsa0k3/384-fix-race-condition-that-sometimes-prevents-subiquity-from-starting)

PR #281, opened 2018-02-11
have initial language match users choice in isolinux screen

mwhudson:


mwhudson:
Tested in a VM, seems to work (with hungarian as the selected language)

PR #289, opened 2018-02-26
do not show full curtin log output by default

mwhudson:
https://asciinema.org/a/fBAnH8LdWTeEDOYjQg5Umiuz1

This changes the final view to just show the events curtin reports to subiquity by default, with a dose of indentation and a spinner that hopefully conveys a feeling of progress.

It also improves the display of install progress in the footer, simplifies some logic in the installprogress controller and switches to storing curtin output in the journal, because it just makes life easier that way.

### DIFF
--- a/scripts/replay-curtin-log.py
+++ b/scripts/replay-curtin-log.py
@@ -17,7 +17,7 @@ logger.addHandler(handler)
 
 json_file = sys.argv[1]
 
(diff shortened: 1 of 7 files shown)


mwhudson:
[showing curtin's logs in full by default is unfriendly](https://trello.com/c/VCqW5IPb/165-showing-curtins-logs-in-full-by-default-is-unfriendly)

PR #292, opened 2018-03-12
add a border to all overlays to make them more readable

mwhudson:
This gives a padding of one row above and below the overlay and 3 characters to the left and right: https://asciinema.org/a/rIYPrq81zOY8kgZE9fF5ajriW


mwhudson:
We can easily tweak the size later if people have Opinions about this, merging now

PR #294, opened 2018-03-13
improve scrolling experience

mwhudson:
Before this change, subiquity has lots of ListBoxes that just contain a single
Pile containing all their contents. This is (a) a bit silly (b) make some parts
of the scrolling experience a bit poor, for example urwid tries to scroll all
of a ListBox element into view when it gets focus but this is defeated by
shoving all the elements into a Pile (this causes
https://bugs.launchpad.net/subiquity/+bug/1750058 and a few other strange
bits).

The fix for this is obvious (don't wrap ListBox elements in a Pile) but this
breaks some aspects of tab cycling (when you shift tab back into a listbox you
want the last element of the box to be both selected and scrolled into view,
that sort of thing). Fixing all these bits of broken behaviour required
rewriting the tab cycling implementation to the point of copy/paste/hack-ing
the Pile.keypress method. Rather than doing the same for Columns, I just
prevent the creation of Columns with more than 1 selectable, which as we want
subiquity to be navigable with up/down/return does not seem so bad.

As penitence for all this, I've added a bunch of commentary explaining what is
going on.

### DIFF
--- a/subiquity/ui/views/filesystem/filesystem.py
+++ b/subiquity/ui/views/filesystem/filesystem.py
@@ -79,22 +79,25 @@ class FilesystemView(BaseView):
         self.model = model
         self.controller = controller
         self.items = []
-        self.body = [
+        body = [
             Text(_("FILE SYSTEM SUMMARY")),
             Text(""),
             Padding.push_4(self._build_filesystem_list()),
             Text(""),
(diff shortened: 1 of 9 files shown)


mwhudson:
This has the potential to break things in obscure ways, testing of views I might have forgotten about would be appreciated!

PR #296, opened 2018-03-13
keyboard fixes: show dialog while applying, re-setup shift tab afterwards

mwhudson:
needs testing in a vm but in the mean time: https://asciinema.org/a/MBWfJvicE8T2L1ynDgeMeLhD4


mwhudson:
now tested in a vm and works after a simple fix

PR #301, opened 2018-03-22
adjust details of partition sizing

mwhudson:
A collection of small hacks to the partition sizing logic. Any two or three of them would probably fix https://bugs.launchpad.net/subiquity/+bug/1750845 but I'm not really interested in minimising the fix...


mwhudson:
[Crash when adding partition with exact max size](https://trello.com/c/11KYG2hJ/728-crash-when-adding-partition-with-exact-max-size)

PR #303, opened 2018-03-26
address confusion around the boot partition

mwhudson:
https://asciinema.org/a/pLJ95ajEDPsOEJQrZVQp3cVbv

Not sure about the UI for moving the boot partitions between drives but can't think of anything much better...

### DIFF
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -217,6 +217,11 @@ class FilesystemController(BaseController):
         back()
 
     def make_boot_disk(self, disk):
+        for p in self.model._partitions:
+            if p.flag in ("bios_grub", "boot"):
+                p.device._partitions.remove(p)
+                disk._partitions.insert(0, p)
+                p.device = disk
         self.partition_disk(disk)
 
     def connect_iscsi_disk(self, *args, **kwargs):
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -216,6 +216,9 @@ class FilesystemController(BaseController):
                 self.model.add_mount(fs, spec['mount'])
         back()
 
+    def make_boot_disk(self, disk):
+        self.partition_disk(disk)
+
     def connect_iscsi_disk(self, *args, **kwargs):
         # title = ("Disk and filesystem setup")
         # excerpt = ("Connect to iSCSI cluster")
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -219,9 +219,13 @@ class FilesystemController(BaseController):
     def make_boot_disk(self, disk):
         for p in self.model._partitions:
             if p.flag in ("bios_grub", "boot"):
+                full = p.device.free == 0
                 p.device._partitions.remove(p)
+                if full:
+                    largest_part = max((part.size, part) for part in p.device._partitions)[1]
+                    largest_part.size += p.size
                 if disk.free < p.size:
-                    largest_part = max((p.size, p) for p in disk._partitions)[1]
+                    largest_part = max((part.size, part) for part in disk._partitions)[1]
                     largest_part.size -= (p.size - disk.free)
                 disk._partitions.insert(0, p)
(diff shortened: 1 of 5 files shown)


mwhudson:
[sort out confusion around the bios_grub partition etc](https://trello.com/c/bEIx34Xp/717-sort-out-confusion-around-the-biosgrub-partition-etc)

vorlonofportland:
Looks pretty good, a few requests/comments:
 - give a more precise description for the bios_grub partition than 'unformatted' in the summary.  maybe 'bios_grub' itself?
 - in the detail screen when clicking on the bios_grub partition, please include a summary as the first line. perhaps:
   Required bootloader partition
   GRUB will be installed onto the target disk's MBR.
   However, on a disk with a GPT partition table, there is not enough space after the MBR for GRUB to store its second-stage core.img, so a small unformatted partition is needed at the start of the disk.  It will not contain a filesystem and will not be mounted, and cannot be edited here.
- Don't present 'Save' and 'Cancel' buttons on this screen that both do the same thing.  Should be a single 'OK' button.
- detail page incorrectly lists a mountpoint of /boot for this, which is confusing.
- 'Make boot disk' is ambiguous.  Suggest 'Select as boot disk'.
- When selecting a different boot disk, should we auto-expand the remaining partitions on the previously-selected disk to take up the space previously reserved for GRUB?  (It took me a moment to figure out why the no-longer-selected disk still had a line for '2.000M'.)
- Is 2MiB the correct size for this bios_grub partition?  I know there was discussion of perhaps doing 4MiB alignment by default for all partitions, which would seem to apply here.  Does 2MiB match the current behavior of d-i?

mwhudson:
Thanks for the review! Here's some more videos:

BIOS/MBR: https://asciinema.org/a/yLMLV24UqH5gdwitTlMjcVZCZ
UEFI: https://asciinema.org/a/75cAvM2fGG0NEN5JRc98gGAnZ

> give a more precise description for the bios_grub partition than 'unformatted' in the summary. maybe 'bios_grub' itself?

Done.

> in the detail screen when clicking on the bios_grub partition, please include a summary as the first line. perhaps:

Used your selected text. Also rewrote the UEFI text a bit which wasn't in the previous video.

> Don't present 'Save' and 'Cancel' buttons on this screen that both do the same thing. Should be a single 'OK' button.

Done.

> detail page incorrectly lists a mountpoint of /boot for this, which is confusing.

Done.

> 'Make boot disk' is ambiguous. Suggest 'Select as boot disk'.

Done.

> When selecting a different boot disk, should we auto-expand the remaining partitions on the previously-selected disk to take up the space previously reserved for GRUB? (It took me a moment to figure out why the no-longer-selected disk still had a line for '2.000M'.)

I wondered about this. I added code to shrink the largest partition on the disk being made the boot disk make way, I guess the parallel would be something like "if the disk that was the boot disk is was full before removing the boot partition, expand the largest remaining partition to take the remaining space". 

I've implemented this logic now.

> Is 2MiB the correct size for this bios_grub partition? I know there was discussion of perhaps doing 4MiB alignment by default for all partitions, which would seem to apply here.

I think Dmitri concluded that MiB alignment is fine.

> Does 2MiB match the current behavior of d-i?

d-i creates a 1MiB partition rather than 2MiB (which is also what the curtin docs suggest). I've changed subiquity to match.

vorlonofportland:
+1, thanks!

raharper:
On Tue, Mar 27, 2018 at 6:27 PM, Michael Hudson-Doyle <
notifications@github.com> wrote:

>
> I think Dmitri concluded that MiB alignment is fine.
>
> Does 2MiB match the current behavior of d-i?
>
> d-i creates a 1MiB partition rather than 2MiB (which is also what the
> curtin docs suggest). I've changed subiquity to match.
>

curtin docs recommend 1MiB

http://curtin.readthedocs.io/en/latest/topics/storage.html#partition-command

mwhudson:
> curtin docs recommend 1MiB

Yeah that's what I meant to say, I didn't phrase my comment very clearly (I blamed teething children who kept me up all night :/)

PR #304, opened 2018-03-27
handle non-latin keyboard layouts more intelligently

mwhudson:
https://asciinema.org/a/IESGmVIzLec6feUMGfKEgGWZH (although as that's recorded in a gnome-terminal of course you can't see the actual effect of changing the keyboard)

The behaviour is entirely stolen from d-i -- when the user chooses a non-latin keyboard layout, they are prompted to choose a key stroke to swap between a latin layout and the one they chose.

I've tested this in KVM, it seems to work.

I spent a bit too long trying to find a way to be notified by the kernel of layout changes so I could display the current layout somewhere but didn't find anything (presumably it's possible by tracking things in /dev/input but brrr)

### DIFF
--- a/subiquity/controllers/keyboard.py
+++ b/subiquity/controllers/keyboard.py
@@ -65,7 +65,7 @@ class KeyboardController(BaseController):
             self._done)
 
     def _done(self, fut):
-        self.signal.emit_signal('next-screen')
+        self.loop.set_alarm_in(0.0, lambda loop, ud: self.signal.emit_signal('next-screen'))
 
     def cancel(self):
         self.signal.emit_signal('prev-screen')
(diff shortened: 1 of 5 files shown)


cyphermox:
Nice, keymaps magic. :+1: 

After being convinced that keymaps switching really does work, and checking for myself in kbd's code; looks nice to me. I especially like the "simplification" of some pieces of the code, though at the cost of reimplementing d-i's keyboard-configuration keymap mangling :/

mwhudson:
[Selecting a keyboard layout may make it unfeasible to type network settings](https://trello.com/c/0Quwuvsa/777-selecting-a-keyboard-layout-may-make-it-unfeasible-to-type-network-settings)

PR #305, opened 2018-03-28
Abbreviated and updated MAAS wording

degville:


evilnick:
@kwm14 - it was easier to edit and runthrough here so we could see it :smiley:

mwhudson:
thanks!

PR #306, opened 2018-04-05
add subiquity information to pollinate user agent

mwhudson:


subiquity/models/subiquity.py
```diff
@@ -114,6 +114,13 @@ def render(self, target, syslog_identifier):
                 'builtin': 'curtin block-meta custom',
                 },
 
+            'pollinate': {
+                'user_agent': {
+                    'subiquity-version': os.environ.get("SNAP_VERSION", 'dry-run'),
+                    'subiquity-revision': os.environ.get("SNAP_REVISION", 'dry-run'),
```
raharper: Would it be possible to combine this into a single entry?

subiquity: "%s_%s" % (snap_version, snap_revisions) 

That way we only have one subiquity entry in the pollinate user_agent string?

mwhudson: Sure, done.

PR #312, opened 2018-04-13
explain that you will not be able to go back once the installation has started

mwhudson:


mwhudson:
[the fact that you can't go back from the identity screen is confusing](https://trello.com/c/MimEQFRm/859-the-fact-that-you-cant-go-back-from-the-identity-screen-is-confusing)

PR #313, opened 2018-04-13
ask cloud-init to configure the locale of the installed system

mwhudson:
This almost seems too easy...


mwhudson:
[should configure locale of installed system](https://trello.com/c/VkU0brzd/861-should-configure-locale-of-installed-system)

vorlonofportland:
Well, eventually we will probably have to deal with languages that don't use UTF-8 encoding by default / standard / convention.  But until then this LGTM.

PR #314, opened 2018-04-16
set wipe: superblock on disks

mwhudson:
installing to a disk that has previously been configured with lvm
fails without this


MathiasMagnus:
When will this PR hit daily builds to test if it indeed solved both related Launchpad bugs?

PR #315, opened 2018-04-17
change scrollbar color if focused

mwhudson:
https://asciinema.org/a/9JRs04vMkpI4ce2f2zpcwhqAV


mwhudson:
I think this or something like this is a good idea, I don't know if we could/should/have time to get feedback from design at this stage.

PR #319, opened 2018-04-20
Offline MAAS installer

xnox:
Compatible with latest livecd-rootfs changes.
Without any install option rename, this will be submitted separately.


subiquity/models/installpath.py
```diff
@@ -40,25 +41,48 @@ def paths(self):
...
+            self.source = '/media/region'
+            self.curtin['debconf_selections'] = {
                 'maas-username': 'maas-region-controller maas/username string %s' % results['username'],
                 'maas-password': 'maas-region-controller maas/password password %s' % results['password'],
-                }
+            }
+            self.curtin['late_commands'] = {
+                '900-maas': ['rm', '-f', '/target/etc/maas/rackd.conf'],
+                '901-maas': ['rm', '-f', '/target/etc/maas/region.conf'],
+                '902-maas': ['curtin', 'in-target', '--', 'maas-rack', 'config', '--init'],
```
vorlonofportland: please add some code comments here to explain why these commands are in the order they are.  E.g. why do you init the config before you reconfigure the package?  Shouldn't we expect dpkg-reconfigure to take care of everything and make the separate 'maas-rack' invocation unnecessary?  Why are two of these commands run through curtin, but others done manually with chroot /target?

subiquity/models/installpath.py
```diff
@@ -40,25 +41,48 @@ def paths(self):
...
                 'maas-secret': 'maas-rack-controller maas-rack-controller/shared-secret password %s' % results['secret'],
-                }
+            }
+            self.curtin['late_commands'] = {
+                '90-maas': ['rm', '-f', '/target/etc/maas/rackd.conf'],
+                '91-maas': ['curtin', 'in-target', '--', 'maas-rack', 'config', '--init'],
+                # maas-rack-controller is broken, and does db_input & go on the password question in the postinst...
+                # when it should have been done in .config
+                # and it doesn't gracefully handle the case of db_go returning 30 skipped
+                '93-maas': ['curtin', 'in-target', '--', 'sh', '-c', 'debconf -fnoninteractive -omaas-rack-controller /var/lib/dpkg/info/maas-rack-controller.postinst configure || :'],
```
vorlonofportland: Can these be factored out to avoid repetition of the same subset of commands for both the rack and region cases?  Certainly it worries me that we only '|| :' one of the two invocations of /var/lib/dpkg/info/maas-rack-controller.postinst configure.

subiquity/models/installpath.py
xnox: `curtin in-target` does a lot of things, specifically, it does mount/unmount `/dev` and `/log`, and it unshares a pid namespace... which would break talking to the postgresql database. Hence only some commands are run via curtin in-target....

`--init` is not run on reconfigure, only on configure. And it initializes uuid, which i'm told is actually unused anymore... but little did I know... hence this matches everything what a udeb/deb based install will do.

subiquity/models/installpath.py
xnox: The rack maintainer scripts, take a different a code path, depending on whether or not the region is installed. Thus in *this* case the maintainer scripts failed and shouldn't. This is to be fixed in their maintainer scripts to handle skipped debconf questions correctly.

xnox:
In short, yes, this is crappy, and once maintainer scripts are improved, I do agree that it should only ever be a simple dpkg-reconfigure call.

mwhudson:
I tested this, realized that the validation for the shared secret was lacking, improved that, tested again and merged.

PR #320, opened 2018-04-20
Maas rename

xnox:


xnox:
After this is merged, we should $ make i18n, and import the updated .pot into launchpad.

PR #321, opened 2018-04-27
move probert into subiquity tree

mwhudson:
I want to make some changes to probert and it's a pain to coordinate them with changes to subiquity, especially now we might want to be able to build revisions of subiquity based on the version that went out with bionic. We could switch to specifying specific commits in subiquity's snapcraft.yaml I suppose, but this just seems easier to me.


xnox:
Was this copy & pasted? or merged with git history using sub-tree merge strategy? it would be nice if `git log -p` would continue to work on the probert code and list useful history/logs/commit messages for it.

mwhudson:
Actually I think I might drop this and just change subiquity's snapcraft.yaml to refer to a commit.

PR #325, opened 2018-05-01
Missing french translation?

Nornort:
All translations are include in subiquity/models/locale.py excepting french.


mwhudson:
I think the French translation was quite incomplete until recently wasn't it? But thanks!

PR #329, opened 2018-05-15
configure proxy for apt and snapd in installed system 

mwhudson:
also emit a signal when proxy config is set

### DIFF
--- a/subiquity/controllers/proxy.py
+++ b/subiquity/controllers/proxy.py
@@ -14,9 +14,9 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import logging
+import os
 
 from subiquitycore.controller import BaseController
-from subiquitycore import utils
 
 from subiquity.ui.views.proxy import ProxyView
 
@@ -43,5 +43,6 @@ class ProxyController(BaseController):
 
     def done(self, proxy):
         self.model.proxy = proxy
+        os.environ['http_proxy'] = os.environ['https_proxy'] = proxy
         self.signal.emit_signal('network-proxy-set')
         self.signal.emit_signal('next-screen')
--- a/subiquity/controllers/proxy.py
+++ b/subiquity/controllers/proxy.py
@@ -43,11 +43,5 @@ class ProxyController(BaseController):
 
     def done(self, proxy):
         self.model.proxy = proxy
-        if proxy:
-            new_env_path = '/etc/environment'
-            if self.opts.dry_run:
-                new_env_path = '.subiquity' + new_env_path
-            with open(new_env_path, 'w') as fp:
-                fp.write(self.model.etc_environment_content())
-            self.signal.emit_signal('network-proxy-set')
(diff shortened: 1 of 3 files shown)


smoser:
Attempting to configure an HTTP proxy via environment variables in /etc/environment has several shortcomings that in my experience make it not a good solution.  The first problem with this is just that virtually no system ever wants a global http_proxy.  As a point of demonstration try this:

    $ export http_proxy=some.host:3128  # as this would be read from /etc/environment
    $ python -m BaseHTTPServer 9999 &
    $ wget http://127.0.0.1:9999/

Do you see the problem here?  No proxy is going to possibly proxy all traffic that is generated on this box.  The 127.0.0.1 address is a good example.  You might think `no_proxy` is a solution for this, but it is not.  Here is a (probably non-comprehensive) list of issues with the general approach:

- support for wildcard or other groups in `http_proxy`, `no_proxy` vary heavily by software package.  GNU tools CIDR based groups `http_proxy=127.0.0.1/8,10.0.0.0/32`, which is great.  However, python standard tools do not support this, so you're left with something like `http_proxy=127.0.0.1,127.0.0.2,10.0.0.1,10.0.0.2....`.  That is just ipv4, think if the fun you'll have putting ipv6 addresses in there!
- For a "server" installation knowing the complete 'no_proxy' set at installation time is not easy.  Try installing lxc or libvirt and configuring a guest network.
- updating environment variables in all programs when you realize you need to add something to /etc/environment requires restarting those programs and maybe even rebooting.
- Modification of /etc/environment requires root access, causing non-root users to do some fiddling/appending/re-setting those variables in their own environment.
- systemd units do not necessarily read /etc/environment (this may be bad memory, but I know that I've fought that before).

My experience is really just that doing this works at first, and then you wish you hadn't, or that at very least you realize that configuring this "globally" didnt really help you out.

The two solutions I've come up with that address many of the above points are:

1. proxy everything to a locally running proxy, set http_proxy=127.0.0.1:3128 and then configure the forwarding rules for that proxy.  Then once you make that change, existing programs start working immediately. [sstack-proxy](https://bazaar.launchpad.net/~smoser/+junk/sstack-proxy/view/head:/sstack-proxy) is a script I use to internally attempt to set proxy globally.
2. add a 'proxy' program that sets the proxy when needed.  Then the user does: proxy wget http://ubuntu.com/ when they want the proxy but otherwise it is not set.

You can certainly do this and it may even work well for some set of users, but my experience is that soon enough you'll find that it isn't as simple as you'd like.  Linux really does need a solution for this.

mwhudson:
Hi, thanks for the comments. Some of them would be better placed in https://bugs.launchpad.net/subiquity/+bug/1766971 perhaps, which might be a better place to decide what to do. My motivation for doing this was that snapd specifically will need to know about a proxy to be able to fetch info about or install snaps and snapd's unit does read /etc/environment. I can do that via a systemd drop-in file of course, so maybe I should do that instead for now. Does that seem reasonable to you?

PR #330, opened 2018-05-17
install_deps target without confirmation

FooBarQuaxx:


mwhudson:
Seems reasonable, thanks!

PR #337, opened 2018-05-21
boot installed system in container to complete configuration

mwhudson:
Not that much code in the end! I've tried to make sure that something at least a bit comprehensible happens if any of the lxc invocations fail.

### DIFF
--- a/subiquity/controllers/installprogress.py
+++ b/subiquity/controllers/installprogress.py
@@ -41,10 +41,11 @@ class InstallState:
     ERROR = -1
 
 
+
 raw_lxc_config = '''\
 lxc.hook.version = 1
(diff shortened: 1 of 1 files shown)


mwhudson:
[create container around /target and pivot to it for instance initialization](https://trello.com/c/bzkDaRS7/951-create-container-around-target-and-pivot-to-it-for-instance-initialization)

raharper:
I generally don't want to pre-boot the system before we actually boot it.  For snap installs, let's defer the cloud-init which supports snap install:

http://cloudinit.readthedocs.io/en/latest/topics/modules.html#snap

mwhudson:
Lets drop this for now then.

PR #338, opened 2018-05-21
Fix/subprocess encoding

raharper:
This fixes running on Xenial and still works on Bionic


mwhudson:
Thanks. This doesn't cover start_command but I think I see how to fix that too, will do it in a followup.

raharper:
It does cover start_command, we drop the encoding parameter and then fix in
the callers, notable the Popen.communicate in the bg_ssh_keys import.

PR #339, opened 2018-05-22
offer featured snaps for installation 

mwhudson:
https://asciinema.org/a/R5FujJOA0p07RJOkawGRAQzA7


mwhudson:
This just shows developer stats for now, will switch to a server section when that appears.

bin/subiquity-tui
```diff
@@ -71,6 +71,7 @@ def parse_options(argv):
     parser.add_argument('--script', metavar="SCRIPT", action='append', dest='scripts', default=[], help='Execute SCRIPT in a namespace containing view helpers and "ui"')
     parser.add_argument('--click', metavar="PAT", action=ClickAction, help='Synthesize a click on a button matching PAT')
     parser.add_argument('--answers')
+    parser.add_argument('--snaps-from-examples', action='store_true', help="Load snap details from sample data, not store")
```
raharper: Is this loading a directory of files? or is there a single index file?

--snaplist  matches closer to the --machine argument where we feed an input file.

examples/snaps/find-output.json
```diff
@@ -0,0 +1,659 @@
+{
```
raharper: How is this generated?  The help section for the --snaps-from-examples could mention that this loads output from 'snap find output' or whatever is needed to fetch this data in this format.

subiquity/controllers/installprogress.py
```diff
@@ -40,10 +49,109 @@ class InstallState:
...
+
+    def __init__(self, controller, download_dir, snap_name, channel):
+        self.controller = controller
+        self.download_dir = os.path.join(download_dir, snap_name)
+        self.snap_name = snap_name
+        self.channel = channel
+
+    def start(self):
+        self.controller._install_event_start(_("downloading {}").format(self.snap_name))
+        os.mkdir(self.download_dir)
```
raharper: This may not be sufficient if download_dir or snap_name includes path separators.    os.makedirs()

subiquity/controllers/installprogress.py
```diff
@@ -40,10 +49,109 @@ class InstallState:
...
+    def start(self):
+        self.controller._install_event_start(_("updating snap seed"))
+
+    def _bg_run(self):
+        # This doesn't really need to be in the background, but as we
+        # have the infrastructure already in place, we may as well.
+        with open(self._seed_yaml) as fp:
+            seed = yaml.safe_load(fp)
+
+        for snap_name in os.listdir(self._tmp_dir):
```
raharper: Is this the "download_dir" from the download task?

subiquity/controllers/installprogress.py
```diff
@@ -40,10 +49,109 @@ class InstallState:
...
+            [snap_path] = glob.glob(os.path.join(self._tmp_dir, snap_name, "*.snap"))
+            [assertion_path] = glob.glob(os.path.join(self._tmp_dir, snap_name, "*.assert"))
+
+            snap_file = os.path.basename(snap_path)
+            assertion_file = os.path.basename(assertion_path)
+            os.rename(snap_path, os.path.join(self._snap_dir, snap_file))
+            os.rename(assertion_path, os.path.join(self._assertions_dir, assertion_file))
+
+            os.rmdir(os.path.join(self._tmp_dir, snap_name))
```
raharper: shutil.move() handles the rename and cleanup.

subiquity/controllers/installprogress.py
```diff
@@ -40,10 +49,109 @@ class InstallState:
...
+            seedinfo = {
+                'name': snap_name,
+                'file': snap_file,
+                'channel': selection.channel,
+                }
+            if selection.is_classic:
+                seedinfo['classic'] = True
+            seed['snaps'].append(seedinfo)
+
+        os.rmdir(self._tmp_dir)
```
raharper: shutil.rmtree() handle subdirs recursively

subiquity/controllers/snaplist.py
```diff
@@ -0,0 +1,236 @@
...
+            log.debug("fetched list of %s snaps", len(self.pending_info_snaps))
+            self._fetch_next_info()
+        self.ongoing[None] = [cb]
+        self.run_in_bg(self._bg_fetch_list, self._fetched_list)
+
+    def stop(self):
+        self.state = SnapInfoLoaderState.STOPPED
+
+    def _bg_fetch_list(self):
+        return self.session.get(self.url_base + 'section=developers', timeout=60)
```
raharper: This seems to bury some configuration, should this be a parameter to the class or a property so callers can pick a different section?

subiquity/controllers/snaplist.py
```diff
@@ -0,0 +1,236 @@
...
+        except requests.exceptions.RequestException:
+            log.exception("loading list of snaps failed")
+            self.state = SnapInfoLoaderState.FAILED
+        else:
+            self.model.load_find_data(response.json())
+        for cb in self.ongoing.pop(None):
+            cb(self.model.get_snap_list())
+
+    def get_snap_list(self, callback):
+        if self.state >= SnapInfoLoaderState.LOADING_INFO:
```
raharper: self.state in [SnapInfoLoaderState.LOADING_INFO, SnapInfoLoaderState.LOADED_INFO]  
This way if someone adds a new state after loaded where it isn't OK to call get_snap_list() it won't break.

subiquity/controllers/snaplist.py
```diff
@@ -0,0 +1,236 @@
...
+        self._maybe_start_new_loader()
+        self.answers = self.all_answers.get('SnapList', {})
+
+    def _maybe_start_new_loader(self):
+        if self.loader:
+            if self.loader.state >= SnapInfoLoaderState.LOADING_INFO:
+                return
+            else:
+                self.loader.stop()
+        if self.opts.snaps_from_examples:
```
raharper: Oh, it's a boolean.  A path here would be more flexible.

subiquity/controllers/snaplist.py
```diff
@@ -0,0 +1,236 @@
...
+            self.loader = SampleDataSnapInfoLoader(
+                self.model,
+                os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), "examples", "snaps"))
+        else:
+            self.loader = SnapdSnapInfoLoader(self.model, self.run_in_bg, '/run/snapd.socket')
```
raharper: the snap socket could be a class attribute or module value.

subiquity/models/snaplist.py
```diff
@@ -0,0 +1,95 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+
```
raharper: whitespace

subiquity/models/snaplist.py
```diff
@@ -0,0 +1,95 @@
...
+@attr.s(cmp=False)
+class SnapSelection:
+    channel = attr.ib()
+    is_classic = attr.ib()
+
+
+class SnapListModel:
+    """The overall model for subiquity."""
+
+    def __init__(self, common):
```
raharper: We take common, shouldn't we hang on to it in the class?

subiquity/models/snaplist.py
```diff
@@ -0,0 +1,95 @@
...
+            self._snaps_by_name[s['name']] = snap
+
+    def load_info_data(self, data):
+        info = data['result'][0]
+        snap = self._snaps_by_name.get(info['name'], None)
+        if snap is None:
+            return
+        channel_map = info['channels']
+        for track in info['tracks']:
+            for risk in ["stable", "candidate", "beta", "edge"]:
```
raharper: Should be in module define

subiquity/ui/views/snaplist.py
```diff
@@ -0,0 +1,367 @@
...
+
+class StarRadioButton(RadioButton):
+    states = {
+        True: SelectableIcon(" *"),
+        False: SelectableIcon("  "),
+        }
+    reserve_columns = 3
+
+
+class NoTabCyclingListBox(ListBox):
```
raharper: Could this go in subiquitycore/ui/ ?  container or lists ?

subiquity/ui/views/snaplist.py
```diff
@@ -0,0 +1,367 @@
...
+    _sizing = frozenset([BOX])
+
+    def __init__(self, parent, snap, cur_channel):
+        self.parent = parent
+        self.snap = snap
+        self.channels = []
+        self.needs_focus = True
+
+        channel_width = max(len(csi.channel_name) for csi in snap.channels) \
+          + StarRadioButton.reserve_columns + 1
```
raharper: Generally you can wrap a new line by putting the expression in parens:

(max() for csi in snap + 
   Button.cols +1)

examples/snaps/find-output.json
mwhudson: I added a examples/snaps/README.md file and mentioned that file in the help.

subiquity/controllers/installprogress.py
mwhudson: Ah so this code makes some assumptions about the directory layout we're working in and I want it to fail if the assumptions fail. I should explain all this a bunch more though... added a long comment

subiquity/controllers/installprogress.py
mwhudson: I've tried to rename some variables for consistency

subiquity/controllers/installprogress.py
mwhudson: I don't think it makes things any clearer here?

subiquity/controllers/installprogress.py
mwhudson: Yes but see above about code making assumptions. I've added a comment.

subiquity/controllers/snaplist.py
mwhudson: Yeah, fair point.

subiquity/controllers/snaplist.py
mwhudson: Yes, but I don't feel super pressed to make this flexible, tbh, it was mostly just something I hacked up so I could work a bit on the UI before I wrote the code to talk to snapd.

subiquity/models/snaplist.py
mwhudson: I tend to follow an ancient convention from launchpad where stdlib imports come first and are separated from third party imports by a blank.

subiquity/models/snaplist.py
mwhudson: Ah it doesn't need to take this parameter any more.

subiquity/models/snaplist.py
mwhudson: If you like.

subiquity/ui/views/snaplist.py
mwhudson: yeah that is usually a better idea, fixed.

subiquity/controllers/snaplist.py
mwhudson: I replaced the state enum with a bunch of booleans that encode the answers to the questions we actually want to know. Better?

subiquity/controllers/installprogress.py
raharper: Really? I trust shutil.move() to do this atomically using context managers. 
Don't we also get to skip the path joining?

shutil.move("path to downloaded snap", "target_dir")
shutil.move("path to downloaded assertion", "target_dir")

subiquity/controllers/snaplist.py
raharper: Sure, but this is about where we store hard coded stuff.  Generally we want to see those at the module or class level, rather than buried in a function.

subiquity/controllers/installprogress.py
```diff
@@ -40,10 +49,168 @@ class InstallState:
...
+    #           [snaps from base seed]
+    #         assertions/
+    #           [assertions from base seed]
+    #         tmp/
+    #           $SNAP1/
+    #             $SNAP1_$REV1.snap
+    #             $SNAP1_$REV1.assert
+    #           $SNAP2/
+    #             $SNAP2_$REV2.snap
+    #             $SNAP2_$REV2.assert
```
raharper: Would it be better to have a tempdir per task?  Rather than share /var/lib/snapd/seed/tmp and determine when it is OK to remove the dir, each "download" task would create one, and once it's relocated the snap/assert files, then clean itself up.?

One thing to note is that /tmp is not safe due to systemd tmpfile removal, we have a wrapper/context manager for using /run/<project name>/tmp  which would be safe from systemd.

https://git.launchpad.net/cloud-init/tree/cloudinit/temp_utils.py#n83

subiquity/controllers/installprogress.py
```diff
@@ -40,10 +49,168 @@ class InstallState:
...
+            [snap_path] = glob.glob(os.path.join(this_snap_download_dir, "*.snap"))
+            [assertion_path] = glob.glob(os.path.join(this_snap_download_dir, "*.assert"))
+
+            snap_file = os.path.basename(snap_path)
+            assertion_file = os.path.basename(assertion_path)
+            os.rename(snap_path, os.path.join(self.snap_dir, snap_file))
+            os.rename(assertion_path, os.path.join(self.assertions_dir, assertion_file))
+
+            # If this directory is not empty, something very
+            # unexpected has happened and we should fail.
```
raharper: Shouldn't we assert that instead?  Do we want a single snap download to be fatal to the entire install?

subiquity/controllers/installprogress.py
```diff
@@ -40,10 +49,168 @@ class InstallState:
...
+                'file': snap_file,
+                'channel': selection.channel,
+                }
+            if selection.is_classic:
+                seedinfo['classic'] = True
+            seed['snaps'].append(seedinfo)
+
+        # If the above loop did not empty this directory, again
+        # something very unexpected has happened and we should fail.
+        os.rmdir(self.tmp_dir)
```
raharper: If we're not fatal but record this, we then have a chance to present the failure to the user to try to download those snaps again.

subiquity/controllers/installprogress.py
mwhudson: Clearly I don't understand what you are suggesting here.

subiquity/controllers/installprogress.py
mwhudson: We can't store the snaps in /tmp or /run/tmp even temporarily because that risks running the live session out of memory (snaps can be big!).

There are clearly other ways we could skin this cat, we could update the seed after each snap download, we could store the snap file name in some data structure rather than in the filesystem or some other things I haven't thought of yet, but I think this is OK? If snap download starts downloading something other than the snap and the assertion, we're definitely going to have to change this code.

subiquity/controllers/installprogress.py
mwhudson: I'd rather fail than have mysterious things happen I think. OTOH, your question about "Do we want a single snap download to be fatal to the entire install?" is fair enough, we should probably retry at least once and let the user request more retries (more complicated control flow incoming :/)

subiquity/controllers/installprogress.py
mwhudson: That's not really what would cause this to fail though. I agree as above that we should allow retrying of failed downloads.

xnox:
The recommendations appear to be graphical games.... and Ubuntu Server default experience is CLI based. Can we recommend CLI apps instead?

vorlonofportland:
On Wed, May 30, 2018 at 09:46:50AM -0700, Dimitri John Ledkov wrote:
> The recommendations appear to be graphical games....  and Ubuntu Server
> default experience is CLI based.  Can we recommend CLI apps instead?

The launch of the server section mentioned 'snap find --section=server'.  Is
this what the subiquity implementation should invoke?

mwhudson:
Ah, the server section exists now? It didn't when I started this branch :) That's clearly what we should show, I'll update to it (and regenerate the example data).

mwhudson:
So I've switched to using cloud-init to install the snaps. It seems to work. It doesn't block boot but it does result in plenty of spam to the console as the snaps are installed. I've pushed my more complex snap-fetching code to another branch in case we decide it's worth the complexity later.

Inclined to merge this soon unless someone objects!

subiquity/controllers/snaplist.py
```diff
@@ -36,6 +36,9 @@ class SampleDataSnapInfoLoader:
     def __init__(self, model, snap_data_dir):
         self.model = model
         self.snap_data_dir = snap_data_dir
+        self.snap_list_fetched = False
+        self.all_snap_info_fetched = False
```
raharper: These two booleans seem to be set in pairs, do we need both?

subiquity/ui/views/snaplist.py
```diff
@@ -102,48 +103,39 @@ def __init__(self, parent, snap, cur_channel):
...
             self.channels.append(Color.menu_button(Columns([
                 (channel_width, btn),
-                (max_version, Text(csi.version)),
-                (max_revision, Text("({})".format(csi.revision))),
-                (max_size, Text(humanize_size(csi.size))),
-                ('pack', Text(notes)),
+                (max_version,   Text(csi.version)),
+                (max_revision,  Text("({})".format(csi.revision))),
+                (max_size,      Text(humanize_size(csi.size))),
+                ('pack',        Text(notes)),
```
raharper: This likely breaks under lint checks, right, the extra spaces after comma?

subiquity/controllers/snaplist.py
mwhudson: Well, there is one place where they're not set together but nothing actually reads all_snap_info_fetched so that one can indeed go.

subiquity/ui/views/snaplist.py
mwhudson: "All checks have passed"

PR #340, opened 2018-05-22
Drop commented out / dead code

raharper:


subiquity/models/locale.py
```diff
@@ -36,7 +36,6 @@ class LocaleModel(object):
         ('fr_FR', 'French'),
         ('de_DE', 'German'),
         ('el_GR', 'Greek, Modern (1453-)'),
-#        ('he_IL', 'Hebrew'), # disabled as it does not render correctly on a vt with default font
```
mwhudson: I want to keep this one for when we start using bterm or something.

raharper: That's fine, we can add a 'noqa:  after the second hash before your disabled comment.

mwhudson: Do you want to do that and then merge this branch then?

raharper:
Updated and force pushed to this branch, after CI says ok, I can merge

PR #343, opened 2018-05-24
Add pep8 and pyflake3 targets and apply all of the fixes

raharper:
Massive pep8 and pyflake3 compliance overhaul.

### DIFF
--- a/scripts/run-pyflakes3
+++ b/scripts/run-pyflakes3
@@ -8,9 +8,19 @@ pycheck_dirs=(
     "tests/"
 )
 
+bin_files=(
+    "bin/console-conf-tui"
+    "bin/subiquity-tui"
+)
+CR="
+"
+[ "$1" = "-v" ] && { verbose="$1"; shift; } || verbose=""
+
 set -f
-if [ $# -eq 0 ]; then
-   files=( "${pycheck_dirs[@]}" )
+if [ $# -eq 0 ]; then unset IFS
+   IFS="$CR"
+   files=( "${bin_files[@]}" "${pycheck_dirs[@]}" )
+   unset IFS
 else
    files=( "$@" )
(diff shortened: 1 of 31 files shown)


scripts/run-pyflakes3
```diff
@@ -0,0 +1,34 @@
...
+   files=( "${bin_files[@]}" "${pycheck_dirs[@]}" )
+   unset IFS
+else
+   files=( "$@" )
+fi
+
+cmd=( "python3" -m "pyflakes" "${files[@]}" )
+
+echo "Running: " "${cmd[@]}" 1>&2
+export PYFLAKES_BUILTINS="_"
```
mwhudson: Ah hah! Thanks for this bit in particular :)

mwhudson:
Wow!
I guess we should not let this sit for very long as it will conflict with absolutely anything (including Scott's Makefile change I just merged).
I have two comments really:
 1. Can we please disable the line length thing? Wrapping at 80 chars in 2018 just seems crazy.
 2. We should run these in travis, perhaps as a separate job? Can't remember how to do that though.

scripts/run-pyflakes3
raharper: Indeed!

raharper:
For (1) I'm generally a fan of the 80 char limit;  If we do away with it, what's the next reasonable limit and why? 
For (2); Yeah, we usually have a separate job for linting.  I've not setup travis jobs, so we can poke around and figure that bit out.

mwhudson:
OK, I'll try to stick to 80 columns for a week or two and see how much I hate it in practice!
For travis we can figure it out together. Can you fix conflicts and merge this asap?

raharper:
Yes, I'll fix and push.

PR #344, opened 2018-05-24
update travis jobs: test on cosmic, run make lint

mwhudson:
There would be something to be said for running the linting in a different "stage" I think but my limited fiddling couldn't get that to work, so I think this will do for now.

### DIFF
--- a/scripts/installdeps.sh
+++ b/scripts/installdeps.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -eux
+apt-get update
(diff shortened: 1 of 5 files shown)


scripts/installdeps.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -eux
+apt-get update
+apt-get -y dist-upgrade
```
raharper: With the -daily release we don't need to dist-upgrade;

PR #346, opened 2018-05-25
Implement zfsroot support

raharper:
Curtin supports a ZFS on Root mode where users can configure their rootfs
as a ZFS filesystem.  This is experimental and restricts use of ZFS only
to the rootfs volume.  This patch adds 'zfsroot' to the supported filesystem
list and enforces the requirements of only one zfsroot and only mounted at
'/'.  Add some code to blank out zfsroot if one is already selected.

This patch also adds a guided answer file for zfsroot.  One small change
to the guided path allowing the answers to supply the filesystem value for
the rootfs.

### DIFF
--- a/examples/zfsroot-answers.yaml
+++ b/examples/zfsroot-answers.yaml
@@ -1,0 +1,24 @@
+Welcome:
+  lang: en_US
+Keyboard:
+  layout: us
+Installpath:
+  path: ubuntu
+Network:
+  accept-default: yes
+Proxy:
+  proxy: ""
+Filesystem:
+  guided: yes
+  guided-index: 0
+  guided-fstype: zfsroot
+Identity:
+  realname: Ubuntu
+  username: ubuntu
+  hostname: ubuntu-server
+  # ubuntu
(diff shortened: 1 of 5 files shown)


mwhudson:
Interesting but do we want to do so something so special cased vs implementing proper ZFS support? I'm hoping (possibly optimistically) that doing it properly won't be incredibly hard after we add RAID support.

I agree with your comments about the way the fstype objects are handled being a bit awkward.

xnox:
having native simplistic zfsroot is nice; cause that's the hardest bit to install by hand. Adding zpools, zfs after that is relatively easy.

zfsroot is similar to e.g. `partman-auto-lvm` in d-i speak, and we do offer that as a one-click/one-tick option in both d-i & ubiquity installers, and it is popular. With "advanced" lvm setup rarely used, and often only tweaked post-install.

xnox:
Hm, i think it would be nice to merge this. But there are conflicts now. And we do have RAID ui now, but it is for the linux md-raid only at the moment. Not sure how reusable that would be to do btrfs/zfs raid UI.

mwhudson:

PR #347, opened 2018-06-01
Add unittests to subiquity.models.filesystem

raharper:
Add about 99% coverage on the filesystem module.  While adding unittest
address a few bugs in the implementation and some of the pep8/pyflakes3
refactoring.

- Add helpers to add CiTestCase which has add_patch method for
  easy mocking
- Add a mocking contextmanager for mocking opens
- Pull _get_system_mounted_disks out of FilesystemModel class
  making it easier to mock
- Refactored add_filesystem to correctly prevent adding filesystems
  to special partitions like bios_grub and boot/efi
- Marked several id elements in the Disk, Partition and other classes
  as not part of the equality statement so we can do equality checks
  on objects that are the same except for uniq-id elements.
- Updated Makefile unit target to run nose with -v and options
- Fixed bug in Disk.reset() where the default didn't match what got set in reset()

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ PYTHONSRC=$(NAME)
 PYTHONPATH=$(shell pwd):$(shell pwd)/probert
 PROBERTDIR=./probert
 PROBERT_REPO=https://github.com/CanonicalLtd/probert
+noseopts ?= -v
 export PYTHONPATH
 CWD := $(shell pwd)
 
@@ -46,9 +47,9 @@ flake8:
 
 unit:
 	echo "Running unit tests..."
-	$(PYTHON) -m nose $(CHECK_DIRS)
+	$(PYTHON) -m nose $(noseopts) $(CHECK_DIRS)
 
-check: unit
+check: lint unit
 
 probert:
 	@if [ ! -d "$(PROBERTDIR)" ]; then \
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ flake8:
 	$(PYTHON) -m flake8 $(CHECK_DIRS)
 
 unit:
-	echo "Running unit tests..."
+	@echo 'tox -e py3' is preferred to 'make unit'
 	$(PYTHON) -m nose $(noseopts) $(CHECK_DIRS)
 
 check: lint unit
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -118,7 +118,7 @@ def asdict(inst):
 @attr.s
 class Disk:
 
-    id = attr.ib(default=id_factory("disk"))
+    id = attr.ib(default=id_factory("disk"), cmp=False)
     type = attr.ib(default="disk")
     ptable = attr.ib(default='gpt')
     serial = attr.ib(default=None)
@@ -140,11 +140,11 @@ class Disk:
     def fs(self):
         return self._fs
 
-    _info = attr.ib(default=None)
+    _info = attr.ib(default=None, cmp=False)
 
     @classmethod
-    def from_info(self, info):
-        d = Disk(info=info)
+    def from_info(cls, info):
+        d = cls(info=info)
         d.serial = info.serial
         d.path = info.name
         d.model = info.model
@@ -152,8 +152,8 @@ class Disk:
 
     def reset(self):
         self.preserve = False
-        self.name = ''
-        self.grub_device = ''
+        self.name = ""
+        self.grub_device = False
         self._partitions = []
         self._fs = None
 
@@ -192,7 +192,7 @@ class Disk:
 @attr.s
 class Partition:
 
-    id = attr.ib(default=id_factory("part"))
+    id = attr.ib(default=id_factory("part"), cmp=False)
     type = attr.ib(default="partition")
     device = attr.ib(default=None)  # Disk
     size = attr.ib(default=None)
@@ -262,6 +262,27 @@ def align_down(size, block_size=1 << 20):
     return size & ~(block_size - 1)
 
 
+def _get_system_mounted_disks():
+    # FIXME: switch to curtin.block.get_installable_blockdevs()
+    # This assumes a fairly vanilla setup. It won't list as
+    # mounted a disk that is only mounted via lvm, for example.
+    mounted_devs = []
+    with open('/proc/mounts', encoding=sys.getfilesystemencoding()) as pm:
+        # mocking the filehandle iterator is hard, so invoke read()
+        for line in pm.read().splitlines():
+            if line.startswith('/dev/'):
+                mounted_devs.append(line.split()[0][5:])
+    mounted_disks = set()
+    for dev in mounted_devs:
+        if os.path.exists('/sys/block/{}'.format(dev)):
+            mounted_disks.add('/dev/' + dev)
+        else:
+            paths = glob.glob('/sys/block/*/{}/partition'.format(dev))
+            if len(paths) == 1:
+                mounted_disks.add('/dev/' + paths[0].split('/')[3])
+    return mounted_disks
+
+
 class FilesystemModel(object):
 
     lower_size_limit = 128 * (1 << 20)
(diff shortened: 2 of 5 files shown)


subiquity/models/tests/test_filesystem.py
```diff
@@ -78,3 +159,702 @@ def test_errors(self):
...
+    def test_partition_number(self):
+        """partition.number derived from device partition index"""
+        disk = Disk()
+        partition = Partition()
+        partition.device = disk
+        disk._partitions.append(partition)
+        self.assertEqual(1, partition._number)
+
+    # FIXME: partition.path does not account for named partition (nvmen0p1)
+    def test_partition_path(self):
```
mwhudson: I wonder if we actually display the partition's path anywhere. We probably shouldn't.

subiquity/models/filesystem.py
```diff
@@ -118,7 +118,7 @@ def asdict(inst):
 @attr.s
 class Disk:
 
-    id = attr.ib(default=id_factory("disk"))
+    id = attr.ib(default=id_factory("disk"), cmp=False)
```
mwhudson: In some branch or other I actually set cmp=False on the whole class. I guess some tests will need to change if we do that.

subiquity/models/tests/test_filesystem.py
```diff
@@ -78,3 +159,702 @@ def test_errors(self):
...
+        swapfs.id = "%s-%s" % (swappart.id, swapfs.fstype)
+        swap_mount = self.fsm.add_mount(swapfs, "")
+        # id: sdb1-mnt-swap
+        swap_mount.id = "%s-mnt-swap" % swappart.id
+
+        # make sure this is a valid config
+        self.assertTrue(self.fsm.can_install())
+        content = yaml.dump({'storage': {'config': self.fsm.render()}},
+                            default_flow_style=False, indent=4)
+        print(content)
```
mwhudson: You probably want to delete this.

subiquity/models/tests/test_filesystem.py
```diff
@@ -78,3 +159,702 @@ def test_errors(self):
...
+        """verify add_swapfile returns False when swap fs is present"""
+        self.fsm.probe()
+        all_disks = self.fsm.all_disks()
+        disk = random.choice(all_disks)
+        partition = self.fsm.add_partition(disk=disk, size=(1 * MB))
+        self.fsm.add_filesystem(partition, 'swap')
+
+        self.assertFalse(self.fsm.add_swapfile())
+
+    def test_render(self):
```
mwhudson: I think this test is probably baking in too many implementation details. Like at the very least, compare the data structures rather than comparing the serialized yaml?

I think what we care about is that all the disks/partitions/etc objects are represented in the config, and that anything referred to by id is defined before the thing that refers to it. All else seems like detail to me, am I forgetting something?

subiquity/models/tests/test_filesystem.py
```diff
@@ -78,3 +159,702 @@ def test_errors(self):
...
+
+        # make sure this is a valid config
+        self.assertTrue(self.fsm.can_install())
+        content = yaml.dump({'storage': {'config': self.fsm.render()}},
+                            default_flow_style=False, indent=4)
+        print(content)
+        self.assertEqual(EXPECTED_FSM_RENDERED_CONTENT, content)
+
+    def test_render_ensures_swap_volume_is_mounted(self):
+        """render includes swap mont when partition.fs='swap' and nomount"""
```
mwhudson: typo ("mont")

subiquity/models/tests/test_filesystem.py
```diff
@@ -78,3 +159,702 @@ def test_errors(self):
...
+        sda = self.fsm.all_disks()[0]
+        sda.id = os.path.basename(sda.path)
+        swappart = self.fsm.add_partition(disk=sda, size=(2 * MB))
+        swappart.id = os.path.basename(swappart.path)
+        swapfs = self.fsm.add_filesystem(swappart, 'swap')
+        swapfs.id = "%s-%s" % (swappart.id, swapfs.fstype)
+
+        config = self.fsm.render()
+        for item in config:
+            print(item)
```
mwhudson: no :)

subiquity/models/tests/test_filesystem.py
```diff
@@ -78,3 +159,702 @@ def test_errors(self):
...
+        sda.id = os.path.basename(sda.path)
+        swappart = self.fsm.add_partition(disk=sda, size=(2 * MB))
+        swappart.id = os.path.basename(swappart.path)
+        swapfs = self.fsm.add_filesystem(swappart, 'swap')
+        swapfs.id = "%s-%s" % (swappart.id, swapfs.fstype)
+        self.fsm.add_mount(swapfs, "")
+
+        config = self.fsm.render()
+        for item in config:
+            print(item)
```
mwhudson: also no

subiquity/models/tests/test_filesystem.py
```diff
@@ -78,3 +159,702 @@ def test_errors(self):
...
+        swapfs.id = "%s-%s" % (sda.id, swapfs.fstype)
+        config = self.fsm.render()
+        for item in config:
+            print(item)
+        mounts = [entry for entry in config if entry['type'] == 'mount']
+        self.assertEqual(1, len(mounts))
+        swapmount = mounts[0]
+        self.assertEqual(swapfs.id, swapmount['device'])
+
+    def test_render_ensures_swap_disk_has_mount_nomount(self):
```
mwhudson: wait how do all these test cases differ? Can you create some helpers to make them easier to read? Do we create Mounts for swap partitions or not? I think we only need to pick one, right?

subiquity/models/tests/test_filesystem.py
```diff
@@ -78,3 +159,702 @@ def test_errors(self):
...
+
+MB = 1 << 30
+GB = 1 << 40
+
+
+class TestFilesystemDisk(CiTestCase):
+    def setUp(self):
+        super(TestFilesystemDisk, self).setUp()
+        self.info = argparse.Namespace()
+        self.info.serial = 'myserial'
```
mwhudson: In general I prefer to have a test factory that makes strings whose content is immaterial to the test, so stuff like

```
disk.serial = self.factory.makeString()
disk.path = self.factory.makeString()
self.assertEqual(disk.label, disk.serial)
```

rather than hardcoding strings where it can be tricky for the reader to know or not if the contents are important. 

I also dislike setUp() methods in general because of http://xunitpatterns.com/Obscure%20Test.html#Mystery%20Guest sort of things, helper methods to make objects of various kinds are much better IMO.

Makefile
```diff
@@ -45,7 +46,7 @@ pyflakes3:
 
 unit:
 	echo "Running unit tests..."
-	nosetests3 $(PYTHONSRC)/tests
+	nosetests3 $(noseopts) $(PYTHONSRC)
```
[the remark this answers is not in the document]
raharper: I'll refactor to master with whatever tox is running.

subiquity/models/filesystem.py
```diff
@@ -262,6 +262,27 @@ def align_down(size, block_size=1 << 20):
     return size & ~(block_size - 1)
 
 
+def _get_system_mounted_disks():
+    # FIXME: switch to curtin.block.get_installable_blockdevs()
```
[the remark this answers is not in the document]
raharper: Possible, maybe we just need a .is_mounted() sort of tag on the probed storage, then the model or controller can apply the filter of removing any mounted disks from being presented as available.

subiquity/models/filesystem.py
raharper: I don't think we need to cmp=False the whole class; the two changing items are the always incrementing id value; which isn't really needed in the Disk class itself, and the private _info which is a copy of the probe data in StorageInfo form.

subiquity/models/tests/test_filesystem.py
raharper: We can; I find that we'll just add it back in when we try to see what's going on if the unittest fails.  The print is swallowed by the harness unless there is a failure.

subiquity/models/tests/test_filesystem.py
raharper: Will fix.

subiquity/models/tests/test_filesystem.py
raharper: I can remove debug prints;

subiquity/models/tests/test_filesystem.py
raharper: I like helper methods as well;  For complex mocking or setup, I think the helper methods are useful, for simple Object creation, using setUp like we would an __init__() saves function overhead and removes repetition in the unittest itself.

subiquity/models/tests/test_filesystem.py
mwhudson: Right but I think we should (perhaps not in this branch) pick two of those and make sure those work. I was confused implementing this about the need for Mounts for swap partitions. Maybe we should implicitly create one (possibly not in model code) when we create a Swap fs? And we should just get the flag right.

subiquity/models/filesystem.py
mwhudson: We really want identity semantics for these objects I think. Adding cmp=False to the id flies in the face of that. For example, if you have two partitions of the same size on the same disk, they compare equal if you add cmp=False to id! That means if the user selects one at to delete and we exectute model._partitions.remove(part), it will always be the first one that gets removed.

So I'm going to be obstinate here I'm afraid and say we'll have to make the tests more awkward (a helper that call asdict on both things, throws out the ids and compares the dicts?) rather than add these cmp=False flags.

subiquity/models/tests/test_filesystem.py
mwhudson: Well I disagree but not enough to demand changes in this branch...

subiquity/models/filesystem.py
raharper: Clearly two partitions of the same size on the same disk should not compare equal; they have other attributes besides id that are different.  What's missing is _eq_ vs _cmp_;  I believe I was dealing with this before switching to attr ; that is partitions can be equivalent, but not _the_ same object.  It's not clear to me that dropping id from the comparison results in having two partitions of the same size being considered the same object.

I'll dig into this a bit more before we merge the branch.

mwhudson:

PR #357, opened 2018-06-07
modernize filesystem views

mwhudson:
This branch changes the filesystem view to the style of interface we want to use to add RAID support, but doesn't actually add RAID support.

Please kick the tires severely, there are a lot of changes here.


mwhudson:
See https://github.com/CanonicalLtd/subiquity/pull/363 instead.

PR #360, opened 2018-06-08
console-conf identity: do not hard-code the core version

sil2100:
Reuse the login_details_tmpl_no_ip template in standalone, use os-release to determine the Ubuntu Core version as we now support both 16 and 18.

### DIFF
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -40,7 +40,7 @@ def get_core_version():
     version = None
     for line in shlex.split(content):
         key, _, value = line.partition("=")
-        if key == "NAME" and value != "Ubuntu Core":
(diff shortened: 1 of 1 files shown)


mwhudson:
Looks fine, the failure in travis looks silly, let me retry it.

console_conf/controllers/identity.py
```diff
@@ -26,6 +27,28 @@
...
+    try:
+        with open(path, "r") as fp:
+            content = fp.read()
+    except FileNotFoundError:
+        return None
+
+    version = None
+    for line in shlex.split(content):
+        key, _, value = line.partition("=")
+        if key == "NAME" and value != "Ubuntu Core":
```
zyga: Should you look at ID instead of NAME?

PR #363, opened 2018-06-11
modernize filesystem views

mwhudson:
A replacement for https://github.com/CanonicalLtd/subiquity/pull/357, code somewhat the same, but this time reviewing commit-by-commit might make some sense. Tyre kicking still gratefully received.

Video: https://asciinema.org/a/u2J4oyjoU6mTmtloPXi4iRoiS


mwhudson:
[adapt the existing filesystem functionality to RAID-ready views](https://trello.com/c/HdwTX2Wa/13-adapt-the-existing-filesystem-functionality-to-raid-ready-views)

subiquity/controllers/filesystem.py
```diff
@@ -85,59 +77,49 @@ def finish(self):
...
+        if fs is None:
+            return
+        self.delete_mount(fs.mount())
+        self.model.remove_filesystem(fs)
 
-    def edit_partition(self, disk, partition):
-        log.debug("Editing partition {}".format(partition))
-        adp_view = PartitionView(self.model, self, disk, partition)
-        self.ui.set_body(adp_view)
+    def create_partition(self, device, spec, flag=""):
```
raharper: create_filesystem? and partition is implicit or do we always create them together?  What if we create a partition but leave it unformated?

subiquity/models/filesystem.py
```diff
@@ -115,8 +116,91 @@ def asdict(inst):
...
+    def constructed_device(self):
+        return self._constructed_device
+
+
 @attr.s
-class Disk:
+class _Device(_Formattable):
+    # Anything that can have partitions, e.g. a disk or a RAID.
+
+    # subclass must implement .size!
```
raharper: Should this be an Abstract  Base class then?

subiquity/models/filesystem.py
```diff
@@ -150,16 +223,49 @@ def from_info(self, info):
         d.model = info.model
         return d
 
+    def info_for_display(self):
+        bus = self._info.raw.get('ID_BUS', None)
```
raharper: Generally shouldn't the items we fetch from _info.raw be attributes on the info object?  That allows it to be mocked or replaced and if where we fetch the ID_BUS or other values from won't brake here if that changes.

subiquity/models/filesystem.py
```diff
@@ -150,16 +223,49 @@ def from_info(self, info):
...
+        major = self._info.raw.get('MAJOR', None)
+        if bus is None and major == '253':
+            bus = 'virtio'
+
+        devpath = self._info.raw.get('DEVPATH', self.path)
+        # XXX probert should be doing this!!
```
raharper: Yes, I thought udev already had this property ?  No, no it doesn't.  Can this be added to the probe method so we fetch that on creation? it won't change over the life of the object.

subiquity/models/filesystem.py
```diff
@@ -356,9 +482,15 @@ def _use_disk(self, disk):
         if disk.path not in self._disks:
             self._disks[disk.path] = disk
 
+    def all_mounts(self):
+        return self._mounts[:]
```
raharper: I've not seen this notation; what does it do ver returning the object name?

subiquity/ui/views/filesystem/delete.py
```diff
@@ -0,0 +1,52 @@
...
+
+from subiquitycore.ui.buttons import danger_btn, other_btn
+from subiquitycore.ui.utils import button_pile
+from subiquitycore.ui.stretchy import Stretchy
+
+
+log = logging.getLogger('subiquity.ui.filesystem.disk_info')
+
+
+class ConfirmDeleteStretchy(Stretchy):
```
raharper: I saw this in the asciinima; since these are constructed items, not real partitions, do we want to impose the confirmation dialog?

subiquity/ui/views/filesystem/partition.py
```diff
@@ -238,54 +213,99 @@ def __init__(self, model, controller, disk, partition=None):
...
+        self.form.remove_field('size')
 
-        super().__init__(None, volume, {}, back)
+        connect_signal(self.form, 'submit', self.done)
+        connect_signal(self.form, 'cancel', self.cancel)
+
+        rows = []
+        if isinstance(device, Disk):
+            rows = [
+                Text(_("Formatting and mounting a disk directly is unusual. "
```
raharper: This seems like an odd comment/ui warning.

subiquity/controllers/filesystem.py
mwhudson: The spec argument contains fields to determine whether to create the filesystem or not. I'm not super happy about this bit of API design, but it's not actually very different from what is in master already.

subiquity/models/filesystem.py
mwhudson: Yeah, I'll change it. Was confused about how to do abstract properties but on looking it up it's really easy.

subiquity/models/filesystem.py
mwhudson: Yes probably, I just moved this code though, can we clean it up later? :)

subiquity/models/filesystem.py
mwhudson: ditto

subiquity/models/filesystem.py
mwhudson: It means make a copy of the list as protection against the caller mutating the returned value. Probably not necessary but for the avoidance of surprises...

subiquity/models/filesystem.py
raharper: Surely, for items deferred, we either open a bug or add a card to our tech debt backlog so we don't forget.

subiquity/ui/views/filesystem/delete.py
mwhudson: I don't know. I'll talk to design I guess. The ideal thing would probably be to not have the dialog but have a way of undoing a mistaken delete. No idea at all how to implement that though!

I was going to include in the dialog a description about the objects that would be implicitly deleted by the deletion (so in the case of a partition that could be filesystem and mount) but apparently I didn't get to that :) I should probably wait until we've decided if this dialog is a good idea before implementing that...

subiquity/ui/views/filesystem/partition.py
mwhudson: Yeah. Currently the button for this functionality has the label "Format or create swap on entire device (unusual, advanced)" which is also pretty odd :)

It's a tricky one because:

1. It *is* a bit strange to mount a disk without partitioning
2. It is *not* strange to mount a RAID without partitioning
3. I can't think of a way to present the option in a way that makes it clear before choosing it that it's a bit of an odd thing to do.

Something about this reminds me that this branch probably doesn't handle formatting things as swap properly.

PR #364, opened 2018-06-14
only add the user to groups that exist in the target system

mwhudson:
for https://bugs.launchpad.net/subiquity/+bug/1775228
("user is in debian-tor group on fresh install")


subiquity/models/subiquity.py
```diff
@@ -41,12 +42,24 @@ def setup_yaml():
 setup_yaml()
 
 
+def get_all_groups(dry_run):
+    command = ['chroot', '/target', 'getent', 'group']
```
raharper: Shouldn't /target be part of one of the models, maybe the base subiquity.model ?

subiquity/models/subiquity.py
```diff
@@ -61,13 +74,15 @@ def __init__(self, common):
...
-            os.path.join(os.environ.get("SNAP", "/does-not-exist"),
+            os.path.join(os.environ.get("SNAP", "."),
                          "users-and-groups"))
         if os.path.exists(users_and_groups_path):
             groups = open(users_and_groups_path).read().split()
         else:
             groups = ['admin']
         groups.append('sudo')
+        all_groups = get_all_groups(self.opts.dry_run)
+        groups = [group for group in groups if group in all_groups]
```
raharper: You can put the call to get_all_groups in the list comprehension unless there is a need to keep the all_groups around.

subiquity/models/subiquity.py
mwhudson: Yes, done, and cleaned up all the other hardcoded paths while I was at it.

subiquity/models/subiquity.py
mwhudson: Sure.

mwhudson:
Thanks for the review. WIll wait for travis and merge.

PR #365, opened 2018-06-14
filter out disks with the ID_CDROM udev property

mwhudson:


subiquity/models/filesystem.py
```diff
@@ -345,12 +345,16 @@ def probe(self):
...
-                    if data["MAJOR"] != "2" and data['attrs'].get('ro') != "1":
-                        # log.debug('disk={}\n{}'.format(
-                        #    path, json.dumps(data, indent=4, sort_keys=True)))
-                        info = self.prober.get_storage_info(path)
-                        self._available_disks[path] = Disk.from_info(info)
+                if data["DEVPATH"].startswith('/devices/virtual'):
+                    continue
+                if data["MAJOR"] == "2" or data['attrs'].get('ro') == "1":
+                    continue
+                if data.get("ID_CDROM", ""):
```
raharper: I think "ID_CDROM" in data is enough since you don't care about the value.

mwhudson: Now I've seen how it gets set, I agree :)

PR #377, opened 2018-07-05
add some raid/not physical disk checks

mwhudson:
Some attempts to avoid unbootable things, and even trying to give the user some hints what to do to progress.

### DIFF
--- a/examples/answers-raid.yaml
+++ b/examples/answers-raid.yaml
@@ -19,16 +19,14 @@ Filesystem:
       data:
         size: 1G
         fstype: null
-    - obj: [disk index 0]
+    - &newpart
+      obj: [disk index 0]
       action: PARTITION
       data:
         size: 1G
-        mount: null
-    - obj: [disk index 0]
-      action: PARTITION
-      data:
-        size: 1G
-        mount: null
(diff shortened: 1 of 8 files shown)


raharper:
Having /boot on a RAID is supported.

One needs the bios_boot partition allocation for GPT setups, or the /boot/EFI for UEFI mode.

vorlonofportland:
> Having /boot on a RAID is supported.

There was discussion of this in Foundations.  It's true that GRUB can boot from /boot on RAID (including RAID5), but Mathieu has pointed out that GRUB's RAID5 implementation is read-only.  This means that, for example, boot menu fallback on failed boot will not work correctly because grub has no persistent env space that it can write variables to.

I believe it is the case that d-i will let you install with /boot on RAID5, but my guidance has been that it is better to be initially conservative in the configurations we allow, limiting ourselves to combinations we're sure work, rather than being bug-for-bug compatible with d-i.

AIUI the GRUB RAID1 driver is read-write, so /boot on RAID1 is ok.  But also, it's more important to land what we have working, and iterate on that as necessary to relax the rules and include more cases.

PR #380, opened 2018-07-07
Bond

xnox:
Everything is done via device sub-menus.
sub-menus are dynamic.
flow starts with "create a new bond", which creates a bond and enslaves the device.
then submenus appear to set this bond as master on other interfaces.
Enslaved devices have option to remove a master.

For modes that allow specifying primary ("Make primary", "make secondary") there is no UI yet, hence automatic primary selection happens.

One can create vlan on top of a bond.

Deleting bond, correctly drops vlans that are on top of it, and clears the state of the underlying devices.

Prerequisite are my two probert branches that fixup probert.

UI needs to be a more dynamic; e.g. on the bond form we should grey out the two extra things, that are not applicable to some of the bond modes, like is done in the model when choosing which options to generate. But I have not looked into signals to wire that up.

Needs testing that end up generating valid configs that netplan can apply.... I've ironed out obvious bugs, but possibly not quite all of them.

ps. Whilst clicking all the things very quickly, i've managed to crash my network card kernel driver and had to hard reboot, as netlink/network stack got stuck.


mwhudson:
I did it my way </sing-song-voice>

PR #387, opened 2018-08-03
agetty needs an explicit -i to not to display the /etc/issue file

sil2100:
Otherwise we get duplicate information (or wrong information), as by principle console-conf and subiquity handle all login-prompt-related printing. This seemed to be the default in the past? Not sure what changed but being more explicit is always good.


mwhudson:
Thanks, the subiquity files in debian/ are not used and should be removed I guess but better to fix them than not :)

PR #388, opened 2018-09-10
Remove the Joliet option to support symlinks

mojodna:
This allows the generated ISO to be used as a cdrom source for apt and quashes the following warnings:

```
libisofs: WARNING : Cannot add /dists/stable to Joliet tree. Symlinks can only be added to a Rock Ridge tree.
libisofs: WARNING : Cannot add /dists/unstable to Joliet tree. Symlinks can only be added to a Rock Ridge tree.
libisofs: WARNING : Cannot add /ubuntu to Joliet tree. Symlinks can only be added to a Rock Ridge tree.
```


xnox:
By default we enable _both_ joilet and rock-ridge extensions i.e. `-J -r`

Thus symlinks are visible on majority of machines that support rock-ridge extensions.

I don't know what would break if we remove joilet extensions =/

vorlonofportland:
Joliet extensions are used for long filenames on Windows.  We don't want to disable Joliet extensions because of this warning, it won't improve the experience for anyone.

PR #393, opened 2018-11-13
Fix default mirror for ports.

xnox:


xnox:
Tested with a launchpad built snap on s390x, works as expected. Also tested locally on amd64, ditto.

subiquity/models/mirror.py
```diff
@@ -14,10 +14,15 @@
...
 import logging
+import platform
 
 log = logging.getLogger('subiquitycore.models.mirror')
 
-DEFAULT_MIRROR = 'http://archive.ubuntu.com/ubuntu'
+# correct default mirror for most arches
+DEFAULT_MIRROR = 'http://ports.ubuntu.com/ubuntu-ports'
+# apart from the two snowflakes
+if platform.machine() in ['i686', 'x86_64']:
```
mwhudson: this information isn't available anywhere else? there's nothing we can query if ppc64el (say) ends up on archive?

xnox: And more importantly, launchpad has no awareness about these, and it publishes unsplit archive by default, and we post-process mangle it to split the tree, without breaking signatures -> hence e.g. there are parts of the APT metadata that refer to ports arches on archive.u.c and vice versa.

We have moved things to/from ports before. But it was done on per-release basis. And generally we add any new architectures on ports, thus special casing x86 arches is ok. And only needs to be revisited on per-series basis, if and when publications change. The only upcoming one I can think off is dropping i386 from archive.ubuntu.com -> but to be fair, I really don't need to specify i686 here either, cause we don't build subiquity for i386.

mwhudson: That all makes sense. Merge away!

PR #394, opened 2018-11-13
Setup colors on serial vt220

xnox:


subiquitycore/core.py
```diff
@@ -398,7 +398,8 @@ def next(loop, user_data):
 
     def run(self):
         if not hasattr(self, 'loop'):
-            if self.common['opts'].run_on_serial:
+            if (self.common['opts'].run_on_serial and
+                    os.environ["TERM"] != "vt220"):
```
vorlonofportland: Why are you hard-coding vt220 as a whitelist for colors?  I think the reasonable default assumption today is that serial console DOES support colors, and it should only be the serial console on z systems, where the serial connection terminates on a java client that the end user has no control over, should be treated as monochrome.

xnox: At the moment `--serial` is passed to subiquity on consoles that systemd generates / determines to be serial. And all of those end up without colors, on all arches.

You can see a good summary of console names here https://www.ibm.com/support/knowledgecenter/en/linuxonibm/com.ibm.linux.z.ludd/ludd_r_console_assure_nodes.html

Note, all of them are detected as serial, but only the `/dev/ttysclp0` is known to support colors, and that one is a `TERM=vt220`. (also kvm is not listed on that page, but `/dev/ttysclp0` is available there as the only console)

So how should this be done?

Should I add a `--force-color` flag to subiquity and then bake this flag as an override on the `serial-getty@ttysclp0.service`? or like add code in subiquity to specifically check for tty names?

We already have a `is_linux_tty` function, do you think that can be used to detect colorful serial consoles? https://github.com/CanonicalLtd/subiquity/blob/master/subiquitycore/core.py#L73

mwhudson:
On Wed, 14 Nov 2018 at 21:57, Dimitri John Ledkov <notifications@github.com>
wrote:

> *@xnox* commented on this pull request.
> ------------------------------
>
> In subiquitycore/core.py
> <https://github.com/CanonicalLtd/subiquity/pull/394#discussion_r233358802>
> :
>
> > @@ -398,7 +398,8 @@ def next(loop, user_data):
>
>      def run(self):
>          if not hasattr(self, 'loop'):
> -            if self.common['opts'].run_on_serial:
> +            if (self.common['opts'].run_on_serial and
> +                    os.environ["TERM"] != "vt220"):
>
> At the moment --serial is passed to subiquity on consoles that systemd
> generates / determines to be serial. And all of those end up without
> colors, on all arches.
>
> You can see a good summary of console names here
> https://www.ibmsccb_get_sclp_send_mask.com/support/knowledgecenter/en/linuxonibm/com.ibm.linux.z.ludd/ludd_r_console_assure_nodes.html
> <https://www.ibm.com/support/knowledgecenter/en/linuxonibm/com.ibm.linux.z.ludd/ludd_r_console_assure_nodes.html>
>
> Note, all of them are detected as serial, but only the /dev/ttysclp0 is
> known to support colors, and that one is a TERM=vt220. (also kvm is not
> listed on that page, but /dev/ttysclp0 is available there as the only
> console)
>
I don't really understand what it means for a serial device to be a vt220
serial device (AIUI, for the kernel a serial device is basically a pipe
that supports some extra bits like an ioctl to configure a character that
should send SIGINT to a bunch of associated processes and the meaning of
the escape codes is a user space thing) but having grepped the kernel a bit
I can see that it definitely is a s390 concept so whatever.

> So how should this be done?
>
Poking around in systemd, it looks to me as if the default TERM for any
ttypath that doesn't match /dev/tty[0-63] is vt220 -- although I'm not in a
position to confirm this by experiment. If I'm right then this change
effectively neuters the run_on_serial flag entirely and I can think of a
different patch :-) I'm not sure we want to go there though, can we assume
that any vt that any user is using to connect to serial on any device
supports vt220? Maybe, but we should think about this.

> Should I add a --force-color flag to subiquity and then bake this flag as
> an override on the serial-getty@ttysclp0.service? or like add code in
> subiquity to specifically check for tty names?
>
> We already have a is_linux_tty function, do you think that can be used to
> detect colorful serial consoles?
> https://github.com/CanonicalLtd/subiquity/blob/master/subiquitycore/core.py#L73
>
 In general terms if there is an ioctl or something you can call to query
if a fd belongs to an vt220 console I would prefer that. But if not, I
guess serial-getty@ttysclp0.service would be OK. But maybe we should think
a little about whether we have to be monochrome on serial in general.

Cheers,
mwh

vorlonofportland:
Right; given that this is checking $TERM in the environment, it's not clear what would be setting that up.  The kernel certainly doesn't, from the kernel's perspective the getty is just a process.  The systemd unit might if there were some magic getty knowledge here; but it wouldn't currently be applying that logic to the subiquity unit, only to the getty unit that we are interposing.  The getty process itself might.  There is no login shell, and nothing being pushed from the remote end of the serial connection that would declare the TERM.

I do think it's sensible for us to assume a fully-featured terminal client on a serial console, in all cases *except* when we know we're talking to one of the limited-capability z serial consoles, because full-featured is the common case and we can extend the blacklist for exceptions as we encounter them (e.g. there might be some POWER serial consoles with similar constraints).

I think implementing that blacklist is going to require subiquity to have knowledge of the different types of kernel serial devices (whether identified by device name or by some other attribute).

In any case, checking TERM=vt220 looks wrong to me.

subiquitycore/core.py
vorlonofportland: (See other comment)

xnox:
I think we do need to pass the console name to subiquity as a flag, or for subiquity to check where it is running (i.e. who am i).

I find it iffy to rely on TERM too. Even if something is vt220, it doesn't have to support colors for example. 

And on s390x, I don't want to be running subiquity at all on the line-based terminals at all.
And on tty_slcp I do want to force running it....

mwhudson:
On Fri, 16 Nov 2018 at 03:14, Dimitri John Ledkov <notifications@github.com>
wrote:

> I think we do need to pass the console name to subiquity as a flag, or for
> subiquity to check where it is running (i.e. who am i).
>
os.ttyname(0)?


> I find it iffy to rely on TERM too. Even if something is vt220, it doesn't
> have to support colors for example.
>
> And on s390x, I don't want to be running subiquity at all on the
> line-based terminals at all.
> And on tty_slcp I do want to force running it....
>

xnox:
@mwhudson @vorlonofportland switched to os.ttyname(0) check. I think it is safe to dismiss vorlonofportland's review and merge this now?

PR #396, opened 2018-11-19
include curtin in snap

mwhudson:
Just about every detail of this was excruciating, but it does work now.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -25,6 +25,8 @@ parts:
     source: git://git.launchpad.net/curtin
     source-branch: ubuntu/devel
     requirements: requirements.txt
+    organize:
+      'lib/python*/site-packages/usr/lib/curtin': 'usr/lib/'
     stage: [ "*", "-lib/python*/site-packages/_yaml.*.so" ]
     stage-packages: [libc6]
   subiquity:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -53,11 +53,11 @@ parts:
     plugin: dump
     build-packages:
       - user-setup
-    prepare: |
+    override-build: |
       echo "get passwd/user-default-groups" | \
         debconf-communicate user-setup | \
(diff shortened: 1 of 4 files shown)


snapcraft.yaml
```diff
@@ -19,6 +19,16 @@ apps:
...
 parts:
+  curtin:
+    plugin: python
+    source-type: git
+    source: git://git.launchpad.net/curtin
+    source-branch: ubuntu/devel
+    requirements: requirements.txt
+    organize:
+      'lib/python*/site-packages/usr/lib/curtin': 'usr/lib/'
+    stage: [ "*", "-lib/python*/site-packages/_yaml.*.so" ]
```
xnox: I wonder why this looks like a syntax error, by oh well.

snapcraft.yaml
```diff
@@ -19,6 +19,16 @@ apps:
     restart-condition: always
 
 parts:
+  curtin:
+    plugin: python
+    source-type: git
+    source: git://git.launchpad.net/curtin
+    source-branch: ubuntu/devel
```
xnox: Do we want to use this, or do we want to use the automatic git import of the packages that are in ubuntu? i.e. https://code.launchpad.net/~usd-import-team/ubuntu/+source/curtin/+git/curtin/+ref/applied/ubuntu/bionic-devel ? Cause this git repo will give us what is in bionic-proposed at the time.

snapcraft.yaml
mwhudson: yaml be yaml

snapcraft.yaml
mwhudson: I think in general we want to use something newer than whatever is in bionic (bionic-updates is 6 months old, although an SRU is currently in progress).

I guess it would be good to ask curtin people (ping @raharper @smoser) what would be a good branch / tag / whatever to use

mwhudson:
Going to merge this now, it's easy to change the curtin branch we use later if we decide to.

PR #397, opened 2018-11-21
Fix alignment of narrow views.

xnox:
This does a few things:
- make `narrow_rows` views to be the same width as the progress bar in the footer
- use `narrow_rows` views in more places, i.e. on simple pages generated from a form

This results in welcome, installpath, proxy, mirror screens all have Header/title text/footer have one alignment, and the form/buttons have a second alignment with the progress bar.

But I was not sure, if we actually want to make this fix. Instead maybe we want to ditch narrow-rows views and have everything left-aligned to the title header?

Before: https://asciinema.org/a/N4XwVYlY3H3k6aDGQZBco8ork
After: https://asciinema.org/a/hwSyQxu1PxktZWvQtNWbrPbSD

E.g. pay attention to the changes on the Welcome, Installpath, Mirror, Proxy.

Notice how before some of the screens/forms are either wider or shorter than the progress bar, but now a few of them are just as wide as the progress bar.

### DIFF
--- a/subiquitycore/ui/utils.py
+++ b/subiquitycore/ui/utils.py
@@ -236,7 +236,7 @@ def screen(rows, buttons, focus_buttons=True, excerpt=None, narrow_rows=False):
     if isinstance(rows, list):
         rows = ListBox(rows)
     if narrow_rows:
-        rows = Padding.center_63(rows)
+        rows = Padding.center_76(rows)
     if buttons is None:
         focus_buttons = False
     elif isinstance(buttons, list):
--- a/subiquity/controllers/installpath.py
+++ b/subiquity/controllers/installpath.py
@@ -66,7 +66,7 @@ class InstallpathController(BaseController):
             "treat physical servers like virtual machines (instances) "
             "in the cloud. Rather than having to manage each server "
             "individually, MAAS turns your bare metal into an elastic "
-            "cloud-like resource. \n\n"
+            "cloud-like resource.\n"
             "For further details, see https://maas.io/."
             )
         self.ui.set_body(MAASView(self.model, self, title, excerpt))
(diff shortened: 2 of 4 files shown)


clagom:
@xnox 
Thank you for the asciinema. Yes the purpose would indeed be to "Instead maybe we want to have everything left-aligned to the title header?" because at the moment there are elements that go more to the left and are not aligned. If we need more space, I think we can move the header title more to the left, and align the content accordingly. We would need to keep also the centred view (e.g. for the screen where users choose the keyboard type), but at least we would have 2 situations and more consistency overall.

screenshot 2018-11-21 at 11 53 30

screenshot 2018-11-21 at 11 53 19

screenshot 2018-11-21 at 11 55 03

xnox:
@clagom argh..... I should have double checked the asciinema recording, it has everything off, and not like actually shown on the display.  I'll retake the screenshots for you to compare better how I started to implement fixing alignments and whether or not, it is the right direction.

mwhudson:
Currently *most* views use the middle 80% (or so, I forget the exact amount) of the view, but for some 80% of 80 characters really is not enough (filesystem being the obvious one) so they use the middle 95% of the view. What I would like to do is to have these views have a minimum width of 76 characters or whatever and have them use the middle 80% if that gives them enough, but not reduce them past their minimum width. urwid doesn't help you do anything like this but I don't think it would be that hard to implement.

subiquitycore/ui/utils.py
```diff
@@ -236,7 +236,7 @@ def screen(rows, buttons, focus_buttons=True, excerpt=None, narrow_rows=False):
     if isinstance(rows, list):
         rows = ListBox(rows)
     if narrow_rows:
-        rows = Padding.center_63(rows)
+        rows = Padding.center_76(rows)
```
xnox: @mwhudson I think you mean this, no? center_63 is what most centered view forms try to use, whilst center_76 is actually the width of the progress-bar below.

I wonder if we can actually re-evaluate the screen estate and make filesystem screens not use so much width. E.g. popups can open to the left, on top of the screen, instead of to the right.

xnox:
abandoning this attempt.

PR #400, opened 2018-11-23
scripts: add s390x support to inject-subiquity-snap

xnox:
This is like copy & paste from debian-cd.


xnox:
@mwhudson not sure if this is an ok way to modify this script, but it does inject the new snap for me.

mwhudson:
I'd like some comments I guess but meh.

PR #401, opened 2018-11-23
filesystem: do not create bios-grub partition on s390x.

xnox:
This makes the installer to correctly install on to /dev/vda1, without any grub-bios nonsense.

### DIFF
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -298,10 +298,10 @@ class FilesystemController(BaseController):
             self.create_filesystem(partition, spec)
             return
 
-        bootable = self.model.has_bootloader_partition()
-        log.debug('model has bootloader partition? {}'.format(bootable))
+        needs_boot = self.model.needs_bootloader_partition()
+        log.debug('model needs a bootloader partition? {}'.format(needs_boot))
         can_be_boot = DeviceAction.MAKE_BOOT in disk.supported_actions
-        if not bootable and len(disk.partitions()) == 0 and can_be_boot:
+        if needs_boot and len(disk.partitions()) == 0 and can_be_boot:
             part = self._create_boot_partition(disk)
 
             # adjust downward the partition size (if necessary) to accommodate
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -22,6 +22,7 @@ import logging
 import math
 import os
 import sys
+import platform
 
 log = logging.getLogger('subiquity.models.filesystem')
 
@@ -439,8 +440,9 @@ class Disk(_Device):
         DeviceAction.PARTITION,
         DeviceAction.FORMAT,
         DeviceAction.REMOVE,
-        DeviceAction.MAKE_BOOT,
         ]
+    if platform.machine() != 's390x':
+        supported_actions.append(DeviceAction.MAKE_BOOT)
     _can_INFO = True
     _can_PARTITION = property(lambda self: self.free_for_partitions > 0)
     _can_FORMAT = property(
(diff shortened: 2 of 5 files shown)


mwhudson:
You'll need to patch the guided flows too I think?

I'd also prefer to not have has_bootloader_partition lie, you could invert the sense of it and call it "still_needs_bootloader_partition" or something.

xnox:
@mwhudson Switched to `needs_bootloader_partition`, and inverted the conditions everywhere. Is it better now? Tested this on s390x only.

mwhudson:
> @mwhudson now, i finally noticed what you mean and fixed up LVM guided flow.

Thanks.

> It installed fine, however, curtin wrote bad root= arg. It wrote a filesystem UUID, instead of the /dev/mapper name for the VG and the LV.

Is that bad? It sounds like what we want, assuming something autoactivates LVM enough to allow finding the filesystem by UUID...

> Also, the root logical volume was quite small to my taste. Only 4GB out of 20GB sounded too low. I was expecting 16GB root with 4GB left as free in the volume group. But that's separate.

https://bugs.launchpad.net/subiquity/+bug/1785321

> So this should finally fully fix "i need no grub-bios partitions" on s390x.

Merge away then!

xnox:
> > It installed fine, however, curtin wrote bad root= arg. It wrote a filesystem UUID, instead of the /dev/mapper name for the VG and the LV.
> 
> Is that bad? It sounds like what we want, assuming something autoactivates LVM enough to allow finding the filesystem by UUID...
> 

Very bad. Like unpredictably bad......
When booting of complex disks, one must not mount root by filesystem UUID, because it is no longer ambiguous. On a raid1 of two disks, the md device and the two disks all have the same filesystem UUID exposed in the userspace. Similarly, logical volume and all snapshots of it have the same filesystem UUID. Ditto, multipath, and all paths to the same filesystem.

Thus when booting of raid / LVM / LUKS / multipath one really ought to specify root=/dev/mapper/something. However, these days I think we should be switching to root=/dev/disk/by-id/* where by-id are udev generated stable, non-conflicting _actually unique_ references to things.


> > Also, the root logical volume was quite small to my taste. Only 4GB out of 20GB sounded too low. I was expecting 16GB root with 4GB left as free in the volume group. But that's separate.
> 
> https://bugs.launchpad.net/subiquity/+bug/1785321
> 

*sigh*

> > So this should finally fully fix "i need no grub-bios partitions" on s390x.
> 
> Merge away then!

Ok.

vorlonofportland:
On Wed, Nov 28, 2018 at 04:09:29AM -0800, Dimitri John Ledkov wrote:

> Thus when booting of raid / LVM / LUKS / multipath one really ought to
> specify root=/dev/mapper/something.  However, these days I think we should
> be switching to root=/dev/disk/by-id/* where by-id are udev generated
> stable, non-conflicting _actually unique_ references to things.

Switching from /dev/mapper/* names to /dev/disk/by-id/* names as the
preferred identifier has far-ranging consequences and should be discussed in
a broader context than this PR.

raharper:
On Wed, Nov 28, 2018 at 6:09 AM Dimitri John Ledkov <
notifications@github.com> wrote:

> It installed fine, however, curtin wrote bad root= arg. It wrote a
> filesystem UUID, instead of the /dev/mapper name for the VG and the LV.
>
> Is that bad? It sounds like what we want, assuming something autoactivates
> LVM enough to allow finding the filesystem by UUID...
>
> Very bad. Like unpredictably bad......
> When booting of complex disks, one must not mount root by filesystem UUID,
> because it is no longer ambiguous. On a raid1 of two disks, the md device
> and the two disks all have the same filesystem UUID exposed in the
> userspace. Similarly, logical volume and all snapshots of it have the same
> filesystem UUID. Ditto, multipath, and all paths to the same filesystem.
>

 https://bugs.launchpad.net/ubuntu/+source/curtin/+bug/1805674

Thus when booting of raid / LVM / LUKS / multipath one really ought to
> specify root=/dev/mapper/something. However, these days I think we should
> be switching to root=/dev/disk/by-id/* where by-id are udev generated
> stable, non-conflicting *actually unique* references to things.
>

FWIW, the UUID is uniq, it's just that the filesystem UUID is replicated on
multiple devices.  And the by-id for the "combined" device, is just another
UUID anyhow.  None-the-less, curtin can look these values up and use that
UUID or /dev/mapper.  Let's
discuss in the above bug w.r.t best-practices and we'll get this fixed in
curtin.

PR #402, opened 2018-11-26
filesystem: whitelist multipath devices.

xnox:
This does a few things

- blacklists individual paths which are part of a multipath device (i.e. `/dev/sd[a-c]` are neither available in guided or manual partitioning)
- blacklists partitions of a multipath device
- whitelists multipath device itself

https://github.com/CanonicalLtd/probert/pull/44/files sets a nice looking serial for multipath disk, i.e. `mpath-36005076306ffd6b60000000000002405` which is clearly a multipath device, and the long WWID is clearly identifyable as to which one, one is installing things onto.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -885,13 +885,18 @@ class FilesystemModel(object):
             if path in currently_mounted:
                 continue
             if data['DEVTYPE'] == 'disk':
-                if data["DEVPATH"].startswith('/devices/virtual'):
+                if (data["DEVPATH"].startswith('/devices/virtual') and
+                        "DM_WWN" not in data):
                     continue
                 if data["MAJOR"] in ("2", "11"):  # serial and cd devices
                     continue
                 if data['attrs'].get('ro') == "1":
                     continue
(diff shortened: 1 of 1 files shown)


xnox:
@mwhudson have you reviewed this? or do you want an access to a multipath setup to test this out?

mwhudson:
The reason I haven't really looked at this, which is admittedly a bit crap, is that I expect everything here to change when reusing existing partitions happens. But that's no reason not to change this if this helps you sooner...

xnox:
@mwhudson with or without reusing existing things, we shouldn't simply ignore multipath devices. Even if this bit of code will disappear. And yes, it will help me a lot on ppc64el and s390x.

mwhudson:
Feel free to land this then.

xnox:
> Feel free to land this then.

But now I am trying to remember why wrote things the way i did.... cause it did matter to expose certain things but not others. And i can't remember it now.

xnox:
So multipath fails to probe now, and this code is obsolete wrt subiquity codebase. Closing and opening a bug report.

PR #403, opened 2018-11-28
my attempt at a check mark

mwhudson:


mwhudson:
It looks like this:

screenshot from 2018-11-28 20-50-00

screenshot from 2018-11-28 20-50-00-1

Adding a different glyph is easy enough now, if you can give me something in an 8x16 grid.

PR #405, opened 2018-12-13
ask the user if they want to install ssh during the install 

mwhudson:
The diff is scary but a lot of that is code moving around.

Demo: https://asciinema.org/a/sFNXS96PxN3IFpBdNxABP3yoC


mwhudson:
Depends on https://code.launchpad.net/~mwhudson/livecd-rootfs/+git/livecd-rootfs/+ref/live-server-no-openssh-server to be able to actually install a system without openssh-server

PR #408, opened 2019-02-04
use more of the screen for content when the terminal is narrow

mwhudson:
I realized that "Padding.center_79(..., min_width=76)" does what I always wanted: it uses most of the available width when the terminal is 80 columns wide (all but two columns on either side) but only uses 80% of the width when the terminal is wider. I really wanted this for the network and filesystem views but I've changed all views to this because consistency is a nice thing.

Video at https://asciinema.org/a/JYQb2QEQ5oeonnOnXQHbVWwCN


mwhudson:
cc @xnox @clagom

PR #417, opened 2019-03-06
add a more systematic way for controllers to record their state

mwhudson:
so we can return the point where the user left off on update

This is a bit more subtle I guess, will leave the PR around for a while in case anyone wants to look :)


xnox:
Nice. This is very similar to how upstart used to serialize state.

PR #422, opened 2019-03-08
subiquity self-refresh

mwhudson:
Allow the user to update to a newer version of the installer, if available.

Offer the upgrade after language selection, and again after mirror selection (i.e. after network/proxy have been configured).

This does not implement showing release notes of the new version to the user; not yet sure how to do that.

### DIFF
--- a/subiquitycore/core.py
+++ b/subiquitycore/core.py
@@ -284,7 +284,7 @@ class Application:
             "signal": Signal(),
             "prober": prober,
             "loop": None,
-            "pool": futures.ThreadPoolExecutor(4),
+            "pool": futures.ThreadPoolExecutor(10),
             "answers": answers,
             "input_filter": input_filter,
             "scale_factor": scale,
--- a/subiquitycore/core.py
+++ b/subiquitycore/core.py
@@ -26,6 +26,7 @@ import tty
 import urwid
 import yaml
 
+from subiquitycore.controller import RepeatedController
 from subiquitycore.signals import Signal
 from subiquitycore.prober import Prober, ProberException
 
@@ -291,6 +292,8 @@ class Application:
         if opts.screens:
             self.controllers = [c for c in self.controllers
                                 if c in opts.screens]
+        else:
+            self.controllers = self.controllers[:]
         ui.progress_completion = len(self.controllers)
         self.common['controllers'] = dict.fromkeys(self.controllers)
         self.controller_index = -1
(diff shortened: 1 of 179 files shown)


subiquitycore/signals.py
```diff
@@ -54,8 +54,8 @@ def connect_signals(self, signal_callback):
             raise SignalException(
                 "Passed something other than a required list.")
         for sig, cb in signal_callback:
-            if sig not in self.known_signals:
-                self.register_signals(sig)
+            # if sig not in self.known_signals:
```
cyphermox: It feels silly to only nitpick this and not find much else; but why comment out code? Git history remains.

subiquitycore/core.py
```diff
@@ -276,13 +277,14 @@ def __init__(self, ui, opts):
...
         self.common = {
+            "application": self,
             "updated": updated,
             "ui": ui,
             "opts": opts,
             "signal": Signal(),
             "prober": prober,
             "loop": None,
-            "pool": futures.ThreadPoolExecutor(4),
+            "pool": futures.ThreadPoolExecutor(10),
```
cyphermox: Does self-refresh require additional threads to work? Seems slightly unrelated (but I don't have much background).

subiquitycore/core.py
```diff
@@ -475,10 +479,21 @@ def run(self):
                 self.run_scripts(self.common['opts'].scripts)
             controllers_mod = __import__('%s.controllers' % self.project,
                                          None, None, [''])
-            for k in self.controllers:
-                log.debug("Importing controller: {}".format(k))
-                klass = getattr(controllers_mod, k+"Controller")
-                self.common['controllers'][k] = klass(self.common)
+            for i, k in enumerate(self.controllers):
```
cyphermox: Is this changing because subiquity isn't restarting to get the new controllers, or because the refresh could cause duplicates to exist (how come? seems like the new rev should /replace/ everything)?

Or is this because the same controller could get shown twice; ie. ordering could change across a refresh?

subiquitycore/signals.py
mwhudson: It's good to call this out because I didn't understand what was going on here at all. Turns out urwid.register_signals(cls, signals) *replaces* the registered set of signals for cls, rather than appending them which the original author of this code clearly (reasonably!) expected. You've prompted me to figure this out so I've fixed it properly now, thanks :)

subiquitycore/core.py
mwhudson: yeah this is dimly related. self-refresh means that more background things run in parallel -- getting the snap list from the snap store, checking for a snap update, sundry other things. In testing i found something was getting blocked because it was waiting for these network requests to complete, which is bogus really. (There's no real reason to run these network requests in a thread vs using something async like asyncio or twisted, but well). There's no real reason to limit the concurrency so maybe I should just bump this to 50 or something silly (assuming threads are still just created as needed, which I think they are).

subiquitycore/core.py
mwhudson: Yeah this is about the refresh controller appearing more than once in the list. If the check shows there is an update available in time, it offers it after language selection. If it didn't offer an update then, it also asks again after mirror selection.

PR #423, opened 2019-03-11
do not run setupcon if the keyboard setup has not changed

mwhudson:
saves a second or so for the US keyboard user


mwhudson:
Thanks for the review.

PR #424, opened 2019-03-18
relax constraint that launchpad usernames must be >1 character long

mwhudson:
just because it feels like I haven't achieved much today :)


mwhudson:
For https://bugs.launchpad.net/subiquity/+bug/1784370

PR #425, opened 2019-03-19
Attempt to add a PReP partition.

xnox:
Bootloader fails to install on POWER because there is no prep partition.

Ensure we create one.

Not sure how this works at all, in neither subiquity or curtin. But it seems like curtin should know what to do with a 'prep' flag partition. I guess it's worth a try.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1032,7 +1032,8 @@ class FilesystemModel(object):
         log.debug("adding %s to %s", fstype, volume)
         if not volume.available:
             if not isinstance(volume, Partition):
-                if (volume.flag == 'prep' or (volume.flag == 'bios_grub' and fstype == 'fat32')):
+                if (volume.flag == 'prep' or (
+                        volume.flag == 'bios_grub' and fstype == 'fat32')):
                     raise Exception("{} is not available".format(volume))
         if isinstance(volume, Disk):
             self._use_disk(volume)
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -483,7 +483,7 @@ class Partition(_Formattable):
         return _("partition {}").format(self._number)
 
     def available(self):
-        if self.flag == 'bios_grub':
+        if self.flag in ['bios_grub', 'prep']:
             return False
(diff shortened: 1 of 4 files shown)


subiquity/ui/views/filesystem/partition.py
```diff
@@ -208,6 +208,13 @@ def validate_mount(self):
     'installed onto this partition, which must be formatted as fat32. The '
     'only aspect of this partition that can be edited is the size.')
 
+prep_partition_description = _(
+    "Required bootloader partition\n"
+    "\n"
+    'This is the PReP partion which is required on POWER systems. Grub will be '
+    'installed onto this partition. The '
+    'only aspect of this partition that can be edited is the size.')
```
raharper: I don't think the size should be editable; or at least we should require a minimum size.  I've seen recommendations for 8, 10, 16 and 32MB in size.

raharper:
Curtin does indeed know what to do with prep flag, curtin will find the prep partition and pass that as the device to install grub.

subiquity/ui/views/filesystem/partition.py
xnox: Yeah, I have no idea what I am doing here. @adconrad is PReP partition always 8MiB? Fixed size? Allow users to twiddle it or not?

raharper:
> This looks basically fine. Are we going to need to add anything like this for s390x?

No special partitions are needed for zipl.

subiquity/models/filesystem.py
```diff
@@ -483,7 +483,7 @@ def short_label(self):
         return _("partition {}").format(self._number)
 
     def available(self):
-        if self.flag == 'bios_grub':
+        if self.flag in ['bios_grub', 'prep']:
```
zyga:
```suggestion
        if self.flag in ('bios_grub', 'prep'):
```

xnox:
> This looks basically fine. Are we going to need to add anything like this for s390x?

s390x is already done a while back, by removing the "make bootable" action / requirements.

mwhudson:
Oh didn't see the comment from @zyga , still that's a minor thing.

PR #426, opened 2019-03-20
Basic zdev view

xnox:
TODO: allow to operate on "built-in" objects in dry-run mode
TODO: skip the screen on !s390x in real mode
TODO: add a lot more actions, as per spec

This is usable to enable/disable devices, if you managed to boot an LPAR zVM with subiquity.
A bit useless on KVM, as all devices should be activated there by default.

Did not run this with dasd passthrough... not sure if needs activation in that case or not.

Also doesn't integrate with preseeded answers stuff (but not sure that's needed at all, given that in KVM all devices are always activated, one simply needs to push "continue" on s390x in that case).

Overall this is not ready to be merged yet.


mwhudson:
Also you will need answers support to the extend of automatically pressing OK if answers are provided.

xnox:
I was hoping to use ZdevModel in the future, to store the output from chzdev there _or_ store a dict of objects there to support `--dry-run --s390x` mode. But i'm not sure if i can about providing fake experience there. I do need to extend this for answers.... however there is no answers to use in qemu/kvm as one can only really do things with these devices on lpar/zvm only.

I guess I at least need to do "go to the next page" answers support.

Focus does jump.... yes, need to wire up focus location correctly.

xnox:
Pushed Zdev key into ubuntu-test-cases-live-server, such that nothing breaks on s390x when this is merged.

mwhudson:
I don't know if you saw my comments on IRC about this but the way this branch sets up the view is that there is single a huge tablepile inside the listbox. Switching to a tablelistbox with the rows directly inside it performs waaaaay better (there is currently no support for dynamically inserting/removing rows from a tablelistbox but this view as currently written doesn't need that feature -- and it wouldn't be hard to add if it turned out to be needed)

PR #428, opened 2019-03-21
Make a unified launcher

xnox:
Currently tty1 subiquity is launched via snap run subiquity-service, which setups up expected PATH, and does a keys&font setup.

At the moment serial subiquity does nto go via the subiquity-service, and results in a different environment. Also serial subiquity does not block on seeding snaps, and does a weird loop of trying to start up like 3 times. Since it's not blocking on snapd via "/usr/bin/snap run subiquity.subiquity-service"

Let's try to make subiquity-service script, a unified entry point into subiquity. And make serial-subiquity@.service to use that.

This merge can go in, anytime. Whilst livecd-rootfs change will need to go in, after this commit is promoted to stable. As otherwise the arguments passed to the script will not be processed in serial-subiquity@ case.

### DIFF
--- a/bin/subiquity-service
+++ b/bin/subiquity-service
@@ -1,6 +1,14 @@
 #!/bin/sh
+port=tty1
+if [ -n "$1" ]; then
+   port=$1
+fi
 /bin/dmesg -n 1
 export PATH=$SNAP/bin:$SNAP/usr/bin:$PATH
-$SNAP/usr/bin/subiquity-loadkeys
-setfont $SNAP/subiquity.psf
-/sbin/agetty -n --noclear -l $SNAP/usr/bin/python3 -o $SNAP/usr/bin/subiquity tty1 $TERM
+if [ "$port" = "tty1" ]; then
+	$SNAP/usr/bin/subiquity-loadkeys
+	setfont $SNAP/subiquity.psf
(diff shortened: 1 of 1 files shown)


xnox:
livecd-rootfs merge: https://code.launchpad.net/~xnox/livecd-rootfs/+git/livecd-rootfs/+merge/364888

xnox:
Actually just tweaking the livecd-rootfs to export the PATH in a similar fashion how the subiquity-service does it, works.

Or maybe subiquity code itself should export the expected PATH.

I kind of feel the more things are in the subiquity entry point, the better. As at the moment, one cannot use /snap/bin/subiquity as a good entry path.

xnox:
@vorlonofportland 
I changed the order of port baud-rate term, which as per manpage, appeared to be bogus to me
```
- /sbin/agetty -n --keep-baud -l ... 115200,38400,9600 %I $TERM
+ /sbin/agetty -n --keep-baud -l ... $port 115200,38400,9600 $TERM
```

As far as I understand the "hvc0 115200,38400,9600 $TERM" is the right order of args, and it works correctly for me.

I have no idea how did "baud_rate port term" managed to work before, when manpage says: agetty [options] port [baud_rate...] [term]

PR #429, opened 2019-03-22
PReP partition must be wiped, and the _disk_ should not have grub_device flag

xnox:
However, we use the grub_device flag on the disk for UX, but we should
either not have it at all for power, or set it on the partition
(rather than the disk), and I guess we need to make our 'can make
boot' smarter for ppc64le... as in check if the disk has any
partitions marked as prep already... not sure.

### DIFF
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -207,8 +207,8 @@ class FilesystemController(BaseController):
         self.delete_mount(fs.mount())
         self.model.remove_filesystem(fs)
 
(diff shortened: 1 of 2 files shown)


subiquity/controllers/filesystem.py
```diff
@@ -247,7 +248,9 @@ def _create_boot_partition(self, disk):
                     fstype=None,
                     mount=None),
                 flag='bios_grub')
-        disk.grub_device = True
+        # should _not_ specify grub defice for prep
```
raharper: s/defice/device

PR #431, opened 2019-03-25
rework network ui, allow offline progress

mwhudson:
This changes how the network UI works to:

 1. apply changes on every edit
 2. mark interfaces that do not autoconfigure as disabled
 3. allow progress past the network screen even if there is no network

Demo at https://asciinema.org/a/LIC8SLjYyZRdsl8lp2vukUh94


mwhudson:
FWIW after this lands I'll propose https://github.com/mwhudson/subiquity/pull/2 against master

PR #432, opened 2019-03-26
use yaml.safe_load instead of yaml.load

mwhudson:
To avoid warnings


xnox:
Thank you! that was causing spam on serial consoles.

PR #433, opened 2019-03-28
better apt configuration during installs

mwhudson:
Configure apt to prefer installs from pool and to not reference network sources at all if there is no network during install.


bin/subiquity-configure-apt
```diff
@@ -0,0 +1,59 @@
...
+    mount -t overlay overlay -o lowerdir=$dir,workdir=$t/work,upperdir=$t/upper $dir
+}
+
+setup_overlay $TARGET_MOUNT_POINT/etc/apt
+
+mount -t tmpfs tmpfs "$TARGET_MOUNT_POINT/run"
+mkdir -p "$TARGET_MOUNT_POINT/run/cdrom"
+mount --bind /cdrom "$TARGET_MOUNT_POINT/run/cdrom"
+
+cat > "/tmp/sources.list" <<EOF
```
raharper: we may want to use /run/subiquity  as /tmp is volatile and systemd-tmpfiles make delete it at any time.

bin/subiquity-configure-apt
```diff
@@ -0,0 +1,59 @@
...
+deb file:///run/cdrom $(lsb_release -sc) main restricted
+EOF
+if [ $HAS_NETWORK = 'true' ]; then
+    cat "$TARGET_MOUNT_POINT/etc/apt/sources.list" >> "/tmp/sources.list"
```
raharper: And we might as well put the path to the target sources.list file in a variable and reference it.

bin/subiquity-configure-apt
```diff
@@ -0,0 +1,59 @@
...
+#     b. if the network is not expected to be working, we replace the
+#        sources.list with a file just referencing the cdrom.
+#
+# 5. If the network is not expected to be working, we also set up an overlay
+#    over /target/var/lib/apt/lists (if the network is working, we'll run "apt
+#    update" after the /target/etc/apt overlay has been cleared).
+
+PY=$1
+HAS_NETWORK=$2
+
```
raharper: Maybe a check that we have a TARGET_MOUNT_POINT otherwise the mount commands may succeed and then things get really strange.

if [ -z "$TARGET_MOUNT_POINT" ]; then
    exit 1;
fi

bin/subiquity-configure-apt
```diff
@@ -0,0 +1,59 @@
...
+cat > "/tmp/sources.list" <<EOF
+deb file:///run/cdrom $(lsb_release -sc) main restricted
+EOF
+if [ $HAS_NETWORK = 'true' ]; then
+    cat "$TARGET_MOUNT_POINT/etc/apt/sources.list" >> "/tmp/sources.list"
+else
+    setup_overlay $TARGET_MOUNT_POINT/var/lib/apt/lists
+fi
+mv /tmp/sources.list "$TARGET_MOUNT_POINT/etc/apt/sources.list"
+$PY -m curtin in-target -- apt-get update
```
raharper: This triggers an apt-get update and there is UI code to run it as well?  Are both needed?

subiquity/controllers/installprogress.py
```diff
@@ -291,18 +294,20 @@ def _bg_install_openssh_server(self):
...
-                sys.executable, "-m", "curtin", "in-target", "-t", "/target",
-                "--", "apt-get", "update",
+            cmds = [
+                ["umount", self.tpath('etc/apt')],
                 ]
-        self._bg_run_command_logged(cmd)
+            if self.base_model.network.has_network:
+                cmds.append([
+                    sys.executable, "-m", "curtin", "in-target", "-t",
+                    "/target", "--", "apt-get", "update",
```
raharper: The shell program also calls apt-get update, are both needed?

bin/subiquity-configure-apt
mwhudson: The one in the UI runs after the install has finished and the /etc/apt overlay unmounted, so that one is needed. I'm not 100% sure this one is needed though, does curtin always run apt-get update itself before using apt? My memory from reading logs is that it does but maybe I found some case in the code where it doesn't.

bin/subiquity-configure-apt
mwhudson: I think this is a bit exaggerated isn't it? We control the environment this runs in entirely, and if this environment really can delete all files in /tmp at any time we should fix that.

When re-reading this code I thought of a different way to do it that avoids using /tmp so I've changed this anyway.

bin/subiquity-configure-apt
mwhudson: Good point.

subiquity/controllers/installprogress.py
mwhudson: See above: this update is run after the apt configuration has been reverted to the configuration we want for the installed system so yes, it's needed.

mwhudson:
Tested the new version with and without network successfully so merging. Can always improve shell style later :)

bin/subiquity-configure-apt
raharper: If you manually invoke apt-get via in-target, then no; but using install_packages or other apt: config level stuff, curtin does handle running update before issuing install/upgrade commands.

What's the need for issuing an update post-install ? Are you installing packages?

subiquity/controllers/installprogress.py
raharper: I see, we're refreshing the cache since the repos have changed.  That makes sense.

bin/subiquity-configure-apt
raharper: I see below that this is done to reset the apt-cache in the target system after removing the pool sources.

snapcraft.yaml
```diff
@@ -58,8 +60,8 @@ parts:
       'bin/console-conf-tui': usr/bin/console-conf
       'bin/subiquity-tui': usr/bin/subiquity
       'bin/subiquity-loadkeys': usr/bin/subiquity-loadkeys
-      'bin/curtin-journald-forwarder': usr/bin/curtin-journald-forwarder
```
xnox: @mwhudson why is that getting deleted? I thought we use it? or not any more?

mwhudson: the actual file referenced here got deleted a year ago, when curtin implemented reporting to the journal natively. One might have expected snapcraft to complain about me forgetting to delete it here too but well. Apparently it didn't!

PR #434, opened 2019-03-29
customize where subiquity refreshes from

mwhudson:
also support passing subquity-channel= on the kernel command line

### DIFF
--- a/subiquity/controllers/refresh.py
+++ b/subiquity/controllers/refresh.py
@@ -114,15 +114,18 @@ class RefreshController(BaseController):
                 log.exception("checking switch")
                 return
             if response.json()["result"]["status"] == "Done":
-                return
+                return True
             time.sleep(0.1)
 
     def _snap_switched(self, fut):
         log.debug("snap switching completed")
         try:
-            fut.result()
+            success = fut.result()
         except:
             log.exception("_snap_switched")
+            return
+        if not success:
+            return
         self.switch_state = SwitchState.SWITCHED
         self._maybe_check_for_update()
 
--- a/subiquity/controllers/refresh.py
+++ b/subiquity/controllers/refresh.py
@@ -16,6 +16,7 @@
 import enum
 import logging
 import os
+import time
 
 import requests.exceptions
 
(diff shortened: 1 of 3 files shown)


subiquity/controllers/refresh.py
```diff
@@ -47,11 +54,93 @@ def __init__(self, common):
...
+                info = (
+                    'Ubuntu-Server 18.04.2 LTS "Bionic Beaver" - '
+                    'Release amd64 (20190214.3)')
+            else:
+                log.debug("failed to find .disk/info file")
+                return
+        else:
+            with fp:
+                info = fp.read()
+        release = info.split()[1]
```
xnox: I don't like any of this. Daily builds might not have digits in the info file at all. And 18.04.2 is too of a restrictive track, no? We probably want just the 18.04 track, no? Imho you should simply inspect `/etc/os-release` the `VERSION_ID="19.04"` field.

mwhudson: I can't see how a daily build could _not_  have digits in the info file, looking at debian-cd, although perhaps I should be a bit more defensive here. And no, 18.04.2 is not too restrictive, it is quite possible for there to be changes that we don't want to go back to a previous release -- for example if we make a change that depends on an addition to the server-ship-live seed between 18.04.3 and 18.04.4 we might not want users of 18.04.3 media to get that as it would break offline installs.

mwhudson: > for example if we make a change that depends on an addition to the server-ship-live seed between 18.04.3 and 18.04.4 we might not want users of 18.04.3 media to get that as it would break offline installs.

Haha this example makes no sense does it. Time for bed.

But the point is that the most likely reason to not want to offer a new subiquity to a user of an older release is that new subiquity depends on some change in how the image is made. Noone has yet suggested a single (non arch specific) feature for subiquity that we would not want to offer to 18.04 users (really, I went through all bugs and cards to check this)

xnox: ok, that's fine. It's just for each point release, we'd then need to open and close a track. And i guess by default we will make the fallback chain to be "stable/18.04.3 -> stable/18.04 -> stable" or maybe simply just "stable/18.04.3 -> stable"

vorlonofportland: We don't have support for creating a non-default fallback chain for closed tracks.  It will be "stable/18.04.3 -> stable"; and if at any point we reopen stable/18.04, we will also need to copy to stable/18.04.{3,4} etc.

PR #435, opened 2019-04-01
make fs action code more generic with some metaprogramming

mwhudson:


xnox:
ok. this looks like valid python, this is as much as i can review this.

PR #439, opened 2019-04-03
render LV sizes with 'B' to match new curtin schema

mwhudson:


mwhudson:
Of course once curtin accepts the declared schema I can just pass size as an integer. But for now...

PR #444, opened 2019-04-05
Use a single list on FilesystemModel to store all the actions

mwhudson:
Another thing looking forward to reusing existing partitions/devices.

One user-visible change here: the devices list now lists constructed
devices, most recently created first, followed by the disks. This feels
better to me.


mwhudson:
Yeah, it's also more broken in other ways. Will get back to it!

>

mwhudson:
This works now I think

PR #446, opened 2019-04-09
show subiquity versions in upgrade screen

mwhudson:
https://asciinema.org/a/ZwsCBvgEXN3QMVwkwPQ06mnck


mwhudson:
We should consider using the ubu.one URL shortener to make a shorter, easier to type URL for the release notes. Any ideas what that should be? ubu.one/live-server-relnotes?

PR #447, opened 2019-04-09
always update all packages as part of installation

mwhudson:
@vorlonofportland are you happy with always doing this?

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -201,6 +201,10 @@ class SubiquityModel:
                 'config': self.filesystem.render(),
                 },
 
+            'system_upgrade': {
+                'enabled': True,
+            },
+
             'write_files': {
                 'etc_default_keyboard': {
                     'path': 'etc/default/keyboard',


xnox:
(this would be curtin doing it, by chrooting into target and upgrading all the things)

vorlonofportland:
@mwhudson what do we do today in other contexts (maas; d-i)?

Always upgrading *all* packages means that two systems installed at different points in time from the same media, on which the user has never manually run an apt upgrade, will end up with different package sets installed (basically, the set of packages that are in updates && !security, since unattended-upgrades only auto-installs security updates by default).

As we move in general towards increased predictability of Ubuntu installs at scale, I think there's an argument to be made that we should not auto-upgrade packages as part of the install, but instead allow the user to opt into this (in the UI if we consider it important enough, or else by running apt upgrade post-install if not).

raharper:
> @mwhudson what do we do today in other contexts (maas; d-i)?
> 

maas runs daily images; they don't send system_upgrade: enabled since they're only a day off of any images.

vorlonofportland:
> > @mwhudson what do we do today in other contexts (maas; d-i)?
> 
> maas runs daily images; they don't send system_upgrade: enabled since they're only a day off of any images.

Fair enough, so although users might pin a particular version of a maas image by mirroring locally and gating the updates to their mirror stream, that is not the expected default behavior.

So I'm ok with this, but reserve the right to revisit based on feedback.

mwhudson:
I can sort of see the argument for offering the choice between installing only security updates or all updates in the UI, but this is much easier to implement obviously (I guess security only could be implemented by using apt pinning to score -updates way down during the install?)

mwhudson:
So I guess the question is, should I merge this now and file a bug for the other feature, or implement the other thing first?

xnox:
We should upgrade and try not to boot systems vulnerable.
Ie. cloud-images update and upgrade on boot, as far as I understand.

If one wants to install systems without updates, they can pull the internet cable out.

raharper:
> If one wants to install systems without updates, they can pull the internet cable out.

That's a good point;  we cannot add that config if we're not network enabled or the install will fail.

mwhudson:
Actually that part is OK, because the only apt source enabled during an offline install is the pool from the iso.

mwhudson:
https://github.com/CanonicalLtd/subiquity/pull/451 supersedes this

PR #448, opened 2019-04-10
allow answers to override the channel subiquity is refreshed from

mwhudson:


xnox:
I like it !

that way i can test things from edge/xnox

PR #451, opened 2019-04-11
(interruptibly) install security updates after install 

mwhudson:
Run u-u in the target after all configuration is done. If the user clicks reboot before it finishes, run  unattended-upgrade-shutdown --stop-only and wait for u-u to exit before actually rebooting.

Dry run version: https://asciinema.org/a/ZwDriz1Oos5e87D2ZuUZ4gCKq

Includes a large refactoring of the way postinstall steps are run, oops.

### DIFF
--- a/subiquity/controllers/installprogress.py
+++ b/subiquity/controllers/installprogress.py
@@ -252,10 +252,10 @@ class InstallProgressController(BaseController):
         self.progress_view.show_complete(True)
         self.default()
 
-    def _bg_run_command_logged(self, cmd):
+    def _bg_run_command_logged(self, cmd, **kwargs):
         cmd = ['systemd-cat', '--level-prefix=false',
                '--identifier=' + self._log_syslog_identifier] + cmd
-        return utils.run_command(cmd)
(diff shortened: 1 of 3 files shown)


mwhudson:
I haven't tested this in a VM yet. I've also changed the message shown while unattended-upgrade-shutdown is running to "deferring update".

mwhudson:
I have now tested this in a VM -- in an offline install, letting u-u run to completion and interrupting it. All seemed to work.

subiquity/controllers/installprogress.py
```diff
@@ -307,76 +452,65 @@ def _bg_cleanup_apt(self):
...
+            self._bg_run_command_logged([
+                sys.executable, "-m", "curtin", "in-target", "-t", "/target",
+                "--", "unattended-upgrades", "-v",
+            ], check=True)
+
+    @task(net_only=True)
+    def abort_uu(self):
+        self._install_event_finish()
 
+    @task(label="deferring update", net_only=True)
```
vorlonofportland: I think "deferring" is unclear.  I suggest "cancelling installation of security updates".

PR #460, opened 2019-04-15
Copy machine-id into target, for it to stay the same machine.

xnox:


xnox:
https://bugs.launchpad.net/subiquity/+bug/1824811

xnox:
all passed, but disco, which is getting fixed soon on the mirrors.

mwhudson:
thumbs up

PR #461, opened 2019-04-15
write hostname/hosts/machine-id earlier

mwhudson:
cloud-init may end up writing them too late


mwhudson:
xnox reviewed this before i made the PR

PR #464, opened 2019-04-17
start a design document

mwhudson:
probably best to get some of this stuff out of my head.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -7,9 +7,13 @@
 1. Subiquity is entirely usable by pressing up, down, space (or return) and the
    occasional bit of typing.
 
-2. The UI never blocks.  If something takes more than about 0.1s, it is done in
-   the background, possibly with some kind of indication in the UI (and the
-   ability to cancel if appropriate).
+2. The UI never blocks.  If something takes more than about 0.1s, it is done
+   in the background, possibly with some kind of indication in the UI and the
+   ability to cancel if appropriate.  (We should consider making sure that if
+   we pop up a progress dialog while something happens -- e.g. applying
+   keyboard configuration, which the user just has to wait for until going on
+   to the next screen -- that the dialog appears for at least, say, 0.5s to
+   avoid flickering the UI).
 
 3. General UX principles that it is worth keeping in mind:
 
@@ -168,7 +172,7 @@ us not use a background thread for absolutely every action we want to run in
 the background (there's no real reason to tie up a thread waiting for a
 subprocess to exit, for example).
 
-Only touch the UI from the main thread.
+A cast-iron rule: Only touch the UI from the main thread.
 
 ### Terminal things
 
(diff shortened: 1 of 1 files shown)


DESIGN.md
```diff
@@ -0,0 +1,225 @@
...
+## UI
+
+### basic ground rules:
+
+1. Subiquity is entirely usable by pressing up, down, space (or return) and the
+   occasional bit of typing.
+
+2. The UI never blocks.  If something takes more than about 0.1s, it is done in
+   the background, possibly with some kind of indication in the UI (and the
+   ability to cancel if appropriate).
```
xnox: I wonder if we should state that showing progress pop-ups for small things results in too fast flicker. Maybe like progress popup should be a minimum of 0.5s, when shown. To prevent too fast flicker.

DESIGN.md
```diff
@@ -0,0 +1,225 @@
...
+controller that listens to netlink events and calls methods on the model and
+view instances in response to, say, a NIC gaining an address.
+
+The views display the model and call methods on the controller to make changes.
+
+Obviously for most screens there is a triple of a model class, controller class
+and a view class for the initial view, but this isn't always true -- some
+controllers don't have a corresponding model class.
+
+### Doing things in the background
```
xnox: this is a very good section!

DESIGN.md
```diff
@@ -0,0 +1,225 @@
...
+does not use, so we can add support for at least a dozen or so more glyphs if
+there's a need.
+
+`subiquity.palette` defines the 8 RGB colors and a bunch of named "styles" in
+terms of foreground and background colors.  `subiquitycore.core` contains some
+rather hair-raising code for mangling these definitions so that using these
+style names in urwid comes out in the right color both in gnome-terminal (using
+ISO-8613-3 color codes) and in the linux tty (using the PIO_CMAP ioctl).
+
+### Testing
```
xnox: URLs to the platform qa test results maybe?

DESIGN.md
mwhudson: Yeah, I wonder about this. Thinking about two examples where we block the UI like this (actually these might be the only two currently?):

 1. applying keyboard configuration -- I think for this we could make an argument for padding the runtime out to 0.5s if the subprocess finished quickly (but it always takes about 1.2s IME)
 2. downloading ssh keys -- I think here the case is a bit less clear, because what follows the download is still part of the same operation in some sense. But OTOH I guess padding out to 0.5s still isn't very long...

So um I dunno. I'll add a comment for now.

DESIGN.md
mwhudson: :champagne:

DESIGN.md
mwhudson: Yeah added, along with the VPN related disclaimer.

mwhudson:
> Overall this is a good document, with accurate and up to date descriptions. Hopefully we will keep it up to date as we progress things.

Yeah, we should try to avoid letting it go stale!

PR #468, opened 2019-04-26
copy /cdrom/.disk/info to /var/log/installer/media-info

mwhudson:
for https://bugs.launchpad.net/bugs/1825576

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -160,6 +160,13 @@ class SubiquityModel:
             os.makedirs(os.path.dirname(path), exist_ok=True)
             write_file(path, content, mode, omode="w")
 
+    def _media_info(self):
+        if os.path.exists('/cdrom/.disk/info'):
+            with open('/cdrom/.disk/info') as fp:
+                return fp.read()
+        else:
+            return "media-info"
+
     def render(self, syslog_identifier):
         config = {
             'apt': {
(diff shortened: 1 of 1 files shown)


mwhudson:
Tested in a VM, seems to work.

subiquity/models/subiquity.py
```diff
@@ -227,6 +234,11 @@ def render(self, syslog_identifier):
                     'content': open('/etc/machine-id').read(),
                     'permissions': 0o444,
                     },
+                'media_info': {
+                    'path': 'var/log/installer/media-info',
+                    'content': self._media_info(),
+                    'permissions': 0o644,
+                    },
```
vorlonofportland: I know this is already merged, but I wonder if we should consider this sufficient.  The purpose of this file is to allow post-install debugging by being able to pinpoint which version of the installer was used to install the system.  But if the user refreshed the subiquity snap, this isn't complete, is it?  Should we include the information about the subiquity snap revision/version here, or in an additional file?

mwhudson: Yes, it would be interesting to have the subiquity revision included in a file that apport picks up as part of making a crash report.  It is included in the pollinate data, so apport could maybe look at that? (I don't actually know where that ends up on disk)

mwhudson: Let's discuss this in the standup cc @murraybd

xnox: We still want the media-info as that describes the squashfs that is installed on target, as that's the primary purpose of the file - rather than just identifying the version of the installer used. Ubiquity too, can be upgraded live.

PR #472, opened 2019-05-01
Flags that do not indicate a bootloader partition are ok_for_raid

mwhudson:
Existing partitions can turn up with flags like "linux" and these are
fine to put into a RAID. I also checked all the other places we inspect
flag to check they weren't being over strict and they seem fine.


mwhudson:
We do have a raid1 job in jenkins.

xnox:
Ah, on amd64 only. Need to add them to ports too.

PR #473, opened 2019-05-02
move special handling of DM_Crypt actions from model to controller

mwhudson:


raharper:
I bet it's this:

https://bugs.launchpad.net/curtin/+bug/1815666

We can test to see if it's safe to include /run for all releases though I think generally it's a specific version of LVM that's triggering an expectation on what's in /run ;  it still seems wrong to expect /run to affect the lvm commands since it had to collect info the first time it runs.

xnox:
> I bet it's this:
> 
> https://bugs.launchpad.net/curtin/+bug/1815666
> 
> We can test to see if it's safe to include /run for all releases though I think generally it's a specific version of LVM that's triggering an expectation on what's in /run ; it still seems wrong to expect /run to affect the lvm commands since it had to collect info the first time it runs.

huh.... initially lvm is executed in host, this is about lvm execution in target. There is /run on the host, but  not in target. But should be, to access udev database. (and resolved state, and networkd state, and the fact that we are systemd booted, and lots of other things).

xnox:
All good now!

PR #474, opened 2019-05-02
Introduce the idea of filesystem action annotations

mwhudson:
This generalizes the way we display if a VG is encrypted and the way we
display the bootloader partitions, and will be how new/existing is
displayed when reusing existing partitions happens.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -320,6 +320,15 @@ class _Formattable(ABC):
     # Base class for anything that can be formatted and mounted,
     # e.g. a disk or a RAID or a partition.
 
+    @property
+    @abstractmethod
+    def label(self):
+        pass
+
+    @property
+    def annotations(self):
+        return []
+
     # Filesystem
     _fs = attr.ib(default=None, repr=False)
     # Raid or LVM_VolGroup for now, but one day ZPool, BCache...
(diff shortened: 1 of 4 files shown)


subiquity/models/filesystem.py
```diff
@@ -553,6 +562,17 @@ class Partition(_Formattable):
     flag = attr.ib(default=None)
     preserve = attr.ib(default=False)
 
+    @property
+    def annotations(self):
+        r = super().annotations
+        if self.flag == "prep":
+            r.append("PReP")
+        elif self.flag == "boot":
+            r.append("ESP")
```
xnox: I thought 'boot' flag is present on bios and prep partitions as well. Meaning that it's not ESP specific....

mwhudson: no, flag is bios_grub or prep on those partitions. It's a string, not a list or a set...

xnox: ah

xnox:
Has conflicts now that dm-crypt stuff is merged.

```
<<<<<<< fs-action-labels
            if device.annotations:
                label = "{} ({})".format(label, ", ".join(device.annotations))
=======
            if device.type == "lvm_volgroup":
                member = next(iter(device.devices))
                if member.type == "dm_crypt":
                    label += _(" (encrypted)")
>>>>>>> master
```

I guess we simply want both hunks, right?

PR #475, opened 2019-05-02
netplan.Config().config should initialize to a dict

xnox:
Normally netplan.Config().config is a dict as produced by
yaml.safe_load(). However, on systems without any netplan configs that
is not true. The subiquity code that calls into this function expects
.config to be a dictionary. When it's not
subiquitycore/models/network.py produces traceback that "list does not
have get method" when trying to get 'network' key.


mwhudson:
thanks

PR #476, opened 2019-05-03
rework how we handle filesystem types a bit

mwhudson:
1. we can't assume we know all filesystems we'll end up seeing, so don't blow up in that case (we assume here that any unknown filesystems are mountable, which given where they come from -- libblkid via udev -- should be the case)
2. get rid of the FS() objects that used to be passed around for filesystem types and just pass strings or None -- works out simpler in practice.

### DIFF
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -161,7 +161,7 @@ class FilesystemController(BaseController):
         raise Exception("could not resolve {}".format(id))
 
     def _action_clean_fstype(self, fstype):
-        return self.model.fs_by_name[fstype]
+        return self.model.get_fs_by_name(fstype)
 
     def _action_clean_devices_raid(self, devices):
         return {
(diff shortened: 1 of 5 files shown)


xnox:
Looks good. Swap, can sort of be mounted though, right? it can be swaped-on/off, and it matters if it's in /etc/fstab or not. As part of reusing partitions, this might matter if one wants or _doesn't_ want swap to be reused. In dual-boot situations, one might legitimately do either. (ie. hybernate each OS into its own swap, or just reuse a single swap with whichever OS is booted). But I guess that's work for later.

mwhudson:
> Looks good. Swap, can sort of be mounted though, right? it can be swaped-on/off, and it matters if it's in /etc/fstab or not. As part of reusing partitions, this might matter if one wants or _doesn't_ want swap to be reused. In dual-boot situations, one might legitimately do either. (ie. hybernate each OS into its own swap, or just reuse a single swap with whichever OS is booted). But I guess that's work for later.

Yes, that does all need to be handled (and is probably going to be a right pain out of all proportion with the use this feature will get) but: not today.

PR #477, opened 2019-05-07
Refactor how model configs are combined

mwhudson:
Basically each model now produces a fragment of a complete curtin config
and they are combined using curtin's merge_config function rather than
SubiquityModel.render knowing the shape of the config produced by each
model.

I've been meaning to do this for a while. It would be easy to mess things up here so please review with a bit of care.

### DIFF
--- a/subiquity/models/installpath.py
+++ b/subiquity/models/installpath.py
@@ -38,26 +38,30 @@ class InstallpathModel(object):
         if self.cmdline_sources:
             cmdline = [(_('Install from cli provided sources'), 'cmdline')]
         return cmdline + [
-            (_('Install Ubuntu'),                 'ubuntu'),
+            (_('Install Ubuntu'),                         'ubuntu'),
             (_('Install MAAS bare-metal cloud (region)'), 'maas_region'),
             (_('Install MAAS bare-metal cloud (rack)'),   'maas_rack'),
         ]
-
-    @property
-    def sources(self):
-        src_map = {
-            'ubuntu': ['cp:///media/filesystem'],
-            'maas_region': ['cp:///media/region'],
-            'maas_rack': ['cp:///media/rack'],
-            'cmdline': self.cmdline_sources}
-        return {self.path + "%02d" % n: u
-                for n, u in enumerate(src_map[self.path])}
 
     def update(self, results):
         self.results = results
 
     def render(self):
-        config = {}
+        src_map = {
+            'ubuntu': ['cp:///media/filesystem'],
+            'maas_region': ['cp:///media/region'],
+            'maas_rack': ['cp:///media/rack'],
+            'cmdline': self.cmdline_sources,
+            }
+        src_list = src_map[self.path]
+
+        sources = {}
+        for n, u in enumerate(src_list):
+            sources[self.path + "%02d" % n] = u
+
+        config = {
+            'sources': sources,
+            }
 
         def t(path):
             return os.path.join(self.target, path)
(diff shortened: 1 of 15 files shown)


xnox:
Note to self:
The interesting pieces to validate are that e.g. `write_files` is correctly merged across all modules and such that e.g. setting proxy doesn't remove the keyboard write_files snipper. Or that apt config has both mirror and proxy. Or that maas install path write_files are still there.

mwhudson:
Yeah, that part is fine: the reason that curtin config has dictionaries in places you might expect lists (like write_files) is exactly so merge_configs dtrt when merging them (or at least, I assume that this is the reason, I haven't asked anyone).

raharper:
I would suggest adding a unittest on this path to ensure that newly added snippets don't clobber things, but generally looking at:
https://git.launchpad.net/curtin/tree/tests/unittests/test_config.py

Should demonstrate expected merging behavior of config.merge_config.

Note, curtin does not have an advanced config for controllering *how* things are merged.
So, generally, "last" key/value pair wins, much like dict.update().

subiquity/models/subiquity.py
```diff
@@ -68,18 +74,30 @@ def __init__(self, common):
...
+
+        # Collect the models that produce data for the curtin config.
+        self._install_models = [
+            self.keyboard,
+            self.installpath,
+            self.network,
+            self.proxy,
+            self.mirror,
+            self.filesystem,
+            ]
```
raharper: this list seems brittle, is it possible to just call render() on all models?  If an new or existing model adds curtin config, then this list needs updating.

mwhudson: I sort of agree but there's no simple way to get a list of "all models" either really?

mwhudson:
> I would suggest adding a unittest on this path to ensure that newly added snippets don't clobber things

This was a good idea, see what you think of the tests I've added now.

subiquity/models/subiquity.py
xnox: One could interate the attributes of the object to see which ones are models.... but at the moment we'd need to like do a fuzzy match that `__class__.endswith('Model')` as all of them are just "objects" and don't inherit from anything. Maybe they should inherit all from a base-class Model, that at least implements `def render(): pass` but meh.

subiquity/models/installpath.py
```diff
@@ -38,26 +38,30 @@ def paths(self):
...
             cmdline = [(_('Install from cli provided sources'), 'cmdline')]
         return cmdline + [
-            (_('Install Ubuntu'),                 'ubuntu'),
+            (_('Install Ubuntu'),                         'ubuntu'),
             (_('Install MAAS bare-metal cloud (region)'), 'maas_region'),
             (_('Install MAAS bare-metal cloud (rack)'),   'maas_rack'),
         ]
 
-    @property
-    def sources(self):
```
xnox: This removed property is referenced by `get_seed_yaml` here https://github.com/mwhudson/subiquity/blob/refactor-config-creation/subiquity/ui/views/snaplist.py#L415

Not sure if we want to keep the sources property, or make the snaplist reference and transverse the full rendered config instead?

Currently it bobms out with assertionerror. So can't merge this.

subiquitycore/i18n.py
```diff
@@ -29,6 +29,7 @@
 
 
 def switch_language(code='en_US'):
+    syslog.syslog('switch_language ' + code)
```
raharper: odd change for this PR

PR #478, opened 2019-05-07
reorganize handling of bootloader partitions

mwhudson:
This includes #477 so go review that first.

The upshot is that we no longer set grub_device on any disks but rather track the boot device as grub_install_device on the FilesystemModel object if necessary -- it's not needed for s390x or UEFI boots. This should be easier to make work with existing partitions.

### DIFF
--- a/subiquity/cmd/tui.py
+++ b/subiquity/cmd/tui.py
@@ -63,9 +63,9 @@ def parse_options(argv):
     parser.add_argument('--machine-config', metavar='CONFIG',
                         dest='machine_config',
                         help="Don't Probe. Use probe data file")
-    parser.add_argument('--uefi', action='store_true',
-                        dest='uefi',
-                        help='run in uefi support mode')
+    parser.add_argument('--bootloader',
+                        choices=['none', 'bios', 'prep', 'uefi'],
+                        help='Override style of bootloader to use')
     parser.add_argument('--screens', action='append', dest='screens',
                         default=[])
     parser.add_argument('--script', metavar="SCRIPT", action='append',
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1187,12 +1187,15 @@ class FilesystemModel(object):
     def needs_bootloader_partition(self):
         '''true if no disk have a boot partition, and one is needed'''
         # s390x has no such thing
-        if platform.machine() == 's390x':
+        if self.bootloader == Bootloader.NONE:
             return False
-        for p in self.all_partitions():
-            if p.flag in ('bios_grub', 'boot', 'prep'):
-                return False
-        return True
+        elif self.bootloader in [Bootloader.BIOS, Bootloader.PREP]:
+            return self.grub_install_device is None
+        elif self.bootloader == Bootloader.UEFI:
+            return self._mount_for_path('/boot/efi') is None
+        else:
+            raise AssertionError(
+                "unknown bootloader type {}".format(self.bootloader))
 
     def _mount_for_path(self, path):
         for mount in self.all_mounts():
@@ -1214,7 +1217,6 @@ class FilesystemModel(object):
         return False
 
     def can_install(self):
-        # Do we need to check that there is a disk with the boot flag?
         return (self.is_root_mounted()
                 and not self.needs_bootloader_partition()
                 and self.is_slash_boot_on_local_disk())
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -555,10 +555,17 @@ class Disk(_Device):
         lambda self: len(self._partitions) == 0 and
         self._constructed_device is None)
     _can_REMOVE = property(_generic_can_REMOVE)
-    _can_MAKE_BOOT = property(
-        lambda self:
-        not self.grub_device and self._fs is None
-        and self._constructed_device is None)
+
+    @property
+    def _can_MAKE_BOOT(self):
+        install_dev = self._m.grub_install_device
+        if install_dev:
+            # For the PReP case, the install_device is the prep partition.
+            if install_dev.type == "partition":
+                install_dev = install_dev.device
+            if install_dev is self:
+                return False
+        return self._fs is None and self._constructed_device is None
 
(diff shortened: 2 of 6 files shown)


subiquity/models/keyboard.py
```diff
@@ -211,3 +211,14 @@ def set_keyboard(self, setting):
             else:
                 scale = os.environ.get('SUBIQUITY_REPLAY_TIMESCALE', "1")
                 run_command(['sleep', str(1/float(scale))])
+
+    def render(self):
```
cyphermox: Seems like this part of refactor to support the reorg for bootloader  should have been a different PR.

subiquity/cmd/tui.py
```diff
@@ -63,9 +63,9 @@ def parse_options(argv):
     parser.add_argument('--machine-config', metavar='CONFIG',
                         dest='machine_config',
                         help="Don't Probe. Use probe data file")
-    parser.add_argument('--uefi', action='store_true',
-                        dest='uefi',
-                        help='run in uefi support mode')
+    parser.add_argument('--bootloader',
+                        choices=['none', 'bios', 'prep', 'uefi'],
+                        help='Override style of bootloader to use')
```
cyphermox: I'm not sure I really understand why this needs to be overridable; surely it's set in stone if you want to boot, that on a particular architecture you need a particular bootloader. No use in having uefi grub on ppc64el, it just won't boot.

subiquity/models/filesystem.py
```diff
@@ -867,14 +880,29 @@ class FilesystemModel(object):
             fs_by_name[fs.label] = fs
     fs_by_name['fat32'] = FS('fat32', True)
 
+    def _probe_bootloader(self):
```
cyphermox: Consider using archdetect. It's on ubuntu-live images, probably could be on ubuntu-server-live images; it's meant to be used (package: archdetect-deb) to automatically gives you architecture and subarch (although it *is* coming from hw-detect, a d-i component; and probably best be extracted into it's own thing). You'd get a string like 'amd64/efi' or 'amd64/generic', that will give you the arch/subarch details that should be enough to tell exactly what bootloader you need; even to the point of telling whether you really need grub-ieee1374 or grub-whatever-else-for-ppc64el, + whatever else you might need for future support of arm/arm64.

xnox: Well. We support a lot less than what hw-detect can do, as we target whatever curtin can do. Going forward we will probably have "uefi+bios", "uefi", "prep", "none". And I wonder if we should simply "patch" platform.machine() from cmdline for testing "different" things. Not relying on hw-detect imho is currently an advantage.

subiquity/cmd/tui.py
mwhudson: It's just for testing, yes (this flag is only inspected if --dry-run is also passed)

subiquity/models/keyboard.py
mwhudson: Hah yes, how did that get here!

mwhudson: Oh wait, this is the change from #477 which is being reviewed separately I hope

mwhudson:
Ah yes, now fixed

xnox:
Github is a confused silly thing, and shows a very bad diff for review. Granted I pre-fetched all of the refs/pulls into my local repository with `fetch = +refs/pull/*:refs/remotes/origin/pull/*` such that I could review this properly under the La Manche / English Channel.

So, pre-recorded from the Eurotunnel:

In Makefile, dryrun target needs adjustment from `--uefi` to `--bootloader=uefi`

In subiquity/models/filesystem.py, class Disk now has an unused grub_device attribute. But the grub config is now moving to the `config['grub']` key. Question arises, if we must now drop the grub_device attribute. As we now generate 'grub_device: false' on the single disk, which is pointed at with the overall 'grub:' key. I expect to use only xor (grub, grub_device), or have them consistent (ie. devices in install_devices, also have the grub_device=True key). My preference is to drop grub_device attribute from the Disk object.

I also noticed that btrfs is excluded as a candidate for swapfiles. New enough kernels (e.g. disco+ hwe) do support swapfiles on btrfs. My plan was to fixup partman-swapfile & i guess curtin to start supporting them. Basically one marks them to be excluded from cow and preallocates blocks to basically have a bit of btrfs volume that is nor realy btrfs and usable for swap. Kind of making a small stylish tear in the btrfs fabric, like the pre-hipster ripped jeans from the 00s.

ps. Live from Eurostar.

xnox:
Please rebase on top of master, to possibly fix github's review diff to be sane.

mwhudson:
Rebased (diff looks sane now) and makefile fixed.

> In subiquity/models/filesystem.py, class Disk now has an unused grub_device attribute. But the grub config is now moving to the `config['grub']` key. Question arises, if we must now drop the grub_device attribute. As we now generate 'grub_device: false' on the single disk, which is pointed at with the overall 'grub:' key. I expect to use only xor (grub, grub_device), or have them consistent (ie. devices in install_devices, also have the grub_device=True key). My preference is to drop grub_device attribute from the Disk object.

This is a mildly interesting meta-point: if I remember correctly I created the action classes by typing up all config mentioned in the curtin config at https://curtin.readthedocs.io/en/latest/topics/storage.html, including config I had no intention of using. So given that we now never set the grub_device flag, should we drop it, or should our action classes remaining a more-or-less complete representation of the curtin schema? My inclination is slightly to the latter. Perhaps we should fiddle things so that grub_device: false doesn't end up in the config though.

> I also noticed that btrfs is excluded as a candidate for swapfiles. New enough kernels (e.g. disco+ hwe) do support swapfiles on btrfs. My plan was to fixup partman-swapfile & i guess curtin to start supporting them. Basically one marks them to be excluded from cow and preallocates blocks to basically have a bit of btrfs volume that is nor realy btrfs and usable for swap. Kind of making a small stylish tear in the btrfs fabric, like the pre-hipster ripped jeans from the 00s.

OK, I didn't know any of this :) Is there better way of checking if this is possible that checking os.uname().release? (Obviously a separate change...)

mwhudson:
Tested in a VM with both BIOS and UEFI boot, good enough for me. Other bits can be fixed up later.

PR #479, opened 2019-05-08
Have the real /run in-target.

xnox:
There is a lot of stuff /run that is relevant to how the system is setup. For example systemd resolved, udevd, networkd states among many other things.

Mounting a clean /run in /target wipes all of that. And things may not operate correctly. For example I see grub-probe -> vgs hanging in /target since udevd is not running in /target and /target/run lacks cached data.

Thus we should be bind-mounting /run in target, and that's what normally all the other installers do.

If we find (or want to prevent) /target stuff affecting the live system, we can bind-mound /run and setup an overlay on top of it. But I don't think we actually need/want that.


xnox:
With this patch in, eoan lvm&luks install no longer hangs for me.

raharper:
Oh, interesting.  I didn't realize subiquity was making a tmpfs /run in the target.

xnox:
@raharper 

this is fairly new, a way to expose /cdrom mounted in /target, and fiddle with in-target sources.list in a non-persistent manner. And i never liked how it was implemented, but now I have a bug i can point this at.

mwhudson:
curtin's ChrootableTarget does not do anything with /run so if this is needed here it's probably needed there too.  It might be better to change this code to put a tmpfs over /tmp and do stuff there instead?

That said, I'm not sure that this is a good idea in general? You wouldn't want the target system to access the host's e.g. snapd.socket would you? Perhaps the issue is that this code makes /run appear to be empty, whereas it actually has content in the ISO and maybe that content is enough to make lvm happy?

raharper:
We worked around this on Centos, but I still consider this to a be a bug in LVM tools, see this comment for what it was hanging on:

https://bugs.launchpad.net/curtin/+bug/1815666/comments/7

Reading udev database data IIUC.

We could make the change for all distros rather than just centos; but it just seems like an LVM bug; previous LVM tool releases do not require this.  I didn't push on the Centos path as we're unlikely to get a Centos LVM fix, however, we can address this in Ubuntu's LVM.

Have we opened a bug on LVM2 yet?

mwhudson:
I've pushed a slightly different version of this. The argument about whether curtin should bind mount /run by default certainly needs to happen but this will unbreak things for now.

I will observe that ubiquity has been bind mounting /run into the target for years (since 2012).

xnox:
@raharper 

lvm tools dropped lvmetada daemon and require udev database now. Thus on eoan and up, a uptodate udev database is required. Why do think it's a bug to link against libudev and use that to query things a bug, rather than a feature? I cannot pin-point that in the NEWS file. And it's normal for libudev to block until stuff about a device appears / finally gets processed by udev. It is abnormal to use lvm tooling with access to /dev but not udevd.

IMHO it is an improvement that lvmetdata daemon is gone for good, and udev database is used to get uptodate state of things.

We can troll through the lvm2 merge diff and upstream, but i think it is fuitile.

Also, e.g. if curtin were to bind-mount /run, lots of other stuff starts "to just work" like /etc/resolv.conf with either resolvconf or resolved. Without fiddling anything.

xnox:
This is now merged / done, in subiquity.

mwhudson:
When I say I've pushed a different version, I meant to this PR :) Nothing has hit master yet

xnox:
@mwhudson ah true... but also, how did you do that?! cause i thought this pull request is from my personal repo. Anyway, let me retest that now and roll with it.

xnox:
After fixing up snapcraft stuff, it builds and installs correctly. Landing this, such that i can review/test the other PRs.

PR #481, opened 2019-05-10
Add generic clear and delete methods to FilesystemController

mwhudson:
The guided "use a whole disk" flow in reusing existing partitions is
going to need to be able to generically destroy the existing users of
a disk, something that subiquity does not currently let you do (you
can't delete a RAID if is part of another RAID, for example). So add
generic ways of clearing and deleting objects.


xnox:
That's nice. Whilst testing this discovered a pre-existing bug https://bugs.launchpad.net/subiquity/+bug/1828661

PR #482, opened 2019-05-10
delete dm_crypts when deleting an encrypted vg

mwhudson:
This includes #474 just to get some test machinery.


xnox:
oh gosh, yeah that's a great bug catch. The behaviour of "stuck" PV in Used devices upon deleting encrypted vg is aweful, and crashes.

And I like the fix too.

PR #483, opened 2019-05-12
fix some bugs in editing and displaying swap partitions/devices

mwhudson:
https://bugs.launchpad.net/subiquity/+bug/1828661


mwhudson:
Eh this is small enough and obviously correct enough that I'm just going to merge it

xnox:
I think i tested this on the train as well. And it was a +1.

PR #484, opened 2019-05-13
canonicalize ptable, size, raidlevel values for filesystem actions

mwhudson:
With bonus mostly cosmetic shuffling.


mwhudson:
Review timed out, merging (if this is broken it should only be in a simple way)

PR #485, opened 2019-05-14
Drop maas, add drop to shell.

xnox:
- remove maas "nobody uses it, just drop it"
- add drop to shell "that is useful"
- move the installpath after network & proxy setup, and do proxy setup in the live system, for convenience.

This way we effectively have the tasteful recovery story done, as we can boot, set language, network, proxy, (not mirror though) and exit to shell.


subiquity/controllers/installpath.py
```diff
@@ -58,40 +60,15 @@ def choose_path(self, path):
         self.model.path = path
         getattr(self, 'install_' + path)()
 
+    def install_execshell(self):
+        if not self.opts.dry_run:
+            utils.disable_subiquity()
+        self.signal.emit_signal('quit')
```
vorlonofportland: ??  This looks like it exits subiquity.  Surely we expect 'drop to shell' to be a subshell like in d-i, that when you exit takes you back to subiquity.  Or is this doing something else that's not obvious from the code?

subiquity/models/installpath.py
```diff
@@ -39,9 +39,8 @@ def paths(self):
         if self.cmdline_sources:
             cmdline = [(_('Install from cli provided sources'), 'cmdline')]
         return cmdline + [
-            (_('Install Ubuntu'),                         'ubuntu'),
-            (_('Install MAAS bare-metal cloud (region)'), 'maas_region'),
-            (_('Install MAAS bare-metal cloud (rack)'),   'maas_rack'),
+            (_('Install Ubuntu'), 'ubuntu'),
+            (_('Exit To Shell'), 'execshell'),
```
vorlonofportland: Right, I would expect like d-i 'Interactive shell' or even the feature's name, 'Drop to shell', and *not* "Exit to shell".

subiquity/controllers/installpath.py
xnox: utils.disable_subiquity() will never retrun, as it stops the unit and starts the getty units. Please see https://github.com/CanonicalLtd/subiquity/blob/f9b9414de1b558daceb2bf9e46552b765f0e8fdc/subiquitycore/utils.py#L170

The quit, is for the dry run mode..... maybe i should invert the code. This is copy & paste from the installprogress crash handler, which already had the 'Exit to shell' button.

subiquity/models/installpath.py
xnox: Exit to shell is existing name / translation available in the installprogress.

There is currently no sensible return-to-subiquity, as we need to start the right subiquity unit..... plus currently we leave the stamp file around to launch the full shell.

Return to subiquity is to be discussed on how to do properly.

I have been pondering if we should `os.execl('/bin/bash', '/bin/bash', '-i')` or whatever. Such that 'ctrl-d' from shell restarts subiquity back where it was.

xnox: @vorlonofportland thoughts?

vorlonofportland: If we are going to provide a 'Run a shell' option in a non-error path, then we definitely need to have "sensible return-to-subiquity" as part of it.

The os.execl() seems like it could be a clever way to achieve this very simply and without having to worry about subiquity stealing any input from the shell, relying on systemd to restart the unit on exit.

mwhudson:
> please separate out the unrelated changes into separate commits. A single commit that adds interactive shell and drops maas is not a logical changeset.

+lots :)

subiquity/models/installpath.py
mwhudson: So my (and Claudio's) thinking for "drop to shell" was that we would have, on every screen, a button in the extreme bottom right with a label like "[ drop to shell ]" that you could tab to[1] that would drop you into a shell and return to exactly where you were where you were when the shell exits, to the point of any unsaved form input still being there, focus still being on the drop to shell button, etc. I'm not 100% sure how to achieve this in terms of terminal modes and file descriptors and so on but I think it's the experience we want so I think it's worth trying for. execl-ing bash and having systemd restart the unit when the shell exits would be a cute hack that we can fall back on if necessary.

[1] I guess you probably can't tab to it when a dialog is open, I think that's OK though

xnox: ack

PR #486, opened 2019-05-14
Drop maas

xnox:
just drop maas, without changing proxy/exec-shell etc.


mwhudson:
Fine as far as it goes, if there is agreement/direction to do this.

Is there any point in retaining the installpath model/view/controller at all? Are other "paths" going to appear, do we think? (FWIW, the "command line source" thing can probably go too). If we think other paths might reappear, we can just drop the Installpath the list of controllers in subiquity/core.py for now.

xnox:
Good question. Looking at the very early installpath it has this:

```
-        (_('Install Ubuntu'),             'installpath:install-ubuntu'),
-        # ('Install MAAS Region Server',  'installpath:maas-region-server'),
-        # ('Install MAAS Cluster Server', 'installpath:maas-cluster-server'),
-        # ('Test installation media',     'installpath:test-media'),
-        # ('Test machine memory',         'installpath:test-memory')
```

Of which MAAS is dead, and the last two only make sense to be done as bootloader options (ie. uefi apps).

I wonder if the cmdline thing is broken anyway, since kernel got moved out of the filesystem.squashfs.....
And I don't think the functionality it was meant to be enabled was ever used.

So yeah, maybe we should kill it out of the controller..... This will break snap UI view again, but i guess now it can just use a hardcoded path then.

xnox:
"if there is agreement/direction to do this." it will become a snap, and be visible on the snaps page (when ready). And those who don't know maas, unlikely know how to answer the questions asked and once installed would be not able to use maas out of the box offline as they will not know how to login into either web UI, or via the cmdline, and they will not have maas images to deploy. So, it's not a good advertisement platform anymore, nor useful for the offline case. So yeah "drop it".

mwhudson:
Oh yeah, I guess the follow up to this is to hack a bunch of stuff out of livecd-rootfs?

xnox:
Yes ubuntu-cd, debian-cd and livecd-rootfs. To stop building maas squash, downloading it, installing it.

vorlonofportland:
On Tue, May 14, 2019 at 09:28:34AM +0000, Michael Hudson-Doyle wrote:
> Fine as far as it goes, if there is agreement/direction to do this.

> Is there any point in retaining the installpath model/view/controller at
> all?  Are other "paths" going to appear, do we think?

Is this the code we would use for supporting installation of other flavors
from a common installer image?  (Say if, in principle, we wanted to support
selecting kubuntu as a flavor to install, which would go download a signed
squashfs from the archive instead of using the one on the install image; or
Ubuntu Desktop's support for minimal installs vs full desktop)

mwhudson:
> Is this the code we would use for supporting installation of other flavors
> from a common installer image?  (Say if, in principle, we wanted to support
> selecting kubuntu as a flavor to install, which would go download a signed
> squashfs from the archive instead of using the one on the install image; or
> Ubuntu Desktop's support for minimal installs vs full desktop)

Er. I guess so? We can always get it back out of git if needed, given I merged this...

PR #487, opened 2019-05-17
 allow /boot to be on any kind of device

mwhudson:
This includes a workaround for the issue causing RAID installs to fail (which will make LVM installs fail), so should not be landed as is.

I also had got the impression that when /boot was on RAID, grub would always show its menu. That didn't happen in my test.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1210,20 +1210,9 @@ class FilesystemModel(object):
     def is_root_mounted(self):
         return self._mount_for_path('/') is not None
 
-    def is_slash_boot_on_local_disk(self):
-        for path in '/boot', '/':
-            mount = self._mount_for_path(path)
-            if mount is not None:
-                dev = mount.device.volume
-                return (
-                    isinstance(dev, Partition)
-                    and isinstance(dev.device, Disk))
-        return False
-
     def can_install(self):
         return (self.is_root_mounted()
-                and not self.needs_bootloader_partition()
-                and self.is_slash_boot_on_local_disk())
+                and not self.needs_bootloader_partition())
 
     def _should_add_swapfile(self):
         mount = self._mount_for_path('/')
--- a/subiquity/ui/mount.py
+++ b/subiquity/ui/mount.py
@@ -46,13 +46,11 @@ LEAVE_UNMOUNTED = object()
 
 
 class MountSelector(WidgetWrap):
-    def __init__(self, mountpoints, ok_for_slash_boot):
+    def __init__(self, mountpoints):
         opts = []
         first_opt = None
         for i, mnt in enumerate(common_mountpoints):
-            if not ok_for_slash_boot and mnt == "/boot":
-                opts.append((mnt, False))
-            elif mnt not in mountpoints:
+            if mnt not in mountpoints:
                 if first_opt is None:
                     first_opt = i
                 opts.append((mnt, True, mnt))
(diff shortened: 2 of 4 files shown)


vorlonofportland:
> I also had got the impression that when /boot was on RAID, grub would always
> show its menu. That didn't happen in my test.

The menu is only force displayed when:
 - grub does not have write access to the backing store of /boot/grub/grubenv, and
 - the grub platform does not have a reliable way for the user to interrupt the boot sequence (i.e.: you're on UEFI not BIOS).

Were you testing on UEFI?

mwhudson:
Ah no, I was testing bios. Can test uefi on Monday.

mwhudson:
> Ah no, I was testing bios. Can test uefi on Monday.

OK, tested UEFI and it works as advertised. Can we really assume that the user has a reliable way to interrupt BIOS boot on a server though?

xnox:
> > Ah no, I was testing bios. Can test uefi on Monday.
> 
> OK, tested UEFI and it works as advertised. Can we really assume that the user has a reliable way to interrupt BIOS boot on a server though?

I believe yes, as there is no such thing as "non-interruptive UEFI fastboot" there.

xnox:
raid1 test now passes in eoan, thanks to fixed up mdadm.
the curtin stuff should be merged & released by curtin too.
I can work on SRUing the mdadm change.

xnox:
Well, this not only allows /boot to be on any device, but / too.
And i think this introduces subtle bugs which was discovered (and still present in d-i). Specifically /boot can be on LVM, as long as it's single/linear device. If it's a multi device lvm, or stripped lvm, grub fails to boot that.

I am concerned about restricting the user too much; versus exposing the user to known unusable configurations.

mwhudson:
> Well, this not only allows /boot to be on any device, but / too.
> And i think this introduces subtle bugs which was discovered (and still present in d-i). Specifically /boot can be on LVM, as long as it's single/linear device. If it's a multi device lvm, or stripped lvm, grub fails to boot that.

Oh ffs. curtin does not allow any configuration of striped/linear fwiw, but I don't know what the default is.

> I am concerned about restricting the user too much; versus exposing the user to known unusable configurations.

Well yes. OTOH, I don't think people have actually been asking for /boot on LVM.

vorlonofportland:
> Well, this not only allows /boot to be on any device, but / too.

Sure, and / should always be less restrictive than /boot.

> And i think this introduces subtle bugs which was discovered (and still present in d-i). Specifically /boot can be on LVM, as long as it's single/linear device. If it's a multi device lvm, or stripped lvm, grub fails to boot that.

I'm unaware of such issues and the update-grub code definitely does not distinguish.  We should certainly not be imposing any restrictions on /boot that are not surfaced by grub itself.

And this is such a corner case, because most users of striping across PVs are going to do this via RAID instead of LVM, and a VG that contains multiple linear PVs is incredibly unlikely to have /boot crossing device boundaries, so even if there IS an issue here it seems very unlikely to ever be hit.  So let's not complicate this more than necessary before we definitively know we need to deal with it.

vorlonofportland:
> > Ah no, I was testing bios. Can test uefi on Monday.
> 
> OK, tested UEFI and it works as advertised. Can we really assume that the user has a reliable way to interrupt BIOS boot on a server though?

As long as you have a keyboard, you are guaranteed to be able to interrupt the boot by holding down the shift key.

If you only have serial console, there's no guaranteed bios interrupt; but that's such an unusual case for BIOS systems, you generally have either a physical console or an iLO.

mwhudson:
OK so I'm going to remove the "use my curtin branch" commit and merge this. When I get back from my bike ride.

PR #490, opened 2019-05-21
summarize device usage in delete, guided views

mwhudson:


mwhudson:
Going to take a slightly different tack.

PR #495, opened 2019-05-28
make the main filesystem view more compact

mwhudson:
I showed some people from the design team the summary of a disk's usage  I made in #490 and they said "why doesn't the main filesystem view look like that?". So now it does.

Before:

Screenshot from 2019-05-28 13-56-38

After:

Screenshot from 2019-05-28 13-55-47

This also includes summaries of the device usage in the delete and guided views, as in #490. It also includes some slightly hairy table rendering tweaks but as penance for that I did write some tests for the table layout algorithm at last.


xnox:
Less zig-zag indents is always better!

subiquity/ui/views/filesystem/filesystem.py
```diff
@@ -378,106 +378,44 @@ def refresh_model_inputs(self):
...
+                    menu = Text("")
+                if obj is device:
+                    start, end = '[', ']'
+                else:
+                    start, end = '', ''
+                cells = [Text(start)] + cells + [menu, Text(end)]
+                if obj is not None:
+                    rows.append(make_action_menu_row(cells, menu))
+                else:
+                    rows.append(TableRow(cells))
```
xnox: Style wise `if obj is not None:` is redundand. One can simply use `if obj:`. Also, I normally preffer avoiding to use the else clause, and instead initialize things.... ie.

```
menu = Text("")
start, end = '', ''
if obj:
    menu = self.....
if obj is device:
    start, end = '[', ']'
```

But that's just style knitpicks. Functionality looks fine.

xnox:
I'd want to capitalise (MD) in the create software RAID button.

I'd want to align the 'No disks or partitions mounted' with the 'Devices'/'unused' grey subtable titles, instead of the `[` button boundary. So pad/notch it by 2 characters to the right. Similarly the 'No used devices' under the USED DEVICES section. That way there is less jiggery between e.g. make bootable & reset

Also I do wonder if we need to indent everything by two from the Headings. Or if we can win 2 characters by shifting everything to be in line with the titles. After all, our '[ ' provide the indent already.

In manual partitioning, after adding a `/boot` on a Crucial drive, I cannot make it "unbootable" i.e. impossible to delete ESP partition. or the `/boot/efi` filesystem.... However, clicking 'reset' did work. So i guess it's ok. Also probably a pre-existing bug/UI.

Looks hot! I think I will propose the further changes I mentioned and will open a bug report about making something "unbootable".

mwhudson:
Thanks for your review/merge!

I agree with your comments about the padding/margins, most of that should go away now I think (I haven't thought about the "outermost" layout of this screen for a while I guess!). Should be easy enough.

I see you found the existing bug about not being able to undo the boot partition setup. I guess my position has softened a bit, maybe it's more consistent to be able to delete the partition. But also don't really care a lot.

PR #500, opened 2019-05-30
enable reuse of existing partitions

mwhudson:
Well here it is. There is a lot here, obviously, but hopefully it is possible to review it commit-by-commit. This includes all the commits in #495 so review that first if it hasn't been merged yet.

I've built a snap and pushed it to the "edge/existing-partitions" channel. I've done some light testing and it seems to work!?


mwhudson:
Oh a tip for a reviewer who wants to play around in dry run mode is to install curtin from [ppa:mwhudson/devirt](https://launchpad.net/~mwhudson/+archive/ubuntu/devirt/+packages) until Ryan's curtin change lands and is SRUed everywhere.

subiquity/controllers/filesystem.py
```diff
@@ -273,15 +285,21 @@ def delete_mount(self, mount):
 
     def create_filesystem(self, volume, spec):
         if spec['fstype'] is None:
-            return
-        fs = self.model.add_filesystem(volume, spec['fstype'])
+            fs = volume.original_fs()
+            if fs is None:
+                return
+            self.model.readd_filesystem(fs)
```
raharper: read_filesystem? re_add_filesystem? looks like a typo;  maybe update_filesystem()?

subiquity/controllers/filesystem.py
```diff
@@ -391,18 +409,30 @@ def clear(self, obj):
         for subobj in obj.fs(), obj.constructed_device():
             self.delete(subobj)
 
+    def reformat(self, disk):
```
raharper: reformat operates on blockdevices, hence the .type check for "disk" UUIC.  Would it be useful to change the parameter here to blockdev or something more generic so the logic for type check becomes more clear?

subiquity/controllers/filesystem.py
```diff
@@ -478,21 +510,48 @@ def volgroup_handler(self, existing, spec):
...
-                largest_part = max(
-                    new_boot_disk.partitions(), key=lambda p: p.size)
-                largest_part.size -= (
-                    boot_partition.size - new_boot_disk.free_for_partitions)
-        self._create_boot_partition(new_boot_disk)
+            if boot_partition.preserve:
+                if self.model.bootloader == Bootloader.PREP:
+                    boot_partition.wipe = None
+                elif self.model.bootloader == Bootloader.UEFI:
+                    self.delete_mount(boot_partition.fs().mount())
```
raharper: Why does this need removing?  Do you create a new mount entry?

subiquity/models/filesystem.py
```diff
@@ -602,7 +690,21 @@ def supported_actions(self):
         return actions
 
     _can_INFO = True
-    _can_PARTITION = property(lambda self: self.free_for_partitions > 0)
+
+    @property
+    def _can_REFORMAT(self):
```
raharper: reformat is only for reusing existing partitions?

subiquity/models/filesystem.py
```diff
@@ -976,14 +1110,77 @@ def _probe_bootloader(self):
...
+        byid = {}
+        objs = []
+        exclusions = set()
+        for action in config:
+            if action['type'] == 'mount':
+                exclusions.add(byid[action['device']])
+                continue
+            c = _type_to_cls.get(action['type'], None)
+            if c is None:
+                # Ignore any action we do not know how to process yet
```
raharper: I wonder what to do when a system has a bcache then?  The backing and caching devices will show up a disks; if they are "preserved" or just ignored (not used by the install path), then the bcache could get in the way, either during the install or after reboot.

subiquity/models/filesystem.py
```diff
@@ -976,14 +1110,77 @@ def _probe_bootloader(self):
...
+        if self._probe_data is not None:
+            config = storage_config.extract_storage_config(self._probe_data)
+            self._actions = self._actions_from_config(
+                config["storage"]["config"],
+                self._probe_data['blockdev'])
+        else:
+            self._actions = []
         self.grub_install_device = None
 
+    def _actions_from_config(self, config, blockdevs):
```
raharper: This could use a comment on how config entries are mapped to actions and what exclusions imply.

subiquity/models/filesystem.py
```diff
@@ -976,14 +1110,77 @@ def _probe_bootloader(self):
...
+            log.debug("exclusions %s", {e.id for e in exclusions})
+            next_exclusions = exclusions.copy()
+            for e in exclusions:
+                next_exclusions.update(itertools.chain(
+                    dependencies(e), reverse_dependencies(e)))
+            if len(exclusions) == len(next_exclusions):
+                break
+            exclusions = next_exclusions
+
+        objs = [o for o in objs if o not in exclusions]
```
raharper: if objs was a set, then this could be:

objs = objs.difference(exclusions)

subiquity/controllers/filesystem.py
mwhudson: Well this is related to the comment thread in that google doc: what does reformat mean for things that are not disks? Currently this branch just wimps out and only defines reformat for disks so the type check is in that sense redundant.

subiquity/controllers/filesystem.py
mwhudson: re-add is definitely the phrase i want to use here (we're putting back a filesystem that has previously been removed. I'll change it to re_add_filesystem though.

subiquity/controllers/filesystem.py
mwhudson: Yes, the scenario here is: there is a disk with an existing ESP. It has been selected to be active, but now the user is making some other ESP the active one. It needs to be unmounted before we create the mount for the new one.

subiquity/models/filesystem.py
mwhudson: I guess there would be no harm in enabling in other cases, indeed. Changed.

subiquity/models/filesystem.py
mwhudson: It's certainly not so great to show a device as "unused" when probert and curtin know it's being used in a bcache. That said, if they are used in the install, curtin will know how to dismantle the bcache. If they are not used in the install, then yes, they will still hang around, but I can't really see how that could be a problem?

I'm not sure there is really scope to do anything different here.

mwhudson: I did realize thinking about this that I also need to ignore actions whose dependencies have been ignored, e.g. the filesystem of a bcache.

subiquity/models/filesystem.py
mwhudson: Added.

subiquity/models/filesystem.py
mwhudson: But (and this is a magnificent example of implicit tight coupling) the order returned actually matters! The devices are presented in the filesystem view in the reverse of the order they appear in _actions, which means that e.g. a RAID appears higher up the list than the disks is is composed of. I think this is quite important as it makes "unpeeling" existing compound structures easy, you just delete the top device until you only have disks left. I've added some comments about this.

mwhudson:
Thanks for the comments. I've made some changes -- so you can see what I've changed I've pushed them as fixup! commits but before merging someone should run rebase --interactive --autosquash to preserve some kind of clean history.

subiquity/controllers/filesystem.py
xnox: we used to ignore user choices like that, and force reuse ESP. or at least that was my experience in like ubiquity.

subiquity/models/filesystem.py
raharper: Yes, I think this is one of my biggest fears;  how do we inform the users *which* devices are being destroyed/cleared.    I somewhat worry that a raid or bcache or some other multi-device element is being selected/used and then curtin will *ensure* that it can use the device, which may require dismantling an associated raid or lvm or bcache.

subiquity/controllers/filesystem.py
mwhudson: Well OK, are you saying we should change here? I think the two cases here are: installing to a partition of a disk which already has an ESP, in which case the default is to reuse it, or installing to a separate disk (e.g. a system with a "windows disk" and a "linux disk") in which case creating a new ESP seems a better default to me. The user can then change this (that's what this code is about) but presumably approximately 0% of users will do that.

subiquity/models/filesystem.py
mwhudson: Well for a compound device subiquity knows about, we make it clear in the UI that it has to be torn down before the disk can be reused, so I think that's fine. For compound devices subiquity doesn't know about (e.g. bcache and zpools but probably some other things too) I'm not sure what we can do. I guess we could add very minimal support for subiquity to display that the disk or partition is a bcache member or used by a zpool. (It would be nice if linux block devices had even _heard_ of the _concept_ of a regular interface here!)

mwhudson:
I've just force pushed to remove the commit that switched to my own curtin branch.

PR #502, opened 2019-06-25
Lots of new things.

xnox:
Update debian packaging to be buildable.

Set magic langpack variable, such that translations are stripped into the ubuntu langpack. That would enable in-ubuntu translations of subiquity.


xnox:
I want to merge & dput this into eoan.

mwhudson:
This looks fine, apart from the commit message which is (obviously?) in no way fine :)

xnox:
```
   * Upgrade consoleconf in Ubuntu archive with i18n support and many other
     improvements since the last archive upload.
```

PR #505, opened 2019-07-12
always wipe a disk when clearing its partition table

mwhudson:
@raharper is any reason to favour on of "wipe: superbloc" or "wipe: superblock-recursive" over the other?


mwhudson:
This definitely helps some cases so I'm merging. Post hoc review welcome.

xnox:
wiping harder is always better.

raharper:
So, wipe: superblock will wipe only the superblock areas, which are 1MB at front and end of the device.  Recursive will the same but for all existing partitions and then the device it self.  If you are changing the partition table, I agree that wipe: superblock-recursive is the best; this prevents "buried" data from showing up later.

PR #506, opened 2019-07-12
relax "don't put all disks in a RAID" check for existing partition tables

mwhudson:
Just a fairly minor thing I noticed playing around.

### DIFF
--- a/subiquity/ui/views/filesystem/compound.py
+++ b/subiquity/ui/views/filesystem/compound.py
@@ -196,9 +196,23 @@ class CompoundDiskForm(Form):
 
     def validate_devices(self):
         if self.model.needs_bootloader_partition():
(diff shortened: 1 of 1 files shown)


xnox:
```
subiquity/ui/views/filesystem/compound.py:209:33: E712 comparison to True should be 'if cond is True:' or 'if cond:'
```

eoan has weird zfs stuff going on.... imho package installation should not try to load modules.... or like do it only when not containarized.

mwhudson:
Fixed the lint. And yes saw the eoan crazy, didn't even have the energy to file a bug though.

xnox:
@mwhudson filed https://bugs.launchpad.net/ubuntu/+source/zfs-linux/+bug/1836902

we can merge this without eoan passing. And/or disable eoan travis for now.

PR #515, opened 2019-08-05
make sure unattended-upgrades always runs

mwhudson:
for https://bugs.launchpad.net/subiquity/+bug/1838520

### DIFF
--- a/subiquity/controllers/installprogress.py
+++ b/subiquity/controllers/installprogress.py
@@ -22,6 +22,7 @@ import signal
 import subprocess
 import sys
 import platform
+import tempfile
 import time
 import traceback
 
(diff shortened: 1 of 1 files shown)


mwhudson:
cc @paride -- not sure how to test this, i don't have a laptop i want to install currently...

subiquity/controllers/installprogress.py
```diff
@@ -567,3 +574,11 @@ def default(self):
             footer = _('Please report this error in Launchpad')
         self.ui.set_body(self.progress_view)
         self.ui.set_footer(footer)
+
+
+uu_apt_conf = """\
+# Config for the unattended-upgrades run to avoid failing on battery power or
+# a metered connection.
+Unattended-Upgrade::OnlyOnACPower "false";
+Unattended-Upgrade::Skip-Updates-On-Metered-Connections "true"
```
paride: missing semicolon here?

paride:
@mwhudson I'm back from holiday and I have a laptop where I can test the installation on battery, but I'm unsure on what the best way to test the installer with this PR merged in is. The only thing I can think of is updating the subiquity snap at install time from an experimental channel with the PR merged, but this requires it to be published first. What do you suggest?

mwhudson:
You can use the `./scripts/test-this-branch.sh` script I think/hope?

paride:
That's convenient! I had to fix the ISO filename in the script (PR coming) and run it from Bionic (it didn't work from Disco). I could install the generated ISO on battery without any problem. Thanks!

mwhudson:
Thanks for testing!

PR #516, opened 2019-08-07
clean up how data is passed around at the centre of subiquity

mwhudson:
Get rid of that 'common' dictionary that had just a bag of stuff in it.

No change in behaviour (hopefully!)

### DIFF
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -240,7 +240,7 @@ class IdentityController(BaseController):
                 login_details_path = '/run/console-conf/login-details.txt'
                 self.model.add_user(result)
         ips = []
-        net_model = self.controllers['Network'].model
+        net_model = self.app.controller_instances['Network'].model
         for dev in net_model.get_all_netdevs():
             ips.extend(dev.actual_global_ip_addresses)
         with open(login_details_path, 'w') as fp:
@@ -258,7 +258,7 @@ class IdentityController(BaseController):
         self.ui.set_header(title)
         self.ui.set_footer(footer)
 
-        net_model = self.controllers['Network'].model
+        net_model = self.app.controller_instances['Network'].model
         ifaces = net_model.get_all_netdevs()
         login_view = LoginView(self.opts, self.model, self, ifaces)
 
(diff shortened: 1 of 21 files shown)


console_conf/controllers/identity.py
```diff
@@ -194,9 +194,9 @@ def write_login_details_standalone():
 
 class IdentityController(BaseController):
 
-    def __init__(self, common):
-        super().__init__(common)
-        self.model = self.base_model.identity
+    def __init__(self, app):
+        super().__init__(app)
+        self.model = app.base_model.identity
```
xnox: This is change of behaviour =) self -> common (now app)

subiquitycore/core.py
```diff
@@ -452,62 +443,66 @@ def next(loop, user_data):
...
+            palette = self.STYLES_MONO
+            screen = urwid.raw_display.Screen()
+        else:
+            screen, palette = setup_screen(self.COLORS, self.STYLES)
 
-            self.common['loop'] = urwid.MainLoop(
-                self.common['ui'], palette=palette, screen=screen,
-                handle_mouse=False, pop_ups=True,
-                input_filter=self.common['input_filter'].filter)
+        self.loop = urwid.MainLoop(
```
xnox: does this have a chance to re-init loop? i.e. if run() is called multiple times, it would create new loops now. Previously it didn't.

Maybe add a test for "if not self.loop" or some such?

I guess, run shouldn't be called multiple times anyway, and it's a bug to do so.

console_conf/controllers/identity.py
mwhudson: Ehhhh

subiquitycore/core.py
mwhudson: Well before "hasattr(self, 'loop'):" was always false before: until this branch nothing ever set self.loop! So I just deleted the check. It could be "self.loop is not None" instead, but ehh. Calling run twice would be an awful bug.

mwhudson:
Thanks for the review.

console_conf/controllers/identity.py
xnox: that was a joke =)

PR #521, opened 2019-08-13
allow real name to be empty

mwhudson:
for https://bugs.launchpad.net/subiquity/+bug/1786381

### DIFF
--- a/subiquity/models/identity.py
+++ b/subiquity/models/identity.py
@@ -41,6 +41,8 @@ class IdentityModel(object):
     def add_user(self, result):
         result = result.copy()
         self._hostname = result.pop('hostname')
+        if not result.get('realname'):
+            result['realname'] = result['username']
         self._user = User(**result)
 
     @property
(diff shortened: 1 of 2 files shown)


xnox:
As per man 5 passwd it seems to me that it is required, it's just loginname is used as realname...  also it seems to be doing that for all other accounts i can see in passwd.

So should we be auto-using loginame, or does it "just happen"?

mwhudson:
Hm, it doesn't seem to be required to me? In /etc/passwd I see several entries with no "comment", for example:

```
syslog:x:102:106::/home/syslog:/usr/sbin/nologin
messagebus:x:103:107::/nonexistent:/usr/sbin/nologin
_apt:x:104:65534::/nonexistent:/usr/sbin/nologin
lxd:x:105:65534::/var/lib/lxd/:/bin/false
uuidd:x:106:110::/run/uuidd:/usr/sbin/nologin
```

That said, having it default to the username is probably more friendly, let's do that.

PR #523, opened 2019-08-15
fix fd leak in run_in_bg

mwhudson:
Notices this while debugging something entirely different.

### DIFF
--- a/subiquitycore/core.py
+++ b/subiquitycore/core.py
@@ -308,6 +308,8 @@ class Application:
         fut = self.common['pool'].submit(func)
 
         def in_main_thread(ignored):
+            self.common['loop'].remove_watch_pipe(pipe)
+            os.close(pipe)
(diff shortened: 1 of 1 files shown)


xnox:
hm.... I'm confused.

in_random_thread we write to a pipe, nothing reads from a pipe, and now you want it closed.... wouldn't writing to a closed pipe fail from now on then?! Also if nothing is reading from the pipe.... why have it at all?

mwhudson:
A pipe created (by urwid, inside watch_pipe) for each invocation of run_in_bg. The watch_pipe docs (http://urwid.org/reference/main_loop.html#urwid.MainLoop.watch_pipe) say:

> If the callback returns False then the watch will be removed from event_loop and the read end of the pipe will be closed. You are responsible for closing the write end of the pipe with os.close(fd).

So that's what I'm doing.

PR #525, opened 2019-08-15
Make the outermost widget an instance of Pile rather than Frame

mwhudson:
Current plans involve putting a focusable widget (a help button) in the
footer, and Pile's behaviour is much more what we want than Frame for
this.


xnox:
Travis is crashing:

```
AttributeError: 'AttrMap' object has no attribute 'options'
The command "sudo ./scripts/test-in-lxd.sh "$IMAGE" "$SCRIPT"" exited with 1.
```
Otherwise looks ok.

I guess note to myself to retest if consoleconf still works... sigh...

mwhudson:
Yeah, pushed too soon. Have a new version. And console_conf still works in dry run mode at least...

PR #531, opened 2019-09-04
refactor some controller stuff

mwhudson:
Nothing very deep here. Rename BaseController.default to BaseController.start_ui.


xnox:
Looks good.

PR #532, opened 2019-09-04
Add guided encrypted partitioning option.

xnox:
https://asciinema.org/a/uOeYHMVkFmBtfl7c9YLUJ2LPm

Not sure about where to store the guided passphrase, hence it's tagged onto the controller. Note that it is cleared any time one goes back. As ultimately it's the property of the volume group inside the model.

Also not sure about the text on the "password" screen and the help text on the choose disk screen.

### DIFF
--- a/subiquity/ui/views/filesystem/guided.py
+++ b/subiquity/ui/views/filesystem/guided.py
@@ -76,8 +76,8 @@ class GuidedFilesystemView(BaseView):
             _("Use An Entire Disk And Set Up LVM"), on_press=self.guided,
             user_arg="lvm")
         luks = ok_btn(
-            _("Use An Entire Disk And Set Up LUKS encrypted LVM"), on_press=self.guided_crypt,
-            user_arg="luks")
+            _("Use An Entire Disk And Set Up LUKS encrypted LVM"),
+            on_press=self.guided_crypt, user_arg="luks")
         manual = ok_btn(_("Manual"), on_press=self.manual)
         back = back_btn(_("Back"), on_press=self.cancel)
         super().__init__(screen(
@@ -116,14 +116,15 @@ It can easily be enlarged with standard LVM command line tools."""),
 partitions on the selected disk: one as required by the bootloader, one \
 for '/boot', and one covering the rest of the disk.
 
-A LUKS encrypted LVM volume group is created containing the large partition. A \
-4 gigabyte logical volume is created for the root filesystem. \
+A LUKS encrypted LVM volume group is created containing the large partition. \
+A 4 gigabyte logical volume is created for the root filesystem. \
 It can easily be enlarged with standard LVM command line tools."""),
 }
 
 
 def _wrap_button_row(row):
     return CursorOverride(Color.done_button(row), 2)
+
 
 loss_warning = _("""Warning: If you lose this security key, all data will be lost. If \
 you need to, write down your key and keep it in a safe place \
@@ -149,6 +150,7 @@ class GuidedPasswordForm(Form):
         if self.password.value != self.confirm_password.value:
             return _("Security keys do not match")
 
+
 class GuidedPasswordView(BaseView):
     title = _("Choose a security key")
     footer = _("")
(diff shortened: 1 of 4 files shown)


xnox:
ascinema thing has old wording, and maybe i should stay on the new pages for longer....

Stealing wording from ubiquity.

mwhudson:
Thanks for working on this!

This makes me think of a dusty old bug about LVM not using the whole disk (https://bugs.launchpad.net/subiquity/+bug/1785321). The proposed fix there is to ask the user how much of the disk to put in the initial VG. The reason for bringing this up is to ask: should that be asked on the same screen as the passphrase for the encrypted case? Should there even just be one screen that asks how much of the disk to use, passphrase and which disk to use and keys whether to use encryption off a passphrase is entered? Just rambling really but interested to see what you think.

(This all reminds me that the guided options would be more useful if we resurrected my old hairy code to be more flexible about allowing objects to be edited even if they are used...)

mwhudson:
Code looks fine btw, modulo some slightly strange wrapping.

subiquity/ui/views/filesystem/guided.py
```diff
@@ -98,13 +111,72 @@ def cancel(self, btn=None):
...
+
+    def validate_confirm_passwor(self):
+        if self.password.value != self.confirm_password.value:
+            return _("Security keys do not match")
+
+class GuidedPasswordView(BaseView):
+    title = _("Choose a security key")
+    footer = _("")
+    excerpt = _("""Disk encryption protects your files in case you lose your \
+computer. It requires you to enter a security key each time the \
```
mwhudson: Wrapping seems odd here.

xnox:
```
>>> print("""long string \
... without leading chars.
... 
... and again long string \
...      but with leading indentation.
... """)
long string without leading chars.

and again long string      but with leading indentation.

```

xnox: i'm not sure how to wrap things, without introducing additional whitespace, whilst also wrapping to a reasonable source code line width.

xnox:
> Thanks for working on this!
> 
> This makes me think of a dusty old bug about LVM not using the whole disk (https://bugs.launchpad.net/subiquity/+bug/1785321). The proposed fix there is to ask the user how much of the disk to put in the initial VG. The reason for bringing this up is to ask: should that be asked on the same screen as the passphrase for the encrypted case? Should there even just be one screen that asks how much of the disk to use, passphrase and which disk to use and keys whether to use encryption off a passphrase is entered? Just rambling really but interested to see what you think.
> 
> (This all reminds me that the guided options would be more useful if we resurrected my old hairy code to be more flexible about allowing objects to be edited even if they are used...)

At the very least changing the default from 4GB to 80% would be a massive improvement.
I'm not sure if we should be asking this, or if it should be amendable on the next screen of manual partitioning / confirmation. Ie. you can change the size of the '/' lvm on the next screen.

vorlonofportland:
> At the very least changing the default from 4GB to 80% would be a massive improvement.

Agreed.  Sensible default is more important than configurability of the LV size during install; and the flow should give you the option of not thinking about the LV size during install, even if you make it configurable.  And 100% is NOT a sensible default, because if you're using LVM it's because you want to have space for snapshots or additional LVs and you can't do an online shrink of a filesystem so an LV using 100% is useless.  But on any modern disk, 80% means you have space that you could lift-and-shift the rootfs to a new LV to shrink it if you decide post-install that you want something different.

mwhudson:
OK, so I think the plan there should be to have a better default (and maybe make changing it afterwards easier). So then for this branch do we ask for the passphrase on the same screen as the disk selection? I guess probbbbbably not.

xnox:
> OK, so I think the plan there should be to have a better default (and maybe make changing it afterwards easier). So then for this branch do we ask for the passphrase on the same screen as the disk selection? I guess probbbbbably not.

So, I think we need design input for this. I have mimicked what is done in Ubiquity. They have one screen to pick the guide options (full disk install, dual-boot install, manual, etc), with tick boxes to "and use LVM" or like "and encrypt", with separate pages to do disk selection, and separate page for the password.

Merging the password and disk selection pages might be doable, but it might make the screen scrollable due to help text for the password and the disk selection......

xnox:
Things to fix.
* create form wrapper for radio button group
* use it
* restructure dialogs

I have started doing some of things things locally already, but not yet fully done.

xnox:
Started doing the new UI as per mpt request.... https://github.com/CanonicalLtd/subiquity/pull/588 and it's a bit hairy.

I wonder, if we can land rebased this code as is, (method+password+disk selection in three steps) and the proceed to rework the guided UI later (method+disk selection in a single step).

mwhudson:
This is all done in separate branches now.

PR #536, opened 2019-09-05
add a [ more... ] button in bottom right and the start of some content

mwhudson:
Ping @clagom 

Demo at https://asciinema.org/a/Mb6B459QQlVZKWR6O8zGfxXAF

Mostly interested in review of the machinery / approach at this point. If everyone likes that then we'll need to spend some time on the words! (Like not promising 5 years of updates when installing eoan...)

### DIFF
--- a/subiquity/core.py
+++ b/subiquity/core.py
@@ -17,6 +17,8 @@ import logging
 import os
 import platform
 
+import urwid
+
 from subiquitycore.core import Application
 
 from subiquity.models.subiquity import SubiquityModel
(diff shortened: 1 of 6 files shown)


mwhudson:
FWIW "drop to shell" after this is https://github.com/mwhudson/subiquity/commit/e3db1421fe834769f281bf0a308779d7c3f0cdca

clagom:
Hi @mwhudson I think this demo covers what we discussed during the call. I can see the fact that we don't always display the 'help on this screen' might be confusing, but I understand we don't have all the content for now.

mwhudson:
I've updated this branch per feedback via email.

mwhudson:
Video of new branch https://asciinema.org/a/SdydUuCJhh7VfSKosnPwgNTT3

mwhudson:
The design for this has changed, closing.

PR #540, opened 2019-09-25
remove footer, make header narrower

mwhudson:
In Paris mpt asked if we could use unicode half-blocks (▄ / ▀) to make the header a bit narrower. As, coincidentally, I'd added these characters to the console font subiquity bundles the week before, this was very easy, and looked good. This also allows us to remove the ever present blank line beneath the header, which gets us another line of usable vertical space, something that is at a bit of a premium in subiquity's smallest target screen of 80x24.

The only problem was that it now looked out of balance with the footer. The footer has been a bit of a sore point for a while: mpt has previously complained that progress bars in installers have been observed to cause confusion in users, and the messages we displayed in the footer were largely a waste of space. So, let's delete the footer entirely! The only useful messaging in the footer was on the filesystem screen, where we provided clues on what had to be done to proceed. So this branch does that more explicitly.

Video at: https://asciinema.org/a/I5av0jfAPxCkmOkwW1YupARm5

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -82,18 +82,9 @@ A subiquity screen consists of:
     2. a scrollable content area
     3. a stack of buttons, including "done"/"cancel" buttons for moving between
        screens
- 3. a footer
 
 The header has a summary line describing the current screen against an "ubuntu
 orange" background.
-
-The footer has a progress bar indicating how far through the install process
-the user was, a blank line and the summary area.  Currently the summary area
-contains static content for the first few screens, but there are vague plans to
-make it specific to the currently focused element. Once the install has started
-but before we get to the final screen, the summary area contains a summary of
-progress made by the installation.  Someday soon the summary area will also
-contain a button that allows you to drop to a shell at any point.
 
 The body area is where most of the action is. It follows a standard pattern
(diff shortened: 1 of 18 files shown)


matthewpaulthomas:
Thanks for the video! Looks good to me.

> mpt has previously complained that progress bars in installers have been observed to cause confusion in users

(For anyone confused by this later: The problem is not progress bars in installers generally, but with lack of visual distinction between (a) your progress through a task _e.g._ shopping checkout or Subiquity steps, (b) something you have to wait for _e.g._ launching a VM, and (c) something that happens in the background _e.g._ the file copying in Ubiquity.)

xnox:
> Thanks for the video! Looks good to me.
> 
> > mpt has previously complained that progress bars in installers have been observed to cause confusion in users
> 
> (For anyone confused by this later: The problem is not progress bars in installers generally, but with lack of visual distinction between (a) your progress through a task _e.g._ shopping checkout or Subiquity steps, (b) something you have to wait for _e.g._ launching a VM, and (c) something that happens in the background _e.g._ the file copying in Ubiquity.)

Yeah, we do not have enough visual aid to do "pulsating / spinner" steps versus something that is blocked on requiring input.

PR #541, opened 2019-09-29
Add help menu

mwhudson:
This adds a help menu to the header, currently containing general about this installer, help on the current screen and help on global shortcut keys.

I'm pretty happy with how this looks (https://asciinema.org/a/AYUYgKvaWPq7kIIJrYKxgEuc6) and the code is OK enough. There are some things I would like opinions on though:

 * The wording for the "about" dialog (I currently have separate content for LTS/non-LTS, maybe I should add content for devel series too? Totally happy to do whatever someone else says here ;-p)
 * More content for help on the current screen (this does not have to be in this branch of course, but we should populate this a bit more before the next release)
 * The behaviour of the help menu when looking at help content. Currently the help menu is disabled/non-focusable when help content is showing. Is that appropriate? Maybe it should still be active, have the current option disabled and selecting a different option shows that instead?

### DIFF
--- a/subiquity/core.py
+++ b/subiquity/core.py
@@ -94,3 +94,10 @@ class Subiquity(Application):
             lambda fut: (
                 fut.result(), self.signal.emit_signal('snapd-network-change')),
             )
+
+    def unhandled_input(self, key):
+        if key == 'f1':
+            if not self.showing_help:
+                self.ui.right_icon.open_pop_up()
+        else:
+            super().unhandled_input(key)
--- a/subiquity/core.py
+++ b/subiquity/core.py
@@ -33,6 +33,7 @@ log = logging.getLogger('subiquity.core')
 class Subiquity(Application):
 
     snapd_socket_path = '/run/snapd.socket'
+    showing_help = False
 
     from subiquity.palette import COLORS, STYLES, STYLES_MONO
 
--- a/subiquity/core.py
+++ b/subiquity/core.py
@@ -33,7 +33,6 @@ log = logging.getLogger('subiquity.core')
 class Subiquity(Application):
 
     snapd_socket_path = '/run/snapd.socket'
-    showing_help = False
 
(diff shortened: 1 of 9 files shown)


cyphermox:
What would be different for a devel release from a non-LTS? I wonder if it would be sufficient to mimick what ubiquity does for the welcome screen, and essentially just say "this is the installer for release X" for most cases; but no particular opinion. I think this is a question for design.

I think the best for help is to have it always focusable, so people can navigate between help screens without having to exit out of it and then display again. 

But yes, if you want to add more help content, please make that on a separate PR.

As for this PR, nothing crazy jumped out at me in the code; ACK on merging.

mwhudson:
> What would be different for a devel release from a non-LTS? I wonder if it would be sufficient to mimick what ubiquity does for the welcome screen, and essentially just say "this is the installer for release X" for most cases; but no particular opinion. I think this is a question for design.

One of the tag lines for Ubuntu server on the website is "five years of free updates", we shouldn't say that for a non-LTS release! But yes, I'll get some input from better qualified here.

> I think the best for help is to have it always focusable, so people can navigate between help screens without having to exit out of it and then display again.

Agreed, I've made this change now.

> But yes, if you want to add more help content, please make that on a separate PR.
> 
> As for this PR, nothing crazy jumped out at me in the code; ACK on merging.

Thanks. Will merge once travis has had its say.

subiquity/core.py
```diff
@@ -93,3 +93,10 @@ def _proxy_set(self):
             lambda fut: (
                 fut.result(), self.signal.emit_signal('snapd-network-change')),
             )
+
+    def unhandled_input(self, key):
+        if key == 'f1':
+            if not self.showing_help:
```
xnox: AttributeError: 'Subiquity' object has no attribute 'showing_help'

Doesn't seem to be ever declared, hence pressing F1 crashes the installer.

PR #545, opened 2019-10-01
add debug shell help menu item / shortcut

mwhudson:


subiquity/core.py
```diff
@@ -30,6 +31,18 @@
...
+
+This shell session is running inside the installer environment.  You
+will be returned to the installer when this shell is exited, for
+example by typing Control-D or 'exit'.
+
+Be aware that this is an ephemeral environment.  Changes to this
+environment will not survive a reboot. If the install has started, the
+installed system will be mounted at /target.""")
+
+
```
xnox: Can we do a confirm here? Ie.

```
-installed system will be mounted at /target.""")
+installed system will be mounted at /target.
+
+Press Enter to start shell...""")
```

subiquity/core.py
```diff
@@ -98,5 +111,12 @@ def unhandled_input(self, key):
...
             if not self.showing_help:
                 self.ui.right_icon.open_pop_up()
+        elif key == 'ctrl z':
+            self.debug_shell()
         else:
             super().unhandled_input(key)
+
+    def debug_shell(self):
+        self.run_command_in_foreground(
+            "clear && echo {} && bash".format(shlex.quote(DEBUG_SHELL_INTRO)),
```
xnox:
```
-            "clear && echo {} && bash".format(shlex.quote(DEBUG_SHELL_INTRO)),
+            "clear && read -p {} start && clear && bash".format(shlex.quote(DEBUG_SHELL_INTRO)),
```

This way, we wait for user to read text and hit enter.
Then clear the screen, then show the prompt.

On 80x24 this is a lot of intro text, that we would want to clear out.

subiquity/core.py
```diff
@@ -98,5 +111,12 @@ def unhandled_input(self, key):
         if key == 'f1':
             if not self.showing_help:
                 self.ui.right_icon.open_pop_up()
+        elif key == 'ctrl z':
```
xnox: `ctrl z` is undo in s390x vt220 ascii terminal (because it's running in a web-browser). Can we have either different or an alternative key binding for this?

Ie. in grub/efi everything is ctrl+ and Fn key. Can we possibly change this to in `('ctrl z', 'f2')` ?

subiquity/core.py
```diff
@@ -98,5 +111,12 @@ def unhandled_input(self, key):
...
+            self.debug_shell()
         else:
             super().unhandled_input(key)
+
+    def debug_shell(self):
+        self.run_command_in_foreground(
+            "clear && echo {} && bash".format(shlex.quote(DEBUG_SHELL_INTRO)),
+            shell=True)
```
xnox: Note to self, this pipe runs in `/bin/sh` to exec bash.

subiquity/core.py
xnox: (also fine with confirm and not clearing intro text)

subiquity/core.py
cyphermox: Why? Won't people have already hit some key to get to it? What they really need is just to acknowledge that popup.

subiquity/core.py
mwhudson: I'm not completely sure I agree with this... it's no larger an amount of text than the default motd is it?

subiquity/core.py
mwhudson: yes, adding f2 as well makes sense.

mwhudson:
> Having only ctrl-z to get to shell is pretty bad; that's without counting how bad Ctrl-Z is as a shortcut by itself.

Well you can also access it via the help menu of course.

> Why do we need this at all when you can switch VTs? 

Because subiquity doesn't always run in a vt?

> In any case, please change shortcut or add another that can be used on different systems where Ctrl-Z will otherwise interfere (as mentioned by xnox). Ctrl-<any F#> would be fine, and I suppose many other shortcuts if they can be caught/redirected on terminal (ctrl-q? ctrl-s?). I think Ctrl-T isn't otherwise mapped to anything?

I am happy for this bikeshed to be painted multiple colours. I think simply adding f2 as another shortcut works for me (Ctrl-F# can't be detected in the linux console it seems).

subiquity/core.py
mwhudson: Printing the message this way is a touch gross, we could also modify run_command_in_foreground to take a hook to run after we have disconnected from stdin or something. This works though.

xnox:
@cyphermox people connected via a serial console only, do not have a liberty of switching to a different tty. They only have a single connection/console, hence we need support for in-place dropping to shell.

Also, exposing UI for this is similar to d-i, which also has exit to shell menu option, and is discoverable. Switching tty's is not discoverable in this instance.

C+z was @vorlonofportland contribution to "background installer" and get "back to shell".

I think a Fn key, a Ctrl key, and visual UI menu item (arrow keys/enter) has all bases covered.

subiquity/core.py
xnox: @mwhudson 

Ok about the length of text.

But what about Enter to confirm the text first, prior to getting shell? Useful, or not really? (the `&& read -p` portion)

PR #548, opened 2019-10-03
keep dialog titles up to date when on screen

mwhudson:
Something I did doing error reporting that has no dependence on the other mounds of stuff.

### DIFF
--- a/subiquitycore/ui/stretchy.py
+++ b/subiquitycore/ui/stretchy.py
@@ -53,7 +53,7 @@ from subiquitycore.ui.container import ListBox, Pile
 
 class Stretchy(metaclass=urwid.MetaSignals):
 
-    signals = ['opened', 'closed']
+    signals = ['opened', 'closed', '_set_title']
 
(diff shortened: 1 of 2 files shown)


xnox:
Are titles out of sync at the moment? How do I trigger that bad behaviour with current code?

mwhudson:
There's no way to trigger this in current code.

The issue is that there's no way to change the title of a dialog once it's on screen. We don't ever change this currently, but I bumped into this viewing an error report: crash reports take a little while to load, so it's done in a background thread. I want to set the title of the crash report dialog to the summary of the report, but if you open the report before it's finished loading...

All very edge casey and possibly me overthinking things.

xnox:
Ok, so this adds APIs to update the title, and refactors things a bit to connect all signals together. I'm ok with merging this. However, might in the future appear redundant if nothing ever tries to update titles dynamically.

mwhudson:
Yes, and I'm not sure what I have here actually works any more :( Let's leave it until a use for this is at least close to being ready to merge.

mwhudson:

PR #550, opened 2019-10-03
Reprobe block devices on debug shell exit

mwhudson:
Forgot about this!

It's not maximally clever really -- maybe it would be better to reprobe on udev block events instead? And we could automatically reprobe until the user makes a configuration change vs showing the storage ui but well. Still an improvement I think.

### DIFF
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -18,6 +18,8 @@ import json
 import logging
 import os
 
+import urwid
+
 from subiquitycore.controller import BaseController
 
 from subiquity.models.filesystem import (
@@ -119,9 +121,16 @@ class FilesystemController(BaseController):
         self.answers.setdefault('guided-index', 0)
         self.answers.setdefault('manual', [])
         self._cur_probe = None
+        self.ui_shown = False
 
     def start(self):
+        urwid.connect_signal(
+            self.app, 'debug-shell-exited', self._maybe_reprobe_block)
         self._start_probe(restricted=False)
+
+    def _maybe_reprobe_block(self):
+        if not self.ui_shown:
+            self._start_probe(restricted=False)
 
     def _start_probe(self, *, restricted=False):
         self._cur_probe = Probe(self, restricted, 5.0, self._probe_done)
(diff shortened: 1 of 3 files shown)


subiquity/controllers/filesystem.py
```diff
@@ -65,72 +120,67 @@ def __init__(self, app):
...
-                                 ProbeState.REPROBING]:
+        if self._cur_probe.state == ProbeState.PROBING:
             self.ui.set_body(SlowProbing(self))
-        elif self._probe_state == ProbeState.FAILED:
+        elif self._cur_probe.state == ProbeState.FAILED:
             self.ui.set_body(ProbingFailed(self))
         else:
+            self.ui_shown = True
+            # Should display a message if self._cur_probe.restricted,
+            # i.e. full device probing failed.
```
cyphermox: As above... Do we have a policy for how to deal with this? Comments in general and what constitutes useful comments; when they have a chance of being addressed assuming they're potential TODOs?

subiquity/controllers/filesystem.py
```diff
@@ -49,11 +51,64 @@
...
+    def _bg_probe(self):
+        if self.restricted:
+            probe_types = {'blockdev'}
+        else:
+            probe_types = None
+        # Should consider invoking probert in a subprocess here (so we
+        # can kill it if it gets stuck).
+        return self.controller.app.prober.get_storage(probe_types=probe_types)
+
+    def _probed(self, fut):
```
cyphermox: I'm sorry, but it's not immediately clear to me what "fut" is; what type of object it is. That points me to a possible refactoring candidate (but it doesn't have to be now). Furthermore, I wonder if it really needs to be passed to the method at all; and couldn't just live in the Probe class.

Maybe it's just hard to tell by looking at the diff.

xnox:
Somehow i was expecting that we are sitting and listening on the udev socket, just like we do for networking netlink. (aka udevmonitor).

This is not just about quiting debug shell, but like USB hot-plugging drives to complete the installation onto.

subiquity/controllers/filesystem.py
mwhudson: fut is a concurrent.futures.Future and is the variable name consistently used for such a thing throughout the codebase.

mwhudson:
> Somehow i was expecting that we are sitting and listening on the udev socket, just like we do for networking netlink. (aka udevmonitor).
> 
> This is not just about quiting debug shell, but like USB hot-plugging drives to complete the installation onto.

Yeah we should do those things but this branch is still better than what is currently in trunk IMO. Merging, we can do better later :)

PR #552, opened 2019-10-04
enable turning colors on and off

mwhudson:
Oops, this wasn't what I was supposed to be working on today...

This adds a way to turn colors on and off (a hot key and a help menu item). It also refactors the screen implementation we use away from the linux console to be more compatible in mono mode.

### DIFF
--- a/subiquity/palette.py
+++ b/subiquity/palette.py
@@ -73,8 +73,8 @@ STYLES = [
 
 
 STYLES_MONO = [
-    ('frame_header_fringe', 'black',   'white'),
-    ('frame_header',        'white',   'black'),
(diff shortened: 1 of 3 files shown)


mwhudson:
https://asciinema.org/a/zIfEpLTe0JQNRBdqApi4ij7lZ

mwhudson:
This also changes subiquity to always start in color mode, the idea being the user can just mash f4 if that looks bad on whatever serial console they are using.

xnox:
Tested this extensively, and this is the right approach i think going forward.

I have further fixes for ASCII mode on top of this.

PR #553, opened 2019-10-04
Add ASCII mode

xnox:
This makes s390x ASCII installer look extra nice!

### DIFF
--- a/subiquity/cmd/tui.py
+++ b/subiquity/cmd/tui.py
@@ -35,12 +35,23 @@ def parse_options(argv):
     parser = argparse.ArgumentParser(
         description='SUbiquity - Ubiquity for Servers',
         prog='subiquity')
+    try:
+        ascii_default = os.ttyname(0) == "/dev/ttysclp0"
+    except OSError:
+        ascii_default = False
+    parser.set_defaults(ascii=ascii_default)
     parser.add_argument('--dry-run', action='store_true',
                         dest='dry_run',
                         help='menu-only, do not call installer function')
     parser.add_argument('--serial', action='store_true',
                         dest='run_on_serial',
                         help='Run the installer over serial console.')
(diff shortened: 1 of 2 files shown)


xnox:
Can run it with --ascii flag to force ascification, can use flag --unicode to force unicodification, and on s390x ASCII console defaults to ascifyied.

subiquitycore/core.py
```diff
@@ -164,6 +164,19 @@ def make_screen(colors, is_linux_tty):
...
+    urwid.escape.DEC_SPECIAL_CHARMAP.update({
+        ord('\N{BLACK RIGHT-POINTING SMALL TRIANGLE}'): '>',
+        ord('\N{BLACK LEFT-POINTING SMALL TRIANGLE}'): '<',
+        ord('\N{BLACK DOWN-POINTING SMALL TRIANGLE}'): 'v',
+        ord('\N{BLACK UP-POINTING SMALL TRIANGLE}'): '^',
+        ord('\N{check mark}'): '+',
+        ord('\N{bullet}'): '*',
+        ord('\N{lower half block}'): '=',
+        ord('\N{upper half block}'): '=',
+    })
```
mwhudson: Haha oh man this is a glorious hack.

subiquity/cmd/tui.py
```diff
@@ -35,12 +35,23 @@ def parse_options(argv):
     parser = argparse.ArgumentParser(
         description='SUbiquity - Ubiquity for Servers',
         prog='subiquity')
+    try:
+        ascii_default = os.ttyname(0) == "/dev/ttysclp0"
+    except OSError:
+        ascii_default = False
+    parser.set_defaults(ascii=ascii_default)
```
mwhudson: Ooh I didn't know about this method. I suspect there is some other stuff that could be made simpler using it but I forget what now!

PR #555, opened 2019-10-07
Asciify color pallete

xnox:
Without custom colors, white on colored backgrounds is hard to read. Thus in ascii mode, use more contrasting colors from the pallete (ie. black on colored).

Also add one more substitution to use box drawing characters in the scrollbar.

This fixes https://bugs.launchpad.net/subiquity/+bug/1847072

### DIFF
--- a/subiquitycore/core.py
+++ b/subiquitycore/core.py
@@ -110,7 +110,7 @@ urwid_8_names = (
 )
 
 
-def make_palette(colors, styles):
+def make_palette(colors, styles, ascii):
     """Return a palette to be passed to MainLoop.
 
     colors is a list of exactly 8 tuples (name, (r, g, b))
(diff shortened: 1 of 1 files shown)


subiquitycore/core.py
```diff
@@ -130,12 +130,18 @@ def make_palette(colors, styles):
 
     urwid_palette = []
     for name, fg, bg in styles:
-        urwid_palette.append((name, urwid_name[fg], urwid_name[bg]))
+        urwid_fg, urwid_bg =  urwid_name[fg], urwid_name[bg]
+        if ascii:
```
mwhudson: Could you add some comments here? I guess what's going on is that there are some palette choices that work ok in 24 bit mode but look crap with the traditional colors?

PR #556, opened 2019-10-08
inject-subiquity-snap: add support for ppc64 iso images

paride:


paride:
This is how I tested that curtin master fixes https://bugs.launchpad.net/subiquity/+bug/1845611. With this change in we can easily add a new job to test the edge subiquity on ppc64.

PR #557, opened 2019-10-08
inject-subiquity-snap: fix the ppc64 edge iso volume name

paride:


paride:
Copy/paste typo :/ Not a big deal but may cause confusion if we happen to log in while testing. Let's get it right.

mwhudson:
haha oops

PR #558, opened 2019-10-08
Multipath

xnox:
When receiving cutin config with multipath data, exclude all but the first seen path per multipath device.

Impove display of drives that are multipath paths, ie. show WWN instead of serial, and indicate that they are multipath drives, and not "local disks".

To test use:
MACHINE=examples/multipath.json make dryrun

use ./tools/build-deb from curtin to build curtin from master, which is needed here.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -611,6 +611,8 @@ class _Device(_Formattable, ABC):
 class Disk(_Device):
     ptable = attributes.ptable()
     serial = attr.ib(default=None)
+    wwn = attr.ib(default=None)
+    multipath = attr.ib(default=None)
     path = attr.ib(default=None)
     model = attr.ib(default=None)
     wipe = attr.ib(default=None)
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -664,7 +664,7 @@ class Disk(_Device):
 
     def desc(self):
         if self.multipath:
-            return "multipath"
+            return "multipath device"
         return _("local disk")
 
     @property
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1172,6 +1172,7 @@ class FilesystemModel(object):
         byid = {}
         objs = []
         exclusions = set()
+        seen_multipaths = set()
         for action in config:
             if action['type'] == 'mount':
                 exclusions.add(byid[action['device']])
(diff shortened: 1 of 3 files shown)


subiquity/models/filesystem.py
```diff
@@ -665,13 +663,13 @@ def annotations(self):
         return []
 
     def desc(self):
+        if self.multipath:
+            return "multipath"
```
raharper: Would it make sense to include either the path or the mpath name or both?

subiquity/ui/views/filesystem/disk_info.py
```diff
@@ -27,9 +27,11 @@
 
 labels_keys = [
     ('Path:', 'devname'),
+    ('Multipath:', 'multipath'),
     ('Vendor:', 'vendor'),
     ('Model:', 'model'),
     ('SerialNo:', 'serial'),
+    ('WWN', 'wwn'),
```
raharper: missing colon on WWN

subiquity/models/filesystem.py
xnox: This is a "type" descriptor column, in short line form. For other things it's "Software RAID", "LVM", "local disk". The full line looks like this "$WWN        multipath" with the second line describing quickly what's on that drive.

mapth[a-z] are random, and WWN is the important bit, which is listed. So "multipath" is here to destinguish from "local disk" which isn't a multipathed one.

xnox: Other desc() are "local disk", "software RAID 1", "LVM volume group", "LVM logical volume"

mwhudson: I wonder if _("multipathed disk") would be better here.

raharper: I've seen 'multipath member device' used (Centos/Fedora).

mwhudson: The thing being labelled is the whole device though, not a member (conceptually speaking, anyway)

xnox: Multipath manpage uses "multipath device" and "multipath device map", as in using multipath as as adjective to describe a device.

Windows world uses MPIO moniker for MultiPath I/O.

Freebsd uses "multipath device" too in its gmultipath util.

I do prefer device, because what's on the other end might not be "a disk".

Solaris seems to call it MPxIO.

I think going with "multipath device" is short enough, and universal enough for our purposes.

PR #559, opened 2019-10-09
Use 19.2-33-gedfdb1d1 curtin from master for all the bugfixes.

xnox:
Otherwise we have no ppc64le installer or s390x one.


xnox:
@paride FYI

PR #565, opened 2019-10-17
WWN is confusing, for non-multipathed (but multipath capable) drives.

xnox:
LP: #1848533

https://bugs.launchpad.net/subiquity/+bug/1848533

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -669,7 +669,9 @@ class Disk(_Device):
 
     @property
     def label(self):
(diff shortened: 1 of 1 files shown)


mwhudson:
Ooops :(

xnox:
@raharper only when something went horribly wrong. If there is no WWN, one is synthesized by udev/multipath tools from the SERIAL, and unless i missunderstand everything is required for multipath to ever work and assemble anything.

But I do wonder if presenting port number and LUN would be better here, especially on s390x..... making a mental note to verify systemd udevd code.

PR #566, opened 2019-10-30
Drop crufty .deb packages from the source package build

vorlonofportland:
The binary packages from the subiquity source are unused with the exception
of console-conf (+subiquitycore) which is input into the core snap build;
and subiquity-tools is particularly bad as it has a hard-coded dependency
on python-all, which is not at all normal for a runtime package.

So just stop building these unmaintained binaries, and only build the ones
we actually use.


vorlonofportland:
now it builds.

mwhudson:
Hooray!

PR #569, opened 2019-10-31
generate crash reports

mwhudson:
This doesn't have any of the nice UI we designed for presenting and uploading crash reports, but it should still be useful. Please ask if anything doesn't make sense, some of the mysterious things might become clearer when I add the UI code but quite possibly some of it is redundant now.

### DIFF
--- a/subiquity/controllers/error.py
+++ b/subiquity/controllers/error.py
@@ -1,0 +1,177 @@
+# Copyright 2019 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import datetime
+import enum
+import json
+import logging
+import os
+
+import apport
+import apport.crashdb
+import apport.hookutils
+
+import attr
+
+from subiquitycore.controller import BaseController
+
+
+log = logging.getLogger('subiquity.controllers.error')
+
+
+class ErrorReportState(enum.Enum):
+    INCOMPLETE = _("INCOMPLETE")
+    DONE = _("DONE")
+    ERROR = _("ERROR")
+
+
+class ErrorReportKind(enum.Enum):
+    BLOCK_PROBE_FAIL = _("Block device probe failure")
+    DISK_PROBE_FAIL = _("Disk probe failure")
+    INSTALL_FAIL = _("Install failure")
+    UI = _("Installer crash")
+    UNKNOWN = _("Unknown error")
+
+
+@attr.s(cmp=False)
+class ErrorReport:
+    controller = attr.ib()
+    base = attr.ib()
+    pr = attr.ib()
+    state = attr.ib()
+    _file = attr.ib()
+
+    meta = attr.ib(default=attr.Factory(dict))
+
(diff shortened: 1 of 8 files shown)


subiquity/controllers/error.py
```diff
@@ -0,0 +1,177 @@
...
+
+class ErrorReportState(enum.Enum):
+    INCOMPLETE = _("INCOMPLETE")
+    DONE = _("DONE")
+    ERROR = _("ERROR")
+
+
+class ErrorReportKind(enum.Enum):
+    BP_FAIL = _("Block device probe failure")
+    RBP_FAIL = _("Disk probe failure")
```
raharper: BLOCK_PROBE_FAIL
DISK_PROBE_FAIL

Would make reading the enums in the rest of the code easier.

subiquity/controllers/error.py
```diff
@@ -0,0 +1,177 @@
...
+
+    meta = attr.ib(default=attr.Factory(dict))
+
+    @classmethod
+    def new(cls, controller, kind):
+        prefix = "installer.{}.{}".format(
+            datetime.datetime.utcnow().isoformat(timespec='seconds'),
+            kind.name.lower())
+        i = 0
+        while 1:
```
raharper: This should have some sort of upper-bound and timeout.

subiquity/controllers/filesystem.py
```diff
@@ -64,6 +65,11 @@ def __init__(self, controller, restricted, timeout, cb):
         self.cb = cb
         self.state = ProbeState.NOT_STARTED
         self.result = None
+        if restricted:
+            self.kind = ErrorReportKind.RBP_FAIL
+        else:
+            self.kind = ErrorReportKind.BP_FAIL
```
raharper: I see two kinds here, but below there are 3 scenarios (bp failed, bp timeout, bp data loading fail)  Should each be a separate enum?

subiquity/controllers/error.py
mwhudson: OK

subiquity/controllers/error.py
mwhudson: How about an upper bound of 10k? :)

subiquity/controllers/filesystem.py
mwhudson: Hmm I tend to think not. It only really makes sense to have a different kind if we want to display them differently in the UI and I don't think that we really will. Can always change this later when I add the UI if my thinking changes here.

subiquity/controllers/error.py
```diff
@@ -61,7 +61,7 @@ def new(cls, controller, kind):
             datetime.datetime.utcnow().isoformat(timespec='seconds'),
             kind.name.lower())
         i = 0
-        while 1:
+        while i < 10000:
```
raharper: I guess I don't understand the failing scenarios.  Is subiquity waiting on something and it keeps trying until it's present?  If we're waiting, shouldn't we sleep in this loop?  If we're not sleeping in this loop will we hit 10k too fast?

mwhudson: No, this is just the usual thing of atomically opening a file without overwriting an existing one. You do know what the 'x' flag means to open()?

raharper: OK, but I don't understand the failure scenario.  Multiple crashes in the same subiquity run?  Surely using a time-based filename should be sufficient, something with micro-second resolution won't end up overlapping multiple calls.

mwhudson: To some extent this is perhaps overthought as the expected number of crash files produced in any given run is at most 2 (of different kinds, to boot). But I do want the filenames to be human-friendly: they are saved to the install media by default, and if the user then pops their usb stick in their desktop to report a bug from there, I want them to be able to find them.

I guess I don't understand your concern here. The fact that this code is complex? That it has undesired consequences?

mwhudson: BTW the most recent version just puts the date in the timestamp.

mwhudson:
I've redone this, putting a nano-second level timestamp into the filename instead of looping. I've rebased rather than made new commits which might make re-review a bit trickier. Sorry!

PR #570, opened 2019-11-04
reprobe block devices on udev events

mwhudson:
This is actually pretty easy! I am a bit worried about the amount of log spam that will result from this, maybe it won't be too bad in practice.

### DIFF
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -17,11 +17,13 @@ import enum
 import json
 import logging
 import os
+import select
 import time
 
 import pyudev
 
 from subiquitycore.controller import BaseController
+from subiquitycore.utils import run_command
 
 from subiquity.models.filesystem import (
     align_up,
@@ -130,7 +132,7 @@ class FilesystemController(BaseController):
         self.answers.setdefault('manual', [])
         self._cur_probe = None
         self._monitor = None
-        self.ui_shown = False
+        self._udev_listen_handle = None
 
     def start(self):
         self._start_probe(restricted=False)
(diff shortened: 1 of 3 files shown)


subiquity/controllers/filesystem.py
```diff
@@ -129,16 +129,24 @@ def __init__(self, app):
...
+        self._monitor = pyudev.Monitor.from_netlink(context)
+        self._monitor.filter_by(subsystem='block')
+        self._monitor.enable_receiving()
+        self._last_udev_event = time.time()
+        self.loop.watch_file(self._monitor.fileno(), self._udev_event)
+
+    def _udev_event(self):
+        action, dev = self._monitor.receive_device()
+        log.debug("_udev_event %s %s", action, dev)
+        if time.time() - self._last_udev_event > 0.1 and not self.ui_shown:
```
raharper: Since we get the event, and we know it's block related, why not use:

udevadm settle --exit-if-exists=dev

This blocks until udev has finished processing the events related to dev?

subiquity/core.py
```diff
@@ -46,8 +44,6 @@
 
 class Subiquity(Application):
 
-    signals = ['debug-shell-exited']
```
raharper: Why is this no longer needed?

mwhudson: It was only used to trigger a reprobe when the debug shell exited. If there's other things we might want to trigger then, we can keep it -- but I can't think of anything.

subiquity/controllers/filesystem.py
mwhudson: My thinking was that udev events tend to come in a flurry and if we just wait for a period of time with no events that's good enough. Waiting for the udevadm queue to settle is probably better though (it's what the network monitoring code does). Whatever we do though, this logic is busted, it will only process the first event of a flurry. Let me fix that...

mwhudson:
> How do we test this is working? Do we inject a usb device while the UI is running?

I tested this by mounting / unmounting an iso and watching the log file.

subiquity/controllers/filesystem.py
```diff
@@ -130,25 +132,44 @@ def __init__(self, app):
...
+            self._udev_listen_handle = None
 
     def _udev_event(self):
-        action, dev = self._monitor.receive_device()
-        log.debug("_udev_event %s %s", action, dev)
-        if time.time() - self._last_udev_event > 0.1 and not self.ui_shown:
-            self._start_probe(restricted=False)
-        self._last_udev_event = time.time()
+        cp = run_command(['udevadm', 'settle', '-t', '0'])
+        if cp.returncode != 0:
```
raharper: Hrm, if we have a non-empty queue (returncode != 0) shouldn't we call udevadm settle and then probe?

mwhudson: That would block the UI, so no. We could do it in a background thread I guess but this seems to fail the effort / benefit ratio

xnox:
I think this is working, but i'm not sure.

> > How do we test this is working? Do we inject a usb device while the UI is running?
> 
> I tested this by mounting / unmounting an iso and watching the log file.

But that doesn't create or delete a block device. One needs to hotplug a drive.

I'm trying this on the mainframe by enabling/disabling dasd drive and it doesn't appear to be reprobed. I'll to simulate this in VM on x86 now. And restarting installer makes it appear.

In the debug log I see 

```
2019-11-06 13:50:06,771 DEBUG subiquitycore.controller.filesystem:169 _udev_even
t change Device('/sys/devices/css0/0.0.0278/0.0.1618/block/dasdb')              
2019-11-06 13:50:06,771 DEBUG subiquitycore.controller.filesystem:169 _udev_even
t add Device('/sys/devices/css0/0.0.0278/0.0.1618/block/dasdb/dasdb1')          
2019-11-06 13:50:06,771 DEBUG subiquitycore.controller.filesystem:169 _udev_even
t change Device('/sys/devices/css0/0.0.0278/0.0.1618/block/dasdb')              
2019-11-06 13:50:06,771 DEBUG subiquitycore.controller.filesystem:169 _udev_even
t change Device('/sys/devices/css0/0.0.0278/0.0.1618/block/dasdb/dasdb1')       
2019-11-06 13:50:06,771 DEBUG block-discover:73 starting probe restricted=False 
2019-11-06 13:50:06,772 ERROR block-discover:108 probing successful restricted=F
alse                                                                            
NoneType: None        
```

xnox:
I wonder if my test is invalid, because i got to the filesystem page already.

Hm.... let's merge this, build it in edge and test interactively.

mwhudson:
> I wonder if my test is invalid, because i got to the filesystem page already.
> 
> Hm.... let's merge this, build it in edge and test interactively.

Hm that NoneType: None thing is strange. Did you get to test this yet?

PR #573, opened 2019-11-06
More accurate estimation of the size of a RAID

mwhudson:
I found the code in mdadm that decides how much of a device is usable
for data and copied it. It's a bit hairy!

For https://bugs.launchpad.net/subiquity/+bug/1816777 and many similar
bugs.

### DIFF
--- a/scripts/raid-size-tests.py
+++ b/scripts/raid-size-tests.py
@@ -119,7 +119,7 @@ def verify_size_ok(level, sizes):
             print("exactly right!")
             r = True
         else:
-            print("underestimated size by", real_size - calc_size)
+            print("subiquity wasted space", real_size - calc_size)
             r = True
     finally:
         cleanup()
--- a/scripts/raid-size-tests.py
+++ b/scripts/raid-size-tests.py
@@ -1,0 +1,127 @@
+#!/usr/bin/python3
+
+import atexit
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import uuid
+
+import attr
+
+from subiquity.models.filesystem import (
+    dehumanize_size,
+    get_raid_size,
+    humanize_size,
+    raidlevels,
+    )
+
+
+tmpdir = tempfile.mkdtemp()
+
+raids = []
+loopdevs = []
+
+_cmdoutput = []
+
+def run(cmd):
+    try:
+        subprocess.run(
+            cmd, check=True,
+            stdout=subprocess.PIPE, stdin=subprocess.DEVNULL,
+            stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:
+        print(e.stdout)
+        raise
+
+
+def cleanup():
+    for raid in raids:
+        subprocess.run(
+            ['mdadm', '--verbose', '--stop', raid])
+    for loopdev in loopdevs:
+        subprocess.run(
+            ['losetup', '-d', loopdev])
+    shutil.rmtree(tmpdir)
+
+
+def create_devices_for_sizes(sizes):
+    devs = []
+    for size in sizes:
+        fd, name = tempfile.mkstemp(dir=tmpdir)
+        os.ftruncate(fd, size)
+        os.close(fd)
+        dev = subprocess.run(
+            ['losetup', '-f', '--show', name],
+            stdout=subprocess.PIPE, encoding='ascii').stdout.strip()
+        devs.append(dev)
+        loopdevs.append(dev)
+    return devs
+
+
+def create_raid(level, images):
+    name = '/dev/md/{}'.format(uuid.uuid4())
+    cmd = [
+        'mdadm',
+        '--verbose',
+        '--create',
+        '--metadata', 'default',
+        '--level', level,
+        '-n', str(len(images)),
+        '--assume-clean',
+        name,
+        ] + images
+    run(cmd)
+    raids.append(name)
+    return name
+
+
+def get_real_raid_size(raid):
+    return int(subprocess.run(
+        ['blockdev', '--getsize64', raid],
+        stdout=subprocess.PIPE, encoding='ascii').stdout.strip())
+
+
+@attr.s
+class FakeDev:
+    size = attr.ib()
+
+
+def verify_size_ok(level, sizes):
+    devs = create_devices_for_sizes(sizes)
+    raid = create_raid(level, devs)
+    devs = [FakeDev(size) for size in sizes]
+    calc_size = get_raid_size(level, devs)
+    real_size = get_real_raid_size(raid)
+    if len(set(sizes)) == 1:
+        sz = '[{}]*{}'.format(humanize_size(sizes[0]), len(sizes))
+    else:
+        sz = str([humanize_size(s) for s in sizes])
+    print("level {} sizes {} -> calc_size {} real_size {}".format(
+        level, sz , calc_size, real_size), end=' ')
+    if calc_size > real_size:
+        print("BAAAAAAAAAAAD", real_size - calc_size)
+        r = False
+    else:
+        print("OK by", real_size - calc_size)
+        r = True
+    run(['mdadm', '--verbose', '--stop', raid])
+    raids.remove(raid)
+    return r
+
+
+fails = 0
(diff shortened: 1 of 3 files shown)


xnox:
i guess i want VM tests for all raid levels at this point.

subiquity/models/filesystem.py
```diff
@@ -218,24 +218,38 @@ def dehumanize_size(size):
...
+    # touch ridiculous. What follows is a translation of the code at:
+    # https://git.kernel.org/pub/scm/utils/mdadm/mdadm.git/tree/super1.c?h=mdadm-4.1&id=20e8fe52e7190b3ffda127566852eac2eb7fa1f7#n2770
+    # (note that that calculation is in terms of 512-byte sectors and
+    # this one is in bytes).
+    #
+    # This makes assumptions about the defaults mdadm uses but mostly
+    # that the default metadata version is 1.2, and other formats use
+    # less space.
+    min_size = min(dev.size for dev in devices)
+    bmspace = 128*1024
```
raharper: I'd expand the variable names here, since it's not kernel code.
bitmap_space, and summarize the decision, we're assuming
most devices are going to be > 200GiB so we select the 128k bitmap_space

subiquity/models/filesystem.py
```diff
@@ -218,24 +218,38 @@ def dehumanize_size(size):
...
+    headroom = 128*1024*1024
```
raharper: What does headroom correspond to?

subiquity/models/filesystem.py
```diff
@@ -218,24 +218,38 @@ def dehumanize_size(size):
...
+    while (headroom << 10) > min_size and headroom > 2*1024*1024:
+        headroom >>= 1
+    # mdadm's Create() can round things a little more so, to be
+    # pessimistic, assume another megabyte gets wasted somewhere.
+    data_offset = align_up(12*1024 + bmspace + headroom) + 1024*1024
```
raharper: I'd be surprised if this ends up being larger than the 8MB overhead you already had.

subiquity/models/tests/test_filesystem.py
```diff
@@ -105,6 +106,12 @@ def test_errors(self):
                 self.assertEqual(expected_error, actual_error)
 
 
+class TestRoundRaidSize(unittest.TestCase):
+
+    def test_lp1816777(self):
+        self.assertLessEqual(round_raid_size(500107862016), 499972571136)
```
raharper: 4.6 TiB drive has 130 MB overhead.  OK.  It would be interesting to see the mdadm --examine superblock dump to match up where the space went (bitmap, alignment, etc).

raharper:
> i guess i want VM tests for all raid levels at this point.

We don't need VM tests so much as cycling values through
the calculation, then running mdadm create on a block device
and using mdadm examine to verify the overhead and that
we can successfully create it.

subiquity/models/filesystem.py
mwhudson: I think there's some value in following the variable names from the mdadm source (maybe I should paste that in a comment) but you're right about where the value for bmspace comes from.

subiquity/models/filesystem.py
mwhudson: No idea really! The mdadm source says "allow for some headroom for reshape".

subiquity/models/filesystem.py
mwhudson: Get ready to be surprised!

subiquity/models/tests/test_filesystem.py
mwhudson: Sadly none of our bug reporters we quite as thorough as to include that :)

mwhudson:
> > i guess i want VM tests for all raid levels at this point.
> 
> We don't need VM tests so much as cycling values through
> the calculation, then running mdadm create on a block device
> and using mdadm examine to verify the overhead and that
> we can successfully create it.

Yeah, indeed. I don't think there's any way to write this test without requiring root so I don't know if we can make anything like this a unit test but it's probably still worth writing these tests...

scripts/raid-size-tests.py
```diff
@@ -89,36 +92,41 @@ class FakeDev:
...
+        if len(set(sizes)) == 1:
+            sz = '[{}]*{}'.format(humanize_size(sizes[0]), len(sizes))
+        else:
+            sz = str([humanize_size(s) for s in sizes])
+        print("level {} sizes {} -> calc_size {} real_size {}".format(
+            level, sz , calc_size, real_size), end=' ')
+        if calc_size > real_size:
+            print("BAAAAAAAAAAAD", real_size - calc_size)
+        else:
+            print("OK by", real_size - calc_size)
```
raharper: print("RAID overhead %s bytes" % real_size - calc_size)

mwhudson: That's not the RAID overhead though, it's the difference between our estimate of the size and the real size. I'll make this a bit clearer.

mwhudson:
I ran some tests and found that my code wasn't right when the sizes of the devices varied. I've updated it to match mdadm's behaviour in that regard and now raid-size-tests.py reports subiquity's estimates as byte accurate (apart from raid10 with odd numbers of devices). It's a bit more involved of a change now, so I'll let it sit for a while before merging...

xnox:
Mdadm doesn't expose this at all does it? Can we like execute mdadm, or its library to give us the numbers? I.e. dryrun on the device?!

mwhudson:
Nope, mdadm doesn't do anything that useful. subiquity *could* do the "loopback devices in a tmpfs" game my test script does, but that sounds insane (it's also quite slow).

mwhudson:
Hmm that's not a great filename.

mwhudson:
And this was the wrong PR of course.

scripts/raid-size-tests.py
```diff
@@ -107,8 +115,14 @@ def verify_size_ok(level, sizes):
...
             print("BAAAAAAAAAAAD", real_size - calc_size)
+            if os.environ.get('DEBUG'):
+                print(raid)
+                input('waiting: ')
+        elif calc_size == real_size:
+            print("exactly right!")
+            r = True
         else:
-            print("OK by", real_size - calc_size)
+            print("underestimated size by", real_size - calc_size)
```
raharper: subiquity wasted space?

subiquity/controllers/filesystem.py
```diff
@@ -419,13 +419,13 @@ def delete_raid(self, raid):
         self.model.remove_raid(raid)
 
     def create_volgroup(self, spec):
-        devices = set()
+        devices = []
```
raharper: One can't add the same disk twice, so set() seems in order here.  What's the reason for the change?

subiquity/models/filesystem.py
```diff
@@ -218,36 +218,63 @@ def dehumanize_size(size):
...
+# We make some assumptions about the defaults mdadm uses but mostly that the
+# default metadata version is 1.2, and other formats use less space.
+#
+# Note that data_offset is computed for the first disk mdadm examines and then
+# used for all devices, so the order matters! (Well, if the size of the
+# devices vary, which is not normal but also not something we prevent).
+#
+# All this is tested against reality in ./scripts/get-raid-sizes.py
+def calculate_data_offset(devsize):
+    devsize >>= 9  # convert to sectors
```
raharper: I think we could add a comment here to the effect that we're using sectors to make it easier to match the kernel mdadm algorithm and we convert back to bytes.

I'd also suggest renaming this function to 'calculate_data_offset_bytes'

scripts/raid-size-tests.py
mwhudson: OK.

subiquity/controllers/filesystem.py
mwhudson: Well the discovery I made yesterday that the order matters to mdadm. But I agree this is conceptually bogus. Probably the better fix is to pass them around as a set and sort them (consistently but arbitrarily) in get_raid_sizes and when rendering the config.

That will let me make this change much smaller, which is nice...

subiquity/models/filesystem.py
mwhudson: Fair on both counts.

PR #575, opened 2019-11-12
stop caching probed storage data

mwhudson:
A recent branch carefully called Prober.get_storage() whenever a udev
block device event was seen. But Prober.get_storage() cached the results
from probert rendering that useless. Rip the caching and some other
complexity out of the prober class.

### DIFF
--- a/subiquitycore/prober.py
+++ b/subiquitycore/prober.py
@@ -15,8 +15,11 @@
 
 import logging
 import yaml
-from probert.network import (StoredDataObserver,
-                             UdevObserver)
+
+from probert.network import (
+    StoredDataObserver,
+    UdevObserver,
+    )
 from probert.storage import Storage
 
 log = logging.getLogger('subiquitycore.prober')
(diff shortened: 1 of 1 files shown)


subiquitycore/prober.py
```diff
@@ -15,45 +15,38 @@
...
     def probe_network(self, receiver):
-        if self.opts.machine_config:
+        if self.saved_config is not None:
             observer = StoredDataObserver(
                 self.saved_config['network'], receiver)
         else:
             observer = UdevObserver(receiver)
         return observer, observer.start()
 
     def get_storage(self, probe_types=None):
```
raharper: Hrm, I do wish we called this probe_storage() to match probe_network().  But that's not a big deal.

mwhudson: One day...

mwhudson:
Tested on real hw and it worked fwiw

xnox:
AHA

PR #576, opened 2019-11-18
if /target exists, run curtin umount /target before first disk probe

mwhudson:
The installer now lets you restart after an install failure, but unless
we do something like this, the target system still being mounted will
make life confusing.

It's more robust to be defensive and try to unmount at startup rather
than trying to unmount before restarting.


mwhudson:
I tested this by making an install fail by SIGINTing it and then choosing to restart the installer, seemed to work fine.

PR #577, opened 2019-11-20
enable sending reports to daisy

mwhudson:
Probably not really ready but might be interesting to play with. Will make a video tomorrow :)

### DIFF
--- a/scripts/installdeps.sh
+++ b/scripts/installdeps.sh
@@ -4,6 +4,6 @@ set -eux
 cat /etc/apt/sources.list | sed -n 's/-updates/-proposed/p' > /etc/apt/sources.list.d/proposed.list
 apt-get update
 DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade
-apt-get install -y --no-install-recommends libnl-3-dev libnl-genl-3-dev libnl-route-3-dev libsystemd-dev python3-distutils-extra pkg-config python3.5 python3-pip git lsb-release python3-setuptools gcc python3-dev python3-wheel curtin pep8 python3-pyflakes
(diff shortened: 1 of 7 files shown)


mwhudson:
I've fixed a few things here and there now and made some videos:

 * A UI crash: https://asciinema.org/a/oe6mVcPyNi2QGiTejY6cGiS5F (in the actual installer, the ui is restarted automatically)
 * An install failure: https://asciinema.org/a/C8eWjF3yD767D99stchuSf8dK
 * A device probe failure: https://asciinema.org/a/XD9GYiz1MzAmbTQYTyOn03c5B
 * A device _and_ disk probe failure: https://asciinema.org/a/8MrhWHGoTpBFmqVjzYxAV9oY4

mwhudson:
> Does it need to be using bson?

Yes, the error reporter expects uploads as bson. To be fair, I just copied what whoopsie does for this, but looking into lp:daisy, I don't see any support for formats other than bson.

> Seems like it might be a better idea to use the json API from python core instead, unless you really specifically need to be using the APIs from bson. I think that would be a trivial change to; since you can still run json.load and whatnot.

I'd happily make this change if it would not break the whole feature :)

matthewpaulthomas:
Sorry I didn’t get time to look at this again earlier. Nice work overall.

If I could change one thing, it would be the sentence “Select [C]ontinue to try again.” This is because “try again” could be referring to three different things: sending the error report again, saving the error report again, or (my assumption, though it’s not suggested anywhere in the UI!) performing the installation again. (A related problem is that the suggestion to “Select [C]ontinue” remains visible during the error report upload process when there is no Continue button to press.)

mwhudson:
https://github.com/CanonicalLtd/subiquity/pull/583

PR #579, opened 2019-12-03
switch to AsyncioEventLoop and start using asyncio APIs

mwhudson:
The thought of implementing autoinstall stuff with our existing concurrency APIs just seemed too painful so I tried moving to first trio (which is neat but requires quite a new Python and isn't packaged for bionic) and then asyncio. There are many more things that could be switched to asyncio but this is enough to prove to my own satisfaction that it's worth it.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -146,22 +146,23 @@ controllers don't have a corresponding model class.
 ### Doing things in the background
 
 If the UI does not block, as promised above, then there needs to be a way of
-running things in the background.  The main way this is done is the `run_in_bg`
-function, which takes two functions: one that takes no arguments and is called
-in a background thread and a callback that takes one argument, and is called in
-the main/UI thread with a `concurrent.futures.Future` representing the result
-of calling the first function. This is a fairly clunky interface that can lead
-to some slightly baffling control flow but, well, it works. I've tried a few
-abstractions that make things clearer -- `subiquity.tasksequence.TaskSequence`
-being the older one that doesn't quite cover all interesting cases and
-`subiquity.controllers.installprogress.StateMachine` being the newer one that I
-hope can be generalized to something useful in many places.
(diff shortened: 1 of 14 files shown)


subiquity/controllers/keyboard.py
```diff
@@ -52,14 +53,13 @@ def start_ui(self):
...
-
-    def _done(self, fut):
-        log.debug("KeyboardController._done next-screen")
+    async def apply_settings(self, setting):
+        await self.model.set_keyboard(setting)
+        log.debug("KeyboardController next-screen")
         self.signal.emit_signal('next-screen')
 
+    def done(self, setting):
+        asyncio.create_task(self.apply_settings(setting))
```
raharper: Would you want some subiquity task abstraction here in case we need to do something different than direct calling of asyncio.create_task?

raharper: Not entirely sure, but this looks like it should use the scheduel_task() method you've defined.

mwhudson: Ah yes good catch.

DESIGN.md
```diff
@@ -146,22 +146,23 @@ controllers don't have a corresponding model class.
...
 
 If the UI does not block, as promised above, then there needs to be a way of
-running things in the background.  The main way this is done is the `run_in_bg`
+running things in the background.  Subiquity uses a few different ways to do
+this but new code should use
+[asyncio](https://docs.python.org/3/library/asyncio.html). `subiquity.async_helpers`
+defines two useful helper functions `run_in_thread` (just a nicer wrapper
+around `run_in_executor`) and `schedule_task` (a wrapper around `create_task`
+that works before the event loop is
+started). [trio](https://trio.readthedocs.io/en/stable/) has nicer APIs bit is
```
xnox: s / bit / but /

mwhudson: thanks

PR #585, opened 2019-12-11
asyncioify network controller

mwhudson:
One of the biggies for slightly confusing control flow.


mwhudson:
Haven't tested this in a VM yet, need to do that soon.

mwhudson:
I fixed a bunch of small things and it seems to work. Configuring a bond segfaults (!) but stable does that too (!).

mwhudson:
This now seems to work as well as the current stable.

mwhudson:
Going to merge this and some other asyncioification branches -- there's a risk of regressions but it's definitely the way I want to go, so I'm going to go for the "land it and fix afterwards" approach

PR #587, opened 2019-12-11
get a fix from upstream urwid

mwhudson:
the symptom of the bug fixed is that exceptions that crash the process
do not get reported sensibly.


mwhudson:
See https://github.com/urwid/urwid/issues/235

PR #588, opened 2019-12-11
Merge guided method & disk selection into one screen.

xnox:
This merges the method & disk selection screens into one.

This is as per mpt sketch in Paris.

This does not yet add encryption tickbox.

Bugs:
* state is not correctly preserved between going forward & back
* answers preseeding is probably broken
* selector just lists all disks, without pretty size, or filtering out drives that can't be used

UX:
* The lvm tickbox is supposed to be indented under the drop down, I have no idea how to make that happen
* the disk drop down, I believe was meant to be inline with disk tickbox, also no idea how to make that happen
* Note that the future luks tickbox needs to be further indented too

@mwhudson please play around with this branch and help me if I should proceed with fixing this branch up, or if we want to add LUKS as a method selector as the current UX stands.


mwhudson:
What do you think about https://github.com/mwhudson/subiquity/tree/new-disk-ux? It's some awful hacking but I think it might be salvageable. (It goes back to displaying all disks in a table, which is probably a mistake, I think a selector that shows more info about each disk is probably better here).

xnox:
@mwhudson i like it!

Let's make it extra shiny. Not sure how to collaborate on that spagetti code.

subiquity/ui/views/filesystem/guided.py
```diff
@@ -48,111 +57,97 @@
...
-
-    def guided(self, btn, method):
-        log.debug("GuidedFilesystemView.guided")
-        self.controller.guided(method)
-
-    def cancel(self, btn=None):
-        self.controller.cancel()
+class GuidedForm(Form):
+
+    # use a proper radio group!!!
```
cyphermox: Hum, please, either do it right away, or remove the comment and file a bug instead. Comments like that will just end up forgotten. Same for the padding comments below.

subiquity/ui/views/filesystem/guided.py
```diff
@@ -175,11 +170,9 @@ def choose_disk(self, btn, disk):
                     size=disk.free_for_partitions,
                     fstype=None,
                     ))
-            vg = self.controller.create_volgroup(
-                spec=dict(
-                    name="ubuntu-vg",
-                    devices=set([part]),
-                    ))
+            spec = dict(name="ubuntu-vg", devices=set([part]))
+            # create volume group on partition
```
cyphermox: Is that a helpful comment? It just describes what the next line already says, rather than clarifying /why/ it's done there or that way. I think a comment is unwarranted here.

subiquity/controllers/filesystem.py
```diff
@@ -255,7 +254,7 @@ def start_ui(self):
             # performed would be tricky.  Possibly worth doing though! Just
             # not today.
```
cyphermox: Can we take the opportunity to make this go; either fix the issue now as a refactor or file a bug to handle later? Is it even relevant with the current rework of the UI?

xnox:
@cyphermox 

In paris, we asked MPT to review our screens of guided installation which had:
- normal
- with lvm
- with luks
- custom

Followed by a page with pick dis
- disk1
- disk2

However, most of the time people want normal install, and most of the time, there is only one disk.

So mpt asked, can you do
- normal
-- pick-disk dropdown
--- use lvm
---- use luks
- custom

Slightly closer to how it is in ubiquity on the desktop.

here is the new screen ascii

https://asciinema.org/a/UsZjzonH0w7lN1x7be97AmXIX?t=21

mwhudson:
Now rebased on https://github.com/CanonicalLtd/subiquity/pull/599

mwhudson:
Aside from the "I like LUKS" option not doing anything, I think this is good now.

mwhudson:
I removed the LUKS stuff for now and polished the code a bunch. Probably best to squash merge this?

PR #591, opened 2019-12-12
asyncioify installation

mwhudson:
Look at that lovely diffstat


mwhudson:
Going to merge this and some other asyncioification branches -- there's a risk of regressions but it's definitely the way I want to go, so I'm going to go for the "land it and fix afterwards" approach

PR #592, opened 2019-12-12
Form tweaks

mwhudson:
Somewhat cleaned up version of the hacks I was creating for @xnox's new-disk-ux branch.


xnox:
I wonder if I should like refactor my radiobutton addition this way too.

mwhudson:
> I wonder if I should like refactor my radiobutton addition this way too.

Yeah, that would make sense I think. I can take a swing at that now.

PR #593, opened 2019-12-12
asyncioify filesystem controller

mwhudson:


subiquity/controllers/filesystem.py
```diff
@@ -139,52 +71,94 @@ def __init__(self, app):
...
+            fname = 'probe-data-restricted.json'
+            key = "ProbeDataRestricted"
+        else:
+            probe_types = None
+            fname = 'probe-data.json'
+            key = "ProbeData"
+        debug_flags = self.debug_flags
+        if 'bpfail-full' in debug_flags and not restricted:
+            await asyncio.sleep(2)
+            1/0
```
xnox: note to self, this is intentional crash, due to crash in debug flags.

mwhudson: I'd actually like to push this crash further down, into the prober but well. One thing at a time.

PR #595, opened 2019-12-13
remove run_in_bg

mwhudson:
This branch, which has #593 and #594 merged into it, removes the run_in_bg style of concurrency completely \o/


xnox:
Not feeling like resolving conflicts.... Might be needed to just take this PR and backout the other two I merged.

PR #596, opened 2019-12-14
Unbreak console-conf in UC20 images

xnox:
storage is not required in consoleconf, and is not installed, ensure we don't import it.

ascii argument is required.


xnox:
With this I can start console-conf on the UC20 image.

mwhudson:
That thing about the ascii argument is a bit unfortunate but well, good enough for now.

PR #605, opened 2019-12-16
make the snaplist code more async

mwhudson:


subiquity/snapd.py
```diff
@@ -146,6 +146,7 @@ def post(self, path, body, **args):
             "Don't know how to fake POST response to {}".format((path, args)))
 
     def get(self, path, **args):
+        time.sleep(1)
```
xnox: Is this ratelimitting talking to snapd? Or some race? Do you want for this to be blocking, or like await sleep?

xnox: Or is it like for debug "fake" mode, and should be pushed further down?

mwhudson: This is the fake stuff, and as far down as it can be pushed I think?

xnox: oh yes.... all of the dirs are fake. Somehow i thought it's constructing a URL.

PR #609, opened 2019-12-16
stop some pointless use of urwid signals

mwhudson:
Signals only make sense when you don't always know who the listener is. For next-screen/prev-screen/quit it's always the app.

### DIFF
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -246,7 +246,7 @@ class IdentityController(BaseController):
     def cancel(self):
         # You can only go back if we haven't created a user yet.
         if self.model.user is None:
-            self.signal.emit_signal('prev-screen')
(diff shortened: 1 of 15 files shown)


mwhudson:
Yeah, we can and should kill all the signals I think (in fact, I'd like to get all use of urwid APIs out of anything but view code -- and maaaaaaaaaaaaaaaaaybe one day even move to a model where even the TUI interface is a http client ......). However the remaining uses of signals are different in character in that the sender does not (conceptually at least) know what the receiver is, so removing them isn't quite so trivial and that's why I didn't remove them in this branch.

We should probably remove the stub code for ceph and iscsi and bcache from the tree though.

xnox:
I have wrapped tty in an http webserver thing with javascript terminal. And it did work nice enough.

PR #614, opened 2019-12-17
 structured reporting

mwhudson:
autoinstalls will need to report their progress in a somewhat structured way. This branch hooks into curtin's reporting machinery to do this. 

The commit history is a mess, the diff is inflated by moving code into with blocks and there are some awful hacks around the formatting and level of log messages (and too much is still at INFO) but... I think this approach has legs.

### DIFF
--- a/subiquity/controllers/error.py
+++ b/subiquity/controllers/error.py
@@ -94,6 +94,7 @@ class ErrorReport(metaclass=urwid.MetaSignals):
     pr = attr.ib()
     state = attr.ib()
     _file = attr.ib()
+    _context = attr.ib()
 
     meta = attr.ib(default=attr.Factory(dict))
     uploader = attr.ib(default=None)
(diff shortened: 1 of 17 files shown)


mwhudson:
I'm actually pretty happy with this now. Best reviewed locally with git diff -b!

subiquity/controllers/error.py
```diff
@@ -158,39 +159,38 @@ def _bg_add_info():
...
+                log.exception("loading problem report failed")
+                self.state = ErrorReportState.ERROR_LOADING
+            else:
+                self.state = ErrorReportState.DONE
         self._file.close()
         self._file = None
         urwid.emit_signal(self, "changed")
 
     def upload(self):
-        log.debug("starting upload for %s", self.base)
```
cyphermox: Why dropping the debug logs for this stuff?

PR #624, opened 2019-12-20
refactor how subiquity is exited after install

mwhudson:
if we want to do things after install has completed (e.g.: run late
commands), we can't have the code that runs the install invoking
/sbin/reboot directly.


mwhudson:
This one probably needs to be tested in a VM (which I am not going to have time to do)

mwhudson:
Have tested in VM now.

PR #625, opened 2019-12-20
autoinstall!

mwhudson:
The branch implements the first cut of autoinstall for subiquity. What is there:

 * support for fully non-interactive and partially interactive installs
 * support for early and late commands

What is not there:

 * any sort of decent error handling
    * this includes any sort of sensible data validation. Get your autoinstall.yaml wrong and things will just blow up in a mess
 * support for the packages, debconf-selections, snaplist, user-data sections of the config file
 * sensible behaviour when more than one subiquity is running (e.g. on tty1 and serial)
 * generation of an autoinstall file at the end of install
 * any kind of sensible storage configuration, it just supports lvm or direct installation on the largest disk in the system

But I have plans for all of these things :)

### DIFF
--- a/examples/autoinstall.yaml
+++ b/examples/autoinstall.yaml
@@ -3,6 +3,10 @@ early-commands:
         - echo a
         - sleep 1
         - echo a
+late-commands:
+        - echo a
+        - sleep 1
+        - echo a
 keyboard:
         layout: gb
 identity:
--- a/examples/autoinstall.yaml
+++ b/examples/autoinstall.yaml
@@ -1,0 +1,12 @@
+version: 1
+early-commands:
+        - echo a
+        - sleep 1
+        - echo a
+keyboard:
+        layout: gb
+identity:
+        realname: ''
+        username: ubuntu
+        password: '$6$wdAcoXrU039hKYPd$508Qvbe7ObUnxoj15DRCkzC3qO7edjH0VV7BPNRDYK4QR8ofJaEEF2heacn0QgD.f8pO8SNp83XNdWG6tocBM1'
+        hostname: ubuntu
--- a/subiquity/autoinstall.py
+++ b/subiquity/autoinstall.py
@@ -1,0 +1,28 @@
+# Copyright 2019 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import yaml
+
+from curtin.config import merge_config
+
+
+def merge_autoinstall_configs(source_paths, target_path):
+    config = {}
+    for path in source_paths:
+        with open(path) as fp:
+            c = yaml.safe_load(fp)
+        merge_config(config, c)
+    with open(target_path, 'w') as fp:
+        yaml.dump(config, fp)
--- a/subiquity/controller.py
+++ b/subiquity/controller.py
@@ -26,8 +26,35 @@ log = logging.getLogger("subiquity.controller")
 
 class SubiquityController(BaseController):
 
-    def deserialize(self, state):
-        self.configured()
+    autoinstall_key = None
+    autoinstall_default = None
+
+    def __init__(self, app):
+        super().__init__(app)
+        self.autoinstall_applied = False
+        if app.autoinstall_config:
+            self.load_autoinstall_data(
+                app.autoinstall_config.get(
+                    self.autoinstall_key,
+                    self.autoinstall_default))
+
+    def load_autoinstall_data(self, data):
+        """Load autoinstall data.
+
+        This is called if there is an autoinstall happening. This
+        controller may not have any data, and this controller may still
+        be interactive.
+        """
+        pass
+
+    async def apply_autoinstall_config(self):
+        """Apply autoinstall configuration.
+
+        This is only called for a non-interactive controller. It should
+        block until the configuration has been applied. (self.configured()
+        is called after this is done).
+        """
+        pass
 
     def interactive(self):
         if not self.app.autoinstall_config:
@@ -44,6 +71,9 @@ class SubiquityController(BaseController):
         if self.model_name is not None:
             self.app.base_model.configured(self.model_name)
 
+    def deserialize(self, state):
+        self.configured()
+
 
 class NoUIController(SubiquityController):
 
@@ -58,6 +88,16 @@ class NoUIController(SubiquityController):
 
 
 class RepeatedController(RepeatedController):
+
+    def __init__(self, orig, index):
+        super().__init__(orig, index)
+        self.autoinstall_applied = False
+
+    async def apply_autoinstall_config(self):
+        await self.orig.apply_autoinstall_config(self.index)
+
+    def configured(self):
+        self.orig.configured()
 
     def interactive(self):
         return self.orig.interactive()
--- a/subiquity/controller.py
+++ b/subiquity/controller.py
@@ -29,6 +29,15 @@ class SubiquityController(BaseController):
     def deserialize(self, state):
         self.configured()
 
+    def interactive(self):
+        if not self.app.autoinstall_config:
+            return True
+        i_sections = self.app.autoinstall_config.get(
+            'interactive-sections', [])
+        if '*' in i_sections or self.autoinstall_key in i_sections:
+            return True
+        return False
+
     def configured(self):
         """Let the world know that this controller's model is now configured.
         """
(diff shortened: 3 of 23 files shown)


mwhudson:
To test this, try something like:

```
$ wget https://people.canonical.com/~mwh/subiquity_autoinstall.snap
$ sudo ~/src/subiquity/scripts/inject-subiquity-snap.sh \
    -c 'cp /home/mwhudson/src/subiquity/examples/autoinstall.yaml new_installer/' \
    ~/isos/focal-live-server-amd64.iso  subiquity_autoinstall.snap ~/isos/custom.iso
```

subiquity/autoinstall.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import yaml
+
+from curtin.config import merge_config
+
+
+def merge_autoinstall_configs(source_paths, target_path):
```
raharper: The order of the source_paths controls what the final config will look like due to how merge_config works with dictionaries and lists.  I suggest documenting that behavior.  Additionally it's important to retain the _order_ of the configs, or the order of the expected precedence of autoinstall file locations.  For example, would you expect the initrd-autoinstall.yaml to override other paths?

subiquity/autoinstall.py
```diff
@@ -0,0 +1,28 @@
...
+    config = {}
+    for path in source_paths:
+        with open(path) as fp:
+            c = yaml.safe_load(fp)
+        merge_config(config, c)
```
raharper: I would include a comment or a value in the merged config indicating the source yamls and the order of their merging so that when debugging a resulting autoinstall we can sort out how a value was modified after merging.

subiquity/controller.py
```diff
@@ -26,6 +26,45 @@
...
+    autoinstall_default = None
+
+    def __init__(self, app):
+        super().__init__(app)
+        self.autoinstall_applied = False
+        if app.autoinstall_config:
+            self.load_autoinstall_data(
+                app.autoinstall_config.get(
+                    self.autoinstall_key,
+                    self.autoinstall_default))
```
raharper: Pre-populating the autoinstall_config dictionary with keys which may be None (due to the default value being set) means that checking for presence of a key w.r.t to whether subiquity uses a default value or the user provided can be difficult.

subiquity/controllers/cmdlist.py
```diff
@@ -0,0 +1,43 @@
...
+class CmdListController(NoUIController):
+
+    autoinstall_default = []
+    cmds = ()
+
+    def load_autoinstall_data(self, data):
+        self.cmds = data
+
+    async def run(self):
+        for i, cmd in enumerate(self.cmds):
```
raharper: This seems to expect that self.cmds will return a list like object; if an autoinstall config does not have any early commands, wont the load_autoinstall_data from the base controller set this to a None which would fail to enumerate?

subiquity/controllers/cmdlist.py
```diff
@@ -0,0 +1,43 @@
...
+
+import asyncio
+
+from subiquity.controller import NoUIController
+
+
+class CmdListController(NoUIController):
+
+    autoinstall_default = []
+    cmds = ()
```
raharper: Should this be a list instead of tuple?

subiquity/controllers/filesystem.py
```diff
@@ -76,7 +78,30 @@ def __init__(self, app):
...
         self._probe_once_task = SingleInstanceTask(
             self._probe_once, propagate_errors=False)
-        self._probe_task = SingleInstanceTask(self._probe)
+        self._probe_task = SingleInstanceTask(
+            self._probe, propagate_errors=False)
+
+    def load_autoinstall_data(self, data):
+        log.debug("load_autoinstall_data %s", data)
+        if not self.interactive() and data is None:
+            data = {
```
raharper: This is curious, what's going on here?

subiquity/controllers/cmdlist.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def load_autoinstall_data(self, data):
+        self.cmds = data
```
raharper: In other controllers, you check if data is None, should you check here as well?

subiquity/controllers/mirror.py
```diff
@@ -38,20 +39,41 @@ class CheckState(enum.IntEnum):
...
         if 'country-code' in self.answers:
             self.check_state = CheckState.DONE
             self.model.set_country(self.answers['country-code'])
         self.lookup_task = SingleInstanceTask(self.lookup)
 
+    def load_autoinstall_data(self, data):
+        if data is None:
+            return
+        self.ai_data = data
+        if 'mirror' in data:
```
raharper: This check fails if users don't provide 'mirror' info in the autoinstall, as the loader sets data['mirror'] = None.
Either the loader shouldn't set default value to None so you can do key checks like this, or this needs to be
if data['mirror'] is not None:

subiquity/controllers/refresh.py
```diff
@@ -58,15 +60,47 @@ def __init__(self, app):
         self.new_snap_version = ""
 
         self.offered_first_time = False
+        self.active = self.interactive()
+
+    def load_autoinstall_data(self, data):
+        if data is not None and data.get('refresh'):
```
raharper: data.get('refresh-installer')  or 'refresh' ?

subiquity/controllers/ssh.py
```diff
@@ -33,12 +33,21 @@ def __init__(self, message, output):
...
         self._fetch_task = None
 
+    def load_autoinstall_data(self, data):
+        if data is None:
+            return
+        self.model.install_server = data.get('install_server', False)
+        self.model.authorized_keys = self.autoinstall_data.get(
+            'authorized-keys', [])
+        self.model.pwauth = self.autoinstall_data.get(
+            'allow-pw', not self.model.authorized_keys)
```
raharper: Policy decision?  Is this documented that one cannot set password and import ssh keys?

subiquity/controllers/welcome.py
```diff
@@ -25,7 +25,11 @@
 
 class WelcomeController(SubiquityController):
 
-    model_name = "locale"
+    autoinstall_key = model_name = "locale"
+    autoinstall_default = 'en_US.UTF-8'
```
raharper: C.UTF-8  is the default in cloud-images;

examples/autoinstall.yaml
```diff
@@ -0,0 +1,16 @@
...
+        - echo a
+        - sleep 1
+        - echo a
+keyboard:
+        layout: gb
+identity:
+        realname: ''
+        username: ubuntu
+        password: '$6$wdAcoXrU039hKYPd$508Qvbe7ObUnxoj15DRCkzC3qO7edjH0VV7BPNRDYK4QR8ofJaEEF2heacn0QgD.f8pO8SNp83XNdWG6tocBM1'
+        hostname: ubuntu
```
raharper: Does this example include all possible keys?  We might want to expand this with comments or examples of other accepted autoinstall keys.

subiquity/autoinstall.py
mwhudson: Yes, we need to make some decisions and document how all this works. I'm going to post to discource, probably tomorrow, about this.

subiquity/autoinstall.py
mwhudson: Good idea.

subiquity/controller.py
mwhudson: Yes, there are a few different ways of looking at this. This API seems reasonably ok. load_install_data has access to self.app.autoinstall_config if it needs it...

subiquity/controllers/cmdlist.py
mwhudson: "autoinstall_default = []" handles this

subiquity/controllers/cmdlist.py
mwhudson: perhaps autoinstall_default should be ()? I don't think it matters a great deal although having them different is indeed odd.

subiquity/controllers/filesystem.py
mwhudson: This is trying to have a non-interactive only default, I think.

subiquity/controllers/mirror.py
mwhudson: I don't understand your comment but I don't think it is correct.

This all needs to change anyway because I want to redo how mirrors are described, see https://github.com/CanonicalLtd/subiquity/pull/626 and https://wiki.ubuntu.com/FoundationsTeam/AutomatedServerInstalls/ConfigReference?action=diff&rev2=40&rev1=38

subiquity/controllers/refresh.py
mwhudson: 'refresh'. The format of the yaml is:

refresh-installer:
   refresh: yes

subiquity/controllers/ssh.py
mwhudson: Uh, the idea is only that the default of allow-pw is False iff a key is supplied. And I think that's what this code does but I've been thinking about udev too much to be sure.

subiquity/controllers/welcome.py
mwhudson: Hmm. subiquity is more targeted towards humans, but maybe not for autoinstall cases. Easy to change later in any case.

examples/autoinstall.yaml
mwhudson: No it certainly does not, and yes providing a more complete example would make sense.

mwhudson:
> I don't see anything that would affect existing interactive installs. I've left some review comments for the autoinstall bits.

Thanks for the review. I don't plan to make many changes to this branch in response to them, but I'll keep them in mind for later. I'll probably make some tweaks and merge this tomorrow.

PR #626, opened 2020-01-24
refactor how mirror data is handled

mwhudson:
This is prep work for allowing autoinstalls to specify a complete apt
config.

### DIFF
--- a/subiquity/models/mirror.py
+++ b/subiquity/models/mirror.py
@@ -41,28 +41,28 @@ DEFAULT = {
         ],
 }
 
-ARCHITECTURE = get_architecture()
-DEFAULT_MIRROR = get_mirror(DEFAULT, "primary", ARCHITECTURE)
-
 
 class MirrorModel(object):
 
     def __init__(self):
         self.config = copy.deepcopy(DEFAULT)
+        self.architecture = get_architecture()
+        self.default_mirror = self.get_mirror()
 
     def set_country(self, cc):
         uri = self.get_mirror()
-        if uri != DEFAULT_MIRROR:
+        if uri != self.default_mirror:
             return
         parsed = parse.urlparse(uri)
         new = parsed._replace(netloc=cc + '.' + parsed.netloc)
         self.set_mirror(parse.urlunparse(new))
 
     def get_mirror(self):
-        return get_mirror(self.config, "primary", ARCHITECTURE)
+        return get_mirror(self.config, "primary", self.architecture)
 
     def set_mirror(self, mirror):
-        config = get_arch_mirrorconfig(self.config, "primary", ARCHITECTURE)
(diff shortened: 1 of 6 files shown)


subiquity/models/mirror.py
```diff
@@ -13,35 +13,61 @@
...
+        {
+            "arches": ["default"],
+            "uri": "http://ports.ubuntu.com/ubuntu-ports",
+        },
+        ],
+}
+
+
+def get_default_mirror():
+    return get_mirror(DEFAULT, "primary", get_architecture())
```
raharper: Could we move the get_mirror() call to the object initialization so we can cache it?  It won't change during the lifetime of the install so we can save some execs and overhead if we keep it around.

raharper: Alternatively, we recently made use of python3's lru_cache() in cloud-init

from functools import lru_cache()

@lru_cache()
def get_default_mirror()

...

subiquity/models/mirror.py
```diff
@@ -13,35 +13,61 @@
...
 
     def __init__(self):
-        self.mirror = DEFAULT_MIRROR
+        self.config = copy.deepcopy(DEFAULT)
 
     def set_country(self, cc):
-        parsed = parse.urlparse(DEFAULT_MIRROR)
+        uri = self.get_mirror()
+        if uri != get_default_mirror():
+            return
```
raharper: Why do we return if the mirror and default mirror don't match?

mwhudson: This is looking ahead to autoinstall: if the user sets the mirror to http://my-private-mirror.internal we don't want to rewrite that to http://CC.my-private-mirror.internal.

subiquity/models/mirror.py
mwhudson: Tweaked a bit, see what you think now.

subiquity/models/mirror.py
```diff
@@ -13,35 +13,59 @@
...
+            "uri": "http://archive.ubuntu.com/ubuntu",
+        },
+        {
+            "arches": ["default"],
+            "uri": "http://ports.ubuntu.com/ubuntu-ports",
+        },
+        ],
+}
+
+ARCHITECTURE = get_architecture()
```
raharper: Doesn't this create a side-effect of executing dpkg if something imports this module?

raharper: AFAICT, nothing outside of this module uses DEFAULT_MIRROR or ARCHITECTURE and the API being used are the method son the MirrorModel itself, could we move these values to the class itself? then they can be properties that are cached after initialization in __init__?

mwhudson: Yeah, that makes more sense, thanks.

PR #627, opened 2020-01-24
stop testing on disco; start testing on eoan and focal

mwhudson:
Clearly we should do something like this, probert isn't installable in lxd containers though (see travis) :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ dist: xenial
 env:
   - IMAGE=ubuntu-daily:bionic SCRIPT="make lint"
   - IMAGE=ubuntu-daily:bionic SCRIPT=./scripts/runtests.sh
-  - IMAGE=ubuntu-daily:disco SCRIPT=./scripts/runtests.sh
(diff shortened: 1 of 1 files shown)


raharper:
looks like a bug in zfs packaging, not probert.

```
● zfs-mount.service - Mount ZFS filesystems
946   Loaded: loaded (]8;;file://tester/lib/systemd/system/zfs-mount.service/lib/systemd/system/zfs-mount.service]8;;; enabled; vendor preset: enabled)
947   Active: failed (Result: exit-code) since Fri 2020-01-24 03:39:51 UTC; 12ms ago
948     Docs: ]8;;man:zfs(8)man:zfs(8)]8;;
949  Process: 5228 ExecStart=/sbin/zfs mount -a (code=exited, status=1/FAILURE)
950 Main PID: 5228 (code=exited, status=1/FAILURE)"
```

xnox:
Trying to add this gain in https://github.com/CanonicalLtd/subiquity/pull/645/files

PR #628, opened 2020-01-30
fix installation hang after refreshing snap

mwhudson:
The installprogress controller waits for various model objects to be
configured before proceeding. But model objects whose state was loaded
from disk after a snap refresh were not marked as configured :( This
meant that, in particular, the postinstall steps waited on the language
being configured indefinitely. The fix is simple: mark model objects as
configured when their state is loaded from disk post-refresh.


mwhudson:
FWIW I've tested this by publishing it to edge/mwhudson and refreshing to that and it worked.

PR #629, opened 2020-02-10
write autoinstall config at end of install

mwhudson:


xnox:
Does this record autogenerated/mangled results, or the manual (original) user input?

I.e. does it record ESP of this many MB, /boot of this many MB, / on LVM of this many MB.

or

does it record Guided LVM partitioning, drive with serial BLAH

Because I'm not sure if want to export one or the other, or both, and comment them both. "This is how to explicitely hard code partitioning" and "This is how to use the default sensible things".

Maybe this is a bad example, because maybe we do only want hardcoded explicit storage yaml.

Maybe stuff like archive URL => encode it commented out, if the default one was used?

mwhudson:
It recorded the fully expanded form. My thinking was that this was the most helpful starting point for editing.

Commenting out the default values is interesting, not completely sure how to implement it. Ditto for recording which guided version was taken.

mwhudson:
Filed a trello card for the improvements, closing now.

mwhudson:
merging, rather!

PR #630, opened 2020-02-11
add guided luks options

mwhudson:
ping @xnox 

The only problem I have with this is that it looks pretty cluttered in 80x24. Unclear how much we should worry about that.


xnox:
i am happy with this UI / UX. Let's have edge channel build for live testing it from boot to complete the install.

PR #637, opened 2020-02-24
debian/console-conf*.service: add core18 services as After deps

anonymouse64:
These are currently applied with hacks in the core18 repo when building the core18 snap, but that was always meant to be a temporary thing until it was upstreamed... here's the upstreaming.

See https://github.com/CanonicalLtd/subiquity/pull/389 and https://github.com/snapcore/core18/pull/74 for a bit of historical context.


anonymouse64:
Closing this in favor of https://github.com/CanonicalLtd/subiquity/pull/638

PR #638, opened 2020-02-25
debian/console-conf.service: ensure correct order with respect to snapd services

bboozzoo:
The PR introduces changes to `(serial-)console-conf@.service` that ensure the right ordering on core20 devices.

The first bit for core18 was already included in https://github.com/CanonicalLtd/subiquity/pull/389 for bionic branch from @mvo5. The relevant service is named `core.start-snapd.service` in core20.
The last patches introduces an ordering dependency on `snapd.recovery-chooser-trigger.service`, which is new in core20. The change was moved from https://github.com/snapcore/core20/pull/23 based on advice from @xnox.


bboozzoo:
included a patch from #637

PR #639, opened 2020-02-26
vtoc support

mwhudson:
Stops trying to reformat dasd drives as gpt. Also displays a hint as to which partition type is being created.


mwhudson:
Merging this so it's easier to test

PR #640, opened 2020-02-29
do not count extended partition as using space

mwhudson:
also show extended/logical partitoin state in ui


okdana:
I tested this with the Focal daily installer and it appears to resolve the issue i reported in LP#1844118. Thanks!

PR #643, opened 2020-03-04
format dasds with cdl not ldl

mwhudson:


xnox:
Hahaha, yes we want cdl =) ldl is the linux one which is bad, we want cdl the compatible one. I don't think anybody actually uses ldl =)

frank-heimes:
You're right, it just there for 'historical reasons', but let still people
assume that this is the right one for Linux - which is not - as you already
pointed out.

PR #650, opened 2020-03-09
Update i18n docs

xnox:


README.md
```diff
@@ -12,6 +12,17 @@ also be discussed in the forum at https://forum.snapcraft.io.
 Our localization platform is Launchpad, translations are managed at
 https://translations.launchpad.net/ubuntu/+source/subiquity/
 
+To update translation template in launchpad:
+ * execute $ make i18n
```
vorlonofportland: I thought this was handled by the default make target, why do we need an explicit make invocation here?

also are there any bits that need committed to the git repo after running this command, or are they all build artifacts not in git?

README.md
```diff
@@ -12,6 +12,17 @@ also be discussed in the forum at https://forum.snapcraft.io.
...
+ * make an upload of subiquity into Ubuntu
+ * wait for it to migrate
```
vorlonofportland: Is this actually required?  I would've thought launchpad imports translations from -proposed, not just from the release pocket.

Also I would make this explicit, "wait for it to migrate to the release pocket".

README.md
xnox: Default target is all, which depends on `dryrun`, `probert`, `i18n`. Thus the default target is a larger superset which requires more things to be installed on the system. The actual minimum is to run the `i18n` target, or any of the others. What it does is simply invoke `python3 setup.py build` which has all the i18n hooks already hooked up. I'm specifying the minimal requirements here.

`.pot` changes timestamp on every build, thus i found it annoying to keep committed in git. Thus both `*.pot` and `*.mo` are ignored in `.gitignore`. The only thing that remains committed in git are the `po` translation files exported from the launchpad and documented on how to update those further down in the text.

README.md
xnox: If we ever do split publishing and build subiquity snap for a given lts series, then translations might need to come from the -updates pocket.

It was not obvious to me, at which point the translation templates are updated, as I see two uploads of translations per package =( once during -proposed and once during migration. And I was not around to request tarball export after it was published in -proposed but didn't migrate yet, to confirm that it is enough for it to just get published in -proposed. So i guess this is a precaution?

README.md
vorlonofportland: should we not perhaps make this a dependency of the source package's clean target, then, so it gets done automatically on a source upload?

README.md
vorlonofportland: OTOH, "wait for it to migrate" before doing what? this is the last thing in that section.  The next part is for translators to translate, which is a lot longer of a wait.  So maybe just omit the last line

xnox: right, i split this into two lists. Maybe it should be the first item in the "export translations & commit them" checklist?

README.md
xnox: oooh, yes.

xnox:
@vorlonofportland everything resolved i think.

vorlonofportland:
LGTM!

PR #654, opened 2020-03-20
wait for cloud init & extract autoinstall config

mwhudson:
I guess I'd like to know if any of the cloud-init developers think this is insane...

### DIFF
--- a/subiquity/cmd/tui.py
+++ b/subiquity/cmd/tui.py
@@ -19,9 +19,11 @@ import logging
 import os
 import fcntl
 import sys
+import time
 
-from subiquitycore.log import setup_logger
 from subiquitycore import __version__ as VERSION
+from subiquitycore.log import setup_logger
+from subiquitycore.utils import run_command
 
 from subiquity.core import Subiquity
 
@@ -97,7 +99,7 @@ AUTO_ANSWERS_FILE = "/subiquity_config/answers.yaml"
 
 
 def main():
-    # Preffer utils from $SNAP, over system-wide
+    # Prefer utils from $SNAP, over system-wide
     snap = os.environ.get('SNAP')
     if snap:
         os.environ['PATH'] = os.pathsep.join([
@@ -116,6 +118,10 @@ def main():
     logger = logging.getLogger('subiquity')
     logger.info("Starting SUbiquity v{}".format(VERSION))
     logger.info("Arguments passed: {}".format(sys.argv))
+
+    ci_start = time.time()
+    run_command(["cloud-init", "status", "--wait"])
+    logger.debug("waited %ss for cloud-init", time.time() - ci_start)
 
     block_log_dir = os.path.join(LOGDIR, "block")
     os.makedirs(block_log_dir, exist_ok=True)
--- a/subiquity/core.py
+++ b/subiquity/core.py
@@ -20,9 +20,12 @@ import platform
 import sys
 import traceback
 import urwid
-import yaml
 
 import apport.hookutils
+
+from cloudinit import safeyaml
+
+import yaml
 
 from subiquitycore.async_helpers import (
     run_in_thread,
@@ -157,7 +160,7 @@ class Subiquity(Application):
             if 'autoinstall' in cfg:
                 atomic_helper.write_file(
                     self.opts.autoinstall,
-                    yaml.dump(cfg['autoinstall']),
+                    safeyaml.dumps(cfg['autoinstall']),
                     mode=0o600)
         if os.path.exists(self.opts.autoinstall):
             with open(self.opts.autoinstall) as fp:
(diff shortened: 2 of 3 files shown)


subiquity/core.py
```diff
@@ -151,6 +151,14 @@ def make_screen(self):
             return s
 
     def run(self):
+        if not os.path.exists(self.opts.autoinstall):
+            from cloudinit import atomic_helper, stages
+            cfg = stages.Init().cfg
+            if 'autoinstall' in cfg:
+                atomic_helper.write_file(
+                    self.opts.autoinstall,
+                    yaml.dump(cfg['autoinstall']),
```
raharper: Unless your yaml is configured to use SafeDumper, you can use cloud-init's

from cloudinit import safeyaml
safeyaml.dumps(cfg)

subiquity/core.py
```diff
@@ -151,6 +151,14 @@ def make_screen(self):
             return s
 
     def run(self):
+        if not os.path.exists(self.opts.autoinstall):
+            from cloudinit import atomic_helper, stages
+            cfg = stages.Init().cfg
```
raharper: cfg = stages.Init().read_cfg()

cfg has some redactions, not clear to me if that will impact where you're looking.

Note the Init cfg property is /etc/cloud , /etc/cloud.cfg.d/* and datasource cfg, but it is NOT user-data.

I think what you want is something like this:
```
from cloudinit import stages
init = stages.Init()
init.read_cfg()
init.fetch(existing="trust")
cloud = init.cloudify()
full_config = cloud.cfg
```
This is the config that's passed to config modules and all of the system config, user-data, vendor-data have all be processed and merged (where processed means downloading from a URL, for example).  As an example, if the user-data were a multi-part mime, then you cannot read /var/lib/cloud/instance/user-data.txt directly (without doing more work)

```
>>> from cloudinit import stages
>>> init = stages.Init()
>>> init.read_cfg()
>>> init.fetch(existing='trust')
>>> cloud = init.cloudify()
>>> cloud.datasource.userdata
<email.mime.multipart.MIMEMultipart object at 0x7f1e4dc88760>
>>> cloud.cfg
{'final_message': 'The system is finally up, after $UPTIME seconds', 'package_update': True, 'package_upgrade': True, 'packages': ['ifupdown'], 'datasource_list': ['NoCloud', 'None'],
```

 

Now, if you from know where the autoinstall cfg is being supplied (for example the kernel cmdline is writing it to a file on /etc/cloud/cloud.cfg.d/ ... then we can make some short cuts here.

subiquity/cmd/tui.py
```diff
@@ -117,6 +119,10 @@ def main():
     logger.info("Starting SUbiquity v{}".format(VERSION))
     logger.info("Arguments passed: {}".format(sys.argv))
 
+    ci_start = time.time()
+    run_command(["cloud-init", "status", "--wait"])
+    logger.debug("waited %ss for cloud-init", time.time() - ci_start)
+
```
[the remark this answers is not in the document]
mwhudson: I think the point is that we don't know, although I expect most use cases will be ds=nocloud-net;s=$url on the kernel command line or using a config drive.

Anyway, this wait is mostly for cosmetic reasons: it looks silly when subiquity starts and then cloud-init final sprays stuff over the console.

subiquity/core.py
mwhudson: OK sure, I didn't realize you needed to be safe when _dumping_ yaml!? (and in any case, the yaml contains shell scripts that are going to be executed as root so if someone controls your autoinstall data it's not like they have to be sneaky at all)

subiquity/cmd/tui.py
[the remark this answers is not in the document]
raharper: OK, that makes sense.

subiquity/core.py
mwhudson: > This is the config that's passed to config modules and all of the system config, user-data, vendor-data have all be processed and merged (where processed means downloading from a URL, for example).

great, that's exactly what I want

> As an example, if the user-data were a multi-part mime, then you cannot read /var/lib/cloud/instance/user-data.txt directly (without doing more work)

yeah, this is the sort of thing I was worried about.

> Now, if you from know where the autoinstall cfg is being supplied (for example the kernel cmdline is writing it to a file on /etc/cloud/cloud.cfg.d/ ... then we can make some short cuts here.

I would prefer not to make any assumptions here if I can help it.

subiquity/core.py
raharper: We've had at least one security review scanning code and scolding for not using safe loaders and dumpers.

mwhudson:
I've pushed an update, please have a look when you can.

> Looking for more info on where/how autoinstall is fed into cloud-init; ie what will subiquity document on how to feed autoinstall to the live installer.

As I said in a comment somewhere, I expect people will use a config drive or nocloud-net, but I want to retain full flexibility.

PR #660, opened 2020-03-29
validate autoinstall data against a schema

mwhudson:
I wrote some documentation about all this: https://wiki.ubuntu.com/FoundationsTeam/AutomatedServerInstalls/ConfigSchema

This also fixes some rough edges around error handling and error-commands are now actually run.

### DIFF
--- a/subiquity/controllers/identity.py
+++ b/subiquity/controllers/identity.py
@@ -26,6 +26,17 @@ log = logging.getLogger('subiquity.controllers.identity')
 class IdentityController(SubiquityController):
 
     autoinstall_key = model_name = "identity"
+    autoinstall_schema = {
+        'type': 'object',
+        'properties': {
+            'realname': {'type': 'string'},
+            'username': {'type': 'string'},
+            'hostname': {'type': 'string'},
+            'password': {'type': 'string'},
+            },
+        'requiredProperties': ['username', 'hostname', 'password'],
+        'additionalProperties': False,
+        }
 
     def load_autoinstall_data(self, data):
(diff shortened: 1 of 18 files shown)


subiquity/controllers/identity.py
```diff
@@ -26,6 +26,17 @@
...
     autoinstall_key = model_name = "identity"
+    autoinstall_schema = {
+        'type': 'object',
+        'properties': {
+            'realname': {'type': 'string'},
+            'username': {'type': 'string'},
+            'hostname': {'type': 'string'},
+            'password': {'type': 'string'},
+            },
+        'required': ['username', 'hostname', 'password'],
```
xnox: Still i dream of not having a password, and installing ssh-in only servers.

subiquity/controllers/keyboard.py
```diff
@@ -29,6 +29,16 @@
...
 
     autoinstall_key = model_name = "keyboard"
+    autoinstall_schema = {
+        'type': 'object',
+        'properties': {
+            'layout': {'type': 'string'},
+            'variant': {'type': 'string'},
+            'toggle': {'type': 'string'},
+            },
+        'required': ['layout'],
```
xnox: I still kind of wish of skipping keyboard question on s390x, and just using C.UTF-8 / US? keyboard layout? I need to figure out if keyboard layout question makes sense on s390x.

mwhudson: Yeah, this would probably make sense. Low priority though? File a bug or something?

PR #661, opened 2020-03-29
be a bit less cavalier about running the install in autoinstall mode

mwhudson:
Currently, if subiquity finds an autoinstall config, it just runs. This would make it too easy to create a USB stick that destroys a machine it is plugged into at boot, so this branch stops and asks unless 'autoinstall' is on the kernel command line.

I added a small bit of docs about this: https://wiki.ubuntu.com/FoundationsTeam/AutomatedServerInstalls?action=diff&rev2=51&rev1=50

This branch also refactors the interactive confirmation so that it is shown even if we are not showing the filesystem screen at all.

### DIFF
--- a/subiquity/core.py
+++ b/subiquity/core.py
@@ -249,6 +249,11 @@ class Subiquity(Application):
                 answer = no
                 if 'autoinstall' in self.kernel_cmdline:
                     answer = yes
+                else:
+                    print(_("Confirmation is required to continue."))
+                    print(_("Add 'autoinstall' to your kernel command line to"
+                            " avoid this"))
+                    print()
                 prompt = "\n\n{} ({}|{})".format(
                     _("Continue with autoinstall?"), yes, no)
                 while answer != yes:
--- a/subiquity/core.py
+++ b/subiquity/core.py
@@ -16,6 +16,7 @@
 import logging
 import os
 import platform
+import shlex
 import signal
 import sys
 import traceback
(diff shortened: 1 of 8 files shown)


subiquity/core.py
```diff
@@ -182,6 +184,31 @@ def report_finish_event(self, name, description, status, level="INFO"):
...
+                    )
+                self.ui.body.show_stretchy_overlay(
+                    InstallConfirmation(self.ui.body, self))
+            else:
+                answer = 'no'
+                if 'autoinstall' in self.kernel_cmdline:
+                    answer = 'yes'
+                while answer != 'yes':
+                    print("\n\nContinue with autoinstall? (yes|no)")
+                    answer = input()
```
xnox: All of the above needs to be translated. Can we use existing translated strings from filesystem steps?

xnox: Also i wonder if we can use `systemd-ask-password` with like all the agents that ask via console, wall, and allow people to ssh in to answer via `systemd-tty-ask-password-agent --query` (or like wrap that into `subiquity-continue-autoinstall`).

xnox: Or like reuse any suitable string we can find in gnome/d-i to import translations.

mwhudson: I've added _()s. Happy to make this cleverer but perhaps not in this branch? In general the thing is going to be people adding autoinstall to the kernel command line. Maybe I should add a hint about that.

xnox: Should we add a boot menu item for it?

mwhudson: Maybe hidden away in an advanced options menu or something. I think it would be confusing to make it too prominent.

PR #662, opened 2020-03-31
apply network autoinstall config

mwhudson:


mwhudson:
> I guess people who want different installer & in-target networking are out of luck.

Yeah, I wonder about that sort of thing. There are always late-commands!

PR #664, opened 2020-03-31
support packages, debconf-selections keys in autoinstall

mwhudson:


mwhudson:
> Given there is no UI for this, and we do need to exercise it I do wonder if we can add a test that sets 'openssh-server/password-authentication' for example via debconf-selections rather than via cloud-init?

That's not a crazy idea I guess. Would be less racy than getting cloud-init to do it! (https://bugs.launchpad.net/cloud-init/+bug/1808072)

> Also, since we do have ssh running in live session, should we copy the host keys into target?

Uhh maybe? Want to ask Colin or Marc or someone? :) Too late at night for me to think about this.

> This thus should reduce our first-boot cloud-init to zero in common case, no?

Well cloud-init creates the user still...

mwhudson:
I should add some tests for this though.

xnox:
please resolve conflicts and then merge.

PR #670, opened 2020-04-01
write netplan directly to the target

mwhudson:
rather than via curtin and then cloud-init

### DIFF
--- a/subiquitycore/models/network.py
+++ b/subiquitycore/models/network.py
@@ -17,6 +17,7 @@ import copy
 import enum
 import ipaddress
 import logging
+import yaml
 from socket import AF_INET, AF_INET6
 
 from subiquitycore import netplan
@@ -253,10 +254,11 @@ class NetworkDev(object):
 class NetworkModel(object):
     """ """
 
-    def __init__(self, support_wlan=True):
+    def __init__(self, project, support_wlan=True):
         self.support_wlan = support_wlan
         self.devices_by_name = {}  # Maps interface names to NetworkDev
         self.has_network = False
+        self.project = project
 
     def parse_netplan_configs(self, netplan_root):
         self.config = netplan.Config()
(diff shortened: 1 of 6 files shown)


mwhudson:
For https://bugs.launchpad.net/subiquity/+bug/1869967

vorlonofportland:
My only reservation is that I have a trace of a memory that we might have changed this intentionally to go via curtin, either for reasons of timing of when things are written out, or sanity of interfaces, or something.  Do we need to dig a bit on what the rationale was for moving to this approach, before we change it back?

raharper:
> My only reservation is that I have a trace of a memory that we might have changed this intentionally to go via curtin, either for reasons of timing of when things are written out, or sanity of interfaces, or something. Do we need to dig a bit on what the rationale was for moving to this approach, before we change it back?

The change came from commit 36bcc539c39bf5083865f3c2273737e13455fb17 .  The commit message says moving to curtin config is a side-effect.

And previously it was written by curtin, including disabling cloud-init networking.

fb995c820918a0a995c8a63522d23d2c0ae6d1bb

subiquitycore/models/network.py
```diff
@@ -350,7 +352,14 @@ def get_all_netdevs(self, include_deleted=False):
     def get_netdev_by_name(self, name):
         return self.devices_by_name[name]
 
-    def render(self):
+    def stringify_config(self, config):
+        return '\n'.join([
+            "# This is the network config written by '{}'".format(
```
raharper, line 357: Maybe @vorlonofportland would like a specific breadcrumb here to indicate that this is _the_ source of truth for netplan config; that it is user editable, etc... maybe in the same form as it was written in the previous installer?

vorlonofportland: FTR I'm ok with this as-is, I think breadcrumbs are most important when we need to communicate to a user /not/ to edit a thing and to look elsewhere.

PR #671, opened 2020-04-01
implement extensions to curtin storage syntax 

mwhudson:
As described in the wiki


mwhudson:
Haven't tested this beyond making sure it doesn't break non-autoinstall scenarios yet!

raharper:
Which wiki is that?

mwhudson:
Oh sorry, it's https://wiki.ubuntu.com/FoundationsTeam/AutomatedServerInstalls/ConfigReference#storage

PR #675, opened 2020-04-01
things to improve debugging launchpad bug 1865027

mwhudson:


mwhudson:
Oh now I remember why I unmount early: we ignore mounted disks when block probing. Needs some thinking.

PR #681, opened 2020-04-03
avoid deleting virtual devices on every network change

mwhudson:
For some reason this also involved sitting harder on networkd to stop it doing things while we reconfigure the NICs.


mwhudson:
tested this in bionic and focal vms, it's an improvement

PR #684, opened 2020-04-06
better experience for partially-interactive installs

mwhudson:
When some screens are interactive and some not, the current behaviour if applying the autoinstall config takes some time is very confusing. This branch shows a progress view if applying autoinstall config takes more than 0.1 s (and ensures it shows progress for at least a second if it shows at all to avoid flickering).

It uses the progress view from the installation, with some added hair because subiquity events can overlap, unlike curtin events.

### DIFF
--- a/subiquity/context.py
+++ b/subiquity/context.py
@@ -1,0 +1,26 @@
+# Copyright 2020 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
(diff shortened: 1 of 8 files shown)


mwhudson:
FWIW you can test with a command like

TTY=$(tty); PYTHONPATH=/home/mwhudson/src/subiquity:/home/mwhudson/src/probert python3  -m subiquity --dry-run    --machine-config examples/existing-partitions.json --autoinstall examples/autoinstall-interactive.yaml --kernel-cmdline "console=${TTY#/dev/}"

I'll fix the cryptic variable names.

curtin does not have translations and we've been exposing curtin descriptions in the UI forever. This does make this aspect worse but I'm not sure what we can really do about that, as you say autoinstalls kind of require english.

PR #692, opened 2020-04-07
console_conf: various recover chooser tweaks

bboozzoo:
A bunch of tweaks for the recovery chooser UI based on the discussion with @niemeyer.

Most important changes:
- drop the ABORT button
- introduce BACK button, goes back to the correct previous screen
- sort actions in the order of run -> recover -> install -> rest
- capitalize action title strings
- simplify current model actions screen
- user hints in the confirm screen for well known modes
- generic hint for unknown modes
- use model's display name in the top banner
- use `(installed)` instead of `(current)` in the all-systems screen for the current system

### DIFF
--- a/console_conf/ui/views/chooser.py
+++ b/console_conf/ui/views/chooser.py
@@ -51,19 +51,17 @@ class ChooserBaseView(BaseView):
     title = "Ubuntu Core"
 
     def __init__(self, current, scr):
+        super().__init__(scr)
+
         if current is not None:
             self.title = current.model.display_name
-
-        super().__init__(scr)
 
 
 def by_preferred_action_type(action):
     """Order action entries by having the 'run' mode first, then 'recover', then
     'install', the rest is ordered alphabetically."""
-    return (action.mode != "run",
-            action.mode != "recover",
-            action.mode != "install",
-            action.title.lower())
+    priority = {"run": 0, "recover": 1, "install": 2}
+    return (priority.get(action.mode, 100), action.title.lower())
 
 
 class ChooserCurrentSystemView(ChooserBaseView):
--- a/console_conf/ui/views/chooser.py
+++ b/console_conf/ui/views/chooser.py
@@ -176,8 +176,8 @@ class ChooserConfirmView(ChooserBaseView):
         "run": "Continue running the system without any changes.",
         "recover": ("You have requested to reboot the system into recovery "
                     "mode."),
-        "install": ("You are about to {action} the system version {version} "
-                    "for {model} from {publisher}.\n\n"
+        "install": ("You are about to {action_lower} the system version "
+                    "{version} for {model} from {publisher}.\n\n"
                     "This will remove all existing user data on the "
                     "device.\n\n"
                     "The system will reboot in the process."),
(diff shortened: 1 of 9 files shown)


bboozzoo:
@xnox please don't merge it yet. I'll let you know when it's ready.

console_conf/ui/views/chooser.py
```diff
@@ -144,56 +151,67 @@ def __init__(self, controller, systems):
...
+        self.controller.back()
 
 
-class ChooserConfirmView(BaseView):
-    title = "Ubuntu Core"
-    excerpt = ("Summary of the selected action.")
+class ChooserConfirmView(ChooserBaseView):
+
+    canned_summary = {
+        "run": "Continue running the system without any changes.",
```
bboozzoo, line 176: @degville would appreciate your input on the messages here. This is presented to the user as a summary of the selected action.

The `default_summary` is a fallback messaage we use for an action that we don't know about.

degville: As I know you want to merge this, I've had a quick glance over the text and it looks good.

The only thing I'd probably tweak is the default summary to something like 'Make sure you understand the consequences of performing this action.`, or maybe add a little more help (or a link) on what those consequences might be.

console_conf/ui/views/welcome.py
```diff
@@ -51,6 +51,13 @@ def confirm(self, result):
...
+        "System mode selection triggered. Proceed to select one of available "
+        "systems and actions."
     )
+
+    def __init__(self, controller):
+        current = controller.model.current
+        if current is not None:
+            self.title = current.model.display_name
+
+        super().__init__(controller)
```
zyga, line 63: Could this init happen before we assign to `self.title`?

console_conf/ui/views/chooser.py
```diff
@@ -47,23 +47,36 @@
...
+        if current is not None:
+            self.title = current.model.display_name
+
+        super().__init__(scr)
+
+
+def by_preferred_action_type(action):
+    """Order action entries by having the 'run' mode first, then 'recover', then
+    'install', the rest is ordered alphabetically."""
+    return (action.mode != "run",
```
zyga, line 63: I find this key function unusual as it returns a tuple (bool, bool, bool, str), I was expecting a (int, str)

```
priority = {"run": 0, "recover": 1, "install": 2}
return (priority.get(action.mode, 100), action.title.lower())
```

console_conf/ui/views/chooser.py
```diff
@@ -47,23 +47,36 @@
...
     title = "Ubuntu Core"
 
-    def __init__(self, controller, current, has_more=False):
-        fmt = "Select one of available actions for \"{}\" by \"{}\"{}."
-        maybe_more = " or view all available systems" if has_more else ""
-        excerpt = fmt.format(current.model.display_name,
-                             current.brand.display_name,
-                             maybe_more)
+    def __init__(self, current, scr):
+        if current is not None:
```
zyga, line 54: Same question, could we call `__init__()` as the first thing and only then assign to `self.title`? I find this style more in line with what ought to happen in a perfect scenario.

console_conf/ui/views/chooser.py
bboozzoo: Nice! I'll push that

console_conf/ui/views/welcome.py
bboozzoo: Fixed this and the other one.

bboozzoo:
@xnox this is ready now.

bboozzoo:
Thanks!

PR #696, opened 2020-04-09
Switch to core18

xnox:
because all the cool kids are switching this week too.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,4 +1,5 @@
 name: subiquity
+base: core18
 version: git
 summary: Ubuntu installer
 description: The Ubuntu server installer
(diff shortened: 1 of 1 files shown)


xnox:
Ok this builds i think. Snaps will be available at https://launchpad.net/~xnox/+snap/subiquity soon

I guess it needs testing?

mwhudson:
+1 assuming it works.

I wouldn't expect subtle failures tbh, if subiquity actually starts it's probably ok

mwhudson:
Does this need to be coordinated with any livecd-rootfs changes?

xnox:
> Does this need to be coordinated with any livecd-rootfs changes?

Well, we should switch if and when lxd switches. livecd-rootfs does the right thing.

mwhudson:
OK, cool, I guess you are staying in touch with the relevant people?

I guess we should test upgrades from 18.04.4 before releasing this to stable. It should work fine but...

xnox:
lxd changed, and so should we.

PR #701, opened 2020-04-12
fix typo at openssh-server installer

m-shibata:
According to ConfigReference document, key name is "install-server" instead of "install_server" to install openssh-server via autoinstall.


mwhudson:
Good spot, thanks!

PR #703, opened 2020-04-13
do not include unreferenced devices in curtin storage config 

mwhudson:
This also includes the fix for including the partition number in the config for preserved partitions, as we will quite often be not including some partitions in the config now.

### DIFF
--- a/scripts/validate-yaml.py
+++ b/scripts/validate-yaml.py
@@ -14,12 +14,11 @@ class StorageChecker:
     def _check_partition(self, action):
         assert 'device' in action
         assert 'size' in action
+        assert 'number' in action
         assert action['device'] in self.actions
         assert 'ptable' in self.actions[action['device']]
         if action['flag'] in ('boot', 'bios_grub', 'prep'):
             assert self.actions[action['device']]['type'] == 'disk'
-        if action.get('preserve', True):
-            assert 'number' in action
 
     def _check_format(self, action):
         assert 'volume' in action
--- a/scripts/validate-yaml.py
+++ b/scripts/validate-yaml.py
@@ -18,6 +18,8 @@ class StorageChecker:
         assert 'ptable' in self.actions[action['device']]
         if action['flag'] in ('boot', 'bios_grub', 'prep'):
             assert self.actions[action['device']]['type'] == 'disk'
+        if action.get('preserve', True):
+            assert 'number' in action
 
     def _check_format(self, action):
         assert 'volume' in action
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1474,6 +1474,8 @@ class FilesystemModel(object):
         def can_emit(obj):
             for dep in dependencies(obj):
                 if dep.id not in emitted_ids:
+                    if dep not in work:
+                        work.append(dep)
                     return False
             if isinstance(obj, Mount):
                 # Any mount actions for a parent of this one have to be emitted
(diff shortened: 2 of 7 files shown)


scripts/validate-yaml.py
```diff
@@ -18,6 +18,8 @@ def _check_partition(self, action):
         assert 'ptable' in self.actions[action['device']]
         if action['flag'] in ('boot', 'bios_grub', 'prep'):
             assert self.actions[action['device']]['type'] == 'disk'
+        if action.get('preserve', True):
+            assert 'number' in action
```
raharper, line 22: My preference would be to always have number present for partition.  There's no reason not to set this value.

subiquity/models/filesystem.py
```diff
@@ -927,7 +928,10 @@ def available(self):
 
     @property
     def _number(self):
-        return self.device._partitions.index(self) + 1
+        if self.preserve:
+            return self.number
+        else:
+            return self.device._partitions.index(self) + 1
```
raharper, line 934: This works because we don't currently allow adding additional partitions on a disk which is preserved right?

subiquity/models/tests/test_filesystem.py
```diff
@@ -1043,3 +1046,29 @@ def test_lv_remaninig(self):
...
+        disk1p1 = make_partition(model, disk1, preserve=True)
+        disk2p1 = make_partition(model, disk2, preserve=True)
+        fs = model.add_filesystem(disk1p1, 'ext4')
+        model.add_mount(fs, '/')
+        rendered_ids = {action['id'] for action in model._render_actions()}
+        self.assertTrue(disk1.id in rendered_ids)
+        self.assertTrue(disk1p1.id in rendered_ids)
+        self.assertTrue(disk2.id not in rendered_ids)
+        self.assertTrue(disk2p1.id not in rendered_ids)
+
```
raharper, line 1063: Does this mean that users can create partitions in the UI and if they don't format or mount them we'd not include that in the storage config?  I suspect users would not be happy about that.  Or are you just constructed probed partitions which were already present but not used?

subiquity/ui/views/filesystem/tests/test_partition.py
```diff
@@ -129,8 +129,8 @@ def test_edit_existing_partition_mountpoints(self):
         partition = model.add_partition(disk, 512*(2**20))
         partition.preserve = True
         fs = model.add_filesystem(partition, "ext4")
-        fs.preserve = True
         model._orig_config = model._render_actions()
+        fs.preserve = True
```
raharper, line 133: Odd change, does it matter?  if so why?

scripts/validate-yaml.py
mwhudson: OK.

subiquity/models/filesystem.py
mwhudson: Yes, either all partitions are old or all are new. We will need to think much harder about partition numbers when we support editing partition tables.

subiquity/models/tests/test_filesystem.py
mwhudson: > Does this mean that users can create partitions in the UI and if they don't format or mount them we'd not include that in the storage config? I suspect users would not be happy about that.

No, those partitions would have preserve=False and so be included. I can add a test to be explicit about this.

> Or are you just constructed probed partitions which were already present but not used?

Right.

subiquity/ui/views/filesystem/tests/test_partition.py
mwhudson: This test is poking around in the guts of things for reasons I don't really remember :/

But the reason for this change is reasonably clear: we want the fs to be in _orig_config and now that render_actions does not include things with preserve = True we need to set that after calling it.

PR #710, opened 2020-04-15
subiquity: add --ssh option to print ssh login details.

xnox:
Add --ssh option, which simply prints ssh login details and quits.

```
$ sudo python3 -m subiquity --ssh --dry-run

It is possible to connect to the installer over the network, which
might allow the use of a more capable terminal.

To connect, SSH to any of these addresses:

    installer@192.168.1.103
    installer@2a01:4b00:85fd:d700:c5d5:d33:cfd5:9062
    installer@2a01:4b00:85fd:d700:f17b:dafd:6e61:e6fb
    installer@fd00::c5d5:d33:cfd5:9062
    installer@fd00::f022:ccd1:f8de:53f5

The password you should use is "rAnd0Mpass".

The host key fingerprints are:

    RSA     SHA256:IcNSi/MT/PE3seH5SWYNKld31Quq8GW6K8Tzqt5gHtE
    ECDSA   SHA256:EvXsL7GMuLSjchGhN1oeZWlAsmIP3nimEazn00mr1sc
    ED25519 SHA256:iRM07uEWhq+uouCMRBSZHbeKkC6mT64JHmlyZE2DTnc
```

This is similar to console_conf write-login-details standalone, but uses (refactored) subiquity messaging.

This will be used on s390x to print this information on the line-based consoles.

### DIFF
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -108,12 +108,7 @@ def write_login_details(fp, username, ips):
                                            first_ip=first_ip,
                                            version=version))
 
-
-def write_login_details_standalone():
-    owner = get_device_owner()
-    if owner is None:
-        print("No device owner details found.")
-        return 0
+def get_ips_standalone():
     from probert.prober import Prober
     from subiquitycore.models.network import NETDEV_IGNORED_IFACE_TYPES
     prober = Prober()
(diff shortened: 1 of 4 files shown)


mwhudson:
Can you move get_ips_standalone to somewhere in subiquitycore please? I don't want subiquity to import console_conf or vv. Also you have some lint somewhere. Otherwise fine!

xnox:
> I guess this will look a bit stupid if get_installer_password returns None, but well whatever. Merge when travis is green.

yes, it does look stupid. but like i think the texts function should say "no password set" and like also print any imported ssh keys.

PR #711, opened 2020-04-16
get apport python package and hooks from focal branch

mwhudson:
This was harder than it should have been!


mwhudson:
I tested this in a vm by triggering an error and there was no traceback and the casper md5 results were reported so good enough.

PR #714, opened 2020-04-16
wipe more devices

mwhudson:
wipe a volume before formatting it, wipe a device that used to be part of a compound device.

### DIFF
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -399,7 +399,8 @@ class FilesystemController(SubiquityController):
 
     def create_filesystem(self, volume, spec):
         if spec['fstype'] is None:
-            volume.wipe = None
+            if getattr(volume, 'flag', None) != 'prep':
+                volume.wipe = None
             fstype = volume.original_fstype()
             if fstype is None:
                 return None
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -399,12 +399,14 @@ class FilesystemController(SubiquityController):
 
     def create_filesystem(self, volume, spec):
         if spec['fstype'] is None:
+            volume.wipe = None
             fstype = volume.original_fstype()
             if fstype is None:
                 return None
             preserve = True
         else:
             fstype = spec['fstype']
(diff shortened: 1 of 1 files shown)


subiquity/controllers/filesystem.py
```diff
@@ -399,12 +399,15 @@ def delete_mount(self, mount):
 
     def create_filesystem(self, volume, spec):
         if spec['fstype'] is None:
+            if getattr(volume, 'flag', None) != 'prep':
+                volume.wipe = None
```
raharper, line 403: we *always* wipe prep, IIRC;  leave a comment here?

subiquity/controllers/filesystem.py
```diff
@@ -480,6 +483,8 @@ def delete_raid(self, raid):
         self.clear(raid)
         for p in list(raid.partitions()):
             self.delete_partition(p)
+        for d in raid.devices | raid.spare_devices:
+            d.wipe = 'superblock'
```
raharper, line 487: This isn't strictly required; curtin already wipes both raid members and spares before using them; but I think it's prudent since the wipe setting will ensure curtin looks to see if they have any holders that need removal.

subiquity/controllers/filesystem.py
```diff
@@ -497,6 +502,7 @@ def delete_volgroup(self, vg):
         for lv in list(vg.partitions()):
             self.delete_logical_volume(lv)
         for d in vg.devices:
+            d.wipe = 'superblock'
```
raharper, line 505: Same comment as above applies here.

subiquity/controllers/filesystem.py
mwhudson: yeah good idea

subiquity/controllers/filesystem.py
mwhudson: Yeah this change was from before I read the bug properly :) It still seems conceptually valid I guess?

subiquity/controllers/filesystem.py
mwhudson: Same reply applies too.

PR #717, opened 2020-04-17
stop staging libc6 now we are core18 based

mwhudson:
Don't think we need this any more, needs testing though.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -36,8 +36,6 @@ parts:
      - "*"
      - "-lib/python*/site-packages/_yaml.*.so"
      - "-lib/python*/site-packages/jsonschema"
-    stage-packages:
-     - libc6
   apport:
     plugin: python
     source-type: bzr
@@ -70,7 +68,6 @@ parts:
       - python3-urwid
     stage-packages:
       - cloud-init
-      - libc6
       - libsystemd0
       - iso-codes
       - lsb-release
@@ -103,7 +100,6 @@ parts:
         debconf-communicate user-setup | \
         cut -d ' ' -f 2- > $SNAPCRAFT_PART_INSTALL/users-and-groups
       cp /usr/lib/user-setup/reserved-usernames $SNAPCRAFT_PART_INSTALL/
-    stage-packages: [libc6]
     stage:
       - users-and-groups
       - reserved-usernames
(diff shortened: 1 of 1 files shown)


xnox:
This is correct. I wonder if we can drop even more deps. Because core snap does have console-conf, and thus has python and many of our deps.

mwhudson:
There is a mighty lot of junk in the snap we could prune out (mostly because I can't figure out how to make things work with python-disutils-extra in build-packages but not in stage-packages). I hadn't thought about using the python from the core18 snap though...

snapcraft.yaml
```diff
@@ -128,7 +123,7 @@ parts:
   probert:
     plugin: python
     build-packages: [python-setuptools, libnl-3-dev, libnl-genl-3-dev, libnl-route-3-dev]
-    stage-packages: [libc6, libnl-3-dev, libnl-genl-3-dev, libnl-route-3-dev]
+    stage-packages: [libnl-3-dev, libnl-genl-3-dev, libnl-route-3-dev]
```
xnox, line 126: We don't need -dev packages staged, just the runtime libs, no? And also, because probert is in the core18 snap, we don't need to stage any of these libnl* libs anyway.

mwhudson: Hm good point, lets try that.

mwhudson:
> Because core snap does have console-conf, and thus has python

Thinking about this, we shouldn't invoke binaries from the base snap in a classic snap, because they will load libraries from the host system. (I guess we could carefully invoke it with /lib64/ld-linux-x86-64.so.2 --library-rpath /snap/core18/current/usr/lib/x86_64-linux-gnu/:... python3, but well...)

xnox:
> > Because core snap does have console-conf, and thus has python
> 
> Thinking about this, we shouldn't invoke binaries from the base snap in a classic snap, because they will load libraries from the host system. (I guess we could carefully invoke it with /lib64/ld-linux-x86-64.so.2 --library-rpath /snap/core18/current/usr/lib/x86_64-linux-gnu/:... python3, but well...)

I think we should be ok to use libraries from our snap, base snap, then host. Or maybe like fail.

It seems we are staging openssl inside our snap, which i feel iffy about. I'd rather use openssl from the core snap.

I'm not sure about the envrionment we are in, but i.e. i think the desktop framework people do have wrappers that setup all that, and allow loading things from "framework" snaps too, and plugins from the host.

mwhudson:
> I think we should be ok to use libraries from our snap, base snap, then host. Or maybe like fail.

Yes, libraries from our snap and base snap are fine. Host is definitely iffy.

> It seems we are staging openssl inside our snap, which i feel iffy about. I'd rather use openssl from the core snap.

Hmm yes, that does sound iffy. Does it get dragged in via Python, I wonder.

> I'm not sure about the envrionment we are in, but i.e. i think the desktop framework people do have wrappers that setup all that, and allow loading things from "framework" snaps too, and plugins from the host.

Hmm.

PR #736, opened 2020-04-28
run "stty sane" in a non-interactive install on the serial line

mwhudson:
for https://bugs.launchpad.net/subiquity/+bug/1874052


mwhudson:
tested this in a vm and it helps

PR #737, opened 2020-04-28
have the lv created by the guided lvm option take up 80% of the vg

mwhudson:
For https://bugs.launchpad.net/subiquity/+bug/1785321

### DIFF
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -750,11 +750,26 @@ class FilesystemController(SubiquityController):
         spec = dict(name="ubuntu-vg", devices=set([part]))
         if lvm_options and lvm_options['encrypt']:
             spec['password'] = lvm_options['luks_options']['password']
-        # create volume group on partition
         vg = self.create_volgroup(spec)
+        # There's no point using LVM and unconditionally filling the
+        # VG with a single LV, but we should use more of a smaller
+        # disk to avoid the user running into out of space errors
+        # earlier than they probably expect to.
+        if vg.size < 10 * (2 << 30):
+            # Use all of a small (<10G) disk.
+            lv_size = vg.size
+        elif vg.size < 20 * (2 << 30):
+            # Use 10G of a smallish (<20G) disk.
+            lv_size = 10 * (2 << 30)
+        elif vg.size < 200 * (2 << 30):
+            # Use half of a larger (<200G) disk.
+            lv_size = vg.size // 2
+        else:
+            # Use at most 100G of a large disk.
+            lv_size = 100 * (2 << 30)
         self.create_logical_volume(
             vg=vg, spec=dict(
-                size=vg.size*8//10,
+                size=lv_size,
                 name="ubuntu-lv",
                 fstype="ext4",
                 mount="/",
--- a/subiquity/controllers/filesystem.py
+++ b/subiquity/controllers/filesystem.py
@@ -752,12 +752,9 @@ class FilesystemController(SubiquityController):
             spec['password'] = lvm_options['luks_options']['password']
         # create volume group on partition
         vg = self.create_volgroup(spec)
-        target_size = dehumanize_size("4G")
-        if target_size > vg.free_for_partitions:
-            target_size = int(vg.size*0.8)
         self.create_logical_volume(
             vg=vg, spec=dict(
-                size=target_size,
+                size=vg.size*8//10,
                 name="ubuntu-lv",
                 fstype="ext4",
                 mount="/",
--- a/subiquity/ui/views/filesystem/guided.py
+++ b/subiquity/ui/views/filesystem/guided.py
@@ -140,9 +140,10 @@ If the platform requires it, a bootloader partition is created on the disk.
 
 If you choose to use LVM, two additional partitions are then created,
 one for /boot and one covering the rest of the disk. An LVM volume
-group is created containing the large partition. A 4 gigabyte logical
-volume is created for the root filesystem. It can easily be enlarged
-with standard LVM command line tools.
+group is created containing the large partition. A logical volume is
+created for the root filesystem, sized using some simple heuristic. It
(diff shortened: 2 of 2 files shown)


subiquity/controllers/filesystem.py
```diff
@@ -752,12 +752,9 @@ def guided_lvm(self, disk, lvm_options=None):
             spec['password'] = lvm_options['luks_options']['password']
         # create volume group on partition
         vg = self.create_volgroup(spec)
-        target_size = dehumanize_size("4G")
-        if target_size > vg.free_for_partitions:
-            target_size = int(vg.size*0.8)
         self.create_logical_volume(
             vg=vg, spec=dict(
-                size=target_size,
+                size=vg.size*8//10,
```
raharper, line 757: Nifty, floor division operator.

vorlonofportland, line 757: Given that this is on LVM, I think preallocating 80% to the rootfs is excessive.  The user can always expand the rootfs later as necessary, but shrinking in order to use the space for something else can't be done online (best case: create a new volume that's <= 20%, copy rootfs data to the other volume, delete the original volume).

I think the default size should be the greater of (50% of VG, 2x the minimum required to copy the squashfs), with provisions for the case where the total VG space is < 2x the squashfs size.

2x gives reasonable "default" headroom for logs, package upgrades, new package installations, etc; and the user can and should adjust to taste from there based on their requirements.

But if you have a 2TB VG, allocating 80% of it to the rootfs is pretty ridiculous in terms of why you use LVM at all.

Though anyway, ZFS solves all of this so it would be more useful to more people in the long term to get the ZFS root story haken out.

xnox: % only makes sense if there is a high/low watermarks.

If disk is less than 10GB, use all of it.
If disk is between 10GB and 20GB, use 10GB.
If disk is between 20GB and 200GB, use 50%.
If disk is larger than 200GB, use 100GB.

Thus we use upto 10GB, on disks upto 20GB in size.
Then we go to 50%, but cap out at 100GB.

Because 1TB / 2TB / 1PB disks do not need or want root to be one lvm volume..... Or at least so I hope.

Can we get info from like bootstrap how the deploy things? or maas?

mwhudson:
@vorlonofportland oked this approach on IRC, landing

trs80:
Will this make it into a 20.04 point release?

mwhudson:
Yes, and it will be available as a snap update in the installer much sooner than that -- maybe tomorrow.

zarej:
I have added 
```
autoinstall:
  refresh-installer:
    update: yes
  storage:
    layout:
      name: lvm
```
But it still creates root partition 4GB, how to get updated subiquity during installation to have this changes?

mwhudson:
Add `channel: edge` to the refresh-installer section. It's taken longer than I'd hoped but this change should still be in stable within the next few days...

mwhudson:
I've released the update containing this to stable now fwiw.

zarej:
Cool, works! What about policy for swap file size how it is calculated?

mwhudson:
Swapfiles just gets curtin's default. [The docs](https://curtin.readthedocs.io/en/latest/topics/config.html#swap) just say this uses "a heuristic" without going into details. I'd sure the default can be tweaked but a closed github PR is probably not the place :) File a bug on launchpad?

enoch85:
IMHO, "Use the entire disk" should be just that. Sure, you could add the full disk in custom settings, but there's a custom disk option for that.

In 18.04 the disk was 100% used, when opting for "Use entire disk", now why this change?

vorlonofportland:
The disk is 100% used, by the volume group.

It is not sensible to create a volume group and fill it 100% with a single volume, because that is a wasteful additional indirection that provides the user with none of the value of LVM.  Because online *growing* of filesystems is supported, online *shrinking* is not.

enoch85:
@vorlonofportland 

> Because online growing of filesystems is supported, online shrinking is not.

I won't argue with that. 

Though maybe you should consider to make it clear during setup and change wording. Right now it's easy to believe you get 100% allocated of the disk, while in reality you need to either customize the disk layout during setup, or do it afterwards.

Personally, I always use 100% of the disk, even with LVM, because LVM makes it easier to extend the storage in a virtual environment (and in a barebone as well ofc).

You seem to have a pretty set opinion based on your [previous post](https://github.com/CanonicalLtd/subiquity/pull/737#discussion_r416829663), but maybe it would be better to:

1. Use the entire disk as advertised.
2. If you don't want that, then use the Custom option.

Novice users will choose the first option, and then realize that they didn't get what they thought and  have no idea how to extend without using their favourite search engine - whilst more advanced users can choose to customize the disk size if they want smaller.

I'm no PRO at all, but I'm talking from experience having worked with mostly novice (Linux) users (coming from Windows) during the past 6 years - providing the VM for both ownCloud and later Nextcloud. Lately I've got many emails regarding this, considering it to be a bug in the installer... "I wanted my full disk, but I only got 50%, what?" 

Even if I'm thankful for the support hours it's generating I think it should be fixed in master.

mwhudson:
The default of not having the lv fill the vg seems the better option because as Steve says, growing the lv is reasonably simple and the reverse really is not. But we could make it easier to get the "fill the vg" behaviour, by e.g. having an option on the guided page perhaps.

mwhudson:
Also closed PRs are not really the place to discuss this sort of thing, maybe a bug on Launchpad?

enoch85:
@mwhudson Sorry, I know this is an old PR and everyone gets pinged each time.

I had a look, and it seems like there's already an issue that relates to this: https://bugs.launchpad.net/subiquity/+bug/1876892/comments/2

mwhudson:
Thanks for the bug. I don't mind the pings, it's more that we do sometimes look through all open subiquity bugs looking for things to work on but funnily enough never look through closed PRs for the same!

PR #738, opened 2020-04-29
Stop all probing during installation

mwhudson:


mwhudson:
> what is the intended state_dir structure?

There has been no coherent plan about this so far. It's just a bag of stuff.

> Instead of blocking probe only during install, would it make more sense to block other subiquity instances from proceeding at all?

That would be nice, see just pushed commits for an implementation of that, but due to subiquity being asynchronous it ends up being mostly independent of blocking probing.

subiquity/core.py
```diff
@@ -152,6 +158,33 @@ def __init__(self, opts, block_log_dir):
...
+                )
+            self.install_lock_file.seek(0)
+            tty = self.install_lock_file.read().strip()
+            self.install_running = InstallRunning(self.ui.body, self, tty)
+            self.ui.body.show_stretchy_overlay(self.install_running)
+            schedule_task(self._hide_install_running())
+
+    async def _hide_install_running(self):
+        await run_in_thread(fcntl.flock, self.install_lock_file, fcntl.LOCK_SH)
+        fcntl.flock(self.install_lock_file, fcntl.LOCK_UN)
```
raharper, line 184: We lock and unlock immediately?

subiquity/ui/views/installprogress.py
```diff
@@ -227,3 +228,36 @@ def cancel(self, sender):
         self.parent.remove_overlay()
         if isinstance(self.parent, ProgressView):
             self.parent.show_continue()
+
+
+running_text = _("""\
+The installer running on {tty} is currently installing the system.
+
+You can wait for this to complete or switch to a shell.
```
raharper, line 236: Nice!

mwhudson:
So I'd like to add the initial bit of this first, I've pushed that as https://github.com/CanonicalLtd/subiquity/pull/742. This might take a bit longer, but your comments are great, I'll get to them soon!

subiquity/controllers/filesystem.py
```diff
@@ -132,36 +134,40 @@ async def _probe_once(self, restricted):
...
-                    block_discover_log.exception(
-                        "block probing failed restricted=%s", restricted)
-                    report = self.app.make_apport_report(
-                        kind, "block probing", interrupt=False)
-                    self._crash_reports[restricted] = report
-                    continue
-                break
-        self.convert_autoinstall_config()
+            await run_in_thread(
+                fcntl.flock, self.app.install_lock_file, fcntl.LOCK_SH)
```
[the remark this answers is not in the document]
mwhudson: Yeah adding logging here probably makes sense.

The state_dir is entirely managed by subiquity code, nothing in snapd or systemd or anything should touch it.

subiquity/controllers/filesystem.py
```diff
@@ -132,36 +134,40 @@ async def _probe_once(self, restricted):
...
+                    except Exception:
+                        block_discover_log.exception(
+                            "block probing failed restricted=%s", restricted)
+                        report = self.app.make_apport_report(
+                            kind, "block probing", interrupt=False)
+                        self._crash_reports[restricted] = report
+                        continue
+                    break
+            finally:
+                fcntl.flock(self.app.install_lock_file, fcntl.LOCK_UN)
```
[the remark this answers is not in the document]
mwhudson: We should probably write some helpers to handle the locking, yes. I don't think the libraries you link to really work for us as they don't integrate into asyncio. I'll cobble something together.

subiquity/core.py
```diff
@@ -122,7 +124,11 @@ def __init__(self, opts, block_log_dir):
         if not opts.bootloader == 'none' and platform.machine() != 's390x':
             self.controllers.remove("Zdev")
 
+        self.journal_fd, self.journal_watcher = journald_listener(
+            ["subiquity"], self.subiquity_event, seek=True)
         super().__init__(opts)
+        self.install_lock_file = open(self.state_path("installing"), 'a+')
```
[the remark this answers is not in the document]
mwhudson: So the file doesn't get truncated in the case where you ssh into the installer while an install is running.

subiquity/core.py
mwhudson: Yes, the lock attempt blocks until the install is finished.

subiquity/ui/views/installprogress.py
mwhudson: :)

subiquity/core.py
```diff
@@ -152,6 +158,33 @@ def __init__(self, opts, block_log_dir):
...
+                    self.install_lock_file, fcntl.LOCK_SH | fcntl.LOCK_NB)
+            except OSError:
+                pass
+            else:
+                fcntl.flock(self.install_lock_file, fcntl.LOCK_UN)
+                return
+            from subiquity.ui.views.installprogress import (
+                InstallRunning,
+                )
+            self.install_lock_file.seek(0)
```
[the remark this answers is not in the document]
mwhudson: The flow is confusing, if we get the lock we immediately release it and return. If we continue into the body of the function, we didn't get the lock.

subiquity/controllers/installprogress.py
```diff
@@ -272,16 +258,20 @@ async def curtin_install(self, context):
...
+        self.app.aio_loop.add_reader(journal_fd, watcher)
 
         curtin_cmd = self._get_curtin_command()
 
         log.debug('curtin install cmd: {}'.format(curtin_cmd))
 
-        cp = await arun_command(
-            self.logged_command(curtin_cmd), check=True)
+        async with self.app.install_lock_file.exclusive():
+            self.app.install_lock_file.set_content(os.ttyname(0))
```
raharper, line 271: Excellent!

subiquity/controllers/refresh.py
```diff
@@ -205,8 +205,7 @@ async def check_for_update(self):
             return CheckState.UNAVAILABLE
 
     async def start_update(self):
-        update_marker = os.path.join(self.app.state_dir, 'updating')
-        open(update_marker, 'w').close()
+        open(self.app.state_path('updating'), 'w').close()
```
raharper, line 208: Not for this PR, but maybe we want a  StateFile with touch() implemented as a helper, very much like your set_content() with empty payload.

subiquity/lockfile.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, path):
+        self.path = path
+        self.fp = open(path, 'a+')
+
+    def read_content(self):
+        self.fp.seek(0)
+        return self.fp.read()
+
+    def set_content(self, content):
```
raharper, line 61: bike-shed color is blye, write_content

subiquity/core.py
```diff
@@ -152,6 +159,28 @@ def __init__(self, opts, block_log_dir):
...
+            schedule_task(self._hide_install_running())
+
+    async def _hide_install_running(self):
+        # Wait until the install has completed...
+        async with self.install_lock_file.shared():
+            pass
+        # And remove the overlay.
+        self.install_running = None
+        self.ui.body.remove_overlay()
+
```
raharper, line 183: Any reason to not set state  within the lock context?

subiquity/controllers/refresh.py
mwhudson: Yes probably no harm in that.

subiquity/lockfile.py
mwhudson: yeah read/write is the correct pair

subiquity/core.py
mwhudson: No i guess not.

PR #739, opened 2020-04-29
fix some autoinstall network oopses

mwhudson:
Mostly this is about allowing the config to look like this:

```
network:
 version: 2
 ethernets:
  eth0: {'dhcp4': True}
```

and not 

```
network:
 network:
  version: 2
  ethernets:
   eth0: {'dhcp4': True}
```

but there are also some fixes around handling netplan apply failures better


mwhudson:
No, this was just reported on discourse.

PR #740, opened 2020-04-29
bin/console-conf-wrapper: don't do anything during install mode

anonymouse64:
For now, we want to just block console-conf from doing anything in install mode, i.e. allowing a user to login etc. but eventually we will probably want to show more useful status information from snapd during the install mode process.

### DIFF
--- a/bin/console-conf-wrapper
+++ b/bin/console-conf-wrapper
@@ -19,6 +19,11 @@ if [ -e /run/snapd-recovery-chooser-triggered ]; then
     fi
 fi
 
+if grep -q 'snapd_recovery_mode=install' /proc/cmdline ; then
+    echo "Installing the system, please wait for reboot"
+    sleep infinity
+fi
+
 if [ "$(snap managed)" = "true" ]; then
     # check if we have extrausers that have no password set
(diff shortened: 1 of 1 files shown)


bin/console-conf-wrapper
```diff
@@ -19,6 +19,11 @@ if [ -e /run/snapd-recovery-chooser-triggered ]; then
     fi
 fi
 
+if grep -q 'snapd_recovery_mode=install' /proc/cmdline ; then
+    echo "Booting install mode, please wait for reboot"
```
bboozzoo, line 23: How about `Installing the system, please wait for reboot` ?

anonymouse64:
> Or was it intentional that one can still get to console-conf during installation?

That was not intentional, thanks for the review I will move it down further.

anonymouse64:
FWIW, I tested the current state of this PR some more when hitting ctrl-c in a serial console and in the qemu GUI window with this PR, it just loops back and outputs the same message again, because AFAICT the trap at the top of the console-conf-wrapper catches ctrl-c and in effect restarts the systemd unit and we loop back to the same spot.

anonymouse64:
That being said, moving the check down to below works just the same, both via serial console and via qemu GUI window. The one other thing to consider about leaving the check above is that with the check run at that point, we don't need the `snap` command to run `snap managed`, which means that we can run slightly sooner to output a message that the system is installing, and that if there are issues with snapd, console-conf doesn't get blocked waiting for a response from snapd via `snap managed`. That being said, still happy to do it below as well.

PR #744, opened 2020-05-01
add and use the concept of a "global overlay"

mwhudson:
a problem with my previous merge was that the top overlay was always
removed when the install finished, but if the user had opened a help
topic in the mean time that was the wrong thing to do. also if the
view was switched under the install notification (which is a bit
unlikely but could happen) the dialog would disappear.

this finally pushed me to generically handle "global overlays" that are
not tied to the underlying, controller-controlled view. there are a few
of these now:

 * help dialogs
 * error reports
 * the "are you sure?" install confirmation
 * now the install notification

this adds an api for adding and removing them, and the api for removing
them includes which overlay to remove.

### DIFF
--- a/subiquity/core.py
+++ b/subiquity/core.py
@@ -35,6 +35,7 @@ from subiquitycore.async_helpers import (
     )
 from subiquitycore.controller import Skip
 from subiquitycore.core import Application
+from subiquitycore.view import BaseView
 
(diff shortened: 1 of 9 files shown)


mwhudson:
going to merge this so i can get it into edge and do some testing but post-hoc reviews most welcome!

PR #746, opened 2020-05-01
copy autoinstall netplan to target

mwhudson:
And some other network fixes


mwhudson:
After testing in VMs and finding and fixing a dose of being much too clever, this finally works.

PR #761, opened 2020-05-08
more work towards being able to provide good translations

mwhudson:
This is not yet finished but it adds support for using pgettext to provide a translation context for strings (doing this it seemed to make most sense to dump build_i18n and intltool from the default build path), uses this in a few places, adds comments for the translator to several more places and also changes several translatable strings to use {} formatting style with named arguments.

https://bugs.launchpad.net/subiquity/+bug/1846434

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ lint: flake8
 
 flake8:
 	@echo 'tox -e flake8' is preferred to 'make flake8'
-	$(PYTHON) -m flake8 $(CHECK_DIRS)
+	$(PYTHON) -m flake8 $(CHECK_DIRS) --exclude gettext38.py
 
 unit:
 	echo "Running unit tests..."
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,0 @@
-[build_i18n]
-domain=subiquity
--- a/setup.py
+++ b/setup.py
@@ -27,16 +27,22 @@ import distutils.command.build
 import distutils.spawn
 import glob
 import os
-import subprocess
 import sys
 
 from setuptools import setup, find_packages
 
 
-class build(distutils.command.build.build):
+class build_i18n(distutils.cmd.Command):
+
+    user_options = []
+
+    def initialize_options(self):
+        pass
+
+    def finalize_options(self):
+        pass
 
     def run(self):
-        super().run()
         data_files = self.distribution.data_files
 
         with open('po/POTFILES.in') as in_fp:
@@ -68,6 +74,12 @@ class build(distutils.command.build.build):
             distutils.spawn.spawn(["msgfmt", po_file, "-o", mo_file])
             targetpath = os.path.join("share/locale", lang, "LC_MESSAGES")
             data_files.append((targetpath, (mo_file,)))
+
+
+class build(distutils.command.build.build):
+
+    sub_commands = distutils.command.build.build.sub_commands + [
+        ("build_i18n", None)]
 
 
 with open(os.path.join(os.path.dirname(__file__),
@@ -110,5 +122,8 @@ setup(name='subiquity',
           ],
       },
       data_files=[],
-      cmdclass={'build': build},
+      cmdclass={
+          'build': build,
+          'build_i18n': build_i18n,
+          },
       )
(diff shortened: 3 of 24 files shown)


xnox:
very nice!

xnox:
having build_i18n as a separate command was nice, as we were able to call just that at debian/rules clean time.

```
        python3 setup.py build_i18n
        dh_auto_clean
```
in debian/rules, to get better subiquity.pot. However, maybe we should just simply commit subiquity.pot in git. This way we will know when we have changed strings, and will not need to "build" subiquity.pot at debian/rules clean time for upload into launchpad.

mwhudson:
> having build_i18n as a separate command was nice, as we were able to call just that at debian/rules clean time.
> 
> ```
>         python3 setup.py build_i18n
>         dh_auto_clean
> ```
> 
> in debian/rules, to get better subiquity.pot. However, maybe we should just simply commit subiquity.pot in git. This way we will know when we have changed strings, and will not need to "build" subiquity.pot at debian/rules clean time for upload into launchpad.

I don't really understand how it is working, but somehow `python3 setup.py build_i18n` still works.

xnox:
```
git rm setup.cfg
sudo apt remove python3-distutils-extra

python3 setup.py build_i18n
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'build_i18n'
```

Also if python3-distutils-extra is installed build_i18n works, and does bad things.

So i think we should drop `*.pot` from .gitignore and just commit `subiquity.pot`

But like to avoid it getting in the way all the time, maybe we can somehow force it to use a consistent timestamp? like call .pot creation under faketime?

mwhudson:
> Also if python3-distutils-extra is installed build_i18n works, 

Ah right. 

> and does bad things.

Oh right, I guess it will generate a pot without pgettext support.

> So i think we should drop *.pot from .gitignore and just commit subiquity.pot

I think this probably makes sense. Do we then switch back to manually uploading the pot to launchpad vs uploading subiquity source packages and having launchpad extract them?

> But like to avoid it getting in the way all the time, maybe we can somehow force it to use a consistent timestamp? like call .pot creation under faketime?

Not sure what you mean here. git doesn't care about timestamps.

xnox:
> > Also if python3-distutils-extra is installed build_i18n works,
> 
> Ah right.
> 
> > and does bad things.
> 
> Oh right, I guess it will generate a pot without pgettext support.
> 
> > So i think we should drop *.pot from .gitignore and just commit subiquity.pot
> 
> I think this probably makes sense. Do we then switch back to manually uploading the pot to launchpad vs uploading subiquity source packages and having launchpad extract them?
> 

I did not figure out a way to upload pot/po for subiquity package into launchpad project without doing a .dsc source package upload.


> > But like to avoid it getting in the way all the time, maybe we can somehow force it to use a consistent timestamp? like call .pot creation under faketime?
> 
> Not sure what you mean here. git doesn't care about timestamps.

The ones inside .po & .pot headers

```
subiquity.pot:"POT-Creation-Date: 2020-05-12 13:10+0100\n"
sv.po:"POT-Creation-Date: 2020-05-06 13:23+0100\n"
sv.po:"X-Launchpad-Export-Date: 2020-05-06 12:01+0000\n"
```

mwhudson:
> I did not figure out a way to upload pot/po for subiquity package into launchpad project without doing a .dsc source package upload.

Well you can do uploads for the subiquity _project_ (vs source package) at https://translations.launchpad.net/subiquity/trunk/+templates

> > Not sure what you mean here. git doesn't care about timestamps.
> 
> The ones inside .po & .pot headers
> 
> ```
> subiquity.pot:"POT-Creation-Date: 2020-05-12 13:10+0100\n"
> sv.po:"POT-Creation-Date: 2020-05-06 13:23+0100\n"
> sv.po:"X-Launchpad-Export-Date: 2020-05-06 12:01+0000\n"
> ```

Oh right, of course. Yes, faketime sounds good for this.

mwhudson:
I think this branch is now landable as-is; we should think about what do with subiquity.pot and where to get translations but it doesn't have to block this.

PR #762, opened 2020-05-11
make language list at snap build time

mwhudson:
Use language names and console-suitability from localechooser with some special cases.

For https://bugs.launchpad.net/subiquity/+bug/1873342

### DIFF
--- a/languagelist
+++ b/languagelist
@@ -1,0 +1,28 @@
+console:en_US:English
+console:en_GB:English (UK)
+ssh:kab:Tamaziɣt Taqbaylit
+console:ca:Català
+ssh:zh_CN:中文(简体)
+console:hr:Hrvatski
+console:cs:Čeština
+console:ast:Asturianu
+console:be:Беларуская
+console:nl:Nederlands
+console:fi:Suomi
+console:fr:Français
+console:de:Deutsch
+console:el:Ελληνικά
+ssh:he:עברית
+console:hu:Magyar
+console:id:Bahasa Indonesia
+console:lv:Latviski
+console:lt:Lietuviškai
+console:pl:Polski
+console:ru:Русский
+console:sr:Српски
+console:es:Español
+console:sv:Svenska
+ssh:bo:བོད་ཡིག
+console:uk:Українська
+console:nb:Norsk bokmål
+ssh:oc:Galés
--- a/scripts/make-language-lists
+++ b/scripts/make-language-lists
@@ -1,30 +1,63 @@
 #!/usr/bin/python3
 
+import gettext
 import glob
 import gzip
 import os
+import sys
 
+DATA_PATH = '/usr/share/localechooser/languagelist.data.gz'
 
 translations = set()
 for fname in glob.glob('po/*.po'):
     translations.add(os.path.splitext(os.path.basename(fname))[0])
 
-print('found translations', translations)
-    
-console_langs = open('languagelist-console', 'w')
-ssh_langs = open('languagelist-ssh', 'w')
-data = gzip.open('/usr/share/localechooser/languagelist.data.gz', mode='rt')
 
-for line in data:
-    level, code, name, desc = line.strip().split(':')
-    if code not in translations:
-        continue
-    level = int(level)
-    new_line = "{}:{}\n".format(code, desc)
-    ssh_langs.write(new_line)
-    if level <= 2:
-        console_langs.write(new_line)
+langs = [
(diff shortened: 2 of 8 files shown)


mwhudson:
This fails on Xenial for reasons I don't particularly understand. I don't think we have to care about that though?

xnox:
It fails on the bionic lxd image no? And it seems we try to print utf8 to output which is in non unicode locale. Do we set test environment to C.UTF-8 before executing?

Hopefully this should result in no change on s390x terminals, that can barely do latin1.

mwhudson:
> It fails on the bionic lxd image no?

Oh yes so it does. So we probably do get to care :/

> And it seems we try to print utf8 to output which is in non unicode locale. Do we set test environment to C.UTF-8 before executing?

Yes, the tests set LANG=C.UTF-8.

> Hopefully this should result in no change on s390x terminals, that can barely do latin1.

Hm good point, it probably does bad things there. Can we identify bad terminals somehow? Doing what d-i does and just using C on serial consoles seems a bit cowardly.

xnox:
> > It fails on the bionic lxd image no?
> 
> Oh yes so it does. So we probably do get to care :/
> 
> > And it seems we try to print utf8 to output which is in non unicode locale. Do we set test environment to C.UTF-8 before executing?
> 
> Yes, the tests set LANG=C.UTF-8.
> 
Weird. Hopefully it is done before the python interpreter opens stdout. Cause it seems like log.debug() fails to print unicode or some such. Or the list of languages is missbuilt. Do we have to rebuild the list of things every time? can we not just have it committed in git without trying to rebuild it on all the series all the time?

> > Hopefully this should result in no change on s390x terminals, that can barely do latin1.
> 
> Hm good point, it probably does bad things there. Can we identify bad terminals somehow? Doing what d-i does and just using C on serial consoles seems a bit cowardly.

Well we already have ascii mode. And i think I made that to replace any non-ascii characters with '?' but like i feel like we want "latin1", "console", "ssh" because no way some of those languages will be at all readable if they are all just "?". But I guess I have to try it.

mwhudson:
> > > It fails on the bionic lxd image no?
> > 
> > 
> > Oh yes so it does. So we probably do get to care :/
> > > And it seems we try to print utf8 to output which is in non unicode locale. Do we set test environment to C.UTF-8 before executing?
> > 
> > 
> > Yes, the tests set LANG=C.UTF-8.
> 
> Weird. Hopefully it is done before the python interpreter opens stdout. Cause it seems like log.debug() fails to print unicode or some such. Or the list of languages is missbuilt. Do we have to rebuild the list of things every time? can we not just have it committed in git without trying to rebuild it on all the series all the time?

Ah I see what's happening: this is when the autoinstall does a pretend refresh and so re-execs itself. By then LANG has been set to en_UK.UTF-8 and on bionic an unrecognized LANG leads to locale.getpreferredencoding(False) being ascii. en_UK.UTF-8 is bogus but en_GB.UTF-8 gets the same reaction unless language-pack-en is installed. So the upshot of all this is that I think subiquity should call locale.setpreferredencoding('UTF-8') very early.

> > > Hopefully this should result in no change on s390x terminals, that can barely do latin1.
> > 
> > 
> > Hm good point, it probably does bad things there. Can we identify bad terminals somehow? Doing what d-i does and just using C on serial consoles seems a bit cowardly.
> 
> Well we already have ascii mode. And i think I made that to replace any non-ascii characters with '?' but like i feel like we want "latin1", "console", "ssh" because no way some of those languages will be at all readable if they are all just "?". But I guess I have to try it.

I wonder if we should start on serial in a very basic mode (ascii, no color) but off people the chance to upgrade to all singing / all dancing before anything else. (ascii mode currently can't be turned on and off dynamically but I'm sure we can fix that).

Color support is orthogonal in some sense to using ascii, but maybe we should just have two modes, basic (ascii, no color) and full (unicode and color)? Well three I guess, we have linux console too, but that's at least a bit more predictable and not something we'd want to toggle on and off.

xnox:
We should

1. fix tests to install language pack
1. test under a valid utf8 locale
1. locale.setpreferredencoding('UTF-8') => is hard, because if one sets locale on the kernel cmdline, i do expect that to be used on tty1 and display russian correctly for example, which will be non-utf8 kuir8 ecoding or some such.

xnox:
Fallback C.UTF-8 is good.

xnox:
Travis passed, this is good to go!

PR #766, opened 2020-05-12
refresh: default to updating the installer when available

xnox:
Way too many good bugfixes are shipped in the new installer, thus
default to update the installer when a new one is available.

Also, it seems like self.active was always interactive, meaning
autoinstall refresh-installer doesn't do anything? Not sure if that
was intentional. Because self.active was only assigned in __init__,
which always has self.ai_data as an empty dict.

### DIFF
--- a/subiquity/controllers/refresh.py
+++ b/subiquity/controllers/refresh.py
@@ -54,6 +54,7 @@ class RefreshController(SubiquityController):
             },
         'additionalProperties': False,
         }
+    autoinstall_default = {'update': True}
 
     signals = [
         ('snapd-network-change', 'snapd_network_changed'),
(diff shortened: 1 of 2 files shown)


xnox:
This is not right. Because i think behaviour is wrong:
 * if implicitely or explicitely enabled, this page needs to be interactive i think
 * if disabled, this page should skip

I don't think above is happening correctly at the moment.

XSpielinbox:
I agree that the default should be to update the installer.
For example when installing Ubuntu Server 22.04.1 currently the installer generates a Netplan configuration that has been deprecated for over a year now. A newer Version of Subiquity has corrected this. Many people reuse installation ISOs and stick to the defaults of the installer, wherefore a installer that updates itself by default would be very beneficial.

PR #767, opened 2020-05-13
default to setting up LVM in guided storage screen

mwhudson:


xnox:
legacy-server.iso defaults to lvm too
d-i mini.iso did not
autoinstall.yaml defaults to lvm as well

paride:
I was a bit surprised by the change, but I guess pros and cons have been weighted already. In general I'm +1.

If LVM is going to be the new default then I think more people will hit https://bugs.launchpad.net/subiquity/+bug/1873328.

xnox:
> I was a bit surprised by the change, but I guess pros and cons have been weighted already. In general I'm +1.
> 
> If LVM is going to be the new default then I think more people will hit https://bugs.launchpad.net/subiquity/+bug/1873328.

"surprised by the change" but in which direction? that subiquity defaulted to ext4 instead of LVM, or that we are changing it from ext4 to LVM, or that autoinstall was different to the interactive?

paride:
Seeing that legacy-server defaults to LVM while subiquity to ext4 I thought it was explicitly decided at some point to stop using LVM by default. Now I see I was probably wrong, that's it!

PR #768, opened 2020-05-13
allow keyboard/toggle to be null

idryzhov:
Documentation allows toggle to be null and it's the default value.

Signed-off-by: Igor Ryzhov <iryzhov@nfware.com>


mwhudson:
Thanks!

PR #769, opened 2020-05-13
write version into autoinstall-user-data

idryzhov:
Signed-off-by: Igor Ryzhov <iryzhov@nfware.com>


mwhudson:
Uh I was sure I'd done this already! Sorry and thanks.

PR #771, opened 2020-05-18
console-conf-wrapper: show snapd log during install

mvo5:
This will help finding some of the current install problems by
showing the snapd log on the terminal during the install.

### DIFF
--- a/bin/console-conf-wrapper
+++ b/bin/console-conf-wrapper
@@ -21,6 +21,9 @@ fi
 
 if grep -q 'snapd_recovery_mode=install' /proc/cmdline ; then
     echo "Installing the system, please wait for reboot"
+    snap watch --last=install-system || true
+    snap change --last=install-system || true
+    journalctl -u snapd.service -f
     sleep infinity
 fi
 
--- a/bin/console-conf-wrapper
+++ b/bin/console-conf-wrapper
@@ -21,8 +21,7 @@ fi
 
 if grep -q 'snapd_recovery_mode=install' /proc/cmdline ; then
     echo "Installing the system, please wait for reboot"
-    snap watch --last=install-system || true
-    snap change --last=install-system || true
+    # XXX: replace with something more user friendly after the beta?
     journalctl -u snapd.service -f
     sleep infinity
 fi


bin/console-conf-wrapper
```diff
@@ -21,6 +21,9 @@ fi
 
 if grep -q 'snapd_recovery_mode=install' /proc/cmdline ; then
     echo "Installing the system, please wait for reboot"
+    snap watch --last=install-system || true
+    snap change --last=install-system || true
+    journalctl -u snapd.service -f
     sleep infinity
```
xnox, line 27: Why `snap change`? previously we had `snap tasks`.

Also, errors are now ignored. This means if getty spawns before install-system task is started, this will not display anything. Or will it? Is there something in place already that ensures that "install-system" change-id has started?

xnox:
Should we not loop all change IDs from 1 to infinity? Wait and show them all?

Also, I thought journal is now connected to the console, and thus no need to print it again? Or am I wrong?

bin/console-conf-wrapper
mvo5: Thanks, those are excellent points. I think for beta the easiest path forward is to display the journal output from snapd. I removed the "snap watch/change" now. Especially with the recent changes in snapd install logging this becomes much more useful than it was before. Something we can reconsider after beta, i.e. if we should show some sort of progress here but that's a larger change.

mvo5:
To your point about "journal is now connected to the console". That's not what I see in the default console at least. I see there is a journal running on one of the ttys but it logs some extra stuff (like stopping/starting gettys and systemd-timesyncd messages). So it's less easy to see/find what is going on.

xnox:
> To your point about "journal is now connected to the console". That's not what I see in the default console at least. I see there is a journal running on one of the ttys but it logs some extra stuff (like stopping/starting gettys and systemd-timesyncd messages). So it's less easy to see/find what is going on.

Let me try again. We currently connect all output from all units to active consoles, when in install mode. If one connects journalctl -f on snapd.service alone, all messages from that will be duplicated twice and still interleaved with all other output.

If you say, that most of journal is noise, but connecting output just from the snapd.service is useful, then maybe we should just do that. As in change  

https://git.launchpad.net/ubuntu-core-initramfs/tree/factory/usr/lib/systemd/system-generators/journald-console

to instead do something like this
```
mkdir -p /run/systemd/system/snapd.service.d
cat <<EOF >/run/systemd/system/snapd.service.d/install-override.conf 
[Service]
StandardOutput=journal+console
StandardError=journal+console
EOF
```
Then only the snapd.service will be connected to the console.

Or we can redirect snapd.service to a file with StandardOutput|StandardError set to file:/run/snapd/installer.log
Then during install instead of journalctl -f, we can do `tail +1f /run/snapd/installer.log` which will print all the produced output, and continue to follow it.
At the end of the install, we can even copy `/run/snapd/installer.log` onto the ubuntu-data partition into `/var/log/installer/` which is the usual path for all of the installer logs from any Ubuntu install media (be it curtin, subiquiity, ubiquity, maas, d-i, etc).

Shall I detach all of journal from console on install? Shall just the snapd.service attached to console? Or redirect to a file instead?

mvo5:
Let me try again: We currently clear the terminal when console-conf-wrapper runs. This means that when the install fails before there is just a blank screen without further information. Here is a video what the current experience with a broken system looks like:
https://photos.app.goo.gl/wFEQbi1MchkGBoeb6

I have no strong opinion how this should be fixed. Doing the redirect of snapd.service to a file and then displaying it in the console-conf wrapper seems reasonable to me.

mvo5:
Fwiw, the error during the install in my test is visible here (from the original proposal): https://photos.app.goo.gl/tf7M7cVRAnUDC1dn8

Having the installer.log or snapd-install-mode.log or something would certainly be nice. I think eventually we need to consider if we show something more user friendly. But for now anything that helps us to diagnose install failures with a screenshot/photo will help :)

xnox:
I kind of feel like setting min image size to 8GiB for the pc gadget somehow, such that ubuntu-image defaults to building something big enough to be launched as a VM straight away.

Cause you either need to qemu-img resize the image to be bigger, or rebuild it again with a larger option.

Doesn't affect pi images, as they self expand to the size of the usbstick / sdcard.

mvo5:
The error is ok for now and well understood, it's actually useful in this case for demonstration purposes.

mvo5:
Can we move forward with this minimal version or do you prefer to work on someting more elaborate? Maybe we go with this initially and then do a followup?

PR #773, opened 2020-05-18
console_conf: fix showing error messages, when snap create-user fails

xnox:


xnox:
https://bugs.launchpad.net/subiquity/+bug/1878493

xnox:
@mwhudson are you ok with this?

I wonder if we want generic error strechy everywhere, or not.

PR #774, opened 2020-05-18
Core/focal

xnox:
- fix wifi
- fix crash upon snap create-user
- fix running consoleconf when not completed, or chooser mode triggered


xnox:
lint errors are not introduced in these cherrypicks, and have been fixed on master.

xnox:
This needs more cherrypicks.

PR #776, opened 2020-05-18
fix the wifi configuration some more

mwhudson:


subiquitycore/ui/views/network_configure_wlan_interface.py
```diff
@@ -151,7 +151,7 @@ def done(self, sender):
         if self.device.configured_ssid[0] is None and self.form.ssid.value:
             # Turn DHCP4 on by default when specifying an SSID for
             # the first time...
-            self.device.dhcp4 = True
+            self.device.config['dhcp4'] = True
```
xnox, line 154: Hm, should we enable dhcp6 too here?

Most of IoT is ipv4 still, but it should at least try ipv6 in 2020, right?

waveform80: A brief data-point on this: Mythic's Pi cloud is entirely IPv6 only (with a proxy to provide web-access to IPv4-only clients) so I'd say it's certainly getting more important in the IoT space.

vorlonofportland:
On Tue, May 19, 2020 at 06:07:44AM -0700, Dimitri John Ledkov wrote:
> @xnox commented on this pull request.
> 
> 
> 
> > @@ -151,7 +151,7 @@ def done(self, sender):
>          if self.device.configured_ssid[0] is None and self.form.ssid.value:
>              # Turn DHCP4 on by default when specifying an SSID for
>              # the first time...
> -            self.device.dhcp4 = True
> +            self.device.config['dhcp4'] = True

> Hm, should we enable dhcp6 too here?

> Most of IoT is ipv4 still, but it should at least try ipv6 in 2020, right?

Most ipv6 networking is SLAAC, not DHCPv6; unless you're in an ipv6-only
environment, you probably don't need DHCPv6.

xnox:
> On Tue, May 19, 2020 at 06:07:44AM -0700, Dimitri John Ledkov wrote: @xnox commented on this pull request. > @@ -151,7 +151,7 @@ def done(self, sender): if self.device.configured_ssid[0] is None and self.form.ssid.value: # Turn DHCP4 on by default when specifying an SSID for # the first time... - self.device.dhcp4 = True + self.device.config['dhcp4'] = True Hm, should we enable dhcp6 too here? Most of IoT is ipv4 still, but it should at least try ipv6 in 2020, right?
> Most ipv6 networking is SLAAC, not DHCPv6; unless you're in an ipv6-only environment, you probably don't need DHCPv6.

right but unless netplan's dhcp6 is set, there is no SLAAC, either right?

vorlonofportland:
On Tue, May 19, 2020 at 12:12:11PM -0700, Dimitri John Ledkov wrote:

> right but unless netplan's dhcp6 is set, there is no SLAAC, either right?

There's no reason this should be the case.  That is certainly not the design
of netplan.

mwhudson:
We do need to straighten out the IPv6 story a bit, see recent subiquity bug reports about that, but AIUI we should not be enabling DHCPv6 by default.

mwhudson:
PS: thanks for merging :)

PR #778, opened 2020-05-20
start in a super basic mode on serial, offer upgrade immediately

mwhudson:
This is mostly a RFC currently, there's some bad code I want to clean up if we think it's a good idea.

Demo at https://asciinema.org/a/3dDg5jPWCmuBLf5AQtoFWpqyj

### DIFF
--- a/subiquity/ui/views/help.py
+++ b/subiquity/ui/views/help.py
@@ -239,6 +239,31 @@ def menu_item(text, on_press=None):
     return Color.frame_button(icon)
 
 
+def get_global_addresses(app):
+    ips = []
+    net_model = app.base_model.network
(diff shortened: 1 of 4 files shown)


xnox:
This makes sense. If there is no network, one should be able to configure zdev+network first, then shown this no?

Or are you assuming serial console == netboot too?

mwhudson:
> This makes sense. If there is no network, one should be able to configure zdev+network first, then shown this no?

So you mean this, or something like this, should be shown as soon as SSH becomes possible? (== we have a global ip address i guess)

> Or are you assuming serial console == netboot too?

No.

mwhudson:
Hm I bet after this that snap refresh will land you back in basic mode. Probably need to fix that.

PR #783, opened 2020-05-25
Normalise messages to consistently use "GitHub"

a4ff7810:
Made changes to consistently use "GitHub" (upper case letter "H") in message strings. Previously, "Github" was used in many cases.

### DIFF
--- a/po/ast.po
+++ b/po/ast.po
@@ -1603,7 +1603,7 @@ msgid "Import Username:"
 msgstr ""
 
 #: ../subiquity/ui/views/ssh.py:73
-msgid "Github Username:"
+msgid "GitHub Username:"
 msgstr ""
 
 #: ../subiquity/ui/views/ssh.py:80
@@ -1619,7 +1619,7 @@ msgid "Import SSH identity:"
 msgstr ""
 
 #: ../subiquity/ui/views/ssh.py:98
-msgid "from Github"
+msgid "from GitHub"
 msgstr ""
 
 #: ../subiquity/ui/views/ssh.py:99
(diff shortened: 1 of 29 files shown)


a4ff7810:
Sorry, only noticed the "to export and update translations in subiquity" section in the README after I had already created this pull request. Will try and follow the instructions from there :).

xnox:
Actually monkey-patching translations like this is good. Because you are hopefully not breaking any existing translations. As I expect one would not translate the word Github inside the translated string.

vorlonofportland:
On Mon, May 25, 2020 at 03:22:07PM -0700, Dimitri John Ledkov wrote:

> Actually monkey-patching translations like this is good.  Because you are
> hopefully not breaking any existing translations.  As I expect one would
> not translate the word Github inside the translated string.

It might be transliterated, but I'm not sure if we should care about
capitalization in such a case? (Cyrillic)

xnox:
> On Mon, May 25, 2020 at 03:22:07PM -0700, Dimitri John Ledkov wrote: Actually monkey-patching translations like this is good. Because you are hopefully not breaking any existing translations. As I expect one would not translate the word Github inside the translated string.
> It might be transliterated, but I'm not sure if we should care about capitalization in such a case? (Cyrillic)

Вай ар ю дуинг зис ту ми!

a4ff7810:
Wow, very good points! Hadn't thought of that. I mean, in its current form, the pull request pretty much only changes instances of "Github" to "GitHub" (which is nitpicky to begin with, I'll admit).

So from what I see in the diff, there currently doesn't seem to be any transliteration happening (cf. "SSH" and "Launchpad" as well, for example). At least none that I can find 🤔...: Message strings either 1) changed to reflect the new spelling, 2) remain empty, indicating they hadn't been translated yet in the first place or 3) already used "GitHub" (e.g. Hungarian and Spanish).

How would you all prefer to proceed :)? I think I'm happy with the changes as they are, so I'll gladly mark the request "ready for review". What I _am_ wondering about is the whole Launchpad thing, though. @xnox: You said the monkey-patching was alright. But then won't the changes be overwritten once/if a new Launchpad translation import happens? I could try and see if I can suggest the changes to https://translations.launchpad.net/ubuntu/+source/subiquity as well.

Thanks for bearing with me on this :)!

PR #784, opened 2020-06-09
Fix incorrect key in user_info

PuffyZA:
The key lock-passwd in the user_info is incorrect, should be lock_passwd.


mwhudson:
Thanks. I'm still a bit confused how this ever worked...

PR #789, opened 2020-06-12
allow starting a shell when connecting to a session running an autoinstall

mwhudson:


mwhudson:
For https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1879284

PR #790, opened 2020-06-15
use the "None" datasource for on-first-boot configuration

mwhudson:
Using the NoCloud source meant that a filesystem label of "cidata"
(probably containing autoinstall config) could override the cloud-init
nocloud seed subiquity wrote and then users would not get created. So
instead write cloud config directly that hardcodes using the
DataSourceNone source and config for it.

https://bugs.launchpad.net/subiquity/+bug/1879103

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -203,10 +203,18 @@ class SubiquityModel:
         # for subsequent boots.
         # (mwhudson does not entirely know what the above means!)
         userdata = '#cloud-config\n' + yaml.dump(self._cloud_init_config())
-        metadata = yaml.dump({'instance-id': str(uuid.uuid4())})
+        metadata = {'instance-id': str(uuid.uuid4())}
+        config = {
+            'datasource_list': ["None"],
+            'datasource': {
+                "None": {
+                    'userdata_raw': userdata,
+                    'metadata': metadata,
+                    },
+                },
+            }
         files = [
-            ('var/lib/cloud/seed/nocloud-net/meta-data', metadata, 0o644),
-            ('var/lib/cloud/seed/nocloud-net/user-data', userdata, 0o600),
+            ('etc/cloud/cloud.cfg.d/installer.cfg', yaml.dump(config), 0o600),
             ('etc/cloud/ds-identify.cfg', 'policy: enabled\n', 0o644),
             ]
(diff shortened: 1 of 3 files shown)


mwhudson:
Would definitely like review from people who understand cloud-init better than me :)

xnox:
I think there was a way to specify in the ds-identify configuration _which_ explicit path of NoNet one wants to use.

I.e. I just hope this current implementation doesn't cause cloud-init to attempt searching all the datasources on the internet.

xnox:
Plus if anybody specified their datasource _after_ `---` on the kernel cmdline during install, that propagates to the installed system, and must be removed too.

Because kernel commandline overrides anything we encode on disk.

subiquity/models/subiquity.py
```diff
@@ -203,10 +203,18 @@ def _cloud_init_files(self):
...
+                "None": {
+                    'userdata_raw': userdata,
+                    'metadata': metadata,
+                    },
+                },
+            }
         files = [
-            ('var/lib/cloud/seed/nocloud-net/meta-data', metadata, 0o644),
-            ('var/lib/cloud/seed/nocloud-net/user-data', userdata, 0o600),
+            ('etc/cloud/cloud.cfg.d/installer.cfg', yaml.dump(config), 0o600),
```
raharper, line 217: I would change this to 99-subiquity.cfg;  using number > 90 will ensure this clobers the built-in 90-dpkg.cfg and I think including subiquity is useful to know what program wrote this (also consider a # comment in the yaml with the signature as well.

mwhudson: That makes sense, I didn't realise that that's what 90_dpkg.cfg had in it. I'm a bit confused as to how this branch helps with the cloud.cfg.d snippets being in the order they are, but it does...

raharper: This is likely due to luck/accident.  ds-identify does not have a full yaml parser, but rather searches for keys while globbing through /etc/cloud/cloud.cfg /etc/cloud/cloud.cfg.d/*.cfg;

mwhudson: Oh yay. I've updated the filename.

mwhudson:
I'll merge this soon -- it may not be perfect but it's definitely an improvement.

PR #791, opened 2020-06-15
Add disk size match smallest

wedi:
In addition to matching the largest disk it would be great to have a matcher for the smallest disk as well. This is helpful if you have a large disk for data and a smaller one for the OS, both being SSD/HDD.


mwhudson:
Thanks!

PR #794, opened 2020-07-02
use snapd apis to check for users

mwhudson:
very much untested

### DIFF
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -76,8 +76,8 @@ def get_device_owner():
         homedir = '/home/' + username
         if os.path.isdir(homedir):
             return {
-                'username': user,
-                'realname': get_realname(user),
+                'username': username,
+                'realname': get_realname(username),
                 'homedir': homedir,
                 }
     return None
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -70,12 +70,15 @@ def get_device_owner():
     """ Get device owner, if any """
     con = SnapdConnection('', '/run/snapd.socket')
     for user in con.get('v2/users').json()['result']:
-        user = user['email'].split('@')[0]
-        if os.path.isdir('/home/' + user):
+        if 'username' not in user:
+            continue
+        username = user['username']
+        homedir = '/home/' + username
+        if os.path.isdir(homedir):
             return {
                 'username': user,
                 'realname': get_realname(user),
-                'homedir': '/home/' + user,
+                'homedir': homedir,
                 }
     return None
 
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -16,11 +16,13 @@
 import json
 import logging
 import os
+import pwd
 import shlex
 import sys
 
 from subiquitycore.controller import BaseController
 from subiquitycore.ssh import host_key_info, get_ips_standalone
+from subiquitycore.snapd import SnapdConnection
 from subiquitycore.utils import disable_console_conf, run_command
 
 from console_conf.ui.views import IdentityView, LoginView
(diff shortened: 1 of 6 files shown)


console_conf/controllers/identity.py
```diff
@@ -52,22 +53,11 @@ def get_core_version():
...
-                'realname': passwd[4].split(',')[0],
-                'username': passwd[0],
-                'homedir': passwd[5],
-                }
-            return result
+    con = SnapdConnection('', '/run/snapd.socket')
+    for user in con.get('v2/users').json()['result']:
+        user = user['email'].split('@')[0]
+        if os.path.isdir('/home/' + user):
+            return user
```
xnox, line 60: It should check if device is managed, then iterate users.

As users may exist, whilst not actually be owners. Thus this should not return a user, when device is unmanged, but has lxd user.

console_conf/controllers/identity.py
```diff
@@ -136,8 +126,8 @@ def start_ui(self):
         device_owner = get_device_owner()
         if device_owner is not None:
             self.model.add_user(device_owner)
-            key_file = os.path.join(device_owner['homedir'],
```
xnox, line 139: at the moment device_owner on unmanaged device, with seeded lxd snap will be `lxd`. It should be empty. Hence get_device_owner() must check for `snap managed`.

console_conf/controllers/identity.py
mwhudson: Isn't the point of checking that /home/$user exists to exclude users like lxd? I assumed that that's what that was for...

mwhudson: Even if the device is managed, the lxd user could still be present and still needs to be excluded...

xnox: If the device is not managed, we should not be skipping the identity page.

And inverse, if device is managed, and there are no users, we must skip identity page and do something else.... but ideally console-conf is prevented from running due to shell code.

And yes, non managing users may have a home directory. The idea of being able to create "user-accounts".

This does duplicate the shell code, but because shell code is supposedly optimization path.

mwhudson: I think you're right though that to match the desired behaviour in the bug, this bit does need to be fixed.

mwhudson:
OK I've pushed a simpler version now that I understand the actual bug a little better. Still not tested though. The handling of the "managed but not owned" case is very bare bones.

console_conf/controllers/identity.py
```diff
@@ -50,24 +52,31 @@ def get_core_version():
...
+        return ''
+    return info.pw_gecos.split(',', 1)[0]
+
+
+def get_device_owner():
+    """ Get device owner, if any """
+    con = SnapdConnection('', '/run/snapd.socket')
+    for user in con.get('v2/users').json()['result']:
+        user = user['email'].split('@')[0]
+        if os.path.isdir('/home/' + user):
```
anonymouse64, line 74: this is wrong, for example for my user, I have this:

```json
{
    "result": [
        {
            "email": "first-name.last-name@canonical.com",
            "id": 1,
            "username": "lp-username"
        }
    ],
    "status": "OK",
    "status-code": 200,
    "type": "sync"
}
```

the `username` key should be what we check for in `/home`, because with the above user created on UC18, I will have `/home/lp-username`, not `/home/first-name.last-name`.

anonymouse64: (and my apologies if the LP comment wasn't specific enough, hopefully the above JSON output is helpful)

console_conf/controllers/identity.py
```diff
@@ -111,17 +120,21 @@ def write_login_details(fp, username, ips):
...
+            print("device managed without user")
+            return 2
+        else:
+            tty_name = os.ttyname(0)[5:]
+            version = get_core_version() or "16"
+            print(login_details_tmpl_no_ip.format(
+                tty_name=tty_name, version=version))
+            return 2
+    if owner is None:
+        print("device managed without user @ {}".format(', '.join(ips)))
```
anonymouse64, line 135: this looks fine to me, thanks

anonymouse64:
I tested this PR with the following patch and it now works to seed the LXD snap:

```patch
diff --git a/console_conf/controllers/identity.py b/console_conf/controllers/identity.py
index ed0e085c..c699567d 100644
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -70,12 +70,13 @@ def get_device_owner():
     """ Get device owner, if any """
     con = SnapdConnection('', '/run/snapd.socket')
     for user in con.get('v2/users').json()['result']:
-        user = user['email'].split('@')[0]
-        if os.path.isdir('/home/' + user):
+        username = user['username']
+        homedir = '/home/' + username
+        if os.path.isdir(homedir):
             return {
-                'username': user,
-                'realname': get_realname(user),
-                'homedir': '/home/' + user,
+                'username': username,
+                'realname': get_realname(username),
+                'homedir': homedir,
                 }
     return None

```

mwhudson:
Thanks for the comments, I've pushed a fix.

console_conf/controllers/identity.py
```diff
@@ -70,12 +70,15 @@ def get_device_owner():
...
     for user in con.get('v2/users').json()['result']:
-        user = user['email'].split('@')[0]
-        if os.path.isdir('/home/' + user):
+        if 'username' not in user:
+            continue
+        username = user['username']
+        homedir = '/home/' + username
+        if os.path.isdir(homedir):
             return {
                 'username': user,
```
anonymouse64, line 79: won't `user` here be an object and not the string value of the username? I'm not super familiar with the rest of this code so perhaps it doesn't matter?

mwhudson: Yes you're right, fixed!

PR #795, opened 2020-07-02
add all documented fields to filesystem model classes

mwhudson:
This lets autoinstall configs set them, even though nothing in subiquity
can set them yet.


xnox:
Somehow this regresses the world

https://bugs.launchpad.net/subiquity/+bug/1889749

PR #796, opened 2020-07-08
Stop subiquity-service on non-installer media.

xnox:
This will prevent subiquity service starting/crashing in a loop, when installed on regular systems.

Separately this allows to e.g. run subiquity in --dry-run mode on regular installed machines.

However, it still hits segmentation fault when run as `$ sudo subiquity --dry-run` But at least it's not piling up crash reports anymore.

Thanks to @anonymouse64 for the know-how!

### DIFF
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+# snapd will always start subiquity.service, but when we are not
+# booted off the server.iso there is not enough things in place to run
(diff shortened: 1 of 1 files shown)


mwhudson:
Makes sense to me, have you tested it?

xnox:
$ snap install --channel latest/edge/xnox --classic

Works here, as in, service is disabled.

I did not test, if the service remains enabled on the live iso!

mwhudson:
Tested in an installer environment, works fine.

PR #797, opened 2020-07-09
backport bug 1881588 fixes to version for core18

mwhudson:


xnox:
I think you want to base on `core/bionic` branch and make merge proposal against that.

mwhudson:
> I think you want to base on core/bionic branch and make merge proposal against that.

Ah!

PR #798, opened 2020-07-10
backport bug 1881588 fixes to version for core18

mwhudson:


xnox:
And for that to land into core18 snap it would need to be uploaded into https://launchpad.net/~canonical-foundations/+archive/ubuntu/ubuntu-image?field.series_filter=bionic PPA as it looks like the last upload went there.

PR #800, opened 2020-07-16
add a s390x-devices section to autoinstall data

mwhudson:


mwhudson:
@xnox can you take a look at this one?

subiquity/controllers/zdev.py
```diff
@@ -633,7 +637,15 @@ def typeclass(self):
 
 class ZdevController(SubiquityController):
 
+    autoinstall_key = "s390x-devices"
```
xnox, line 640: zdev would be better, as it is `lszdev` `chzdev` not `s390x-devices`.

subiquity/controllers/zdev.py
```diff
@@ -633,7 +637,15 @@ def typeclass(self):
...
+    autoinstall_default = ()
+    autoinstall_schema = {
+        'type': 'array',
+        'items': {'type': 'string'},
```
xnox, line 644: Somewhere we did document schema, and wanted `enable`, `disable`, `ignore`. For chzdev enable, czdev disable, and cio_ignore. As on large LPARs cio_ignore=all makes the boot so much faster.

Disable is needed to disable autoactivated devices, or like call enable all followed by disable one or two things one doesn't want. The latter is very rare.

subiquity/controllers/zdev.py
mwhudson: OK.

subiquity/controllers/zdev.py
mwhudson: I don't remember anything about a more detailed schema I'm afraid, so I'm happy to believe anything you come up with is a design I've already agreed to :)

frank-heimes:
I think for consistency reasons the section in the autoinstall yaml about ccw devices should be named like in the subiquity UI, so I agree with @xnox that just "zdev" would be a good and coherent naming (of course following the rule that all top level scheme keys are lower-case).

"enable" is obviously the most important option, but I agree that "disable" would be needed too - for the reason @xnox explained above.

I am a bit unsure about "ignore", since I believe there are cases where cio_ignore (to hide ranges of ccw devices aka zdev) needs to be added (usually as kernel parameter) already to the initial install system - like the basic netboot config that one has to put into the parm file for casper).
Hence it could just be _to late_ to have it in the autoinstall yaml only, no?

Imho it _could_ be sufficient to "carry forward" a potential cio_ignore kernel arg, that was passed to the initial install system, to the target system too - otherwise people would need to specify cio_ignore twice if doing autoinstalls - at the initial boot as well as in the autoinstall yaml, which would not be ideal.

frank-heimes:
Just one (relatively new) aspect that I think need to be considered:
We recently we noticed that more and more s390x installations are done on systems that do not have any ccw-devices (the z-specific ones 'zdev'), but have PCIe devices instead (mlx5 and NVMe) - LP#1987236.
So a check (e.g. using 'lszdev --online' at the end of the installation may lead to an empty result, which would be totally valid - and nothing would need to be added to the autoinstall yaml in such a case.

xnox:
@mwhudson I agree with all the comments from frank above, but also do not have time to rebase this branch, who is going to work on implementing this?

PR #805, opened 2020-07-29
snaplist: fix snaplist with filesystem without snaps

xnox:
When no snaps are seeded in the filesystem, empty string is returned,
which yaml safe load interprets as None.

But snapd seed.yaml is defined as a dictionary with snaps key. Ensure
that when no seed.yaml is present (or readable), that at least an
empty dictionary is returned, such that later calls to `.get("snaps",
[])` succeed.

LP: #1889429

### DIFF
--- a/subiquity/ui/views/snaplist.py
+++ b/subiquity/ui/views/snaplist.py
@@ -426,7 +426,7 @@ class SnapListView(BaseView):
     file: lxd_59.snap
 '''
         seed_location = '/media/filesystem/var/lib/snapd/seed/seed.yaml'
-        content = ""
+        content = "{}"
         try:
             with open(seed_location, encoding='utf-8', errors='replace') as fp:
                 content = fp.read()


xnox:
> The `FileNotFoundError` (line 433) exception also has to be handled more gracefully, by printing a warning or info message. Currently `log.exception()` prints an error and the full traceback:
> 
> ```
> 2020-07-29 12:53:46,903 ERROR subiquity.views.snaplist:434 could not find source at '/media/filesystem/var/lib/snapd/seed/seed.yaml'
> Traceback (most recent call last):
>   File "/snap/subiquity/1938/lib/python3.6/site-packages/subiquity/ui/views/snaplist.py", line 431, in get_seed_yaml
>     with open(seed_location, encoding='utf-8', errors='replace') as fp:
> FileNotFoundError: [Errno 2] No such file or directory: '/media/filesystem/var/lib/snapd/seed/seed.yaml'
> ```
> 
> which is not what we want.

What do we want?

paride:
I'd replace
```
log.exception("could not find source at %r", seed_location)
```
with something like:
```
log.info("seeded snaps source not found at %r", seed_location)
```
or `log.warn()`.

mwhudson:
Eh I don't think one more traceback in the logs is a real problem.

paride:
@mwhudson well, it doesn't prevent subiquity from working but we're logging an ERROR + Traceback message for something we decided to not consider an error (otherwise we should have patched `livecd-rootfs` to make it create an empty `seed-yaml`.)

Some time ago I was trying to make the ISO testing jobs parse the subiquity and curtin logs looking for non-fatal tracebacks. Bogus one would add noise to this kind of check. Anyway as this is merged already let's rediscuss should this come back as an issue one day.

mwhudson:
That's an interesting idea but subiquity already .exceptions non-fatal errors in other places (e.g. if downing a nic fails) so yes, needs discussion :)

PR #812, opened 2020-08-13
introduce some explicitly typed objects for eventual use by the client/server API 

mwhudson:
And use them in the views.

This includes https://github.com/CanonicalLtd/subiquity/pull/806


mwhudson:
I think all of the useful bits of this landed in other branches.

PR #817, opened 2020-08-23
remove all UI code from "core" application object, add tuiapplication

mwhudson:
The subiquity server process will not have a UI, so we need to start splitting fundamental from UI code.

### DIFF
--- a/subiquity/controller.py
+++ b/subiquity/controller.py
@@ -20,8 +20,10 @@ import jsonschema
 from subiquitycore.context import with_context
 from subiquitycore.controller import (
     BaseController,
+    )
+from subiquitycore.tuicontroller import (
     RepeatedController,
-    Skip,
+    TuiController,
     )
 
 log = logging.getLogger("subiquity.controller")
@@ -38,6 +40,9 @@ class SubiquityController(BaseController):
         self.autoinstall_applied = False
         self.context.set('controller', self)
         self.setup_autoinstall()
+
+    def interactive(self):
+        return False
 
     def setup_autoinstall(self):
         if self.app.autoinstall_config:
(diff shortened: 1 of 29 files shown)


subiquity/controller.py
```diff
@@ -90,15 +86,15 @@ def make_autoinstall(self):
...
-    def cancel(self):
-        pass
+class SubiquityTuiController(SubiquityController, TuiController):
 
     def interactive(self):
+        if not self.app.autoinstall_config:
+            return True
+        i_sections = self.app.autoinstall_config.get(
+            'interactive-sections', [])
+        if '*' in i_sections or self.autoinstall_key in i_sections:
```
patviafore-canonical, line 96: You can just return this conditional rather than using an if blah return True else return False construct

subiquitycore/tui.py
```diff
@@ -0,0 +1,279 @@
...
+    def initial_controller_index(self):
+        if not self.updated:
+            return 0
+        state_path = self.state_path('last-screen')
+        if not os.path.exists(state_path):
+            return 0
+        with open(state_path) as fp:
+            last_screen = fp.read().strip()
+        controller_index = 0
+        for i, controller in enumerate(self.controllers.instances):
```
patviafore-canonical, line 264: This loop can be replaced by a call to index, with an except returning zero

subiquitycore/tuicontroller.py
```diff
@@ -0,0 +1,120 @@
...
+            if c is None:
+                c = getattr(self, '_action_clean_{}'.format(k), lambda x: x)
+            field = getattr(form, k)
+            from subiquitycore.ui.selector import Selector
+            v = c(v)
+            if isinstance(field.widget, Selector):
+                field.widget._emit('select', v)
+            field.value = v
+            yield
+        yield
```
patviafore-canonical, line 79: Do you mind throwing a comment around the second yield and why it's needed?

subiquitycore/tui.py
mwhudson: Not directly it can't? We're searching by name.

subiquitycore/tuicontroller.py
mwhudson: It probably isn't needed really. This code is all a bit crazy, it hasn't been touched really since I converted everything else to asyncio, it could perhaps be more sensible now.

patviafore-canonical (approved):

PR #818, opened 2020-08-23
examples/autoinstall-interactive: remove duplicate "snaps" section

Forst:
Fixes a duplicate `snaps` key in `examples/autoinstall-interactive.yaml`.


mwhudson:
Thanks!

PR #823, opened 2020-09-11
add api definition and support code for client server split

mwhudson:
Nothing uses this yet, but it might be interesting to review!

### DIFF
--- a/subiquity/common/api/client.py
+++ b/subiquity/common/api/client.py
@@ -36,17 +36,17 @@ def _wrap(make_request, path, meth, serializer):
 
     async def impl(*args, **kw):
         args = sig.bind(*args, **kw)
-        params = {
-            k: json.dumps(serializer.serialize(meth_params[k].annotation, v))
-            for (k, v) in args.arguments.items() if k != payload_arg
-            }
-        if payload_arg in args.arguments:
-            v = args.arguments[payload_arg]
-            data = serializer.serialize(payload_ann, v)
-        else:
-            data = None
+        query_args = {}
+        data = None
+        for arg_name, value in args.arguments.items():
+            if arg_name == payload_arg:
+                data = serializer.serialize(payload_ann, value)
+            else:
+                query_args[arg_name] = json.dumps(
+                    serializer.serialize(
+                        meth_params[arg_name].annotation, value))
         async with make_request(
-                meth.__name__, path, json=data, params=params) as resp:
+                meth.__name__, path, json=data, params=query_args) as resp:
             resp.raise_for_status()
             return serializer.deserialize(r_ann, await resp.json())
     return impl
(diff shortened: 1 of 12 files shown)


subiquity/common/api/client.py
```diff
@@ -0,0 +1,82 @@
...
+
+    async def impl(*args, **kw):
+        args = sig.bind(*args, **kw)
+        params = {
+            k: json.dumps(serializer.serialize(meth_params[k].annotation, v))
+            for (k, v) in args.arguments.items() if k != payload_arg
+            }
+        if payload_arg in args.arguments:
+            v = args.arguments[payload_arg]
+            data = serializer.serialize(payload_ann, v)
```
patviafore-canonical, line 45: Is there any sort of lurking bug if payload_ann doesn't get set in the _wrap call? Or do we feel confident that Payload will always be one of the parameter's annotations

subiquity/common/api/client.py
```diff
@@ -0,0 +1,82 @@
...
+    return C
+
+
+def make_client_for_conn(
+        endpoint_cls, conn, resp_hook=lambda r: r, serializer=None):
+    @contextlib38.asynccontextmanager
+    async def make_request(method, path, *, params, json):
+        async with aiohttp.ClientSession(
+                connector=conn, connector_owner=False) as session:
+            async with session.request(
```
patviafore-canonical, line 77: What does the "a" in the "http://a" represent? Do you mind adding a comment?

subiquity/common/api/server.py
```diff
@@ -0,0 +1,172 @@
...
+                args[data_arg] = serializer.deserialize(
+                    data_annotation, payload)
+            for arg, ann, default in query_args_anns:
+                if arg in request.query:
+                    v = serializer.deserialize(
+                        ann, json.loads(request.query[arg]))
+                elif default != inspect._empty:
+                    v = default
+                else:
+                    1/0
```
patviafore-canonical, line 116: Why division by zero here? Is it just to throw an exception if we get here?

subiquity/common/api/client.py
mwhudson: No, this is actually fine because if there is no Payload argument, then payload_arg will be None and so cannot be in arg.arguments, which is a string-keyed dictview. But it is a bit unclear, and I was already a bit unhappy with it, so I'll try a rewrite.

subiquity/common/api/client.py
mwhudson: I've added a comment about this.

subiquity/common/api/server.py
mwhudson: Yeah "1/0" is my lazy way of writing "assert False" and I should have cleaned this up before sending it for review!

This is the case of failing to provide a required argument in the api call, something the client side of this won't let you do but it should probably still be handled a little bit gracefully. I'll add a test case and make this more sensible.

mwhudson:
Thanks for the review, I've just pushed an update with I think addresses your comments.

mwhudson:
I removed the commit that added the full API -- it wasn't actually importable as all the types aren't in the tree yet and I think adding it bit by bit is going to be better.

PR #857, opened 2020-10-20
welcome: fix crash on serial ip addresses

xnox:
LP: #1900707

https://bugs.launchpad.net/subiquity/+bug/1900707


xnox:
I wonder if we need function annotations or something. Cause surely this could have been automatically detected by something?

mwhudson:
Yeah, mypy or something I guess. I fixed this in the client-server-6 branch, I guess I didn't realize it was broken in main already.

PR #859, opened 2020-10-28
use -report_el_torito as_mkisofs option of xorriso to find command to repack ISO

mwhudson:
This doesn't work because the quoting goes wrong. But we can probably make something like this work.

### DIFF
--- a/scripts/inject-subiquity-snap.sh
+++ b/scripts/inject-subiquity-snap.sh
@@ -186,5 +186,5 @@ mksquashfs new_installer new_iso/casper/installer.squashfs
     md5sum ./casper/installer.squashfs >> md5sum.txt
 )
 
-opts=$(xorriso -indev ${OLD_ISO} -report_el_torito as_mkisofs 2>/dev/null)
-xorriso -as mkisofs $opts -o ${NEW_ISO} -V Ubuntu\ custom new_iso
+eval set -- $(xorriso -indev ${OLD_ISO} -report_el_torito as_mkisofs 2>/dev/null)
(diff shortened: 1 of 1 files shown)


paride:
@mwhudson looks so simple to do, but it is not. This is the best I came up with:

```
opts=$(xorriso -indev "${OLD_ISO}" -report_el_torito as_mkisofs 2>/dev/null)
readarray -t opts < <(xargs -n1 <<<"$opts")
xorriso -as mkisofs "${opts[@]}" -o "${NEW_ISO}" -V "Ubuntu custom" new_iso
```

which uses `xargs` to do the splitting interpreting the quotes in `opts`. I don't know if it's doable with pure bash, maybe with an ugly `eval` trick.

Unfortunately the script still fails with:

```
xorriso : FAILURE : Will not graft-in the whole local filesystem by path '/'
```

mwhudson:
> @mwhudson looks so simple to do, but it is not. This is the best I came up with:
> 
> ```
> opts=$(xorriso -indev "${OLD_ISO}" -report_el_torito as_mkisofs 2>/dev/null)
> readarray -t opts < <(xargs -n1 <<<"$opts")
> xorriso -as mkisofs "${opts[@]}" -o "${NEW_ISO}" -V "Ubuntu custom" new_iso
> ```
> 
> which uses `xargs` to do the splitting interpreting the quotes in `opts`. I don't know if it's doable with pure bash, maybe with an ugly `eval` trick.

Yikes! I tried an "ugly eval trick" and I don't think it's as bad as this :laughing: It works for me on random focal and groovy ISOs I have lying around, so it's probably good enough?

PR #864, opened 2020-11-20
fixes for FBA, unformatted, and passed-via virtio dasds

mwhudson:
Together with the referenced branches (which should be merged before this branch lands I guess), this should fix all the bugs I know about with dasds.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -636,12 +636,6 @@ class _Device(_Formattable, ABC):
     def ptable_for_new_partition(self):
         if self.ptable is not None:
             return self.ptable
-        for action in self._m._orig_config:
-            if action['id'] == self.id:
-                if action.get('ptable') == 'vtoc':
-                    return action['ptable']
-        if self.dasd() is not None:
-            return 'vtoc'
         return 'gpt'
 
     def partitions(self):
@@ -781,6 +775,17 @@ class Disk(_Device):
         }
         return dinfo
 
+    def ptable_for_new_partition(self):
+        if self.ptable is not None:
+            return self.ptable
+        dasd_config = self._m._probe_data.get('dasd', {}).get(self.path)
+        if dasd_config is not None:
+            if dasd_config['type'] == "FBA":
+                return 'msdos'
+            else:
+                return 'vtoc'
+        return 'gpt'
+
     @property
     def size(self):
         return align_down(self._info.size)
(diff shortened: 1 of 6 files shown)


mwhudson:
i should add dasdview output for a FBA drive but the s390x system is still inaccessible due to the data centre move.

subiquity/models/filesystem.py
```diff
@@ -781,9 +775,31 @@ def info_for_display(self):
...
+            return align_down(self._info.size)
+        # An unformatted (ECKD) dasd reports a size of 0 via e.g. blockdev
+        # --getsize64. So figuring out how big it is requires a bit more work.
+        data = self._m._probe_data.get('dasd', {}).get(self.path)
+        if data is None or data['dasd_type'] != 'ECKD':
+            return 0
+        tracks_per_cylinder = data['tracks_per_cylinder']
+        cylinders = data['cylinders']
+        blocksize = 4096  # hard coded for us!
+        blocks_per_track = 12  # just a mystery fact that has to be known
```
xnox, line 801: Scooby Dooby Doo!

subiquity/models/filesystem.py
```diff
@@ -848,7 +864,9 @@ def _can_PARTITION(self):
             return False
         if self.free_for_partitions <= 0:
             return False
-        if self.ptable == 'vtoc' and len(self._partitions) >= 3:
+        # we only use msdos partition tables with FBA dasds, which
+        # only support 3 partitions.
+        if self.ptable in ['vtoc', 'msdos'] and len(self._partitions) >= 3:
```
xnox, line 869: Hm.... if we are resuing partitions, on any drive on any architecture we may hit msdos. Which may have 4 or more partitions (thanks to the extended partitions).

I guess that would be a TODO to fix, if we ever want to support msdos more fully.

xnox (approved):

subiquity/models/filesystem.py
raharper: Don't we know DASD type? if not, you could scope the 3 partition limit to arch=s390x to avoid tripping existing MSDOS layouts on non-s390x.

xnox: there is `self.dasd`.... but i can't tell if it is set when this code runs. @mwhudson ? is self.dasd live at this point, and can we do `if self.dasd` check instead of `self.ptable in ['vtoc', 'msdos']` given that only these two are allowed on dasd. gpt just blows up on them.

PR #867, opened 2020-12-08
check udev data directly when processing match: specs

mwhudson:
also allow matching on ID_VENDOR

### DIFF
--- a/subiquity/models/tests/test_filesystem.py
+++ b/subiquity/models/tests/test_filesystem.py
@@ -851,18 +851,25 @@ class TestFilesystemModel(unittest.TestCase):
         self.assertTrue(dos_esp.is_esp)
 
 
+def fake_up_blockdata_disk(disk, **kw):
+    model = disk._m
+    if model._probe_data is None:
+        model._probe_data = {}
+    blockdev = model._probe_data.setdefault('blockdev', {})
+    d = blockdev[disk.path] = {
+        'DEVTYPE': 'disk',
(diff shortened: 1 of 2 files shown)


subiquity/models/tests/test_filesystem.py
```diff
@@ -856,6 +856,8 @@ def fake_up_blockdata(model):
     for disk in model.all_disks():
         bd[disk.path] = {
             'DEVTYPE': 'disk',
+            'ID_SERIAL': disk.serial,
+            'ID_MODEL': disk.model,
```
raharper, line 860: Did you want a separate disk where you match ID_VENDOR?  

Also, do we know what happens if multiple disks match? both VENDOR and MODEL are likely to hit duplicates in a multi-disk environment.

mwhudson: > Did you want a separate disk where you match ID_VENDOR?

Good point, done.

> Also, do we know what happens if multiple disks match? both VENDOR and MODEL are likely to hit duplicates in a multi-disk environment.

This is documented! https://ubuntu.com/server/docs/install/autoinstall-reference -> 

    Any disk action is assigned a matching disk – chosen arbitrarily from the set of unassigned disks if there is more than one, and causing the installation to fail if there is no unassigned matching disk

and

    size: largest|smallest: take the largest or smallest disk rather than an arbitrary one if there are multiple matches (support for smallest added in version 20.06.1)

PR #869, opened 2020-12-17
append, do not overwrite logs

mwhudson:
The log file names have pids in now, but when subiquity re-execs itself
to fake a snap refresh the pid doesn't change. Having the pre "refresh"
logs get overwritten does not help anything and is sometimes very
annoying.


mwhudson:
failure was a glitch

PR #873, opened 2020-12-18
smoosh InstallState into ApplicationState

mwhudson:
Having these be separate was slightly confusing, particular as I need to
add an "ERROR" state that isn't necessarily due to the install step
itself failing. Now we have one enum for the overall status of the
installer and a separate field that indicates whether the client should
be in interactive or non-interactive mode (or "not-yet-known" which is
handled more or less like non-interactive mode).


mwhudson:
We need to get off travis

PR #879, opened 2021-01-22
wait for cloud-init to complete after starting api server

mwhudson:
Currently the api server waits for cloud-init to complete very very
early, before starting the API server. This can leave the client at a
message that just says "connecting...." for quite a few seconds. This
branch adds a couple more states to ApplicationState to use during
startup and has the client print appropriate messages for each one.

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -234,21 +234,46 @@ class SubiquityClient(TuiApplication):
                 self.exit()
 
     async def connect(self):
-        print("connecting...", end='', flush=True)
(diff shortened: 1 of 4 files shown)


mwhudson:
Here's what this looks like with an artificial delay in waiting for cloud-init: https://asciinema.org/a/YtxLgM0Z7o6zlJx9omqexXPLQ

mwhudson:
For anyone watching along at home, this was a bit broken but I fixed with a direct push to main.

PR #882, opened 2021-01-26
client: lookup machine IPs via API call

dbungert:
Attempts to look them up directly fail due to an empty app.base_model.
Add an API call invocation to grab the public IPs.


mwhudson (approved):
Perfect thanks :champagne:

PR #885, opened 2021-02-03
Corrected oc language name to "Occitan"

elcste:
The language with the code `oc` is is [Occitan](https://en.wikipedia.org/wiki/Occitan_language) (same name in English and in Occitan). Galés means ["Welsh" in Occitan](https://oc.wikipedia.org/wiki/Gal%C3%A9s)(‽), so "oc:Galés" is like having "en:Welsh".

### DIFF
--- a/languagelist
+++ b/languagelist
@@ -25,4 +25,4 @@ console:sv:Svenska
 ssh:bo:བོད་ཡིག
 console:uk:Українська
 console:nb:Norsk bokmål
-ssh:oc:Galés
+ssh:oc:Occitan


xnox:
`scripts/make-language-lists` is the one that generates this file.

and "name" variable was not reset at all, thus when things are not in the languagelist we eneded up calling name on the last thing that was processed, which happened to be welsh.

as per 639-3 the "oc" language name is `Occitan (post 1500)` will "ssh:oc:Occitan (aprèp 1500)" do? or is that weird as well?

If that's weird, i can add an override for OC to be just "Occitan"

xnox:
Hopefully fixed slightly more generically in https://github.com/CanonicalLtd/subiquity/pull/887/files

elcste:
> as per 639-3 the "oc" language name is `Occitan (post 1500)` will "ssh:oc:Occitan (aprèp 1500)" do? or is that weird as well?
> 
> If that's weird, i can add an override for OC to be just "Occitan"

I'm not an Occitan speaker, but my university degree is in linguistics, so I have some area knowledge :-) I do think the "(aprèp/post 1500)" is weird for this use, even if it is correct as per the ISO code name. The best analogies I think are Greek and Hebrew, since they are also languages with modern, spoken versions and widely-studied historical varieties with a large literature. Both entries here have simply "Greek" and "Hebrew" in the respective languages.

PR #886, opened 2021-02-03
The native name of "kab" language corrected to " Taqbaylit"

SelyanKab:
the native name of kabyle language "kab" corrected.

### DIFF
--- a/languagelist
+++ b/languagelist
@@ -1,6 +1,6 @@
 console:en_US:English
 console:en_GB:English (UK)
-ssh:kab:Tamaziɣt Taqbaylit
+ssh:kab:Taqbaylit
 console:ca:Català
 ssh:zh_CN:中文(简体)
 console:hr:Hrvatski


xnox:
Hi,

Althought languagelist is what is committed in git, it is made by the `./scripts/make-language-lists`

And that's where we have
```
    ('ssh',     'kab',   'Tamaziɣt Taqbaylit'),
```
https://github.com/CanonicalLtd/subiquity/blob/main/scripts/make-language-lists#L26

So a patch against that file would be nice, such that the bug doesn't come back next time we regenerate the languagelist.

Based on https://en.wikipedia.org/wiki/Kabyle_language "Taqbaylit" is the native name of the language, where as "Tamaziɣt" seems to be a name of many languages in the language group that kab is part of https://en.wikipedia.org/wiki/Berber_languages . Or something like that.

Thus this change looks correct.

xnox:
Fixed more generically in https://github.com/CanonicalLtd/subiquity/pull/887/files

Looks like kab language doesn't ship 639-3 iso gettext translations, but it does have 639 ones. So let's use those, without special case.

PR #887, opened 2021-02-04
scripts: fix make-language-lists for non-languagelist languages

xnox:
Fixes #885

### DIFF
--- a/languagelist
+++ b/languagelist
@@ -1,6 +1,5 @@
 console:en_US:English
 console:en_GB:English (UK)
-ssh:kab:Tamaziɣt Taqbaylit
 console:ca:Català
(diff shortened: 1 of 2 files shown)


mwhudson:
I guess you developed this on groovy or hirsute? Unfortunately the iso-codes in focal (let alone bionic, which is where subiquity snaps are still built) does not know about Kabyle / Taqbaylit :( So I guess we can merge this, as it looks better, but we'll still need to special case kab.

mwhudson:
pushed some changes, let me know what you think

mwhudson:
Going to merge this, there is another change I want to make that will conflict if i don't merge this first.

PR #890, opened 2021-02-10
do not append .UTF-8 to language in cloud-init user data

mwhudson:
"locale-gen fr" succeeds, "locale-gen fr.UTF-8" does not. possibly we should be writing
"fr_FR.UTF-8" but this is a simple fix.

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -168,7 +168,7 @@ class SubiquityModel:
             'growpart': {
                 'mode': 'off',
                 },
-            'locale': self.locale.selected_language + '.UTF-8',
+            'locale': self.locale.selected_language,
             'resize_rootfs': False,
         }
         if self.identity.hostname is not None:


paride (approved):
LGTM, but apparently a test needs to be updated (CI failure).

xnox (commented):
does this end up generating UTF-8 locale too, in addition to all the others? (i.e. latin1 and regional encoding ones?)

otherwise i would want to pass the full `fr_FR.UTF-8`

cause if we don't generate a UTF-8 one, things will remain broken.

paride:
I tried it (on Focal) and it generated *only* UTF-8 locales:

```
root@paride-f:~# locale-gen fr
Generating locales (this might take a while)...
  fr_BE.UTF-8... done
  fr_CA.UTF-8... done
  fr_CH.UTF-8... done
  fr_FR.UTF-8... done
  fr_LU.UTF-8... done
Generation complete.
```

By contrast:

```
root@paride-f:~# locale-gen fr_FR
Generating locales (this might take a while)...
  fr_FR.ISO-8859-1... done
Generation complete.
```

mwhudson:
so it turns out that while "locale-gen fr" generates UTF-8 locales, "locale-gen en_US" generates the 8859-1 one :( I think we should fix this vaguely properly by having the language list file contain a full "<language>_<locale>.<charset>" identifier and pushing that through. We can pull the default locale for a language from the data in localechooser-data. 

For autoinstall compat we should probably a be a little bit smarter about what we write to the userdata (e.g. only append ".UTF-8" if the supplied value looks like "<language>_<locale>")

PR #891, opened 2021-02-12
put a full locale (language/location/codeset) in languagelist

mwhudson:


xnox (approved):

subiquity/models/subiquity.py
```diff
@@ -164,11 +164,14 @@ def get_target_groups(self):
...
     def _cloud_init_config(self):
+        locale = self.locale.selected_language
+        if '.' not in locale and '_' in locale:
+            locale += '.UTF-8'
         config = {
             'growpart': {
                 'mode': 'off',
                 },
-            'locale': self.locale.selected_language + '.UTF-8',
+            'locale': locale,
```
xnox, line 174: shouldn't this also install the right language pack for the language? that way one will get translations to be available.

mwhudson: Yes, I guess so -- there doesn't seem to be anything in cloud-init to install them. I guess this gets into issues around whether the langpacks should be included in server-ship-live and/or whether the install had network access... something for another branch I guess.

PR #893, opened 2021-02-18
switch to github actions

mwhudson:


mwhudson:
Seems simple enough, a couple of tweaks here and there.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,24 @@
...
+        image:
+          - ubuntu-daily:bionic
+          - ubuntu-daily:focal
+          - ubuntu-daily:groovy
+          - ubuntu-daily:hirsute
+    steps:
+    - uses: actions/checkout@v2
+    - name: run
+      run: sudo ./scripts/test-in-lxd.sh ${{ matrix.image }} ./scripts/runtests.sh
+  lint:
```
dbungert, line 19: I've heard it suggested before to lint before running tests, as the lint step should be quicker and might save some test cycles on a build that's going to fail on lint anyhow.
Otherwise LGTM.  A shame we can't test it yet until we change what org it's on.

mwhudson: That sounds like a good idea but something for the next branch maybe :)

For testing, the actions ran on push to my repo at least.

PR #900, opened 2021-02-23
inject-subiquity-snap: log xorriso errors

paride:


mwhudson:
I wonder why i put the redirect there in the first place. Anyway, this makes sense, thanks.

PR #901, opened 2021-03-01
create and use an API for guided disk usage

mwhudson:
Not completely sure of the approach here. The types in the API in this PR expose pretty much exactly the information needed to implement subiquity's current UI for the guided storage config screen which is fine for a start I guess.

The part that makes me uncertain going on from this is how much to do server side vs client side. For example: when a RAID device is created, the disks or partitions that go into it should gain a label saying 'part of RAID mdX'. If we keep labels something the server decides, then there needs to be an API call to create the RAID device that then somehow indicates that the label of the partition should change. If we keep it client side, as currently implemented, there is more chance that the clients will diverge in behaviour, which is _probably_ bad?

Probably in general things should be done more server side, because I still want to implement the behaviour where you can plug a USB stick in while looking at the device screen and have it appear in the UI, and it wouldn't feel right for the client to be listening to udev events itself.

So I guess this means a RESTish API for manipulating storage configuration? That sounds quite fun to do but also like a lot of work!

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -28,6 +28,8 @@ from subiquity.common.types import (
     ApplicationState,
     ApplicationStatus,
     ErrorReportRef,
+    GuidedChoice,
+    GuidedStorageResponse,
     KeyboardSetting,
     KeyboardSetup,
     IdentityData,
@@ -184,6 +186,14 @@ class API:
             def GET(dev_name: str) -> str: ...
 
     class storage:
+        class guided:
+            def GET(wait: bool = False) -> GuidedStorageResponse:
+                pass
+
+            def POST(choice: Optional[GuidedChoice]) \
+                    -> StorageResponse:
+                pass
+
         def GET(wait: bool = False) -> StorageResponse: ...
         def POST(config: Payload[list]): ...
 
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -223,6 +223,7 @@ class Disk:
     size: int
     usage_labels: List[str]
     partitions: List[Partition]
+    ok_for_guided: bool
 
 
 @attr.s(auto_attribs=True)
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -209,10 +209,41 @@ class Bootloader(enum.Enum):
 
 
 @attr.s(auto_attribs=True)
+class Partition:
+    size: int
+    number: int
+    annotations: List[str]
+
+
+@attr.s(auto_attribs=True)
+class Disk:
+    id: str
+    label: str
+    type: str
+    size: int
+    usage_labels: List[str]
+    partitions: List[Partition]
+
+
+@attr.s(auto_attribs=True)
+class GuidedChoice:
+    disk_id: str
+    use_lvm: bool = False
+    password: Optional[str] = attr.ib(default=None, repr=False)
+
+
+@attr.s(auto_attribs=True)
+class GuidedStorageResponse:
+    status: ProbeStatus
+    error_report: Optional[ErrorReportRef] = None
+    disks: Optional[List[Disk]] = None
+
+
+@attr.s(auto_attribs=True)
 class StorageResponse:
     status: ProbeStatus
-    bootloader: Optional[Bootloader] = None
     error_report: Optional[ErrorReportRef] = None
+    bootloader: Optional[Bootloader] = None
     orig_config: Optional[list] = None
     config: Optional[list] = None
     blockdev: Optional[dict] = None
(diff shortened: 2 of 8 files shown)


paride:
I may not have the full picture in mind here, but in general I think the storage config logic belongs more to the server side.

subiquity/ui/views/filesystem/guided.py
```diff
@@ -173,34 +197,33 @@ def _toggle_guided(self, sender, new_value):
...
-                self.form.as_screen(focus_buttons=False, excerpt=_(subtitle)))
-        elif found_disk:
-            super().__init__(
-                screen(
-                    [Text(rewrap(_(no_big_disks)))],
-                    [other_btn(_("OK"), on_press=self.manual)]))
+        if disks:
+            found_ok_disk = False
+            for disk in disks:
+                if disk.size > 6*(2**30):
```
dbungert, line 210: Do we aspire that this code would be used in ubuntu core like cases where even smaller disks would be permissible?  At minimum I suggest that we pull this value out as a constant.
Further this could change over time.  I suppose it'd be nice to have the ISO carry an indicator of what the minimum should be, though maybe it changes infrequently enough that that is overkill and we can just bump it manually.

subiquity/server/controllers/filesystem.py
```diff
@@ -108,36 +113,122 @@ async def apply_autoinstall_config(self, context=None):
...
+                ))
+        spec = dict(name="ubuntu-vg", devices=set([part]))
+        if lvm_options and lvm_options['encrypt']:
+            spec['password'] = lvm_options['luks_options']['password']
+        vg = self.create_volgroup(spec)
+        # There's no point using LVM and unconditionally filling the
+        # VG with a single LV, but we should use more of a smaller
+        # disk to avoid the user running into out of space errors
+        # earlier than they probably expect to.
+        if vg.size < 10 * (2 << 30):
```
dbungert, line 148: And we have some sizes here too.  I suggest pulling all the sizes into one spot so that we can concisely set policies of minimum/use all/use half/etc.

subiquity/ui/views/filesystem/guided.py
mwhudson: Better calculation of the minimum disk space is on the desktop team's wish list already but yeah, having it as a duplicated magic constant here is silly. I'm also not sure why it's as high as this tbh. server needs / to be somewhere between 1 and 2 gigs currently, I guess desktop probably has slightly higher requirements. I moved the determination into the server -- at some point the server part of the installer will need to know if it is installing desktop or server or core and can use that to decide. But not today!

subiquity/common/apidef.py
```diff
@@ -168,6 +170,14 @@ class info:
             def GET(dev_name: str) -> str: ...
 
     class storage:
+        class guided:
+            def GET(wait: bool = False) -> GuidedStorageResponse:
```
MarcusTomlinson, line 174: Just checking I understand this correctly. I call GET on 'storage/guided' and get a list of disks back in GuidedStorageResponse. I then POST a GuidedChoice containing the disk I want to install to. Simple as that?

subiquity/common/types.py
```diff
@@ -140,11 +140,43 @@ class Bootloader(enum.Enum):
...
+    usage_labels: List[str]
+    partitions: List[Partition]
+    ok_for_guided: bool
+
+
+@attr.s(auto_attribs=True)
+class GuidedChoice:
+    disk_id: str
+    use_lvm: bool = False
+    password: Optional[str] = attr.ib(default=None, repr=False)
```
MarcusTomlinson, line 165: I assume this is for password protected storage devices. How do I tell before populating a GuidedChoice whether I the disk I've selected requires a password or not?

mwhudson: It's the other way around: you pass a password if you want the device to be encrypted. We don't have any support for unlocking existing dm-crypt volumes (maybe one day...)

subiquity/common/apidef.py
mwhudson: Yes, exactly. If pass don't pass wait or pass wait=false as a url param, you might get a result with probe_status == PENDING, but if you pass wait=true the method will block until the list of disks is known. In desktoppy cases I really wouldn't expect the probe to take more than a trivial amount of time though. (It can be slow with virtual ISOs and stuff like that).

subiquity/common/types.py
MarcusTomlinson: Ah I see!

subiquity/common/apidef.py
MarcusTomlinson: Awesome, thanks for clarifying.

MarcusTomlinson (approved):
This looks good to me!

mwhudson:
OK let's merge this and we can iterate if needed!

PR #902, opened 2021-03-03
Schema generation fix and test

dbungert:
It seems schema generation was broken during the client/server refactor.
Address that.
Add a test to help enforce the current schema against unintended changes.


mwhudson (approved):
very nice, thanks for adding this to the CI

PR #904, opened 2021-03-09
do not try to reuse a name of an existing vg in guided_lvm

mwhudson:
for https://bugs.launchpad.net/bugs/1905412

### DIFF
--- a/subiquity/common/filesystem.py
+++ b/subiquity/common/filesystem.py
@@ -398,7 +398,12 @@ class FilesystemManipulator:
                 size=disk.free_for_partitions,
                 fstype=None,
                 ))
-        spec = dict(name="ubuntu-vg", devices=set([part]))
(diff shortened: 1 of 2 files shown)


mwhudson:
Haven't tested this at all yet!

mwhudson:
I've now tested this a bit and it seems to work. Will test it again tomorrow after a nice fresh cup of coffee (unless someone beats me to it)

dbungert (approved):
LGTM

dbungert:
I did a test of this as well and it seems fine.  My test was to run a sample install in kvm, create a second disk, run install a second time hitting that second disk, and then boot it and analyze the actual VGs present.  In this configuration the system booted the first install but can see both VGs.

mwhudson:
Thanks for testing!

PR #905, opened 2021-03-09
add a couple of super hacky scripts for quickly testing subiquity changes

mwhudson:
The first of these scripts mashes the current branch's code over that of
a pre-existing subiquity snap.

The second uses the first to mash the current branch's code over the
subiquity snap from an ISO and then creates a new ISO with this new
snap.

It's all a bit terrifying but being able to get your changes into a
bootable ISO in about 30s is quite a large improvement on building the
snap from scratch.

### DIFF
--- a/scripts/slimy-update-snap.sh
+++ b/scripts/slimy-update-snap.sh
@@ -55,6 +55,6 @@ add_overlay old new
 rm -rf new/lib/python3.6/site-packages/subiquity
 rm -rf new/lib/python3.6/site-packages/subiquitycore
 
-git -C $src archive HEAD subiquity subiquitycore | tar -C new/lib/python3.6/site-packages/ -x
+rsync -a $src/subiquity $src/subiquitycore new/lib/python3.6/site-packages
 
 mksquashfs new $new -comp gzip -Xcompression-level 3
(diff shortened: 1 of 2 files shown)


scripts/slimy-update-snap.sh
```diff
@@ -0,0 +1,60 @@
...
+trap cleanup EXIT
+
+do_mount $old old
+
+add_overlay old new
+
+rm -rf new/lib/python3.6/site-packages/subiquity
+rm -rf new/lib/python3.6/site-packages/subiquitycore
+
+git -C $src archive HEAD subiquity subiquitycore | tar -C new/lib/python3.6/site-packages/ -x
```
dbungert, line 58: It looks like this command won't pick up uncommitted changes.  Is it your intention that a commit be performed before being able to use slimy-update?

dbungert, line 58: It looks like this command won't pick up uncommitted changes. Is it your intention that a commit be performed before being able to use slimy-update?

What's your thoughts on this?

```suggestion
rsync -a --delete --exclude '*.pyc' --delete-excluded subiquity new/lib/python3.6/site-packages
rsync -a --delete --exclude '*.pyc' --delete-excluded subiquitycore new/lib/python3.6/site-packages
```

Or maybe leave in the rm -fr commands and we can simplify the delete directives to rsync.

dbungert (approved):

dbungert:
Nice scripts - I used them as part of my test of your lvm change and it's a really great workflow improvement.

PR #912, opened 2021-03-16
move most keyboard logic to server

mwhudson:
This could be lots of more incremental branches but keeping track of all that was getting a bit annoying. I can split it up again if that would make life easier -- I definitely recommend reading through this commit by commit.


scripts/make-kbd-info.py
```diff
@@ -0,0 +1,93 @@
...
+            if v not in pc105tree.steps:
+                raise Exception(
+                    f"step {index} {step} references missing step {v}")
+    elif isinstance(step, StepKeyPresent):
+        for v in step.yes, step.no:
+            if v not in pc105tree.steps:
+                raise Exception(
+                    f"step {index} {step} references missing step {v}")
+
+with open(os.path.join(tdir, 'pc105.json'), 'w') as out:
```
dbungert, line 92: Given that pc105.json is the final output from this, do we care if it's human readable?  ('no' is an acceptible answer)
If so we can pretty-print it https://docs.python.org/3/library/json.html

snapcraft.yaml
```diff
@@ -122,19 +122,20 @@ parts:
...
+  keyboard-data:
     plugin: nil
     build-packages:
       - console-setup
       - locales
       - xkb-data-i18n
-    override-build: |
-      /usr/share/console-setup/kbdnames-maker /usr/share/console-setup/KeyboardNames.pl > $SNAPCRAFT_PART_INSTALL/kbdnames.txt
+      - python3-attr
+    override-build: PYTHONPATH=$SNAPCRAFT_PROJECT_DIR/ $SNAPCRAFT_PROJECT_DIR/scripts/make-kbd-info.py
```
dbungert, line 132: I'm wondering if we want an integration test step which fails the build if the committed version of any file doesn't match the generated version, sort of like what I did with the schema file but applied more generically.
Put differently, if running the build is known to not change any committed file, then I think we could safely ignore steps like doing make-kbd-info.py at snapcraft time.  What do you think?

dbungert (commented):
Just a few superficial questions

scripts/make-kbd-info.py
mwhudson: I actually tried this and the output is almost as unreadable because the bulk of the data are fairly short lists which get spread over 1000 or so lines. If there was a way of pretty printing that did lists compactly and objects more verbosely that would be worth trying but I don't think there is.

snapcraft.yaml
mwhudson: It's an interesting idea. One thing that's a bit different here is that the keyboard data comes from the console-setup -- if the data is different between focal and hirsute, say, then the test you're proposing would fail on one of these systems. This is something that is actually a change in this branch: previously subiquity parsed the pc105.tree file from the live installer environment, but I think this is a good thing as it seems more proof against that file changing (or moving or disappearing) i a future release of Ubuntu and breaking us.

Of course all of this is a bit theoretical as I don't think any of this stuff is very likely to change any time soon.

PR #913, opened 2021-03-16
add an action to build each PR into a snap for testing

mwhudson:
Not at all sure I have the bits set up on github to get this working quite right yet but let's give it a go.


mwhudson:
Hmm that doesn't work, and the UI for approving "deployments" is a bit inappropriate for this. I think an action triggered by a comment from a maintainer that builds and publishes a snap is probably a more useful thing.

PR #914, opened 2021-03-17
Update NAME_REGEX as per bug #1916325

jmhunter:
The NAME_REGEX currently used is "[a-z_][a-z0-9_-]*" which does correctly prevent hostnames starting with a hyphen.. but also prevents otherwise perfectly valid hostnames that start with a number. (Interestingly, an underscore is not defined as a valid hostname character, but is included as valid in the regex, I'm not sure why this is included but am leaving it in for backwards compatibility)

This patch adds 0-9 as valid characters at the start of a hostname, as per https://man7.org/linux/man-pages/man7/hostname.7.html#:~:text=Valid%20characters%20for%20hostnames%20are,to%20an%20address%20for%20use.
       Each element of the hostname must be from 1 to 63 characters long
       and the entire hostname, including the dots, can be at most 253
       characters long. Valid characters for hostnames are ASCII(7)
       letters from a to z, the digits from 0 to 9, and the hyphen (-).
       A hostname may not start with a hyphen.

### DIFF
--- a/subiquity/ui/views/identity.py
+++ b/subiquity/ui/views/identity.py
@@ -107,8 +107,8 @@ class IdentityForm(Form):
                 "Server name too long, must be less than {limit}"
                 ).format(limit=HOSTNAME_MAXLEN)
 
-        if not re.match(r'[a-z_][a-z0-9_-]*', self.hostname.value):
-            return _("Hostname must match NAME_REGEX, i.e. [a-z_][a-z0-9_-]*")
+        if not re.match(r'[a-z0-9_][a-z0-9_-]*', self.hostname.value):
+            return _("Hostname must match NAME_REGEX, i.e. [a-z0-9_][a-z0-9_-]*")
 
     def validate_username(self):
         username = self.username.value
--- a/subiquity/ui/views/identity.py
+++ b/subiquity/ui/views/identity.py
@@ -40,9 +40,11 @@ from subiquity.common.types import IdentityData
 log = logging.getLogger("subiquity.views.identity")
 
 HOSTNAME_MAXLEN = 64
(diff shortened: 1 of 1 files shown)


jmhunter:
(This is my first pull request here - let me know if any issues!)

jmhunter:
Looks like the 'lint' check fails due to one line being two characters over- length.

Is this a problem - does the line need to be split to pass the check? In my view it's fine but I don't know if this would cause a problem later on - in which case I can of course split the text string into two parts..

subiquity/ui/views/identity.py
```diff
@@ -107,8 +107,8 @@ def validate_hostname(self):
                 "Server name too long, must be less than {limit}"
                 ).format(limit=HOSTNAME_MAXLEN)
 
-        if not re.match(r'[a-z_][a-z0-9_-]*', self.hostname.value):
-            return _("Hostname must match NAME_REGEX, i.e. [a-z_][a-z0-9_-]*")
+        if not re.match(r'[a-z0-9_][a-z0-9_-]*', self.hostname.value):
```
dbungert, line 110: With the duplication between these two lines, maybe we could fix the lint and the redundancy by pulling the regex out to a variable when we can then use for both lines.  What do you think?

jmhunter: I agree, thanks. I had initially thought to keep the changes as small as possible but I do agree.

I've now checked out the code locally (rather than direct editing on github UI!) and am also testing using 'make dryrun' now. I'm making the change for username as well as hostname, since these are both structured in the same way - I'll update shortly.

mwhudson:
Thanks!

PR #918, opened 2021-03-24
have subiquity set password for installer user in live session

mwhudson:
We used to have cloud-init do this but it hides the password better now.

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -50,7 +50,7 @@ from subiquity.common.types import (
 from subiquity.journald import journald_listen
 from subiquity.ui.frame import SubiquityUI
 from subiquity.ui.views.error import ErrorReportStretchy
-from subiquity.ui.views.help import HelpMenu
+from subiquity.ui.views.help import HelpMenu, ssh_help_texts
 from subiquity.ui.views.installprogress import (
     InstallConfirmation,
     )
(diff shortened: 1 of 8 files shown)


mwhudson:
Hm, this will cause the password to be re-set when refreshing from a version that does not have this change to a version that does. Should probably still look in the cloud-init log file to avoid that.

xnox:
This is good for when there is no custom cloud-init metadata.

Can you check that somebody didn't provide userdata for the live session and set custom password and/or ssh keys? Ideally, if ssh key was provided for the installer@ user we should not be setting password auth at all. Or if a non-randome password was already supplied, we shouldn't be resetting it.

mwhudson:
Yeah, that's a good point. Detecting if the user has keys is easy enough I guess, there must be a way of telling if the user has no password set...

mwhudson:
OK I think this implements the right behaviour: if the default user for cloud-init has no pasword or ssh keys, a random password is set and shown to the user. If the user has a password already, the user is invited to log in using "the preconfigured password passed to cloud-init"), and the fingerprints of any authorized keys are shown.

Not 100% sure each permutation is shown to the user with ideal amounts of vertical whitespace but well.

mwhudson:
Going to merge this now -- if it's not perfect, it's definitely an improvement and we can iterate as needed.

PR #919, opened 2021-03-25
fix submitting error reports

mwhudson:
this is a bit embarassing!


dbungert (approved):
Confirmed per docs that pipe() returns in order (r, w)

PR #920, opened 2021-03-26
add setting policy for post-install updates via autoinstall

dbungert:
This can be controlled with autoinstall
updates: security or all

Also an API for controlling this:
curl --silent --unix-socket .subiquity/socket a/updates ->
"security"
curl -d '"all"' --unix-socket .subiquity/socket a/updates


subiquity/models/subiquity.py
```diff
@@ -81,6 +82,7 @@ def setup_yaml():
     "packages",
     "snaplist",
     "ssh",
+    # "updates", # FIXME hey mwhudson - do I need this?
```
mwhudson, line 85: Not unless you want to force the client to provide a value for this -- which I don't think we do.

subiquity/common/apidef.py
```diff
@@ -44,11 +44,12 @@
...
 class API:
     """The API offered by the subiquity installer process."""
+    identity = simple_endpoint(IdentityData)
     locale = simple_endpoint(str)
-    proxy = simple_endpoint(str)
     mirror = simple_endpoint(str)
-    identity = simple_endpoint(IdentityData)
+    proxy = simple_endpoint(str)
     ssh = simple_endpoint(SSHData)
+    updates = simple_endpoint(str)
```
mwhudson, line 52: I'd ever so slightly prefer to have this API method take an enum rather than a string.

mwhudson (commented):
Hey, this looks pretty good! Just a couple of comments. I guess we should have a conversation about whether we really do want to allow autoinstalls to disable security updates.

subiquity/server/controllers/updates.py
```diff
@@ -0,0 +1,58 @@
...
+        self.model.updates = data
+
+    def make_autoinstall(self):
+        return self.model.updates
+
+    def serialize(self):
+        return self.model.updates
+
+    def deserialize(self, data):
+        self.model.updates = data
```
mwhudson, line 51: oh, forgot: you don't need these methods

mwhudson: oh actually maybe it makes sense to keep them. i'm sure I'll have made up my mind by tomorrow :)

subiquity/models/subiquity.py
dbungert: Agree, removing this line

subiquity/server/controllers/updates.py
dbungert: SubiquityController wants deserialize.  serialize might be not required at the API level.

subiquity/common/apidef.py
dbungert: I messed with putting enum on the API boundary and it made a mess of things.  I'm willing to try again but at a certain point I cut my losses and pushed a simpler version.

mwhudson (approved):
nice

PR #922, opened 2021-03-29
catch all errors getting application status

mwhudson:
When refreshing the snap, it is possible for the client to make a "get
current status" request to the new server before it gets restarted and
the new server can return values the client does not know about. Rather
than crashing the client when this happens, just ignore all errors when
getting application status.

### DIFF
--- a/subiquity/client/controllers/progress.py
+++ b/subiquity/client/controllers/progress.py
@@ -77,7 +77,8 @@ class ProgressController(SubiquityTuiController):
             try:
                 app_status = await self.app.client.meta.status.GET(
                     cur=self.app_state)
-            except aiohttp.ClientError:
+            except Exception:
+                log.exception("getting status failed")
                 await asyncio.sleep(1)
                 continue
             self.app_state = app_status.state


subiquity/client/controllers/progress.py
```diff
@@ -77,7 +77,8 @@ async def _wait_status(self, context):
             try:
                 app_status = await self.app.client.meta.status.GET(
                     cur=self.app_state)
-            except aiohttp.ClientError:
+            except Exception:
```
dbungert, line 80: catch Exception in a while True loop makes me nervous.
What if we at least log the cases where we caught an exception but it's not ClientError?
If someone gets stuck we should get a log file positively filled with whatever they are stuck on.

dbungert (changes requested):
see other comment

subiquity/client/controllers/progress.py
mwhudson: yes that's fair

subiquity/client/controllers/progress.py
```diff
@@ -77,7 +77,8 @@ async def _wait_status(self, context):
...
+                log.exception("getting status failed")
```
mwhudson, line 81: Oh wait, this line will be filling up the log :)

dbungert: Let me back up a bit - if catching Exception gets someone stuck, I want a usable log of that event.  What do you think?

mwhudson:
Going to fix this properly, withdrawing this change

PR #925, opened 2021-03-31
add all documented fields to filesystem model classes

mwhudson:
This lets autoinstall configs set them, even though nothing in subiquity
can set them yet.

Trying again now we've removed the multipath coalescing code that caused this to break the world last time.


dbungert:
So I see the comment about "nothing in subiquity can set them yet" - but it also looks like nothings acts on them yet?  Is this just laying a groundwork for followup?

mwhudson:
No the point of this change is that the fields can be propagated from the autoinstall specification to the curtin config (the code that converts the autoinstall spec into model objects iterates over all the fields the class defines, look around here: https://github.com/canonical/subiquity/blob/main/subiquity/models/filesystem.py#L1544-L1550). Let me bodge a test of this into the integration test.

dbungert (approved):
LGTM

PR #927, opened 2021-04-01
do not crash when subshell exits via signal

mwhudson:
Here is something you can try with a live server installer today: drop
into a subshell and type "kill -9 $$". The installer will crash with an
'Input/output error' from tcsetattr.

What's going on is some deep unix arcana: when the subshell exits via
signal somehow the shell's process group is still in the foreground
(even though there are no processes in it?) and calling tcsetattr() from
a background process is not allowed. So the fix for this is reasonably
simple (or at least: short): call tcsetpgrp() to put our process group
back into the foreground. A background process doing this gets sent
SIGTTOU, but if we ignore that, all is well.

Frankly this is all very strange and I would like a lie down now.

### DIFF
--- a/subiquitycore/tui.py
+++ b/subiquitycore/tui.py
@@ -16,9 +16,12 @@
 import asyncio
 import inspect
 import logging
-import yaml
+import os
+import signal
 
 import urwid
+
+import yaml
 
 from subiquitycore.async_helpers import schedule_task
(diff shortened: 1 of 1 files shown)


dbungert:
Closing this since it's superseded by #928, please reopen if I'm wrong on this.

PR #928, opened 2021-04-01
kill any foreground subprocess before restarting

mwhudson:
This fixes a problem where you drop to a shell and refresh subiquity
from that shell -- the client tries to restart but it is running in the
background and so crashes trying to modify the terminal settings. So
this kills the subprocess before restarting. This required the extremely
angry PR I sent before: forcefully killing the subprocess also crashes
the client before restart in a similar way.


dbungert (approved):
LGTM

PR #929, opened 2021-04-01
Fixed minimum number of component devices for RAID5 and RAID10

enzolis:
This fix allows creation of RAID10 with two or three component devices (See also https://bugs.launchpad.net/subiquity/+bug/1880023). Additionally the minimum number of components for RAID5 is changed to two.

**Reasoning:**
In general, the numbers should reflect what is possible to do with mdadm on the command-line. Additionally there are some non-obvious reasons:
- RAID10 using two disks offers the advantage of roughly RAID0 read speed in a single stream scenario. In comparison, RAID1 will only deliver the speed of a single component. For multi-stream reads (number streams >= number components) the speed is comparable.
- While RAID5 with two component devices is only relevant for special purposes (e.g. growing to more components, different IO semantics), a three component RAID5 is probably not uncommon.

**References:**
Unfortunately, documentation is distributed on the raid mailinglist, but the mdadm sourcecode gives some hints:
- [RAID5](https://github.com/neilbrown/mdadm/blob/5f4184557a98bb641a7889e280265109c73e2f43/Create.c#L155)
- RAID10 is a little bit more complex. The layout default is: [0x102](https://github.com/neilbrown/mdadm/blob/5f4184557a98bb641a7889e280265109c73e2f43/Create.c#L58) and the [minimum requirement for RAID10](https://github.com/neilbrown/mdadm/blob/5f4184557a98bb641a7889e280265109c73e2f43/Create.c#L226) is 
 `(layout&255) * ((layout>>8)&255) == number of disks`

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -163,9 +163,9 @@ raidlevels = [
     RaidLevel(_("0 (striped)"),  "raid0",  2, False),
     # for translators: this is a description of a RAID level
     RaidLevel(_("1 (mirrored)"), "raid1",  2),
-    RaidLevel(_("5"),            "raid5",  3),
+    RaidLevel(_("5"),            "raid5",  2),
     RaidLevel(_("6"),            "raid6",  4),
-    RaidLevel(_("10"),           "raid10", 4),
+    RaidLevel(_("10"),           "raid10", 2),
     ]
 
 


xnox:
aren't RAID10 with 2 drives & RAID5 with 3 drives a dangerous combination, i.e. failure of any drive results to loss of data.

Whilst one can make those, one maybe shouldn't?

enzolis:
Short answer: No.

Long answer:
In RAID5 you always have one "drive" (actually the parity is distributed equally over all component devices to improve performance) redundancy (and in two component configuration half capacity). If one drive fails you will not lose any data.

In RAID10 case (please remind yourself we are talking about a special mode of linux raid, not a conventional RAID0-over-RAID1s) you select the redundancy yourself using the layout parameter, e.g. n2 --> near2 layout (one drive redundancy), f3 --> far3 layout (two drives redundancy). Minimum number of component devices is two, which gives you one drive redundancy.

mwhudson:
This is one of those things where there is a conflict between making something easy and making something possible. While you *can* make a RAID5 with two drives, it's definitely a bit unusual and not something that people should be able to do by mistake, I think? (iirc, we've also had requests to allow creation of degraded RAID1 arrays with a single drive, so more drives can be added later). So if we can design a sensible UI for "no, really, this is actually want I want to do", we could allow RAID5 with two disks.

I think for RAID10 we should just change it though.

I hate how terrible the documentation is for all this stuff :(

enzolis:
In general we should remember this is about server installation and not desktop installation. Thus I think limiting people from what they want to do might be more often annoying than helpful. But this is just a personal opinion.

In this specific case I would probably agree regarding creation of degraded RAIDs, but in case of RAID5 with two disks the "unintentional user" would not suffer from big disadvantages (Same redundancy, roughly same write speed, same read speed on single stream, half read speed on multi stream). On the other hand I see valid use-cases, e.g. home-servers which grow over time and get additional HDDs on demand (even though I would strongly recommend RAID6 in this case).

Nevertheless, my main interest is correcting RAID10, since we use a lot of bare metal machines with two (NVMe) drives and we would like to move from debian installer autoinstallation (bypassing partman completely) to a out of the box subiquity installation.
If you think RAID10 is OK, please merge this part.

mwhudson:
https://github.com/canonical/subiquity/pull/932 <- RAID10

For autoinstall, does this matter? I don't think subiquity enforces these minimums in this case... oh but [curtin does](https://git.launchpad.net/curtin/tree/curtin/block/mdadm.py#n529). Sigh.

mwhudson:
Made a curtin merge request too: https://code.launchpad.net/~mwhudson/curtin/+git/curtin/+merge/400931

enzolis:
Thanks for that, I was not aware there is a second place for this check.
Actually I assumed it is a good idea to first install manually to get the yaml before fiddling with the details of the autoinstall.

PR #930, opened 2021-04-02
Leverage code from Ubiquity to get rst presence

dbungert:
Example:
curl --silent --unix-socket .subiquity/socket a/storage/has_rst ->
false

Origin commit:
https://git.launchpad.net/ubiquity/commit/?id=e5c7a06d97f017296bedf593c5966e931e9a8aec

Note that ubiquity believed this detection to be incomplete, so we
should explore the possibility that there is more to do here.

But then again, the comment predates the actual commit, so maybe it's ok?

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -206,7 +206,6 @@ class API:
             def GET() -> RstResponse:
                 pass
 
-
     class snaplist:
         def GET(wait: bool = False) -> SnapListResponse: ...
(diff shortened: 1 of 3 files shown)


mwhudson:
I know last time I asked you to make the API more typed, but this time, surely this method can just return bool?

Otherwise, completely fine. At some point I guess we'll want a way to force this true for testing but that can wait?

mwhudson (approved):
Thanks! Merge at will.

PR #932, opened 2021-04-11
reduce the minimum number of devices for a RAID10 to 2

mwhudson:


mwhudson:
Argh but curtin has its own checks, so this isn't enough.

dbungert:
I think the curtin conversation has settled, shall we wake these two MPs up?

PR #933, opened 2021-04-12
remove some dead code

mwhudson:
No particular reason to sneak this in before release, but otoh it should be pretty harmless.


dbungert (approved):
LGTM

PR #934, opened 2021-04-12
replace use of urwid signals in server and fix some proxy bugs

mwhudson:
Although I said #933 didn't need to be in the release, perhaps this makes it worth it. It's a refactor, so a bit risky, but it also fixes a bug.


dbungert (approved):
LGTM

PR #935, opened 2021-04-12
guided_GET: allow caller to specify min size

dbungert:
Caller to guided_GET can now specify a minimum size disk that is
acceptable for guided partitioning.  The unit is bytes.
The default value if this is unspecified is 6GiB.

Note that this default value is wholly inadequate for Desktop, which by
my measurements needs at least 8GiB to get the install completed if you
tell it to install everything, even though later it settles to a lower
installed size of 6.6GiB.


mwhudson:
Hmm I had a slightly different model for how this would work, which is that the server would provide the client with a list of "sources" that can be installed from, the client would choose one of these to use and then the server would know how much space that choice required. That's all a bunch more work though so this is probably OK for a start...

dbungert:
Agree - this was meant to be the simple stop-gap solution.

mwhudson (approved):

PR #936, opened 2021-04-12
fix keyerror: updates

dbungert:
While playing with interactive-sections I found that I could trigger a keyerror exception where 'updates' was not found.
Here's a quick fix for now and I want to ponder adding a test case.


mwhudson:
Yeah, I've seen this too in some situations I could never pin down.

PR #939, opened 2021-04-19
make sure freshly allocated action ids are really unique

mwhudson:
Because the server and any clients can create action objects now,
per-process global counters are not going to cut it.

For https://bugs.launchpad.net/subiquity/+bug/1925063.


mwhudson:
fwiw this was exposed by moving the guided stuff into the server but it could have happened before that if you used two clients to complete the configuration (for example if your ssh connection dropped and you reconnected)

dbungert (approved):

PR #941, opened 2021-04-19
Drop the confirm_ssh_keys call if UI not as expected

dbungert:
There is a larger problem here of the UI body changing underneath of us,
but this is at least a baby step in the right direction.
Also this matches existing client/controllers/ssh.py behavior.
At least log the event when it happens.
LP: #1925048


mwhudson (approved):
Thanks

PR #942, opened 2021-04-20
Use networkctl delete/down

dbungert:
By using networkctl delete/down, we should be able to let networkctl
decide if more drastic operations are needed, and drop the forced
network restart.

Marking draft because I haven't tested this to my liking yet.

### DIFF
--- a/subiquitycore/controllers/network.py
+++ b/subiquitycore/controllers/network.py
@@ -212,24 +212,19 @@ class BaseNetworkController(BaseController):
     def apply_config(self, context=None, silent=False):
         self.apply_config_task.start_sync(context=context, silent=silent)
 
-    async def _down_devs(self, devs):
+    async def _operate_on_devs(self, op, devs):
         for dev in devs:
-            try:
-                log.debug('downing %s', dev.name)
-                self.observer.rtlistener.unset_link_flags(dev.ifindex, IFF_UP)
-            except RuntimeError:
-                # We don't actually care very much about this
-                log.exception('unset_link_flags failed for %s', dev.name)
-
-    async def _delete_devs(self, devs):
-        for dev in devs:
-            # XXX would be nicer to do this via rtlistener eventually.
-            log.debug('deleting %s', dev.name)
-            cmd = ['ip', 'link', 'delete', 'dev', dev.name]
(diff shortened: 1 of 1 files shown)


mwhudson:
Hi this looks very nice but unfortunately I don't think it's going to work with the networkctl in focal (lacks "down") or bionic (doesn't have "down" or "delete"). I guess it would be OKish to not offer future updates with this change to bionic but we should keep improving focal for a while at least :wink:. I guess this means some conditionals and not being able to delete the old code (sadly)

mwhudson (changes requested):
See previous reply

slyon (approved):
This LGTM from a code POV :eyes: 
And it is a nice cleanup of code, using different technologies (rtlistener, iproute2, systemctl), unifying to just `networkctl`!

I agree with Michael's stance on Focal compatibility, though... looking it up in the systemd repo, this seems to be a rather small & simple change. Maybe we could get this SRUed into the Focal systemd, as a targeted fix, instead of keeping the old code. What do you think?
https://github.com/systemd/systemd/commit/c30ffcee9b4960987bc8cf201138aaf0d7b26b14

This should only be merged once the Focal compatibility issue is resolved in one way or another.

xnox:
> This should only be merged once the Focal compatibility issue is resolved in one way or another.

not really. We offer subiquity snap refresh, but that will not pull in new core (still core18) nor update the systemd package on the host (which can be 20.04.0 release).

The change is quite small and is scoped to networkctl itself, maybe we can vendor patched networkctl binary into subiquity snap?

dbungert:
Thank you all for the feedback.  I see now that I can't assume that networkctl has all the capability I require.
I lean toward the 'vendored' solution - that should keep the code tidy, and we can document why we're vendoring networkctl so that we can remember in a few years to remove it.

mwhudson:
Oh if it's just the networkctl binary let's do that. I'd assumed it would involve dbus messages and so on.

mwhudson:
Although! Given that the code in networkctl appears to do exactly the same as the code being removed here from subiquity, I'm a touch confused as to why the messing around with the systemd-networkd service was necessary? (I wrote all this code, I just don't remember why).

I'm still in favour of using networkctl's code fwiw.

PR #943, opened 2021-05-02
have console-conf-write-login-details exit with code 2 when no IPs found

mwhudson:
Also make the log files resulting from console-conf-write-login-details
execution a bit easier to spot.


mwhudson:
The logic in console-conf-wrapper expects this behaviour but it wasn't hooked up right. Will need to be backported to core/bionic and core/focal.

dbungert (approved):

PR #944, opened 2021-05-02
fix logging unconfigured models

mwhudson:
debugging Launchpad bug 1926616 would have been much easier if this was
working!


mwhudson:
Example of what log message now looks like:

2021-05-02 23:19:14,884 DEBUG subiquity.models.subiquity:152 model keyboard for install is configured, to go {'network', 'proxy', 'mirror', 'filesystem'}

dbungert (approved):
I was wondering about those 'to go' lines, now I understand.

PR #946, opened 2021-05-03
set trusted=yes,check-date=no for cdrom pool during installation

mwhudson:
The pool is loaded off the same media as the installer and the
filesystem being installed so there's no real loss of security doing
this and it makes it easier for people to master custom ISOs with extra
packages in the pool and systems with inaccurate clocks (that cannot
reach ntp.ubuntu.com).

### DIFF
--- a/bin/subiquity-configure-apt
+++ b/bin/subiquity-configure-apt
@@ -61,7 +61,7 @@ else
 fi
 
 cat > "$TARGET_MOUNT_POINT/etc/apt/sources.list" <<EOF
-deb file:///cdrom $(lsb_release -sc) main restricted
+deb [trusted=yes,check-date=no] file:///cdrom $(lsb_release -sc) main restricted
 EOF
 
 $PY -m curtin in-target -- apt-get update


julian-klode:
Makes me wonder if I should make apt trusted=yes also turn off the date checks as a big hammer. There's little point in having them if we don't have the signature check.

But on whether the check is useful: If this persists in the installed system, it should not be set. You don't want someone else to swap your cd out under you and then install their malware from there.

I'd prefer to check signatures during install as well, so people can't insert a second disc without you knowing, but you probably need the correct day on your machine at the very least and we don't configure time before installing I guess.

mwhudson:
> Makes me wonder if I should make apt trusted=yes also turn off the date checks as a big hammer. There's little point in having them if we don't have the signature check.

TBF, I didn't actually check if the check-date=no is required when you have trusted=yes.

> But on whether the check is useful: If this persists in the installed system, it should not be set. You don't want someone else to swap your cd out under you and then install their malware from there.

It's not persistent. (Also the cdrom pool is mounted at /cdrom in the target which doesn't exist by default so you can't put a malicious cd there without already being root).

> I'd prefer to check signatures during install as well, so people can't insert a second disc without you knowing, but you probably need the correct day on your machine at the very least and we don't configure time before installing I guess.

I'm not sure I can see an attack that doesn't require you already have root in the installer (hardly an achievement if you have any kind of access to the installer environment at all).

dbungert (approved):

julian-klode (approved):
Given that it's used during install time only, that's a safe change

xnox:
The livefs editor is probably best to simply generate a new signing key on the fly and added as trusted.gpg.d one because anything else looks very very very odd.

julian-klode:
I'm sorry, I forgot about netbooting

mwhudson:
> When booting remotely over http, the fielsystem.squashfs.gpg is verified

As we discussed elsewhere, I don't think this is actually true. It probably should be true though, and you're right that the trusted=yes bit should probably be reverted until we have a better solution.

PR #947, opened 2021-05-04
mark the locale model as configured at start up

mwhudson:
The subiquity client does not ask about language in non-rich mode on a
serial port, and that's OK. But we still need the install to complete :)

There are other ways to fix this I guess -- we could not wait on the
locale model to get configured, for example, or explicitly select a
C.UTF-8 locale when "Continue in basic mode" is selected, or probably
some other things. But this works and seems OK.

### DIFF
--- a/subiquity/client/controllers/serial.py
+++ b/subiquity/client/controllers/serial.py
@@ -29,6 +29,13 @@ class SerialController(SubiquityTuiController):
         ssh_info = await self.app.client.meta.ssh_info.GET()
         return SerialView(self, ssh_info)
 
+    def run_answers(self):
+        if 'rich' in self.answers:
+            if self.answers['rich']:
+                self.ui.body.rich_btn.base_widget._emit('click')
+            else:
+                self.ui.body.basic_btn.base_widget._emit('click')
+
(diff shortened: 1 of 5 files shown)


mwhudson:
I've checked that the integration test I've just added fails without the corresponding change, which is nice.

subiquity/client/controllers/serial.py
```diff
@@ -29,6 +29,13 @@ async def make_ui(self):
         ssh_info = await self.app.client.meta.ssh_info.GET()
         return SerialView(self, ssh_info)
 
+    def run_answers(self):
+        if 'rich' in self.answers:
+            if self.answers['rich']:
+                self.ui.body.rich_btn.base_widget._emit('click')
```
dbungert, line 35: is there a downside to, instead of emit('click'), calling down to rich_mode/basic_mode?

mwhudson: Well my vague goal for "answers" is that the answers should operate the GUI as much as possible (not that we're very consistent about this). Possibly this is going a bit far...

dbungert (approved):

PR #949, opened 2021-05-05
have console-conf-write-login-details exit with code 2 when no IPs found

mwhudson:
Also make the log files resulting from console-conf-write-login-details
execution a bit easier to spot.

I'd missed when fixing this on main that the text to display is never recalculated when console-conf-write-login-details exits with code 0.

### DIFF
--- a/console_conf/cmd/write_login_details.py
+++ b/console_conf/cmd/write_login_details.py
@@ -14,14 +14,22 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import logging
 import sys
-from console_conf.controllers.identity import write_login_details_standalone
+
 from subiquitycore.log import setup_logger
+from subiquitycore import __version__ as VERSION
+
+from console_conf.controllers.identity import write_login_details_standalone
 
 
 def main():
-    setup_logger(dir='/var/log/console-conf')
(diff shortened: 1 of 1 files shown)


mwhudson:
Relevant part of the core20 snap build process https://github.com/snapcore/core20/blob/master/snapcraft.yaml#L12-L46

dbungert (commented):
Does the caller care about the different states?  We could give it different error codes in response if so.  Otherwise this looks fine.

mwhudson:
The tortuous logic that's in play here is in this file: https://github.com/canonical/subiquity/blob/main/bin/console-conf-wrapper

The intent is that console-conf-write-login-details is called every 5s until the system has an IP. console-conf-write-login-details was intended to exit with code 0 if there is an IP, code 2 if there is no IP and some other code (probably 1) if something else went wrong. Because of the bug this PR is fixing, console-conf-write-login-details exited with code 0 even if there was no IP, so the wrapper script thought that the file it wrote was final and so doesn't bother calling console-conf-write-login-details again.

Revisiting all of this it seems a bit fragile and overcomplicated and possibly due for a rethink and redesign but well. I think this change will make it work, at least.

bugraaydogar:
I can confirm that the suggested changes fix the mentioned problem.

dbungert (approved):

PR #950, opened 2021-05-10
offer to run a shell after an autoinstall fails

mwhudson:
otherwise, especially on a serial console, there is not really any way
to debug what is going on.

also, when non-interactive defer setting the applicationstate to ERROR
until the information in the error report has been collected and the
error-commands have run.

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -234,6 +234,11 @@ class SubiquityClient(TuiApplication):
             elif confirm_task is not None:
                 confirm_task.cancel()
                 confirm_task = None
+            if app_state == ApplicationState.ERROR:
+                await asyncio.sleep(1)
+                print("An error occurred. Press enter to start a shell")
+                await run_in_thread(input)
+                os.execvp("/bin/bash", ["/bin/bash"])
             try:
                 app_status = await self.client.meta.status.GET(
                     cur=app_state)
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -308,7 +308,12 @@ class SubiquityServer(Application):
         await report._info_task
         Error = getattr(self.controllers, "Error", None)
         if Error is not None and Error.cmds:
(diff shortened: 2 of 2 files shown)


dbungert (commented):
I love the idea and had a crash handy, so I rebased a copy of my branch to this but found that I didn't see the terminal on serial.
Let me retest again and see if I get different results - I assume we should get the shell prompt in the exception case?

mwhudson:
I tried this in dry-run mode (using SUBIQUITY_DEBUG=install-fail) but I haven't tried in a VM. Seems like that would be a good thing to do...

dbungert:
Did a new version of this test and it's working for me.  I'll think about the other case.

Interesting detail - schema validation failure on serial produces helpful output, whereas that output is sent elsewhere in the non-serial case.  Example serial output follows:

connecting...  
waiting for cloud-init...  
An error occurred. Press enter to start a shell
finish: subiquity/Updates/load_autoinstall_data: 'justno' is not one of ['security', 'all']

Failed validating 'enum' in schema:
    {'enum': ['security', 'all'], 'type': 'string'}

On instance:
    'justno'

dbungert (approved):

mwhudson:
> Interesting detail - schema validation failure on serial produces helpful output, whereas that output is sent elsewhere in the non-serial case

What do you mean by "non-serial case" -- it should be the same in all clients, or at least, that's the idea...

dbungert:
So I used a modified version of my kvm-test script that I have started in another PR, and intentionally gave it bad autoinstall data as a method to try to trigger the shell behavior in this PR, and it did so.
If I told kvm to give me a serial connection (-nographic -append console=ttyS0), I get both the shell and the schema validation failure output on the serial terminal.
If I don't run in serial, the schema validation output is not shown on the same terminal as the one that offers me a shell prompt.
I can write up more concrete steps if this is interesting.

PR #953, opened 2021-05-12
include the traceback in the response when a server method fails

mwhudson:
example:

$ curl  --unix-socket .subiquity/socket a/dry_run/crash
  File "/home/mwhudson/src/subiquity/subiquity/common/api/server.py", line 122, in handler
  File "/home/mwhudson/src/subiquity/subiquity/server/dryrun.py", line 24, in crash_GET
    1/0
ZeroDivisionError: division by zero

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -400,6 +400,17 @@ All of these errors result in an error report (using the same format
 as apport) being generated, before the error-commands (if any) are
 run.
 
+Error reports are represented over the API by the `ErrorReportRef`
+type. This contains enough information to find the error report and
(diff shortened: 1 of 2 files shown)


dbungert:
Do we want a more sophisticated response packet?
Like maybe we could return an object and have traceback be one of the values present and we can add more stuff later.

mwhudson:
There already is more sophisticated stuff in the headers. I added a blurb to DESIGN.md about how that works.

This is just a hack for development really.

dbungert (approved):

PR #956, opened 2021-05-14
Split unit / integration tests a bit

dbungert:
Unit tests should be quick and run frequently.
Integration tests can be longer and run less often.
Split them up.


mwhudson (approved):
Makes sense, :+1:

PR #957, opened 2021-05-20
Don't redundantly remove from the overlay list

dbungert:
There are two places that remove from this list, and the current arm64
crash is running into a scenario where we try to remove from the
overlays list only to find that it's not actually there.
So don't remove if it's already gone.

LP: #1925068

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -400,7 +400,8 @@ class SubiquityClient(TuiApplication):
             self.ui.body.show_stretchy_overlay(overlay)
 
     def remove_global_overlay(self, overlay):
-        self.global_overlays.remove(overlay)
+        if overlay in self.global_overlays:
+            self.global_overlays.remove(overlay)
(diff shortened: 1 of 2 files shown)


mwhudson (approved):
In part I don't like this -- it's a bug for an overlay not in global_overlays to be removed! -- but that shouldn't stand in the way of fixing an actual problem affecting actual users. So feel free to merge!

Just spitballing now, but probably we should handle global overlays in a more, er, global way by having subiquitycore.ui.frame.SubiquityCoreUI (or the subiquity-specific subclass) have a list of overlays and render them over the currently selected body (done sufficiently generically, subiquitycore.view.BaseView could use the same code and then the horrific logic in that class' remove_overlay method can go away).

dbungert:
I agree completely on this being the wrong answer, which is why I didn't move forward on a PR like this a week or two ago.  I do think it helps make a crash less bad though.

PR #959, opened 2021-05-25
Open the permissions on the socket

dbungert:
Installer clients can be non-root.
Installer clients run in an environment where root is just a
password-less sudo away.

### DIFF
--- a/subiquity/common/api/server.py
+++ b/subiquity/common/api/server.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import inspect
+import os
 import traceback
 
 from aiohttp import web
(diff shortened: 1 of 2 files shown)


dbungert:
@mwhudson Would we want to do something similar for the socket started by subiquity/client/controllers/network.py?

mwhudson (approved):
TIL there's no way (beyond umask) to set the permissions of a unix socket when it is created.

It probably makes sense to do this for the "reverse" connections too (or do the other thing and switch that stuff over to websockets).

dbungert:
>> Would we want to do something similar for the socket started by subiquity/client/controllers/network.py
> It probably makes sense to do this for the "reverse" connections too (or do the other thing and switch that stuff over to websockets).

Does this refer to the client thing above or is this something eles?

mwhudson:
Yes, that's what I meant sorry.

PR #960, opened 2021-05-25
Rearrange makefile targets

dbungert:
Skip a recursive make step and simplify a bit.
Add dryrun-server.
Also an unrelated whitespace fix.


mwhudson (approved):
I don't really use the Makefile much tbh, so fine :)

PR #964, opened 2021-06-02
handle some networking errors better

mwhudson:
Before this, having a bug in a method that the network controller calls in the client would crash the server and fail to record the traceback anywhere!

### DIFF
--- a/subiquity/server/controllers/network.py
+++ b/subiquity/server/controllers/network.py
@@ -238,7 +238,10 @@ class NetworkController(BaseNetworkController, SubiquityController):
             if conn.closed:
                 log.debug('closed')
                 return
-            await getattr(client, meth_name).POST(*args)
+            try:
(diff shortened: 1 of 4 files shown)


subiquity/server/controllers/network.py
```diff
@@ -239,7 +238,10 @@ async def _call_client(self, client, conn, lock, meth_name, *args):
             if conn.closed:
                 log.debug('closed')
                 return
-            await getattr(client, meth_name).POST(*args)
+            try:
+                await getattr(client, meth_name).POST(*args)
+            except aiohttp.ClientError:
```
dbungert, line 243: Is it interesting to grab the text of the clienterror?  Otherwise LGTM.

mwhudson: yeah probably actually

dbungert (approved):

PR #966, opened 2021-06-02
move storage description functionality to singledispatch methods

mwhudson:
This has the nice side-effect that subiquity.models.filesystem now contains very few translatable strings, which seems appropriate somehow.

### DIFF
--- a/subiquity/common/filesystem/labels.py
+++ b/subiquity/common/filesystem/labels.py
@@ -124,12 +124,12 @@ def _desc_lv(lv):
 
 
 @functools.singledispatch
-def label(device):
+def label(device, *, short=False):
     raise NotImplementedError(repr(device))
 
 
 @label.register(Disk)
-def _label_disk(disk):
+def _label_disk(disk, *, short=False):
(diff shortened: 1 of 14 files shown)


subiquity/common/filesystem/labels.py
```diff
@@ -0,0 +1,254 @@
...
+            r.append(_("backup ESP"))
+        else:
+            r.append(_("unused ESP"))
+    elif partition.flag == "bios_grub":
+        if partition.preserve:
+            if partition.device.grub_device:
+                r.append(_("configured"))
+            else:
+                r.append(_("unconfigured"))
+        r.append(_("bios_grub"))
```
dbungert, line 81: at risk of sounding excessively picky, I suggest capitalizing BIOS and removing the underscore

mwhudson: I did this and added another word too!!

dbungert (approved):
The groovy builder was timing out, so I changed over to impish and things seem happier.
LGTM

PR #967, opened 2021-06-02
write any wifi config to a file only readable by root

mwhudson:
Perhaps a bit trivial for a PR to itself but oh well.

### DIFF
--- a/subiquity/models/network.py
+++ b/subiquity/models/network.py
@@ -33,11 +33,13 @@ class NetworkModel(NetworkModel):
             return super().render_config()
 
     def render(self):
-        return {
+        netplan = self.render_config()
+        wifis = netplan['network'].pop('wifis', None)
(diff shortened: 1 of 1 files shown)


dbungert (changes requested):
What about wired 802.1x?
Should we just 0600 the whole thing?

mwhudson:
> What about wired 802.1x?

Hm on the one hand subiquity doesn't support this, on the other, yes, maybe we should be more generic here.

> Should we just 0600 the whole thing?

Or do that. I feel like it would be nice not to do this but I don't know if it's worth the effort.

dbungert (approved):
Given that wired 802.1x is not supported yet, downgrading my change request to an eventual suggestion.

mwhudson:
Thanks. I'll add a comment about that.

PR #968, opened 2021-06-02
move kernel package selection to its own controller/model

mwhudson:
This is partly prep for the oem kernel support that we'll get to eventually but it also allows the autoinstall file to override the choice of kernel, something I'm sure at least one person asked for once.

### DIFF
--- a/subiquity/server/controllers/kernel.py
+++ b/subiquity/server/controllers/kernel.py
@@ -1,0 +1,82 @@
+# Copyright 2021 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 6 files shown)


subiquity/server/controllers/kernel.py
```diff
@@ -0,0 +1,82 @@
...
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+
+from subiquitycore.lsb_release import lsb_release
+
+from subiquity.server.controller import NonInteractiveController
```
dbungert, line 20: Well I needed to know about that sooner.
I've got a few changes to make.

dbungert (approved):

PR #970, opened 2021-06-02
Cleanup example data

dbungert:
These examples were crashing dryrun if supplied as a machineconfig due
to the missing blockdev.


mwhudson:
Wow these must have been old. Thanks!

PR #971, opened 2021-06-02
Add groovy back if it is working

dbungert:


mwhudson:
So fwiw this boils down to these commands failing:

```
lxc launch ubuntu-daily:groovy wtf-snapd  -c security.privileged=true
(wait for networking to come up in container)
lxc exec wtf-snapd -- bash -c "apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade"
```

PR #974, opened 2021-06-07
test-in-lxd: remove privileged container usage

dbungert:


mwhudson (approved):
Looks good. Can we reenable groovy now?

PR #975, opened 2021-06-08
make_cloudconfig - start moving cloud config generation to models

dbungert:
Add make_cloudconfig as a function that models should offer that allows
for generation of the cloud-config snippet.
Move snaplist and locale to this mechanism.
Add a test for snaplist output.


subiquitycore/model.py
```diff
@@ -0,0 +1,24 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from abc import ABC
+import logging
+
+log = logging.getLogger("subiquitycore.model")
+
+
+class CloudConfigModel(ABC):
+    def make_cloudconfig(self):
```
mwhudson, line 23: So I'm not sure this adds anything. There are a few thoughts it makes bounce around in my head:

 1. if you have a base class like this, then the check in subiquity.models.subiquity should be "isinstance" not "getattr"
 2. there's not much point inheriting from ABC if you don't have an abstractmethod 
 3. as you've probably noticed while doing this, not all postinstall models contribute to cloud-init config and some are not custom types at all.
 4. this should be in subiquity not subiquitycore

all that said, I think I would support just dropping this baseclass and continuing to to the getattr based test in subiquity.models.subiquity, but I'm also not opposed to having this baseclass so long as it's in the subiquity package, has an abstractmethod and you use isinstance in subiquity.models.subiquity

subiquity/models/identity.py
```diff
@@ -13,10 +13,10 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import logging
-
 import attr
+import logging
```
mwhudson, line 17: so the import pattern I use in subiquity (which I don't think I've written down anywhere, but iirc I got this habit from launchpad) is:

```
<stdlib modules, in alphabetical order>
<blank line>
<third party modules, in alphabetical order>
<blank line>
<other packages from the same project, in alphabetical order>
<blank line>
<imports from the same package>
```

'attr' is a third party module, so that's why it was where it was.

subiquity/models/subiquity.py
```diff
@@ -215,17 +211,12 @@ def _cloud_init_config(self):
...
-                cmd.append(selection.name)
-                cmds.append(' '.join(cmd))
-            config['snap'] = {
-                'commands': cmds,
-                }
+        for model_name in POSTINSTALL_MODEL_NAMES:
+            model = getattr(self, model_name)
+            if getattr(model, 'make_cloudconfig', None):
+                fragment = model.make_cloudconfig()
+                if fragment:
```
mwhudson, line 218: I don't think you need to check the return value here

subiquity/models/subiquity.py
```diff
@@ -215,17 +211,12 @@ def _cloud_init_config(self):
                 config['ssh_authorized_keys'] = self.ssh.authorized_keys
         if self.ssh.install_server:
```
mwhudson, line 212: it's unfortunate that I don't think the identity / ssh stuff can be handled completely orthogonally here. Oh well.

mwhudson (changes requested):

subiquitycore/model.py
dbungert: An earlier version of this PR removed "packages" and "userdata" from POSTINSTALL_MODEL_NAMES, and thus could skip the getattr(model, 'make_cloudconfig') check.  Does POSTINSTALL_MODEL_NAMES need those two entries?

mwhudson: in principle yes, because we don't want to move on to the postinstall stages until they have been configured. We could divide up the postinstall steps to be more granular -- so only the coroutine that's installing the packages blocks on packages being configured etc but ehhhh. Basically "postinstall" is wider than just cloud-init config generation whereas "install" really is just making the curtin config, I guess.

In practice I'm sure everything works if they are not in POSTINSTALL_MODEL_NAMES because both are always noninteractive and so get configured very early.

DESIGN.md
```diff
@@ -191,7 +191,9 @@ classes live in `subiquity.models`.  An instance of each model class is
 attached as an attribute to the `SubiquityModel` class and the name of the
 attribute added to `INSTALL_MODEL_NAMES` or `POSTINSTALL_MODEL_NAMES` as
 appropriate. Models that go into `INSTALL_MODEL_NAMES` need to define a
-render() method that returns a fragment of curtin config.
+render() method that returns a fragment of curtin config.  POSTINSTALL models
+should define a make_cloudconfig() method that returns a cloud config
+fragment.
```
mwhudson, line 196: I think this could be slightly more accurately phrased as "POSTINSTALL models that contribute to cloud-init config should define ..."

Thanks for remembering to update this file btw :)

mwhudson (approved):
Just a small wording change in DESIGN.md but this looks good now, thanks!

PR #977, opened 2021-06-08
Show source list from test containers

dbungert:
Builds on github are randomly failing with an error that sounds like a
bad sources.list.  Start showing the contents of that file in the build
logs.

### DIFF
--- a/scripts/test-in-lxd.sh
+++ b/scripts/test-in-lxd.sh
@@ -39,6 +39,9 @@ then
     done
 fi
 
+# debugging a github issue where the source.list is possibly malformed
+lxc file pull tester/etc/apt/sources.list -
+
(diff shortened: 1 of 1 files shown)


mwhudson:
Oh! I know this one. The problem is that cloud-init rewrites the sources.list file at some point and if we run apt update while that is happening, it can fail. We should jam a cloud-init status --wait in somewhere.

mwhudson:
https://github.com/canonical/subiquity/pull/978 -- there's a chance this change won't actually show anything useful because of the racy nature of all this.

dbungert:
Given that you have a theory on what's happening, I'm closing this one for now and we can retry later if the wait doesn't help.

PR #979, opened 2021-06-09
install wpasupplicant when a wifi nic is seen

mwhudson:
This should (together with seed changes) should be enough to support wifi in subiquity. Haven't tested it yet though!

### DIFF
--- a/subiquity/models/network.py
+++ b/subiquity/models/network.py
@@ -23,7 +23,7 @@ log = logging.getLogger('subiquity.models.network')
 class NetworkModel(NetworkModel):
 
     def __init__(self):
-        super().__init__("subiquity", support_wlan=False)
+        super().__init__("subiquity")
         self.override_config = None
 
     def render_config(self):
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -267,7 +267,7 @@ class InstallController(SubiquityController):
         await self.configure_cloud_init(context=context)
         packages = []
         if self.model.ssh.install_server:
-            packages = ['openssh-server']
+            packages.append('openssh-server')
         if self.model.network.needs_wpasupplicant:
             packages.append('wpasupplicant')
         packages.extend(self.app.base_model.packages)
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -268,6 +268,8 @@ class InstallController(SubiquityController):
         packages = []
         if self.model.ssh.install_server:
             packages = ['openssh-server']
+        if self.model.network.needs_wpasupplicant:
+            packages.append('wpasupplicant')
         packages.extend(self.app.base_model.packages)
         for package in packages:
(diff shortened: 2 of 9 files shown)


mwhudson:
So with https://github.com/canonical/subiquity/pull/980 and https://github.com/canonical/probert/pull/106 (and https://code.launchpad.net/~mwhudson/casper/+git/casper/+merge/403846 and https://code.launchpad.net/~mwhudson/ubuntu-seeds/+git/ubuntu/+merge/403848) this works, at least as far as configuring networking in the installer environment! (I don't have a machine to hand I can actually install to to test that end of things, although that is a fair bit simpler)

xnox:
```
Testing WiFi in VM
* Boot to Try Ubuntu
* modprobe virt_wifi module
* Use `ip link add link en0 name wlan0 type virt_wifi` to wrap wifi in ethernet
```

subiquity/models/network.py
```diff
@@ -23,7 +23,7 @@
 class NetworkModel(NetworkModel):
```
dbungert, line 23: Wait, that works?  That looks super strange.
I know this isn't part of the PR but what's your thoughts on a minor rename?

subiquity/server/controllers/install.py
```diff
@@ -268,6 +268,8 @@ async def postinstall(self, *, context):
         packages = []
         if self.model.ssh.install_server:
             packages = ['openssh-server']
```
dbungert, line 270: Again, not part of the PR, but maybe we should packages.append() instead of assigning directly.

subiquitycore/ui/views/network.py
```diff
@@ -409,9 +431,31 @@ def _create_bond(self, sender=None):
         stretchy.attach_context(self.controller.context.child("add_bond"))
         self.show_stretchy_overlay(stretchy)
 
+    def update_for_wlan_support_install_state(self, state):
+        if state == "NOT_NEEDED":
+            return
+        if self.error_showing:
+            b = 2
```
dbungert, line 438: What about instead of these b/i variables, we define some start and end points?  I think that would scan better.  What do you think?  Also I had no idea I could define a slice ahead of time, really cool.

dbungert (changes requested):
I do have some suggestions but they are mostly on code not even in this PR, so do what you will with that.

subiquity/models/network.py
mwhudson: Yeah it works fine (welcome to Python!). It might be a bit strange but can we leave it for now? :)

subiquity/server/controllers/install.py
mwhudson: yeah i thought that when writing it. will change.

subiquitycore/ui/views/network.py
mwhudson: yeah that would be clearer, I'll change this too.

dbungert (approved):
Thanks!

PR #982, opened 2021-06-11
add python3-yaml to build-packages (make-kbd-info.py now indirectly r…

mwhudson:
(make-kbd-info.py now indirectly requires this)


mwhudson:
tested that this fixes the build in launchpad so just going to merge this asap

PR #983, opened 2021-06-11
Add EventCallback, and move mirror task things to GeoIP

dbungert:
Move mock_app to common location.
Move run_coro to subiquitycore so that subiquitycore doesn't have to
reference things in subiquity, even for test.
Move task tracking things from mirror to geoip.
Server app owns the geoip instance.
Create EventCallback as an alternative to MessageHub that should
hopefully express clearer intermodule dependencies.


subiquity/common/geoip.py
```diff
@@ -14,19 +14,53 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
mwhudson, line 14: I failed to notice this last time, but the idea behind subiquity.common is for things imported by both client and server and this module is only used by the server. Can you move it and its tests over please?

subiquity/server/controllers/mirror.py
```diff
@@ -70,32 +57,20 @@ def load_autoinstall_data(self, data):
 
     @with_context()
     async def apply_autoinstall_config(self, context):
+        async def cc_is_set():
```
mwhudson, line 60: This local function is pointless right? You can just put "asyncio.wait_for(self.cc_event.wait())" below?

subiquitycore/pubsub.py
```diff
@@ -33,3 +33,21 @@ async def abroadcast(self, channel):
 
     def broadcast(self, channel):
         return asyncio.get_event_loop().create_task(self.abroadcast(channel))
+
+
+class EventCallback:
```
mwhudson, line 38: I'm still not completely happy about this as an abstraction but it's an improvement and will do for now :-)

It's not a problem for your usage here, but if one controller wants to subscribe to another controller's EventCallback it's not obvious how to do it without risking races today. Clearly you can create the EventCallback objects in the controller's __init__ methods but when do you subscribe? You can't do that in __init__ because other controller might not even exist yet, but if you do it in start() the other controllers start() method might already have been called and you might miss events. I guess there could be yet another method in the controller interface like "do_subscriptions" or something but that just makes things more complicated. Anyway, you've heard me go on like this enough already :)

mwhudson (approved):
Looks good. One file move request, one small fix and one whinge (most of which you have heard before).

subiquitycore/pubsub.py
dbungert: Sure thing.  Let's discuss next steps.  I put this on agenda for the week.

PR #986, opened 2021-06-16
TimeZone: autoinstall and API (FR-1184)

dbungert:
Add support for Get/Set timezone methods.  Get means that we inquire
with GeoIP as to which timezone is suggested.  Non-availability of
GeoIP, or a previous explicit Set, means that we return the system
timezone.  Set of timezone by Post results in set of the live system
timzeone, and queuing a set of the target system by way of cloud-init.


dbungert:
Note that this go around there are some additional calls to timedatectl that might be interesting.

dbungert:
Marking draft until we get the stable build out

subiquity/models/timezone.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class TimeZoneModel(object):
+    """ Model representing timezone"""
+
+    timezone = ''
+
+    def __init__(self):
+        self._request = None
+        self.timezone = ''
```
mwhudson, line 28: I don't understand the distinction between _request and timezone. Can you add a comment about this?

subiquity/server/controllers/timezone.py
```diff
@@ -0,0 +1,120 @@
...
+        if data is None:
+            return
+        if data not in self.possible:
+            raise ValueError(f'Unrecognized time zone request "{data}"')
+        self.model.set(data)
+        if self.model.detect_with_geoip and self.app.geoip.timezone:
+            self.model.timezone = self.app.geoip.timezone
+            self.model.got_from_geoip = True
+        else:
+            self.model.got_from_geoip = False
```
mwhudson, line 100: Why is this logic not part of the model.set() method? (Not meant to be a leading question!)

mwhudson (approved):
Seems fine, sorry for being slow. Just a couple of questions.

subiquity/server/controllers/timezone.py
dbungert: data validation feels model-y to me, that could be there.
geoip interaction felt controller-y to me, so I had it here.
Actually setting system timezone also felt like a controller operation to me.
It feels borderline to me.  I'm happy to modify if preferred.

subiquity/models/timezone.py
dbungert: Splitting it this way allows reporting `timezone: geoip` in /var/log/installer/autoinstall-user-data.  Users who want a specific timezone can set that instead, and it'll be persisted with their selected timezone.
Adding code comment to match - I can easily see myself forgetting the above in a few months.

mwhudson (approved):
Ah thanks for that comment update,I think this is good to go now. And sorry for the slow reviews.

PR #987, opened 2021-06-16
Workaround phase mismatch of systemd packages

dbungert:
At the moment, hirsute systemd packages have a mix of phased values,
likely related to LP: #1925745, which is causing the hirsute test suite
to fail.  Fix by Michael Hudson-Doyle.


dbungert:
Redundant with #988, which is a better version.

PR #991, opened 2021-06-24
install curtin at commit specified in snapcraft.yaml for CI

mwhudson:
also probert


mwhudson:
I'm going to add an imsm test soon which depends on a change that I landed to curtin a few minutes ago. This branch will keep CI working but I'm not sure what to do for developer systems...

dbungert (commented):
Fine for CI, I'm always a fan of not repeating ourselves.
As for dev systems, we've got the `installdeps.sh` script and `make install_deps`, what if we somehow have `make installdeps` trigger this logic and have `installdeps.sh` leverage that?

mwhudson:
Yeah that makes sense. We should move to a world where probert and curtin are handled the same way, I guess.

mwhudson:
There is more that could be done to make ./scripts/installdeps.sh and the Makefile less redundant but this is at least a step in the right direction.

scripts/checkout-part.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/python3
```
dbungert, line 1: A small part of me feels bad that we're reimplementing submodules, but then I remember that I don't particularly like submodules.

I will console myself by merely modifying snapcraft.yaml instead of remembering submodules commands.

Makefile
```diff
@@ -59,12 +59,17 @@ integration:
 
 check: unit integration
 
+curtin:
```
dbungert, line 62: Thanks for helping bring this stuff to the Makefile, I like where this is going.
Under this plan, curtin and probert are going to have different results if they need an update and have already been checked out.
I suggest making these calls idempotent.
Because probert does the directory check, it can get stuck on an old version, where curtin will refuse to clone when there is one already.

dbungert (changes requested):

dbungert:
also we'll want a gitignore tweak for the new /curtin directory

scripts/checkout-part.py
mwhudson: I've certainly never entirely 'got' how submodules are expected to work. And given that we need to specifiy the commits in snapcraft.yaml for snapcraft, maybe this is OK...

mwhudson: Although hmm it does seem that snapcraft natively supports submodules. So we could move to that instead if we wanted to?

Makefile
mwhudson: I updated my script to be more idempotent, see what you think. It won't quite do the right thing when we are tracking a branch and the branch moves but we've been using full hashes for a while now and I don't want to perform network access to check if the branch has moved every time we do anything so this is ok for now?

dbungert: Thanks!

scripts/checkout-part.py
dbungert: I will be fine with submodules when we choose to move to them, really I'm just disappointed with how non-transparent the git submodule implementation is.

dbungert (approved):

PR #993, opened 2021-06-27
allow a mount action to not have a device

mwhudson:
for https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1933711

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -356,11 +356,11 @@ class attributes:
     # Just a namespace to hang our wrappers around attr.ib() off.
 
     @staticmethod
-    def ref(*, backlink=None):
+    def ref(*, backlink=None, default=attr.NOTHING):
         metadata = {'ref': True}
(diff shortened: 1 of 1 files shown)


subiquity/models/filesystem.py
```diff
@@ -846,8 +846,8 @@ def _available(self):
 
 @fsobj("mount")
 class Mount:
-    device = attributes.ref(backlink="_mount")  # Filesystem
     path = attr.ib()
+    device = attributes.ref(backlink="_mount", default=None)  # Filesystem
```
dbungert, line 850: Is the distinction between None and NOTHING intentional?
If not we can omit this change.

mwhudson: yes, default=attr.NOTHING means there is no default, so the attribute must be supplied when the instance is constructed, default=None means the attribute defaults to None.

dbungert (approved):

PR #994, opened 2021-07-01
Unhardcode path to subiquity-loadkeys

dbungert:
With multiple users of subiquity in different ways, this may not be
always available at this specific location.


mwhudson:
This looks fine, I think perhaps we can go further and remove the subiquity-loadkeys "app" from snapcraft, and possibly do something similar for subiquity-configure-apt (remove the absolute path from the generated curtin config, remove the app from snapcraft.yaml)? Do you want to give that a shot or shall we merge this and go further in a subsequent branch?

dbungert:
I think I want to merge this for now so I can get the ubuntu-desktop-team farther, and we can pursue the more general solution in the next branch.

PR #995, opened 2021-07-01
Default selected_language to en_US if LANG is not helpful

dbungert:


mwhudson:
Hm there is code in livecd-rootfs to set LANG to C.UTF-8 by default, but only for a few projects:

```

        case $PROJECT:${SUBPROJECT:-} in
		ubuntu-server:*|ubuntu-cpc:*|ubuntu:desktop-preinstalled)
			# Set locale to C.UTF-8 by default. We should
			# probably do this for all images early in the
			# 18.10 cycle but for now just do it for
			# server and cpc products.
			echo "LANG=C.UTF-8" > chroot/etc/default/locale
			;;
	esac
```

Would there be any downsides to doing that for more, like at least ubuntu:canary?

Anyway, this patch is fine I guess :)

PR #996, opened 2021-07-01
remove some apps from the snapcraft.yaml

mwhudson:


mwhudson:
Definitely want to do a full snap build and a test in a vm before merging this one.

mwhudson:
I did the test and they passed so that'll do.

PR #998, opened 2021-07-06
Abstract resource file lookup relative to SNAP

dbungert:
Using SNAP to find the resource files is a bit of an odd requirement for
non-subiquity clients.  Start abstracting that away so it's easier to
change later.

Also move loadkeys to lookup from this, since the previous solution is
still a problem for ubuntu-desktop-installer.

### DIFF
--- a/subiquity/server/controllers/keyboard.py
+++ b/subiquity/server/controllers/keyboard.py
@@ -23,6 +23,7 @@ from subiquitycore.context import with_context
 from subiquitycore.utils import arun_command
 
 from subiquity.common.apidef import API
+from subiquity.common.resources import resource_path
 from subiquity.common.serialize import Serializer
 from subiquity.common.types import (
     AnyStep,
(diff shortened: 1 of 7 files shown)


subiquity/server/controllers/keyboard.py
```diff
@@ -199,7 +200,7 @@ async def set_keyboard(self):
             fp.write(self.model.render_config_file())
         cmds = [
             ['setupcon', '--save', '--force', '--keyboard-only'],
-            ['subiquity-loadkeys'],
+            [resource_path('usr/bin/subiquity-loadkeys')],
```
mwhudson, line 203: Can you update https://github.com/canonical/subiquity/blob/main/subiquity/models/subiquity.py#L283 as well please?

mwhudson (approved):
Hi, this is clearly a nice refactoring, thanks for doing it. One more thing to update, I think.

Would be interesting to see if we can drop https://github.com/canonical/subiquity/blob/main/subiquity/cmd/common.py#L33 after a bit more of this.

subiquity/server/controllers/keyboard.py
dbungert: good idea

mwhudson:
Have you tried a snap build with this branch? It looks good but I'd like to run through an install before merging just to be sure (I can do this).

dbungert:
I have indeed - loadkeys is easy to trigger by changing keyboard layout

PR #999, opened 2021-07-07
Initialize locale to C

dbungert:
As ubuntu-desktop-installer is able to show currently, there are still
code flows where if you call 'GET /keyboard', it fails due to the None
selected language in locale.
It would be a weird codependency for 'GET /keyboard' to be required to
be after 'POST /locale'.

### DIFF
--- a/subiquity/models/locale.py
+++ b/subiquity/models/locale.py
@@ -24,7 +24,7 @@ class LocaleModel(object):
     XXX Only represents *language* selection for now.
     """
 
-    selected_language = None
+    selected_language = 'C'
 
     def switch_language(self, code):
         self.selected_language = code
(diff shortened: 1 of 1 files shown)


mwhudson:
Is this the thing where if cloud-init is not present, we don't call start on the controllers? We should probably fix this more generically by not actually starting the api server until we've called the start methods, it's not a valid situation imo.

dbungert:
Right, yes, I stopped thinking about this once I realized that having cloud-init missing broke everything.
I think the next step there is that either
1) missing cloud-init is a fatal error and we just shut subiquity server down
2) we catch that exception at the right spot, complain loudly in the log files, but allow the other `start()` methods to do their thing

I lean toward option 2

mwhudson:
yeah 2 makes sense to me. going to close this now.

PR #1001, opened 2021-07-13
Lookup resources from SUBIQUITY_ROOT

dbungert:


bin/subiquity-server
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+# the backend part of subiquity
+export PYTHON=$SNAP/usr/bin/python3
+export SUBIQUITY_ROOT=$SNAP
+$PYTHON -m subiquity.cmd.server
```
mwhudson, line 5: Rather than adding this wrapper, I think I would prefer if this was done either in in the setup_environment method partly because shell wrappers are gross but also because there's a particularly well hidden entry point here: https://git.launchpad.net/livecd-rootfs/tree/live-build/ubuntu-server/includes.chroot.ubuntu-server.installer/usr/bin/subiquity-shell that this change doesn't cover (this is what runs when you ssh to installer@).

You might also be able to set these variables in `apps.<app-name>.environment` in snapcraft.yaml, I don't know if you can say:

```yaml
apps:
  subiquity:
    ...
    environment:
      SUBIQUITY_ROOT: $SNAP
```

in there or if that will just set the value of the variable to a literal '$SNAP'.

mwhudson (changes requested):
100% the right idea but I think we can do it a bit more cleanly. Given that the desktop team want this asap I might try and do it myself if you're EOD now.

bin/subiquity-server
dbungert: I'm working a bit later, here's my suggestion:
* you reimplement this as you have in mind
* I'll fix the problems I'm looking at (make check fails on my machine, I bisected it to "Merge pull request #997 from mwhudson/imsm")
* I've got a fix pending where I can do a better schema check which should be more tolerant of timezone churn
* once this is reimplemented and merged, I want to point to it for https://github.com/canonical/ubuntu-desktop-installer/pull/167

dbungert:
Nice

PR #1002, opened 2021-07-13
Add new schema comparison, robust to timezone changes

dbungert:
Improve the schema comparison - mostly I'm worried about the
non-timezone items, spot check the timezone list for a few that should
be there.

### DIFF
--- a/scripts/schema-cmp.py
+++ b/scripts/schema-cmp.py
@@ -9,8 +9,7 @@ import sys
 def load(filename):
     with open(filename, 'r') as f:
         data = json.load(f)
-    tz = data['properties']['timezone']['enum']
-    del data['properties']['timezone']['enum']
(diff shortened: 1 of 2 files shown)


scripts/schema-cmp.py
```diff
@@ -0,0 +1,35 @@
...
+# schema comparison should match, except timezones which we just spot
+# check for a few expected values and expect that reality is a superset
+import json
+import sys
+
+def load(filename):
+    with open(filename, 'r') as f:
+        data = json.load(f)
+    tz = data['properties']['timezone']['enum']
+    del data['properties']['timezone']['enum']
```
mwhudson, line 13: you can write this as tz = data['properties']['timezone'].pop('enum') if you like

mwhudson (approved):

PR #1003, opened 2021-07-14
Add missing configure on timezone POST

dbungert:


dbungert:
Verified by rebuild and test in VM that it doesn't cause a regression for at least a basic subiquity install

PR #1004, opened 2021-07-14
Verify that machine-config arg is a file

dbungert:
If machine-config is pointed to a non-existant file, dryrun can kind of
unhelpfully hang without much indication that anything is happening.


mwhudson:
Sure, I'll take it on trust that this helps :)

dbungert:
Basically, I changed branches to a branch that didn't have my machine config, and trying to run dryrun with this non-existent file it will sit at the 'connecting...' prompt forever.
Adding this means that it tells you that the file doesn't exist and shows usage information.

PR #1005, opened 2021-07-15
Workaround dryrun test actually setting timezone

dbungert:
Until a proper fix can be finished, remove the set of timezone from the
integration tests.


mwhudson:
I guess this is ok for a workaround

PR #1006, opened 2021-07-15
fix "make check" on uefi systems

mwhudson:
an ESP can be on a RAID now. Also always run the integration tests in UEFI mode.


dbungert (commented):
Works for me, thanks!

dbungert (approved):

PR #1007, opened 2021-07-15
timezone: dryrun, GET with less side effects

dbungert:
* Remove the workaround for timezone set escaping dryrun
* Replace above workaround with a fix
* Reenable tests for above 
* Change behavior of GET /timezone, to no longer trigger timezone set to system
* Add tests for dryrun escape and GET not setting timezone

LP: #1936310


mwhudson:
Looks good, thanks.

PR #1008, opened 2021-07-19
use livefs-editor to make customized isos

mwhudson:
a while ago I rewrote inject-subiquity-snap in python, generalized it
and put it at https://github.com/mwhudson/livefs-editor.  TBH, it's
always been better than the shell version but now there's a reason to
switch to it: the impish live server ISO use layers, which the current
shell scripts do not support and livefs-editor now does.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ venv
 # top level dependencies fetched from git, ignore top level only.
 /probert/
 /curtin/
+/livefs-editor/
 
 # installer isos
 *.iso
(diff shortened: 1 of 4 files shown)


dbungert:
Yes, deprecating the duplicate logic in the shell scripts is great.
Would you expect this to work against pending?
On both this branch and main, when I try to run `sudo ./scripts/quick-test-this-branch.sh impish-pending.iso pending-test.iso`, it fails like so:

```
+ mount old_iso/casper/installer.squashfs installer
mount: /tmp/tmp.N6upppStqN/installer: special device old_iso/casper/installer.squashfs does not exist.
```

mwhudson:
Ah I forgot that script apparently. Fixed now (I committed an optimization to livefs-edit that is related but not necessary)

dbungert (approved):

PR #1009, opened 2021-07-21
In dryrun, run simulated curtin with same python

dbungert:
Currently, attempting to follow ubuntu-desktop-installer 'flutter run' steps results in the simulated curtin failing with exit code 1.


seb128:
Confirmed, the fix makes the desktop installer go through a working installation in dry mode on hirsute!

PR #1010, opened 2021-07-23
Adjust /reboot signature to return True 

dbungert:
Ubuntu-desktop-installer project requested a return value on /reboot,
just before the reboot occurrs.


dbungert:
This PR removes rebooting_event - I couldn't come up with a reason for why it was needed, and it made the event ordering not what was desired.  Does it do something that I missed?

mwhudson (commented):
The point of rebooting_event is to make the POST block until the reboot is actually happening. The TUI client exits as soon as the http request returns. You can see the difference if you run it in dry-run mode and press restart while unattended upgrades is (fake) running: with this branch it exits immediately but on main it waits until the unattended upgrade has been cancelled, then exits.

There are other ways we could do this (via adding some more application states, maybe) and it only really matters for dry-run mode but I don't think this should be landed as is.

mwhudson (changes requested):
.

dbungert:
I did a full snap build and tested cancellation in a VM - there is a crash, but main does it also.   I think cancellation is worth a revisit but I don't think this branch changes it for the worse.
This branch addresses your feedback about dry run.

subiquity/server/controllers/reboot.py
```diff
@@ -83,9 +84,12 @@ async def copy_logs_to_target(self, context):
...
-            run_command(["/sbin/reboot"])
+
+        def f():
+            if self.opts.dry_run:
+                self.app.exit()
+            else:
+                if platform.machine() == 's390x':
+                    run_command(["chreipl", "/target/boot"])
+                run_command(["/sbin/reboot"])
+        self.app.aio_loop.call_soon(f)
```
mwhudson, line 95: So this trick means that we can be sure that the POST() method returns before /sbin/reboot is invoked but it's not clear to me that we can be sure that the results of the POST() method are actually written to the socket the client is listening to before /sbin/reboot is invoked -- in fact I would be surprised if general aiohttp machinery *didn't* yield control to the event loop before any socket action happens and if it does, the event loop may well run `f` here before aiohttp machinery gets run again. We'd need to call reboot on some kind of signal that the response is done, but I don't think aiohttp provides such a signal.

However! What is this branch for again? :) The general idea is/was that you POST to /reboot and that blocks until reboot is invoked. So it seems to me the only case where what happens after it returns matters at all is dry-run mode, where the client should exit when the call returns. What does the desktop installer think it can do with the return value of this method? Do they just want to display a different screen before the reboot fully kicks in? Should we reboot a desktop install session in a more desktop environment friendly way?

mwhudson (commented):
> I did a full snap build and tested cancellation in a VM - there is a crash, but main does it also.

This seems bad. We should fix that?

> I think cancellation is worth a revisit but I don't think this branch changes it for the worse.
> This branch addresses your feedback about dry run.

I'm sorry to be a pain, but I'm now confused about how this branch can do anything useful I'm afraid. More detail in another comment.

PR #1012, opened 2021-07-29
Initial WSL OOBE implementation in Subiquity

patrick330602:
This is  the WSL OOBE implementation according to WSL OOBE specification (WS004), which includes the following: 

1. WSL Server
2. WSL Client (TUI)
3. Reconfiguration Mode with TUI


dbungert:
Are you trying to merge this to main now, or to a different branch and merge to main later?

system_setup/server/server.py
```diff
@@ -0,0 +1,47 @@
...
+class SystemSetupServer(SubiquityServer):
+
+    from system_setup.server import controllers as controllers_mod
+    controllers = [
+        "Reporting",
+        "Error",
+        "Userdata",
+        "Locale",
+        "Identity",
+        "WSLConfiguration1",
```
jpnurmi, line 30: WSLConfiguration2 went missing

jpnurmi: just out of curiosity, is it necessary to alter the list of available controllers based on the options? it makes testing a bit inconvenient because we have to restart the server in between.

patrick330602: > just out of curiosity, is it necessary to alter the list of available controllers based on the options? it makes testing a bit inconvenient because we have to restart the server in between.

It is necessary, since wslconf1 and wslconf2 have different API implementations.

patrick330602:
> Are you trying to merge this to main now, or to a different branch and merge to main later?

Yes, I am trying to merge this to main. Should I do a single commit Pull Request instead? @dbungert

system_setup/server/server.py
jpnurmi: I mean, is it necessary to hide the wslconf2 endpoint or could it be visible in dry-runs at least?

patrick330602: > I mean, is it necessary to hide the wslconf2 endpoint or could it be visible in dry-runs at least?

It is visible in dry-run actually

jpnurmi: For (automatic) testing of both `wslconf1` and `wslconf2`, the server needs to be restarted in between. Only either endpoint is visible at a time. `wslconf1` is visible when running without `--reconfigure`, whereas `wslconf2` is visible when running with `--reconfigure`.

If both endpoints were visible at the same time _in dry-run mode_, it would speed up and simplify testing on the UI side. Notice that the level of testing is not more than that if "the communication works and an answer is received with the expected data structure". This ensures that any endpoint/protocol changes, later on, would not go unnoticed.

In practice, if `WslConfiguration2` was listed here and not overridden below, it would no longer be necessary to restart the server in tests:

```diff
diff --git a/system_setup/server/server.py b/system_setup/server/server.py
index 6d6486d2..06d21bac 100644
--- a/system_setup/server/server.py
+++ b/system_setup/server/server.py
@@ -28,10 +28,11 @@ class SystemSetupServer(SubiquityServer):
         "Locale",
         "Identity",
         "WSLConfiguration1",
+        "WSLConfiguration2",
     ]
 
     def __init__(self, opts, block_log_dir):
-        if opts.reconfigure:
+        if opts.reconfigure and not opts.dry_run:
             self.controllers = [
                 "Reporting",
                 "Error",
```
According to a quick test, this could work, but I'm not sure if querying both `wslconf1` and `wslconf2` breaks the server's internal state even if they are sending dummy responses?

patrick330602: Interesting. I will try to change the logic for this part.

jibel:
@patrick330602 The test failure is caused by a missing class in subiquity/common/types.py

```
@attr.s(auto_attribs=True)
class TimeZoneInfo:
    timezone: str
    from_geoip: bool

```

It's certainly because you merged main and didn't pick the change instead of doing a rebase. The tests are passing once the wsl branch is rebased on main.

Could you please proceed with the rebase on main (**not a git merge**) and update this branch so we can move on and the review can start?

Thanks.

didrocks:
IMHO commit #2f3dc90b7e565e9592fc27f9748e2ca1ae7c65f5 should be reverted. We intentionally used composition over copying or inheritance to avoid code duplication. This commit defeats the purpose and doesn’t explain what the fix is nor what the error was.

I’m pretty sure we can minimize the amount of code duplication while still fixing the bug.

patrick330602:
> IMHO commit [2f3dc90](https://github.com/canonical/subiquity/commit/2f3dc90b7e565e9592fc27f9748e2ca1ae7c65f5) should be reverted. We intentionally used composition over copying or inheritance to avoid code duplication. This commit defeats the purpose and doesn’t explain what the fix is nor what the error was.
> 
> I’m pretty sure we can minimize the amount of code duplication while still fixing the bug.

So basically the current inherited code is not working at all including: 

1. The form is not validating everything at all; no error is shown if there is an invalid input
2. Done button is not properly greyed out; the Done button should be greyed out if not everything is filled or any one of them failed validation.

patrick330602:
@mwhudson Hi Michael,

This PR is now ready to be reviewed. The Server mode can be run using `sudo make dryrun-system-setup-server` and the client mode can be run using `sudo make dryrun-system-setup`. There is one problem I am unable to solve though: In the TUI (client) mode, the Progress Page will get stuck at the end. Could you help us with that part? Thanks!

system_setup/client/controllers/identity.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def done(self, identity_data):
+        log.debug(
+            "IdentityController.done next_screen user_spec=%s",
+            identity_data)
+        if self.opts.dry_run:
+            username = "dryrun_user"
+        else:
+            username = identity_data.username
+        with open('/var/run/ubuntu_wsl_oobe_assigned_account', 'w') as f:
```
dbungert, line 48: Please protect this operation from dry run.
It should be possible with dry run to do so without running in 'sudo'
One alternate possibility here is that you could write this to a temp file or similar.

dbungert (changes requested):

dbungert:
In the makefile is a variable CHECK_DIRS
If I modify the makefile like so and add system_setup:
-CHECK_DIRS := console_conf/ subiquity/ subiquitycore/
+CHECK_DIRS := console_conf/ subiquity/ subiquitycore/ system_setup/
It exposes a few lints.  Could those lints be addressed?

system_setup/ui/views/overview.py
```diff
@@ -0,0 +1,47 @@
...
+            user_name = f.read()
+        os.remove('/var/run/ubuntu_wsl_oobe_assigned_account')
+        complete_text = _("Hi {username},\n"
+                          "You have complete the setup!\n\n"
+                          "It is suggested to run the following command to update your Ubuntu "
+                          "to the latest version:\n\n\n"
+                          "  $ sudo apt update\n  $ sudo apt upgrade\n\n\n"
+                          "You can use the builtin `ubuntuwsl` command to manage your WSL settings:\n\n\n"
+                          "  $ sudo ubuntuwsl ...\n\n\n"
+                          "* All settings will take effect after first restart of Ubuntu.").format(username=user_name)
```
dbungert, line 34: Is this phrasing correct, on 'will take effect after first restart'?
Since this tool can be run multiple times, maybe we ant to say 'will take effect after the next restart'?

system_setup/client/controllers/identity.py
dbungert: Also, generally speaking, we want the server to be doing things like writing values and the client just sends requests to the server.
This will be important with the plans for using ubuntu-desktop-installer as a frontend - that client does not require root permissions to operate as the desktop installer with baseline subiquity, and if this were a client responsibility we would have to have this code that writes out the magic file both the UI and TUI versions of the client.

dbungert (changes requested):
On the topic of the hang toward the end:

In systemsetup-server-debug.log, you can see this line:
2021-08-11 15:35:37,159 DEBUG system_setup.models.system_server:104 model wslconf1 for install is configured, to go {'identity'}                                                                                                                              

The identity controller is awaiting configuration.
But you're not using it, right?  Either it needs removed from the list, or marked configured.

system_setup/ui/views/overview.py
patrick330602: WSL version will only require one restart, so I think it would be okay to use "first restart". What do you think?

patrick330602:
> In the makefile is a variable CHECK_DIRS
If I modify the makefile like so and add system_setup:
-CHECK_DIRS := console_conf/ subiquity/ subiquitycore/
+CHECK_DIRS := console_conf/ subiquity/ subiquitycore/ system_setup/
It exposes a few lints. Could those lints be addressed?

This is now addressed and fixed.

system_setup/ui/views/overview.py
```diff
@@ -0,0 +1,47 @@
...
+log = logging.getLogger("ubuntu_wsl_oobe.ui.views.overview")
+
+
+class OverviewView(BaseView):
+    title = _("Setup Complete")
+
+    def __init__(self, controller):
+        self.controller = controller
+        user_name = ""
+        with open('/var/run/ubuntu_wsl_oobe_assigned_account', 'r') as f:
```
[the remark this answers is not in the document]
patrick330602: I am not sure there is an API in the UI view for accessing another model's data. Can you give me a pointer?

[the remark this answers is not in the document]
dbungert: subiquity/server/controllers/install.py is an example of a file that goes poking around the data from multiple models.  It's somewhat conceptually similar to the overview, as it's a late controller and does a variety of things to finalize the install.

mwhudson:
Hi, I still need to take a proper look at all this, a million apologies for taking so long. I would still like your stuff to avoid some of the complexity of the subiquity package, which does require some refactoring. I won't block merging this on that though.

I broke some of your things with a recent merge that redid the guts of SubiquityModel. https://github.com/mwhudson/subiquity/tree/wsl fixes that, at least to some extent.

jpnurmi:
I tried to run the server in a newly installed Ubuntu 20.04 WSL. Is it supposed to run timedatectl? Is my WSL installation missing something?

```bash
jpnurmi@DESKTOP-KDL15B3:~/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity$ sudo make dryrun-system-setup-server
python3 -m system_setup.cmd.server --dry-run
starting server
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to create bus connection: Host is down
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/jpnurmi/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity/system_setup/cmd/server.py", line 89, in <module>
    sys.exit(main())
  File "/home/jpnurmi/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity/system_setup/cmd/server.py", line 47, in main
    from system_setup.server.server import SystemSetupServer
  File "/home/jpnurmi/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity/system_setup/server/server.py", line 16, in <module>
    from subiquity.server.server import SubiquityServer
  File "/home/jpnurmi/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/server/server.py", line 185, in <module>
    class SubiquityServer(Application):
  File "/home/jpnurmi/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/server/server.py", line 203, in SubiquityServer
    from subiquity.server import controllers as controllers_mod
  File "/home/jpnurmi/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/server/controllers/__init__.py", line 33, in <module>
    from .timezone import TimeZoneController
  File "/home/jpnurmi/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/server/controllers/timezone.py", line 69, in <module>
    class TimeZoneController(SubiquityController):
  File "/home/jpnurmi/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/server/controllers/timezone.py", line 73, in TimeZoneController
    possible = generate_possible_tzs()
  File "/home/jpnurmi/Projects/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/server/controllers/timezone.py", line 29, in generate_possible_tzs
    list_tz_out = subprocess.check_output(tzcmd, universal_newlines=True)
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['timedatectl', 'list-timezones']' returned non-zero exit status 1.
make: *** [Makefile:57: dryrun-system-setup-server] Error 1
```

jibel:
On 18/08/2021 15:51, J-P Nurmi wrote:
>
> I tried to run the server in a newly installed Ubuntu 20.04 WSL. Is it 
> supposed to run timedatectl? Is my WSL installation missing something?
>
No that's something I asked Patrick to fix. timedatectl cannot run 
because there is no systemd in WSL by default.

mwhudson:
Having had a bit of a look finally, I think this is OK to merge and keep working on it in main. Two requests:

 1. you rebased on main but need to do something like the branch I linked or it doesn't work at all
 2. I think we need some integration tests to make sure I don't keep on breaking it :)

system_setup/server/server.py
patrick330602: @jpnurmi I have fixed this bug, it should work as intented

mwhudson:
This can be closed now, right?

jibel:
Yes, it can be closed.

PR #1016, opened 2021-08-03
Update probert for bitlocker changes

dbungert:


dbungert:
If this looks reasonable I can do another round of full build and vm test before merge.
(I did so earlier today, based on the branches before merge)

mwhudson (approved):

PR #1019, opened 2021-08-04
Make install_deps depend on gitdeps

jpnurmi:
On a clean system running subiquity_client tests would throw

    ModuleNotFoundError: No module named 'curtin'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ endif
 
 all: dryrun
 
-install_deps:
+install_deps: gitdeps
 	sudo apt-get install -y python3-urwid python3-pyudev python3-nose python3-flake8 \
 		python3-yaml python3-coverage python3-dev pkg-config libnl-genl-3-dev \
 		libnl-route-3-dev python3-attr python3-distutils-extra python3-requests \


dbungert:
@jpnurmi
Curtin in the archive will usually not be the version that we want.
`make gitdeps` will grab the exact curtin/probert.
`make install_deps` probably doesn't even need probert at this point.

What if we do this:
* subiquity: `make install_deps` depends on `make gitdeps`
* u-d-i: use pythonpath to point to the curtin & probert subdirectories that get created by `make gitdeps`

Makefile
```diff
@@ -26,7 +26,8 @@ install_deps:
 		python3-yaml python3-coverage python3-dev pkg-config libnl-genl-3-dev \
 		libnl-route-3-dev python3-attr python3-distutils-extra python3-requests \
 		python3-requests-unixsocket python3-jsonschema python3-apport \
-		python3-bson xorriso isolinux python3-aiohttp probert cloud-init ssh-import-id
+		python3-bson xorriso isolinux python3-aiohttp probert cloud-init ssh-import-id \
+		curtin
```
dbungert, line 30: By itself this is fine, but see comment for why I think this isn't what you want.

dbungert (changes requested):

jpnurmi:
Thanks, I hadn't even noticed the `gitdeps` target. The installer's instructions listed `make install_deps` so I was used to running that.

Personally, I would be fine setting up `PYTHONPATH`, but I wonder how we could make it as convenient as possible for external contributors. There are no makefiles in u-b-i, but Flutter devs will open the project in IDE, hit the run button, and wonder why things don't work.

jpnurmi:
Hmm, looks like the CI runs `make install_deps && sudo apt install python3-curtin`, so it could start failing if `curtin` is outdated.

https://github.com/canonical/ubuntu-desktop-installer/blob/main/.github/workflows/flutter-ci.yml#L37

dbungert:
Yes, we will need to ensure that your CI command can get dependencies, and that it's easy for flutter contributors to work with.
The makefile dependency I mention above should get the right version of curtin/probert for these use cases, without having to adjust your install instructions or change your CI config.
As far as PYTHONPATH, I suppose we could set environment variables in app.dart before invoking subiquity.  In the snap this is handled already with the wrapper script.
What do you think?

jpnurmi:
Thanks, I misunderstood the original proposal. This approach works indeed. :)

dbungert (approved):

jpnurmi:
This didn't work as expected. Should it be reverted? Also, would it be better to start using `scripts/installdeps.sh` like in the CI?

jpnurmi:
Ohh, sorry. I haven't actually tried the latest main.

I went straight for `installdeps.sh` this time because of the dependency issue we saw in the CI. Wasn't it so that gitdeps depends on libsomething-dev, which is installed by install_deps?

PR #1020, opened 2021-08-04
have slimy-update-snap.sh update curtin in the snap too

mwhudson:
Also tweak scripts/update-part.py to work better when switching
repositories.

### DIFF
--- a/scripts/slimy-update-snap.sh
+++ b/scripts/slimy-update-snap.sh
@@ -52,9 +52,12 @@ do_mount $old old
 
 add_overlay old new
 
+rm -rf new/lib/python3.6/site-packages/curtin
 rm -rf new/lib/python3.6/site-packages/subiquity
 rm -rf new/lib/python3.6/site-packages/subiquitycore
 
-rsync -a $src/subiquity $src/subiquitycore new/lib/python3.6/site-packages
+(cd "${src}" && ./scripts/update-part.py curtin)
+
+rsync -a $src/subiquity $src/subiquitycore $src/curtin/curtin new/lib/python3.6/site-packages
 
 mksquashfs new $new -comp gzip -Xcompression-level 3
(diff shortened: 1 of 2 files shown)


scripts/slimy-update-snap.sh
```diff
@@ -52,9 +52,12 @@ do_mount $old old
 
 add_overlay old new
 
+rm -rf new/lib/python3.6/site-packages/curtin
```
dbungert, line 55: What's your thoughts on doing probert also?  Does it get weird with the compiled libs?
If it's not too hard that would be nice.

dbungert (approved):
While the changes look reasonable, slimy still fails for me on this branch with a crash on start of the UI, with both the pending and non-pending impish isos.  Any thoughts?

mwhudson:
> While the changes look reasonable, slimy still fails for me on this branch with a crash on start of the UI, with both the pending and non-pending impish isos. Any thoughts?

Are you running on an ISO that only has a stable version of the snap in it? That won't work, you need an ISO that has a newer snap (edge works).

scripts/slimy-update-snap.sh
mwhudson: Hm. Yes compiled libs would be weird, although I guess it ought to be possible to replace the python code, which is all we usually care about. Maybe one for the next branch? :)

dbungert:
> > While the changes look reasonable, slimy still fails for me on this branch with a crash on start of the UI, with both the pending and non-pending impish isos. Any thoughts?
> 
> Are you running on an ISO that only has a stable version of the snap in it? That won't work, you need an ISO that has a newer snap (edge works).

Ah, ok.  Presumably I can do a full snap build, inject, use that as my base iso, slimy, inject, and that sounds like it should work.  Trying it out.

dbungert:
Yes, the above plan is working.  I can now script my way to a slimy-friendly daily base iso.
Ship it!

PR #1021, opened 2021-08-06
redo how install variant determines which models to wait for

mwhudson:
This is probably a bit overkill, but the thrust here is to allow the source variant (i.e. desktop or server) to change more than once. This isn't going to happen for a long time obviously but I wanted to clean this up sooner rather than later. There are ways this could be made still more general of course, but well. I think this is OK.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -345,7 +345,10 @@ The server code proceeds in stages:
     commands.
  2. Then it waits for all the model objects that feed into the curtin
     config to be configured.
- 3. It waits for confirmation.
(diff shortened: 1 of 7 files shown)


dbungert (approved):
LGTM

dbungert:
FYI if we land this before WSL, there will be some nasty merge conflicts waiting for that branch.

mwhudson:
I really need to make time to digest that work :/ I think possibly this can wait until after we make a release for 20.04.3 too, it's a bit risky and brings no immediate benefit.

mwhudson:
Going to merge this now that we did a stable release and the world didn't explode.

PR #1023, opened 2021-08-09
Add /storage/has_bitlocker convenience API

dbungert:
* Adds bitlocker info query API
* Adds first API test to runtests for above


dbungert:
@jpnurmi - FYI

mwhudson (approved):
Hi looks good. I wonder if we can do something a little more ... structured? ... with the API tests but this is good enough for a start :)

PR #1026, opened 2021-08-10
make: fix install_deps order

dbungert:
Getting the apt packages needs to be first, as processing probert is
more than a trivial git clone and needs some of those apt packages.
Drop probert as a requirement since it's redundant with the git
checkout.  Also +build-essential to the list.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,13 +21,16 @@ endif
 
 all: dryrun
 
-install_deps: gitdeps
+aptdeps:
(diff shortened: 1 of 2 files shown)


dbungert:
u-d-i CI was using 'make install_deps', I have changed it over to the installdeps.sh script instead to match subiquity CI, this should fix the Makefile verison.

mwhudson (approved):

PR #1032, opened 2021-08-19
update translations from launchpad

mwhudson:
A while ago I set up subiquity for translations at
https://translations.launchpad.net/subiquity by uploading the pot and po
files that were current at the time and set it to export translations to
lp:~canonical-foundations/subiquity/translations-export (unfortunately,
launchpad translations only support exporting to bzr).  This commit
contains the po files from that branch. The diff is noisy but I don't
think there are many actual changes (although there are a few so some
people must have found this).

I haven't publicized this or asked the ubuntu-translators team to work
on the translations because I am still uncertain about what the best way
to manage translations for subiquity is. But this still seems better
than nothing.

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -1,23 +1,23 @@
-# Latvian translation for subiquity
-# Copyright (c) 2019 Rosetta Contributors and Canonical Ltd 2019
+# Arabic translation for subiquity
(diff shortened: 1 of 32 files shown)


dbungert (approved):
Do we need to touch the languagelist file?

PR #1033, opened 2021-08-23
make the snaps produced by slimy-update-snap.sh pass review

mwhudson:
I mean we /probably/ shouldn't be using the snaps this makes for
anything serious but it can be useful for testing emergency fixes...


dbungert (approved):
Yea, uploading slimy-produced snaps is probably a poor choice, but getting the snap closer to what a real snap looks like is a clear improvement.

PR #1036, opened 2021-08-30
Fixed a typo

joeljose350:


mwhudson (approved):
I don't think this is a typo exactly, I think it was deliberate to spell the name of the product for that for a while, but I don't think we do that any more so thanks :)

xnox:
It was meant to be Server [Ubiquity](https://launchpad.net/ubiquity) in it self was a nod that it was a single iso that worked as ephemeral live/test environment and as an installer at the same time. When it was developed, doing that was ground-breaking.

PR #1037, opened 2021-08-31
system_setup: add workaround for configure for test automation

dbungert:
In integration test with an answers file, this step runs way too quickly
for the rest of the software to cope with.  A sleep here seems harmless
for real usage, is only in the system_setup code path, and lets us test
automation happen while we ponder a better fix.

### DIFF
--- a/system_setup/server/controllers/configure.py
+++ b/system_setup/server/controllers/configure.py
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import asyncio
 import logging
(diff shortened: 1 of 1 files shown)


didrocks:
Thanks Daniel for looking and finding a workaround!

We looked again with fresh eyes a little bit deeper and found the real issue, which was the configure (equivalent to your install) controller which was not waiting on the models to be configured. We fixed it in https://github.com/canonical/subiquity/commit/3eb5231fcab83a93efea928fa3bd23824f4718dd. which is now in the new wsl oobe PR.
I think we can thus close that PR and merge the other one in the main branch if everything is ok. Thanks again!

PR #1038, opened 2021-09-01
OOBE and configuration for WSL

jibel:
This implements the client and server of the OOBE for WSL. 
It covers 2 modes:
1. Initial setup
2. Reconfiguration

At the moment only the initial setup is implemented.
The code is commented with "TODO WSL" to indicate the part that remains to implement or complete.

There is also an initial "answer" test of the OOBE in "setup" mode to prevent regressions.

It addresses all the comments in [now superseded branch 1012](https://github.com/canonical/subiquity/pull/1012) and should be good to merge.

Then we'll proceed with separate PRs for each item in the TODO list.


system_setup/server/controllers/shutdown.py
```diff
@@ -0,0 +1,51 @@
...
+        # Do not copy logs to target
+        self.server_reboot_event.set()
+        self.app.aio_loop.create_task(self._run())
+
+    @with_context(description='mode={self.mode.name}')
+    def shutdown(self, context):
+        self.shuttingdown_event.set()
+        if self.opts.dry_run:
+            self.app.exit()
+        else:
```
dbungert, line 43: We should answer the question - what is desired today for non-dry-run until reboot/shutdown is handled?
Subiquity server process exit might be a good start.  Or is it OK if it stays around?

system_setup/server/controllers/wslconfadvanced.py
```diff
@@ -0,0 +1,190 @@
...
+            }
+        }
+    }
+
+    def __init__(self, app):
+        super().__init__(app)
+
+        # load the config file
+        data = {}
+        if path.exists('/etc/wsl.conf'):
```
dbungert, line 99: do we want to substitute a placeholder config file when in dry run?

system_setup/ui/views/wslconfadvanced.py
```diff
@@ -0,0 +1,200 @@
+""" WSLConfigurationAdvanced View
```
dbungert, line 1: I keep wondering about this view from a UX standpoint.
When we have trouble giving a name to a UI element, that sometimes indicates that the UI element has too many unrelated things.  Would it make sense to resplit some of these values into different screens?

system_setup/ui/views/wslconfadvanced.py
```diff
@@ -0,0 +1,200 @@
...
+                                        "allows you to see the MOTD."))
+
+    def validate_custom_path(self):
+        p = self.custom_path.value
+        if p != "" and (re.fullmatch(r"(/[^/ ]*)+/?", p) is None):
+            return _("Mount location must be a absolute UNIX path"
+                     " without space.")
+
+    def validate_custom_mount_opt(self):
+        o = self.custom_mount_opt.value
```
dbungert, line 114: There are a few here with short variable names.  Longer variable names would be nice.

system_setup/ui/views/wslconfbase.py
```diff
@@ -0,0 +1,125 @@
...
+                 r"(fs|def|root)?context=\w+", r"(no)?dev", r"(no)?diratime",
+                 r"dirsync", r"(no)?exec", r"group", r"(no)?iversion",
+                 r"(no)?mand", r"_netdev", r"nofail", r"(no)?relatime",
+                 r"(no)?strictatime", r"(no)?suid", r"owner", r"remount",
+                 r"ro", r"rw", r"_rnetdev", r"sync", r"(no)?user", r"users"]
+        # DrvFs filesystem mount option
+        drvfsmo = r"case=(dir|force|off)|metadata|(u|g)id=\d+|(u|f|d)mask=\d+|"
+        fso = "{0}{1}".format(drvfsmo, '|'.join(fsimo))
+
+        if o != "":
```
dbungert, line 73: Duplicate code with the other view?

dbungert (commented):
I've got a few suggestions here, but don't consider them blocking for merge.

system_setup/ui/views/wslconfbase.py
jibel: Yes, this is identified and in the TODO list of things to fix once this merged.

system_setup/ui/views/wslconfadvanced.py
jibel: We want to have these variable names fixed together with the code duplication removal.

system_setup/ui/views/wslconfadvanced.py
jibel: Agreed. However this is the implementation of the approved design specification.

system_setup/server/controllers/wslconfadvanced.py
jibel: This is part of the request to read and write in .subiquity in dryrun mode and part of the TODO list of fixes after this PR is merged.

system_setup/server/controllers/shutdown.py
jibel: +1, will fix.

jibel:
This update removes the progress page on WSL which is always blank on non-systemd systems and make the summary page the last page of the setup. 
It also addresses Daniel's comment about always exiting the server.

system_setup/server/controllers/shutdown.py
jibel: done in last update

subiquitycore/utils.py
```diff
@@ -143,3 +144,8 @@ def disable_subiquity():
                  "snap.subiquity.subiquity-service.service",
                  "serial-subiquity@*.service"])
     return
+
+
+def is_wsl():
```
mwhudson, line 149: This function seems to be unused.

dbungert (changes requested):

subiquity/server/server.py
```diff
@@ -112,14 +113,15 @@ async def mark_configured_POST(self, endpoint_names: List[str]) -> None:
                 controller.configured()
 
     async def client_variant_POST(self, variant: str) -> None:
```
mwhudson, line 115: So I have a PR that removes this API... we can just make it only supported in the system_setup code I guess. It also feels a bit wrong to me to have the client decide whether this is setup or reconfiguration but well.

subiquity/models/subiquity.py
```diff
@@ -146,6 +147,13 @@ def set_source_variant(self, variant):
                 self._confirmation_task.cancel()
         else:
             self._install_event.set()
+        unconfigured_postinstall_model_names = \
+            self._cur_postinstall_model_names - self._configured_names
+        if unconfigured_postinstall_model_names:
+            if self._postinstall_event.is_set():
+                self._postinstall_event = asyncio.Event()
+        else:
+            self._postinstall_event.set()
```
mwhudson, line 156: Conceptually speaking, this new code should only run if self._confirmation is set. I guess  `SystemSetupModel.__init__` should just set its _confirmation event as soon as it is created.

subiquity/server/controllers/timezone.py
```diff
@@ -19,20 +19,27 @@
 from subiquity.common.apidef import API
 from subiquity.common.types import TimeZoneInfo
 from subiquity.server.controller import SubiquityController
+from shutil import which
+import os
 
 log = logging.getLogger('subiquity.server.controllers.timezone')
 
+TIMEDATECTLCMD = which('timedatectl')
```
mwhudson, line 27: I guess this sort of thing is a good example of why we shouldn't do things at import time...

subiquity/models/subiquity.py
mwhudson: In general... all this code currently falls in a kind of uncanny valley between being generic and specific. It allows for complete freedom in model names -- which is good for WSL -- but it assumes that all these model names fall into an install / confirm / postinstall flow -- which is not remotely appropriate for WSL. It feels like we could extract a base class from SubiquityModel that defers to subclasses the definition of this flow somehow, and then SystemSetupModel could inherit from this base class and not SubiquityModel. But this has flow on implications for the things like the values the ApplicationState enum can take so ehhhhh. Yet another thing to try to clean up after merge.

subiquity/server/server.py
```diff
@@ -263,9 +270,20 @@ def __init__(self, opts, block_log_dir):
...
+            self.snapd = AsyncSnapd(connection)
+        else:
+            log.info("no snapd socket found. Snap support is disabled")
+            reload_needed = False
+            for controller in ["Refresh", "Snaplist"]:
+                if controller in self.controllers.controller_names:
+                    self.controllers.controller_names.remove(controller)
+                    reload_needed = True
+            if reload_needed:
+                self.controllers.load_all()
```
mwhudson, line 285: When is this bit actually needed?

system_setup/client/client.py
```diff
@@ -0,0 +1,85 @@
...
+        # 3. provide an API for this for the flutter UI to know about it
+        # 4. Add Configuration Base page before Advanced
+        # 5. Add language page
+        # self.variant = "wsl_configuration"
+        if opts.reconfigure:
+            self.controllers = [
+                "WSLConfigurationBase",
+                "WSLConfigurationAdvanced",
+                "Summary",
+            ]
```
mwhudson, line 54: In general my approach for conditional screens has been for the client to be prepared to show all of them and the server to tell the client not to show a particular screen by sending an x-override-status: skip header (this is 100% not something you need to fix before merge!)

system_setup/client/controllers/identity.py
```diff
@@ -0,0 +1,44 @@
...
+import logging
+
+from subiquity.client.controllers import IdentityController
+from subiquity.common.types import IdentityData
+from system_setup.ui.views import WSLIdentityView
+
+log = logging.getLogger('system_setup.client.controllers.identity')
+
+
+class WSLIdentityController(IdentityController):
```
mwhudson, line 25: I don't think this gains anything from inheriting from subiquity's identity controller.

system_setup/client/controllers/summary.py
```diff
@@ -0,0 +1,98 @@
...
+            if app_status.error is not None:
+                if self.crash_report_ref is None:
+                    self.crash_report_ref = app_status.error
+                    if self.summary_view:
+                        self.ui.set_body(self.summary_view)
+                    self.app.show_error_report(self.crash_report_ref)
+
+            if self.app_state == ApplicationState.NEEDS_CONFIRMATION:
+                if self.showing:
+                    self.app.show_confirm_install()
```
mwhudson, line 72: Is there any reason you expect to add a confirmation step part way through the wsl config? If not, I'd rather see this stuff removed before merge tbh.

mwhudson (commented):
Hi, so I've gone through this all now. I think it's almost OK to merge, I've made some comments and feel free to use your judgement as to which to deal with now vs later.

In general I really do want to move to a world where the system_setup apps and the subiquity apps are less entangled, for example I think it would be much better if they had separate apidef files. I know this isn't possible now, but I do hope we can have some agreement that this is where we should be moving towards.

This brings me to a question actually: do you plan to have autoinstall support in the system_setup application? How do you see that working, will it be triggered via cloud-init still (I can't see how that would work though?). If this actually works already, can you add an integration test for the automated mode? If not, can you at lease write down some notes about your plans in the area somewhere?

subiquitycore/utils.py
jibel: Removed.

subiquity/models/subiquity.py
```diff
@@ -146,6 +147,13 @@ def set_source_variant(self, variant):
                 self._confirmation_task.cancel()
         else:
             self._install_event.set()
+        unconfigured_postinstall_model_names = \
+            self._cur_postinstall_model_names - self._configured_names
+        if unconfigured_postinstall_model_names:
+            if self._postinstall_event.is_set():
+                self._postinstall_event = asyncio.Event()
```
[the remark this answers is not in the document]
jibel: It's the same code than lines 144,145 for the install event. It should be maybe better to fix both in a separate PR. WDYT?

system_setup/client/client.py
```diff
@@ -0,0 +1,85 @@
...
+            self.restarting = True
+            self.ui.block_input = True
+            self.aio_loop.create_task(self._restart_server())
+            return
+        if self.urwid_loop is not None:
+            self.urwid_loop.stop()
+        cmdline = sys.argv
+        if self.opts.dry_run:
+            cmdline = [
+                sys.executable, '-m', 'system_setup.cmd.tui',
```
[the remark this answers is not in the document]
jibel: Done, there were 2 places where this code was duplicated.

system_setup/cmd/tui.py
```diff
@@ -0,0 +1,152 @@
...
+            print("running server pid {}".format(server_proc.pid))
+        elif opts.server_pid is not None:
+            print("reconnecting to server pid {}".format(opts.server_pid))
+        else:
+            opts = parser.parse_args(args)
+    else:
+        opts = parser.parse_args(args)
+        if opts.socket is None:
+            opts.socket = '/run/subiquity/socket'
+    os.makedirs(os.path.basename(opts.socket), exist_ok=True)
```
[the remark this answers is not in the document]
jibel: TODO Added

system_setup/client/controllers/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from .identity import WSLIdentityController
+from .wslconfbase import WSLConfigurationBaseController
+from .summary import SummaryController
+from .wslconfadvanced import WSLConfigurationAdvancedController
+
+from subiquity.client.controllers import (WelcomeController)
```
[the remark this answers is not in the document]
jibel: Done.

system_setup/server/controllers/shutdown.py
```diff
@@ -0,0 +1,50 @@
...
+import logging
+
+from subiquitycore.context import with_context
+from subiquity.common.types import ShutdownMode
+from subiquity.server.controllers import ShutdownController
+
+log = logging.getLogger("system_setup.controllers.restart")
+
+
+class SetupShutdownController(ShutdownController):
```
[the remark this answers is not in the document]
jibel: All the event management is inherited and in line with what subiquity is doing, we would prefer to keep the inheritance rather than reimplementing this code.

system_setup/server/server.py
```diff
@@ -0,0 +1,59 @@
...
+        "wslconfbase",
+    },
+    wsl_setup={
+        "identity",
+    },
+    wsl_configuration={
+        "wslconfadvanced",
+    })
+
+POSTINSTALL_MODEL_NAMES = ModelNames({})
```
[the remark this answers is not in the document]
jibel: Done.

subiquity/models/subiquity.py
```diff
@@ -76,7 +76,8 @@ def __init__(self, default_names, **per_variant_names):
         self.per_variant_names = per_variant_names
 
     def for_variant(self, variant):
-        return self.default_names | self.per_variant_names.get(variant, set())
+        return set(self.default_names) | self.per_variant_names.get(variant,
```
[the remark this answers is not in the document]
jibel: Done, but note that 4 lines below in the method all() there is exactly the same casting to set()

subiquity/server/server.py
jibel: Agreed, the architecture the server will decide which variant it is executing.
Lets merge this, then adapt the PR to the new API  both subiquity and system_setup afterwards.

subiquity/models/subiquity.py
jibel: Ack. This is a longer term goal.

subiquity/server/controllers/timezone.py
jibel: Done, the entire test about a working timedatectl and systemd enabled has been moved to a function called in the if statement.

system_setup/client/client.py
jibel: There is a TODO to finish the implementation of the  reconfiguration mode. The change you request will be part of this work.

system_setup/client/controllers/identity.py
jibel: Ack. WSLIdentityController is now directly inheriting from SubiquityTuiController.

system_setup/client/controllers/summary.py
jibel: Removed.

subiquity/server/server.py
jibel: Removed. This was a leftover from when we were still loading these 2 controllers on WSL.

jibel:
All the comments have been addressed and it should good to merge.

Regarding autoinstall, it is very on our TODO list but without cloud-init which is not supported in WSL. This is something we will start investigating once this initial PR is merged and current "TODO WSL" list is done.

dbungert (approved):

mwhudson (approved):
Hi, thanks for the changes. OK to merge from my POV. Just one more question! Do you want to rebase this to get a better commit history or should we just squash merge?

jibel:
Rebase done and ready to merge.

system_setup/models/wslconfadvanced.py
```diff
@@ -0,0 +1,135 @@
...
+        d['interop_appendwindowspath'] = result.interop_appendwindowspath
+        d['gui_theme'] = result.gui_theme
+        d['gui_followwintheme'] = result.gui_followwintheme
+        d['legacy_gui'] = result.legacy_gui
+        d['legacy_audio'] = result.legacy_audio
+        d['adv_ip_detect'] = result.adv_ip_detect
+        d['wsl_motd_news'] = result.wsl_motd_news
+        d['automount'] = result.automount
+        d['mountfstab'] = result.mountfstab
+        self._wslconfadvanced = WSLConfigurationAdvancedModel(**d)
```
jpnurmi, line 71: is there a typo here? shouldn't it create a `WSLConfigurationAdvanced` instance?

```
  File "/home/jpnurmi/Projects/canonical/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/common/api/server.py", line 123, in handler
  File "/home/jpnurmi/Projects/canonical/ubuntu-desktop-installer/packages/subiquity_client/subiquity/system_setup/server/controllers/wslconfadvanced.py", line 189, in POST
    self.model.apply_settings(data, self.opts.dry_run)
  File "/home/jpnurmi/Projects/canonical/ubuntu-desktop-installer/packages/subiquity_client/subiquity/system_setup/models/wslconfadvanced.py", line 71, in apply_settings
    self._wslconfadvanced = WSLConfigurationAdvancedModel(**d)
TypeError: __init__() got an unexpected keyword argument 'custom_path'
```

jibel: Are you in first run mode or reconfiguration mode? POST in reconfigure is not fully implemented and should be fixed by the work Patrick is doing.

jpnurmi: I'm running some simple POST/GET tests in first run mode just to verify that the data structure is in the expected format.

PR #1039, opened 2021-09-02
pubsub: Obsolete EventCallback

dbungert:
* Adjust existing usage of EventCallback to use MessageHub instead.
* Add constants for use with MessageHub.
* Adapt EventCallback test for MessageHub instead.

### DIFF
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -25,7 +25,6 @@ from typing import Dict, List, Optional, Union
 import attr
 
 from subiquitycore.models.network import NetDevInfo
-from subiquitycore.pubsub import CoreChannels
 
 
 class ErrorReportState(enum.Enum):
@@ -336,12 +335,6 @@ class TimeZoneInfo:
 class ShutdownMode(enum.Enum):
     REBOOT = enum.auto()
     POWEROFF = enum.auto()
-
-
-class InstallerChannels(CoreChannels):
-    NETWORK_PROXY_SET = 'network-proxy-set'
-    SNAPD_NETWORK_CHANGE = 'snapd-network-change'
-    GEOIP = 'geoip'
 
 
 @attr.s(auto_attribs=True)
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -25,6 +25,7 @@ from typing import Dict, List, Optional, Union
 import attr
 
 from subiquitycore.models.network import NetDevInfo
+from subiquitycore.pubsub import CoreChannels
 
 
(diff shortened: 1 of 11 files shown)


subiquitycore/pubsub.py
```diff
@@ -14,9 +14,17 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import asyncio
+import enum
 import inspect
 
 
+class MessageChannels(enum.IntEnum):
```
mwhudson, line 21: This should be in the subiquity package somewhere I think.

subiquity/server/controllers/mirror.py
```diff
@@ -65,9 +66,9 @@ async def apply_autoinstall_config(self, context):
         except asyncio.TimeoutError:
             pass
 
-    def on_countrycode(self, cc):
+    def on_geoip(self):
```
mwhudson, line 69: This reminds me of the fact that I think I made a wrong call in MessageHub: you can call subscribe with arguments that will be passed to the callback, but I think it would be better to allow arguments to be passed to broadcast, which can then be passed to subscribers. Then the GEOIP message could be sent with the geoip data. Just a thought, certainly doesn't have to be changed in this branch.

mwhudson (approved):
I like this a lot but MessageChannels needs to live in the subquity, not subiquitycore, package. Move that and merge!

mwhudson:
FWIW here's a branch with the follow on changes I was talking about earlier:

 https://github.com/mwhudson/subiquity/tree/broadcast-configured

as you can see from the comments, it's not *quite* as neat as I hoped.

subiquitycore/pubsub.py
dbungert: That will make things a little weird for the pubsub broadcast in subiquitycore/controllers/network.py.
Otherwise I agree.

subiquity/server/controllers/mirror.py
dbungert: Yes, I debated adding that or not.  Follow-on branch.

subiquity/common/types.py
```diff
@@ -337,6 +338,12 @@ class ShutdownMode(enum.Enum):
     POWEROFF = enum.auto()
 
 
+class InstallerChannels(CoreChannels):
```
mwhudson, line 341: Sorry for doing this but can this be in subiquity/server please? common should only be for things that might be imported by the client.

dbungert: Thanks for the reminder

PR #1040, opened 2021-09-06
Fix advancing from snap list screen before snaps have loaded

lambda-propaganda:
when done() is called before the snap has loaded this causes an attribute error on the following line

'log.debug("snaps to install %s", self.selections_by_name)'

### DIFF
--- a/subiquity/ui/views/snaplist.py
+++ b/subiquity/ui/views/snaplist.py
@@ -116,6 +116,7 @@ class SnapInfoView(WidgetWrap):
         self.parent = parent
         self.snap = snap
         self.needs_focus = True
+        self.selections_by_name = {}
 
         self.description = Text(snap.description.replace('\r', '').strip())
         self.lb_description = ListBox([self.description])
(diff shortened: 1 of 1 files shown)


lambda-propaganda:
Sorry was in quite a rush when I was submitting the PR. I've had issues running the installer in an air gapped network. The issue is if the connection to the snapstore was still pending when the submit button was pressed, the variable 'selections_by_name' attribute had not been created yet causing the installation to crash with an AttributeError. 

This PR just adds a better error message, I'm not able to test the change as I made the PR through the web browser.

dbungert:
This is a step in the right direction but not quite correct.
Upon testing this, without the change I can see the crash, but with it we get stuck in a loop that also can produce a crash.

dbungert (changes requested):
See comment -  the change isn't quite enough to fix.

dbungert:
@lambda-propaganda - are you interested in pursuing this further, or should we take it from here?  I didn't want to interrupt if you were looking to continue on this, but at the same rate this is pretty nasty for offline users and I'd like to see it resolved.

lambda-propaganda:
Hi Sorry for the delay, feel free to take over. I discovered this bug in an offline environment, so it makes submitting a PR very difficult as I have to do it through the web browser on another machine.

dbungert (approved):
Marking approve because it looks deceptive otherwise in the list.
@mwhudson - please review, should be trivial

mwhudson (approved):

PR #1041, opened 2021-09-08
DE-97 - system_setup: reduce WSLConfigurationAdvanced size

patrick330602:
This fixes the TODO 1 for the WSL OOBE
1. Remove duplicated fields between `wslconfadvanced` and `wslconfbase`.

### DIFF
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -346,7 +346,7 @@ class WSLConfigurationBase:
 
 
 @attr.s(auto_attribs=True)
-class WSLConfigurationAdvanced(WSLConfigurationBase):
+class WSLConfigurationAdvanced:
     gui_theme: str = attr.ib(default='default')
     gui_followwintheme: bool = attr.ib(default=True)
     legacy_gui: bool = attr.ib(default=False)
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -346,7 +346,7 @@ class WSLConfigurationBase:
 
 
 @attr.s(auto_attribs=True)
-class WSLConfigurationAdvanced:
+class WSLConfigurationAdvanced(WSLConfigurationBase):
     gui_theme: str = attr.ib(default='default')
     gui_followwintheme: bool = attr.ib(default=True)
     legacy_gui: bool = attr.ib(default=False)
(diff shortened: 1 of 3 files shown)


subiquity/common/types.py
```diff
@@ -346,7 +346,7 @@ class WSLConfigurationBase:
 
 
 @attr.s(auto_attribs=True)
-class WSLConfigurationAdvanced:
+class WSLConfigurationAdvanced(WSLConfigurationBase):
```
didrocks, line 349: WSLConfigurationBase and WSLConfigurationAdvanced are 2 independents models and controllers.

Where in reconfigure mode, we show both base and advanced screen, each one showing its own model.

Do no heritate from it, WSLConfigurationAdvanced are additional properties that are not in WSLConfigurationBase, or said differently, being in setup or reconfigure mode, custom_path, custom_mount_opt, gen_host and gen_revolvconf are always set in WSLConfigurationBase.

system_setup/models/wslconfadvanced.py
```diff
@@ -18,14 +18,15 @@
...
+from .wslconfbase import WSLConfigurationBase
 
 log = logging.getLogger('subiquity.models.wsl_configuration_advanced')
 
 # TODO WSL: Remove all common attributes with wslconfbase
 
 
 @attr.s
-class WSLConfigurationAdvanced(object):
+class WSLConfigurationAdvanced(WSLConfigurationBase):
```
didrocks, line 29: ditto, do not inherite here

didrocks (changes requested):

subiquity/common/types.py
patrick330602: Both are fixed

system_setup/models/wslconfadvanced.py
patrick330602: Both are fixed

didrocks:
This LGTM, thanks! I’ll let upstream review now.

didrocks (approved):

dbungert (approved):

dbungert:
I took the liberty of deleting the branch, if it is still needed we can restore it easily.

PR #1042, opened 2021-09-08
DE-98/DE-99 - system_setup: Wsl oobe auto reconf mode

patrick330602:
This fixes TODO 2 and TODO 3 (with the last commit): 
2. Implement an API in the server to know if we are in reconfigure or setup mode based on existing UID user >= 1000. Fetch that API from the client and set the variant to wsl_setup or wsl_configuration.
3. In wsl_configuration mode the UI shows ”base”/”advanced” screens. (required models list is already updated automatically)

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -114,6 +114,8 @@ class SubiquityClient(TuiApplication):
         "Progress",
         ]
 
+    variant_to_controllers = {}
+
     def __init__(self, opts):
         if is_linux_tty():
             self.input_filter = KeyCodesFilter()
@@ -321,6 +323,16 @@ class SubiquityClient(TuiApplication):
                             ])
                     print(line)
                 return
+
+            # Get the variant from the server and reload desired
+            # controllers if an override exists
+            variant = await self.client.meta.client_variant.GET()
+            if variant != self.variant:
+                self.variant = variant
+                controllers = self.variant_to_controllers.get(variant)
+                if controllers:
+                    self.load_controllers(controllers)
+
             await super().start()
             # Progress uses systemd to collect and display the installation
             # logs. Although some system don't have systemd, so we disable
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -78,6 +78,8 @@ class API:
         class client_variant:
             def POST(variant: str) -> None:
                 """Choose the install variant - desktop/server"""
+            def GET() -> str:
+                """Get the install variant - desktop/server"""
 
         class confirm:
             def POST(tty: str) -> None:
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -77,9 +77,11 @@ class API:
 
         class client_variant:
             def POST(variant: str) -> None:
-                """Choose the install variant - desktop/server"""
+                """Choose the install variant -
+                desktop/server/wsl_setup/wsl_reconfigure"""
             def GET() -> str:
-                """Get the install variant - desktop/server"""
+                """Get the install variant -
+                desktop/server/wsl_setup/wsl_reconfigure"""
 
         class confirm:
             def POST(tty: str) -> None:
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -105,6 +105,8 @@ class SubiquityModel:
         if root != '/':
             self.target = root
 
+        self.client_variant = ''
+
         self.debconf_selections = DebconfSelectionsModel()
         self.filesystem = FilesystemModel()
         self.identity = IdentityModel()
@@ -135,6 +137,7 @@ class SubiquityModel:
         self._postinstall_event = asyncio.Event()
 
     def set_source_variant(self, variant):
+        self.client_variant = variant
         self._cur_install_model_names = \
             self._install_model_names.for_variant(variant)
         self._cur_postinstall_model_names = \
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -105,8 +105,6 @@ class SubiquityModel:
         if root != '/':
             self.target = root
 
-        self.client_variant = ''
-
         self.debconf_selections = DebconfSelectionsModel()
         self.filesystem = FilesystemModel()
         self.identity = IdentityModel()
@@ -137,7 +135,6 @@ class SubiquityModel:
         self._postinstall_event = asyncio.Event()
 
     def set_source_variant(self, variant):
-        self.client_variant = variant
         self._cur_install_model_names = \
             self._install_model_names.for_variant(variant)
         self._cur_postinstall_model_names = \
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -123,7 +123,7 @@ class MetaController:
             self.app.controllers.Source.configured()
         else:
             self.app.base_model.set_source_variant(variant)
-        self.app.variant = variant
+        self.app.set_source_variant(variant)
 
     async def client_variant_GET(self) -> str:
         return self.app.variant
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -123,9 +123,10 @@ class MetaController:
             self.app.controllers.Source.configured()
         else:
             self.app.base_model.set_source_variant(variant)
+        self.app.variant = variant
 
     async def client_variant_GET(self) -> str:
-        return self.app.base_model.client_variant
+        return self.app.variant
 
     async def ssh_info_GET(self) -> Optional[LiveSessionSSHInfo]:
         ips = []
@@ -250,6 +251,7 @@ class SubiquityServer(Application):
 
     def __init__(self, opts, block_log_dir):
         super().__init__(opts)
+        self.set_source_variant(self.supported_variants[0])
         self.block_log_dir = block_log_dir
         self.cloud = None
         self.cloud_init_ok = None
@@ -296,6 +298,9 @@ class SubiquityServer(Application):
         self.hub.subscribe(InstallerChannels.NETWORK_PROXY_SET,
                            self._proxy_set)
         self.geoip = GeoIP(self)
+
+    def set_source_variant(self, variant):
+        self.variant = variant
 
     def load_serialized_state(self):
         for controller in self.controllers.instances:
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -124,6 +124,9 @@ class MetaController:
         else:
             self.app.base_model.set_source_variant(variant)
 
+    async def client_variant_GET(self) -> str:
+        return self.app.base_model.client_variant
+
     async def ssh_info_GET(self) -> Optional[LiveSessionSSHInfo]:
         ips = []
         if self.app.base_model.network:
--- a/subiquitycore/core.py
+++ b/subiquitycore/core.py
@@ -72,9 +72,16 @@ class Application:
         self.hub = MessageHub()
         self.aio_loop = asyncio.get_event_loop()
         self.aio_loop.set_exception_handler(self._exception_handler)
-        self.controllers = ControllerSet(
-            self.controllers_mod, self.controllers, init_args=(self,))
+        self.load_controllers(self.controllers)
         self.context = Context.new(self)
+
+    def load_controllers(self, controllers):
+        """ Load the corresponding list of controllers
+
+        Those will need to be restarted if already started """
+        self.controllers = ControllerSet(
+            self.controllers_mod, controllers,
+            init_args=(self,))
 
     def _exception_handler(self, loop, context):
         exc = context.get('exception')
--- a/system_setup/client/client.py
+++ b/system_setup/client/client.py
@@ -15,7 +15,7 @@
 
 import logging
 import sys
-from system_setup.common.helpers import is_reconfigure
+import os
 
 from subiquity.client.client import SubiquityClient
 
@@ -37,14 +37,43 @@ class SystemSetupClient(SubiquityClient):
         "WSLIdentity",
         "WSLConfigurationBase",
         "Summary",
-        ]
+    ]
 
-    def __init__(self, opts):
-        if is_reconfigure(opts.dry_run):
-            self.variant = "wsl_configuration"
-            self.controllers = [
-                "WSLConfigurationBase",
-                "WSLConfigurationAdvanced",
-                "Summary",
-            ]
-        super().__init__(opts)
+    variant_to_controllers = {
+        "wsl_setup": controllers,
+        "wsl_configuration": [
+            "WSLConfigurationBase",
+            "WSLConfigurationAdvanced",
+            "Summary",
+        ]
+    }
+
+    def restart(self, remove_last_screen=True, restart_server=False):
+        log.debug(f"restart {remove_last_screen} {restart_server}")
+        if self.fg_proc is not None:
+            log.debug(
+                "killing foreground process %s before restarting",
+                self.fg_proc)
+            self.restarting = True
+            self.aio_loop.create_task(
+                self._kill_fg_proc(remove_last_screen, restart_server))
+            return
+        if remove_last_screen:
+            self._remove_last_screen()
+        if restart_server:
+            self.restarting = True
+            self.ui.block_input = True
+            self.aio_loop.create_task(self._restart_server())
+            return
+        if self.urwid_loop is not None:
+            self.urwid_loop.stop()
+        cmdline = sys.argv
+        if self.opts.dry_run:
+            cmdline = [
+                sys.executable, '-m', 'system_setup.cmd.tui',
+                ] + sys.argv[1:] + ['--socket', self.opts.socket]
+            if self.opts.server_pid is not None:
+                cmdline.extend(['--server-pid', self.opts.server_pid])
+            log.debug("restarting %r", cmdline)
+
+        os.execvp(cmdline[0], cmdline)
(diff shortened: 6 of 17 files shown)


subiquity/common/apidef.py
```diff
@@ -77,6 +77,8 @@ def POST(endpoint_names: List[str]) -> None:
         class client_variant:
             def POST(variant: str) -> None:
                 """Choose the install variant - desktop/server"""
+            def GET() -> str:
+                """Get the install variant - desktop/server"""
```
didrocks, line 81: Please add wsl_setup and wsl_reconfigure.

subiquity/common/apidef.py
```diff
@@ -77,6 +77,8 @@ def POST(endpoint_names: List[str]) -> None:
         class client_variant:
             def POST(variant: str) -> None:
                 """Choose the install variant - desktop/server"""
```
didrocks, line 79: Please add wsl_setup and wsl_reconfigure in the comment.

subiquity/server/server.py
```diff
@@ -117,6 +117,9 @@ async def client_variant_POST(self, variant: str) -> None:
             raise ValueError(f'unrecognized client variant {variant}')
         self.app.base_model.set_source_variant(variant)
 
+    async def client_variant_GET(self) -> str:
+        return self.app.base_model.client_variant
```
didrocks, line 121: I would not store it on the base_model (above), but rather on the application directly. Then, the method is used to adapt the base_model, without storing it.

subiquity/models/subiquity.py
```diff
@@ -105,6 +105,8 @@ def __init__(self, root, install_model_names, postinstall_model_names):
         if root != '/':
             self.target = root
 
+        self.client_variant = ''
```
didrocks, line 108: see comment on server.py. I don’t think we should store the variant here but on the application from client_variant_POST.

system_setup/client/client.py
```diff
@@ -38,16 +39,10 @@ class SystemSetupClient(SubiquityClient):
...
-        #    (or default user set in wsl.conf?)
-        # 3. provide an API for this for the flutter UI to know about it
-        # 4. Add Configuration Base page before Advanced
-        # 5. Add language page
-        # self.variant = "wsl_configuration"
-        if opts.reconfigure:
+        if is_reconfigure(opts.dry_run):
+            self.variant = "wsl_configuration"
             self.controllers = [
+                "Welcome",
```
didrocks, line 45: We shouldn’t have the Welcome controller loaded in reconfigure mode (see the spec)

system_setup/client/client.py
```diff
@@ -38,16 +39,10 @@ class SystemSetupClient(SubiquityClient):
...
-        # TODO WSL:
-        # 1. remove reconfigure flag
-        # 2. decide on which UI to show up based on existing user UID >=1000
-        #    (or default user set in wsl.conf?)
-        # 3. provide an API for this for the flutter UI to know about it
-        # 4. Add Configuration Base page before Advanced
-        # 5. Add language page
-        # self.variant = "wsl_configuration"
-        if opts.reconfigure:
+        if is_reconfigure(opts.dry_run):
```
didrocks, line 42: We shouldn’t detect reconfiguration mode in the client. The model is basically:

1. Let the server detects based on UID rules if we are in setup or reconfiguration mode.
2. Get it from the server API in the client in which mode we are
3. A little bit silly with the current implementation, but needed… Post the mode back to the server for it to load the expected server models…

Even if this seems counter-intuitive, because of the 3. this allows:
- the server to do the initial detection, as this is where the actual code is running (client could be remote)
- let the client override it if needed (for the tests for instance)

system_setup/client/controllers/summary.py
```diff
@@ -87,10 +89,21 @@ async def _wait_status(self, context):
...
+            tmp_variant = onsite_variant
+
         identity = getattr(self.app.client, "identity")
         if identity is not None:
             data = await identity.GET()
             real_name = data.realname
+        if tmp_variant == "wsl_configuration":
+            tmp_uid = os.getuid()
+            if tmp_uid >= 1000 and tmp_uid < 65534:
+                real_name = pwd.getpwuid(tmp_uid)[0]
```
didrocks, line 105: I think we shuld rather change the summary message without using the user name here. Then the whole variant os.getuid is not needed here as we don’t need real_name.

system_setup/client/controllers/wslconfbase.py
```diff
@@ -12,7 +12,11 @@ class WSLConfigurationBaseController(SubiquityTuiController):
 
     async def make_ui(self):
         data = await self.endpoint.GET()
-        return WSLConfigurationBaseView(self, data)
+        variant = "wsl_setup"
+        onsite_variant = getattr(self.app, "variant")
+        if onsite_variant is not None:
+            variant = onsite_variant
+        return WSLConfigurationBaseView(self, data, variant)
```
didrocks, line 19: Same, variant should not be needed in the Base View, it’s the same in both setup and reconfiguration

system_setup/common/helpers.py
```diff
@@ -0,0 +1,31 @@
...
+def is_reconfigure(is_dryrun):
+    is_dryrun_reconfigure = is_dryrun and \
+        os.getenv("DRYRUN_RECONFIG") == "true"
+    count = 0
+    with open('/etc/passwd', 'r') as f:
+        for line in f:
+            # check every normal user except nobody (65534)
+            if int(line.split(':')[2]) >= 1000 and \
+               int(line.split(':')[2]) != 65534:
+                count += 1
```
didrocks, line 29: Rather than counting the user, you can stop as soon as you have found one matching user and return true

system_setup/common/helpers.py
```diff
@@ -0,0 +1,31 @@
...
+    is_none_dryrun_normaluser = not is_dryrun and count != 0
+    return is_dryrun_reconfigure or is_none_dryrun_normaluser
```
didrocks, line 31: Return early in the function if is_dryrun_reconfigure is true just below DRYRUN_RECONFIG environment get.

system_setup/ui/views/wslconfbase.py
```diff
@@ -95,9 +95,11 @@ class WSLConfigurationBaseView(BaseView):
     excerpt = _(
         "In this page, you can configure Ubuntu WSL options to your needs.\n")
 
-    def __init__(self, controller, configuration_data):
+    def __init__(self, controller, configuration_data, variant):
         self.controller = controller
-
+        self.is_conf = variant == "wsl_configuration"
+        if self.is_conf:
+            self.title = _("WSL configuration - Base options")
```
didrocks, line 102: We should always use that title without using the variant.

didrocks (changes requested):

system_setup/ui/views/wslconfbase.py
patrick330602: This one is for In wsl_configuration mode the UI shows ”base”/”advanced” screens though, People would be confused that there are two configuration pages with the same title, otherwise we should change the description

system_setup/client/controllers/wslconfbase.py
patrick330602: This one is for In wsl_configuration mode the UI shows ”base”/”advanced” screens.

system_setup/client/client.py
patrick330602: I am not sure how should I do this since we should then call an async function in __init__, which of course crashes. What is the proper way I should implement this?

system_setup/ui/views/wslconfbase.py
didrocks: Yes. but my suggestion was to always use that title rather. Those are the base options and can show them as such in setup mode.

system_setup/client/controllers/wslconfbase.py
didrocks: See my other comment below, which should make this uneeded :)

system_setup/client/client.py
didrocks: This is a good question as this is triggered by the fact of having static class member with the base inherited/+meta class making magic on controllers, replacing the objects. I think of multiple hackish solutions, but maybe @mwhudson or @dbungert has a more upstream-compatible idea?

didrocks:
Something else I didn’t spot yesterday: please target `main` branch instead of `wsl_oobe_reduce_dup`. This has no dependency on `wsl_oobe_reduce_dup` and should not include this commit. As you changed `wsl_oobe_reduce_dup`, you would need otherwise to due a useless rebase (and same later for other branches that depends on this one)

system_setup/client/client.py
mwhudson: So I guess there are two broad approaches here.

1. Have all controllers loaded in the client always but have the ones that are inappropriate for the current mode skip themselves (the generic tui client code has support for this, just raise subiquitycore.tuicontroller.Skip from make_ui)
2. Change the set of controllers to be loaded after seeing what the server says

The first is probably easier, tbh, although the second might be aesthetically more pleasing. Nothing in the installer tui client touches the controllers instance until the first status API call has returned so while the ControllerSet class wasn't written with this in mind it looks like assigning `controller_names` between the instance being constructed and calling `load_all` will work fine (another approach would be to construct the ControllerSet later, probably).

mwhudson: WRT the POSTing the variant back... surely something in the system_state server startup code can call set_source_variant on the model during startup?

subiquity/common/apidef.py
patrick330602: This is fixed.

subiquity/server/server.py
patrick330602: Updated.

system_setup/client/client.py
patrick330602: Removed.

system_setup/client/client.py
patrick330602: Updated, and tested locally. It should now work properly.

system_setup/client/controllers/summary.py
patrick330602: The logic is changed for this, needs check whether should be improved.

system_setup/ui/views/wslconfbase.py
patrick330602: Addressed

system_setup/client/controllers/wslconfbase.py
patrick330602: addressed

system_setup/client/controllers/summary.py
didrocks: I don’t see any changes there, please have a look before I make a second review pass.

didrocks:
There is still one non outdated comment that needs to be fixed IMHO and some merge conflicts, do you mind fixing those before requesting a new review?

system_setup/client/controllers/summary.py
patrick330602: Sorry for that, fixed now

didrocks (changes requested):
Some changes that were requested in this PR are still not here. Also, the branch needs a rebase and not a merge as already stated.

patrick330602:
@didrocks I just checked and I have fixed everything that was requested, including rebase to the main. Am I missing anything here?

didrocks:
From what I see in a quick check:
- you merged main, not do a `git rebase`
- WSLConfigurationAdvanced still inherites from WSLConfigurationBase (we removed it and asked to remove it multiple times, is the diff not up to date with your local checking? :/)
- There are some comments added like
`# controllers = {
    #    "server": [
`
- In the diff, there are some empty lines added/removed

Please do a rebase, and then do a real serious check that all previously changes are fixed, that there is no leftover of debug and no uneeded diffs.

patrick330602:
> From what I see in a quick check:
> 
> * you merged main, not do a `git rebase`
> * WSLConfigurationAdvanced still inherites from WSLConfigurationBase (we removed it and asked to remove it multiple times, is the diff not up to date with your local checking? :/)
> * There are some comments added like
> 
> ```
> # controllers = {
>     #    "server": [
> ```
> 
> * In the diff, there are some empty lines added/removed
> 
> Please do a rebase, and then do a real serious check that all previously changes are fixed, that there is no leftover of debug and no uneeded diffs.

Updated and fixed. Could you check it again? Thanks!

subiquity/client/client.py
```diff
@@ -112,7 +112,9 @@ def make_ui(self):
         "SSH",
         "SnapList",
         "Progress",
-        ]
```
didrocks, line 115: Remove this ] in diff and keep existing alignement.

subiquity/server/server.py
```diff
@@ -247,6 +251,7 @@ def make_model(self):
 
     def __init__(self, opts, block_log_dir):
         super().__init__(opts)
+        self.variant = self.supported_variants[0]
```
didrocks, line 254: This should call `self.set_source_variant()`

subiquitycore/core.py
```diff
@@ -32,13 +32,15 @@ class Application:
     # A concrete subclass must set project and controllers attributes, e.g.:
     #
     # project = "subiquity"
-    # controllers = [
+    # controllers = {
+    #    "server": [
```
didrocks, line 36: Please remove all this file diff, not correct anymore with the current implementation (this was already a comment in the previous review)

system_setup/ui/views/summary.py
```diff
@@ -12,7 +12,7 @@
 from subiquitycore.ui.buttons import (
     cancel_btn,
     ok_btn,
-    )
```
didrocks, line 15: Remove this and keep existing alignement.

system_setup/ui/views/summary.py
```diff
@@ -60,7 +64,7 @@ def update_for_state(self, state):
             btns = [
                 self.view_error_btn,
                 self.reboot_btn,
-                ]
```
didrocks, line 63: Remove this and keep existing alignement.

system_setup/ui/views/wslconfbase.py
```diff
@@ -97,7 +97,6 @@ class WSLConfigurationBaseView(BaseView):
 
     def __init__(self, controller, configuration_data):
         self.controller = controller
-
```
didrocks, line 100: Remove this not relevant in the current PR.

didrocks (changes requested):
Some small typos and alignements fixes needed. Then, it will be ready for upstream to review.

didrocks:
You should remove https://github.com/canonical/subiquity/pull/1042/commits/e0884e9943c23c21994c8cc4bedd50f8e87fa612. This commit doesn’t do anything apart from introducing a diff.

didrocks:
You reintroduce the inheritance in https://github.com/canonical/subiquity/pull/1042/commits/675e6afeec3d9e59e194492810cd9752b28cf36c. Please fix this and remove last commit.

didrocks:
Ok, all comments are now resolved, the tests pass. @mwhudson @dbungert, mind having a review of the non system_setup part?

dbungert (approved):

system_setup/server/controllers/configure.py
```diff
@@ -33,6 +33,7 @@ class ConfigureController(SubiquityController):
 
     def __init__(self, app):
         super().__init__(app)
+        self.app = app
```
mwhudson, line 36: This line isn't necessary, surely?

system_setup/client/client.py
```diff
@@ -36,21 +37,43 @@ class SystemSetupClient(SubiquityClient):
...
+    variant_to_controllers = {
+        "wsl_setup": controllers,
+        "wsl_configuration": [
+            "WSLConfigurationBase",
+            "WSLConfigurationAdvanced",
+            "Summary",
         ]
+    }
+
+    def restart(self, remove_last_screen=True, restart_server=False):
```
mwhudson, line 51: Wait, didn't I ask you to parameterize the superclass implementation of this method in a previous PR?

subiquity/server/server.py
```diff
@@ -123,6 +123,10 @@ async def client_variant_POST(self, variant: str) -> None:
             self.app.controllers.Source.configured()
         else:
             self.app.base_model.set_source_variant(variant)
+        self.app.variant = variant
```
mwhudson, line 126: It would be more consistent for this to call self.app.set_source_variant would it not?

mwhudson (commented):
Seems ok, a few questions. Happy for Dan to merge this if you can convince him your responses to my questions are ok :)

system_setup/client/client.py
patrick330602: I am not clear about this.... could you elaborate more?

system_setup/server/controllers/configure.py
mwhudson: Well OK we can argue about how and whether it's needed in that branch then :)

system_setup/client/client.py
mwhudson: Having read things a bit more, I don't think you need to add the restart method to this class at all. Can you explain why you did that?

patrick330602: Now I check again, yep you are correct, the restart is not needed here.

subiquity/server/server.py
patrick330602: done for the fixes

system_setup/client/client.py
patrick330602: done for the fixes here.

mwhudson (approved):
LGTM, thanks for the changes

PR #1046, opened 2021-09-10
actions: trigger -push

dbungert:
The push trigger is semi redundant.  It is nice in that someone about to
open a PR can first verify on their own fork that the CI passes, but
other than that it just causes the CI to run twice.  This is
particularly visible on the system_setup branhces, which has double the
number of CI steps expected.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   test:


mwhudson:
Hmm i do sort of like the way things are (tbh i would like a way to optionally run CI on _each commit_ of a branch sometimes). Is the problem with the wsl branches that they are developed in the canonical repo and not some user repo?

dbungert:
Yea, it's not a super bad problem, and it only really applies to the ones
where the branch work is done directly on canonical/subiquity.

-Dan

PR #1047, opened 2021-09-10
actions: +accept-existing-contributors for CLA

dbungert:
Canonical employees who have not signed the CLA, and sign commits with
their ubuntu.com alias, fall into a weird gap where their commits are
covered (as employees) but there is no great way for the CLA check to
determine that.
https://github.com/canonical/subiquity/pull/1044#issuecomment-916764603

### DIFF
--- a/.github/workflows/cla-check.yaml
+++ b/.github/workflows/cla-check.yaml
@@ -8,3 +8,5 @@ jobs:
     steps:
       - name: Check if CLA signed
         uses: canonical/has-signed-canonical-cla@v1
+        with:
+          accept-existing-contributors: true


mwhudson:
So this prompts two questions:

1. What do we do for the next Canonical employee to have this problem? (I guess we just merge over the cla check's protestations?)
2. We have already accepted a few, fairly trivial, contributions from external people. We should probably check if these people have signed the CLA, or at least check that before merging anything more from them...

didrocks:
> What do we do for the next Canonical employee to have this problem? (I guess we just merge over the cla check's protestations?)
This is what we did in some other projects. Accept the first merge despite CLA check's protestation. Then, they are OK.

mwhudson (approved):
This isn't great, but it's better than not having it IMO

PR #1048, opened 2021-09-10
autoinstall: skip autoinstall on empty arg

dbungert:
If autoinstall is set on the command line, don't override that with a
pre-baked file or from cloud-init.  The intent here is to allow
ubuntu-desktop-installer to temporarily say --autoinstall="" to disable
autoinstall until UI support for autoinstall is in place.


mwhudson:
:+1:

mwhudson (approved):

PR #1051, opened 2021-09-14
DE-101 - Move all “Commits/Write to file” in the configure controller

patrick330602:
fixes **just** "Move all “Commits/Write to file” in the configure controller"

### DIFF
--- a/system_setup/server/controllers/configure.py
+++ b/system_setup/server/controllers/configure.py
@@ -14,6 +14,9 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import logging
+import os
+import subprocess
+from subiquitycore.utils import run_command
 
 from subiquitycore.context import with_context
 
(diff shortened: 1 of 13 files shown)


didrocks:
CI is failing due to lack of dependency of ubuntu-wsl-integration. I think to avoid fixing this (once rebased with DE-100 accepted) and then remove it in DE-102, let’s merge DE-101 and DE-102. What do you think?

Side note: please wait for the CI to pass (or analyse why it’s failing if it’s a false positive) before requesting reviews.

didrocks (changes requested):
Changing state according to feedback

system_setup/server/controllers/configure.py
```diff
@@ -68,13 +72,135 @@ async def configure(self, *, context):
...
+                run_command(["/usr/sbin/useradd", "-m", "-s", "/bin/bash",
+                             "-p", wsl_identity.password,
+                             wsl_identity.username])
+                run_command(["/usr/sbin/usermod", "-a",
+                             "-c", wsl_identity.realname,
+                             "-G", self.get_userandgroups(),
+                             wsl_identity.username])
+
+                wslconf_base = self.model.wslconfbase
+                run_command(["/usr/bin/ubuntuwsl", "update",
```
dbungert, line 87: There's a lot of repetition here.  A utility function taking the args  "WSL.automount.root" and wslconf_base.custom_path could really help shorten this.  About 22 usages of this same construct.

system_setup/server/controllers/wslconfbase.py
```diff
@@ -43,6 +47,32 @@ class WSLConfigurationBaseController(SubiquityController):
         'additionalProperties': False,
         }
 
+    def __init__(self, app):
+        super().__init__(app)
+
+        # load the config file
+        data = {}
+
+        if path.exists('/etc/wsl.conf'):
```
dbungert, line 56: Looks similar to the construct in wsl_advanced, can we refactor out a utility function?

dbungert (changes requested):

system_setup/server/controllers/wslconfbase.py
patrick330602: Gotcha. but the PR is now moved to the other one now.

system_setup/server/controllers/configure.py
didrocks: This is not what I suggested. I suggested merging TODO DE101 and DE102, only 104 can stay aside (for dry-run mode). However, 102 is about changing ubuntu-wsl to write directly on system wsl.conf

PR #1053, opened 2021-09-20
Have SubiquityController.configured broadcast rather than set an event

mwhudson:
I'm not _sure_ the change to make configured async is necessary but it probably makes things more flexible...

### DIFF
--- a/subiquity/server/controllers/locale.py
+++ b/subiquity/server/controllers/locale.py
@@ -37,7 +37,7 @@ class LocaleController(SubiquityController):
     def start(self):
         self.model.selected_language = os.environ.get("LANG") \
                 or self.autoinstall_default
-        self.configured()
+        self.app.aio_loop.create_task(self.configured())
 
     def serialize(self):
         return self.model.selected_language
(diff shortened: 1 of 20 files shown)


subiquity/server/controllers/locale.py
```diff
@@ -37,7 +37,7 @@ def load_autoinstall_data(self, data):
     def start(self):
         self.model.selected_language = os.environ.get("LANG") \
                 or self.autoinstall_default
-        self.configured()
+        self.app.aio_loop.create_task(self.configured())
```
dbungert, line 40: Application.start is already async, it would be in future work to see if the controller.start() methods could be async also.
I guess this is more of an asyncio question, but is there any risk that multiple tasks are trying to modify the _configured_names set?

mwhudson: My mental model for the controller start() methods is that it is when to start doing async things, but shouldn't really be async themselves -- we don't want to block here. But maybe that's not helpful.

SubiquityModel._configured is not async (or, more precisely, doesn't await on anything) so it's always executed serially.

dbungert (approved):

PR #1054, opened 2021-09-20
write server state to disk when it is updated

mwhudson:
and check this in client when getting status over api fails

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -250,6 +250,15 @@ class SubiquityClient(TuiApplication):
                 app_status = await self.client.meta.status.GET(
                     cur=app_state)
             except aiohttp.ClientError:
+                try:
+                    fp = open(self.state_path("server-state"))
+                except FileNotFoundError:
+                    pass
+                else:
+                    with fp:
+                        state = getattr(ApplicationState, fp.read(), None)
+                    if state == ApplicationState.EXITED:
+                        self.exit()
                 await asyncio.sleep(1)
                 continue
 
@@ -258,9 +267,6 @@ class SubiquityClient(TuiApplication):
             print('start: ' + event["MESSAGE"])
         elif event['SUBIQUITY_EVENT_TYPE'] == 'finish':
             print('finish: ' + event["MESSAGE"])
-            context_name = event.get('SUBIQUITY_CONTEXT_NAME', '')
-            if context_name == 'subiquity/Shutdown/shutdown':
-                self.exit()
 
     async def connect(self):
 
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -303,8 +303,7 @@ class SubiquityClient(TuiApplication):
                 "waiting for cloud-init", self._status_get(cur=status.state))
         if status.state == ApplicationState.EARLY_COMMANDS:
             print("running early commands")
-            status = await spinning_wait(
-                "waiting for cloud-init", self._status_get(cur=status.state))
+            status = await self._status_get(cur=status.state)
             await asyncio.sleep(0.5)
         return status
 
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -229,6 +229,22 @@ class SubiquityClient(TuiApplication):
             answer = await run_in_thread(input)
         await self.confirm_install()
 
+    async def _status_get(self, cur=None):
+        while True:
+            try:
+                return await self.client.meta.status.GET(cur=cur)
+            except aiohttp.ClientError:
+                try:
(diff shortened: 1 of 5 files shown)


jibel (commented):
This partially addresses the point discussed in PR #1044.
It only checks the "EXIT" state when early command. This fixes a crash in the client when the server exits and the client is still waiting on it. 
However if the server exits after the early_commands check in the client and we don't have journal support, it introduces a race and the client will never see the state and will never exit.
This is why in PR #1044 the state check is done in `noninteractive_watch_app_state(self, initial_status)`

mwhudson:
How about this version then?

jibel:
It looks good to me. I've a couple of question though:

* How do you envision integration tests to validate this code path?
* For integration tests, how do you test the fallback (ie not systemd booted) on systems where systemd is running?

subiquity/cmd/tui.py
```diff
@@ -87,6 +87,8 @@ def main():
         opts, unknown = parser.parse_known_args(args)
         if opts.socket is None:
             os.makedirs('.subiquity', exist_ok=True)
+            if os.path.exists('.subiquity/run/subiquity/server-state'):
+                os.unlink('.subiquity/run/subiquity/server-state')
```
jibel, line 91: This must also be done in system_setup/cmd/tui.py

mwhudson: oh yes

jibel:
Thanks. Approved!

subiquity/client/client.py
```diff
@@ -227,6 +227,22 @@ async def noninteractive_confirmation(self):
...
+            except aiohttp.ClientError:
+                try:
+                    fp = open(self.state_path("server-state"))
+                except FileNotFoundError:
+                    pass
+                else:
+                    with fp:
+                        state = getattr(ApplicationState, fp.read(), None)
+                    if state == ApplicationState.EXITED:
+                        self.exit()
```
dbungert, line 243: do we need handling for the non-exited case here?
presumably we're only in this part if the server was running, and the process has exited without cleaning transitioning to an EXITED state first - a server process crash most likely.
FWIW, I tried to create this scenario by running dryrun and stepping thru a few screens then killing the server process, and created a different client crash instead
```
  File "/home/dbungert/dev/github/canonical/subiquity/subiquity/common/api/client.py", line 82, in make_request
    async with session.request(
  File "/usr/lib/python3/dist-packages/aiohttp/client.py", line 1117, in __aenter__
    self._resp = await self._coro
  File "/usr/lib/python3/dist-packages/aiohttp/client.py", line 520, in _request
    conn = await self._connector.connect(
  File "/usr/lib/python3/dist-packages/aiohttp/connector.py", line 535, in connect
    proto = await self._create_connection(req, traces, timeout)
  File "/usr/lib/python3/dist-packages/aiohttp/connector.py", line 1197, in _create_connection
    raise ClientConnectorError(req.connection_key, exc) from exc
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host a:80 ssl:default [Connection refused]
```

mwhudson: We also have to consider that the server might be restarting. We could be more explicit about that by adding a RESTARTING state and explicitly not-handling that here I suppose.

It's a bit poor that killing the server makes the client crash but I'm not sure it's that bad either. That's a different codepath of course and so best handled in a different branch?

dbungert: Yes, new branch is fine

dbungert: Or the client crash thing on a possible future branch I mean

dbungert (approved):

jibel (approved):

jibel:
I ran the testsuite locally and it fails with:
+ timeout --foreground 60 sh -c 'LANG=C.UTF-8 python3 -m subiquity.cmd.tui --autoinstall examples/autoinstall.yaml                                --dry-run --machine-config examples/existing-partitions.json --bootloader bios                                --kernel-cmdline '\''autoinstall'\'''
running server pid 11100
connecting...  
running early commands
generating crash report
report saved to .subiquity/var/crash/1632463956.191208601.ui.crash
  File "/root/subiquity/subiquity/client/client.py", line 390, in run
    super().run()
  File "/root/subiquity/subiquitycore/tui.py", line 374, in run
    super().run()
  File "/root/subiquity/subiquitycore/core.py", line 135, in run
    raise exc
  File "/root/subiquity/subiquity/client/client.py", line 312, in start
    self.interactive = status.interactive
AttributeError: 'coroutine' object has no attribute 'interactive'

PR #1055, opened 2021-09-20
WSL OOBE - no dry mode

jibel:
2 fixes after running in real mode (no dry run) on a WSL system:
1. The server must be started by the client when it is not socket activated
2. When there is no cloud init, the client was waiting on it forever


didrocks:
Ack. Only impacting `system_setup/`, we tested it together, let’s merge it.

didrocks (approved):

PR #1056, opened 2021-09-20
schema: update timezone list

dbungert:


mwhudson (approved):
I mean, sure, but where does this list come from? Why update it now?

dbungert:
I get the list of possibles from timedatectl list-timezones.
If this is proving more aggravation than value, checking valid timezone strings could be made to be just a non-schema runtime check (I've been thinking this already).
When other schema changes are made and the timezone list is out of date, doing `make schema` thus updates the timezones.
So doing this now is to get the list updated because other things are touching the schema (curtin apt stuff, and depending on how we solve autoinstall roundtrip schema might get adjusted there)

dbungert:
will change to not have timezone list in schema instead

PR #1059, opened 2021-09-22
pubsub: optional argument at broadcast time

dbungert:


mwhudson (changes requested):
I'd rather we got rid of the arguments on subscribe (nothing uses that, right?) and add *args, **kw to the broadcast methods.

dbungert:
That makes sense. I guess the private_data variable in a callback makes more sense in C, but we have access to callbacks that can reference self so it doesn't seem overwelmingly helpful and I'd rather have access to a rich set of arguments on the broadcast side.

mwhudson (approved):
nice

PR #1060, opened 2021-09-22
autoinstall: fix issues in autoinstall roundtrip

dbungert:
Fix autoinstall output for kernel & snaplist.
This is necessary but not sufficient to roundtrip an autoinstall file.

### DIFF
--- a/subiquity/server/controllers/kernel.py
+++ b/subiquity/server/controllers/kernel.py
@@ -75,8 +75,4 @@ class KernelController(NonInteractiveController):
         self.model.metapkg_name = package
 
     def make_autoinstall(self):
-        return {
-            'kernel': {
-                'package': self.model.metapkg_name,
-                },
-            }
+        return {'package': self.model.metapkg_name}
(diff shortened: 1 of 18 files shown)


mwhudson:
Argl let's just fix how the snap stuff is generated please?

mwhudson:
Argh sorry to be a pain here but I think we probably need to keep compatibility in the answers file here, there is an automated test case that uses is_classic. Or I guess we can fix this as soon as this branch moves to stable... maybe that's OK.

mwhudson (approved):

mwhudson:
Pushed an attempt at that, feel free to merge if you like that.

subiquity/server/controllers/kernel.py
```diff
@@ -75,8 +75,4 @@ def load_autoinstall_data(self, data):
         self.model.metapkg_name = package
 
     def make_autoinstall(self):
-        return {
-            'kernel': {
```
RobQuistNL, line 79: Thank you :) came here to fix this bugger..

PR #1061, opened 2021-09-23
set wider permissions on log directory

mwhudson:


subiquitycore/log.py
```diff
@@ -13,12 +13,16 @@
...
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import grp
 import logging
 import os
 
 
 def setup_logger(dir, base='subiquity'):
-    os.makedirs(dir, exist_ok=True)
+    os.makedirs(os.path.basename(dir), exist_ok=True)
```
seb128, line 22: Why this change? It seems not needed for the permission and probably wrong (or it relies on the helper to run in the right location?)

subiquitycore/log.py
```diff
@@ -13,12 +13,16 @@
...
+    if os.getuid() == 0:
+        os.chmod(dir, 0o775)
+        os.chown(dir, 0, grp.getgrnam('adm').gr_gid)
```
seb128, line 25: the documentation states 'To leave one of the ids unchanged, set it to -1', wouldn't it be better than forcing 0 as uid?

seb128 (commented):
Thanks. It should allow for the graphical installer to write to the log but only after the subiquity service is started, it still feel like to me that it would be more robust to create the dir with the right permission on the iso or in casper.

subiquitycore/log.py
mwhudson: Because I shouldn't write code late at night :) (I had a version that passed a mode to mkdir but that ran into issues with umask)

subiquitycore/log.py
mwhudson: Ah TIL.

seb128 (approved):
Thanks, it should help at least in cases where the service is active before the frontend tries to create the log

PR #1062, opened 2021-09-24
Add '--quiet' to 'lszdev' to omit additional progress msg - lp1944516

frank-heimes:
Adding "--quiet" as additional option to "lszdev" omits additional progress status messages that "lszdev" might produce in case a significant amount of ccw devices are present (in the range of thousands of devices), since such progress messages will eventually cause parsing issues, like mentioned in LP#1944516.

### DIFF
--- a/scripts/zdev-generate.py
+++ b/scripts/zdev-generate.py
@@ -12,6 +12,6 @@ for row in rows:
     record = dict([k.split('=',1) for k in row])
     if record['COLUMN'] not in ('TYPE', 'ATTR:', 'ATTRPATH:'):
         desired_columns.append(record['COLUMN'].lower())
-cmd = ['lszdev','--pairs','--columns',','.join(desired_columns)]
+cmd = ['lszdev', '--quiet', '--pairs','--columns',','.join(desired_columns)]
 print(cmd)
 print(' '.join(cmd))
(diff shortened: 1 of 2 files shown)


scripts/zdev-generate.py
```diff
@@ -12,6 +12,6 @@
     record = dict([k.split('=',1) for k in row])
     if record['COLUMN'] not in ('TYPE', 'ATTR:', 'ATTRPATH:'):
         desired_columns.append(record['COLUMN'].lower())
-cmd = ['lszdev','--pairs','--columns',','.join(desired_columns)]
+cmd = ['lszdev', '--quiet', '--pairs','--columns',','.join(desired_columns)]
```
dbungert, line 15: Thanks for looking into this!
If I can nitpick, it would be nice to be consistent on the spaces.  The rest of the array doesn't use them.  Go with a space after the comma for the rest of the array please, that way it's a little closer to passing the linter.

dbungert (changes requested):

scripts/zdev-generate.py
frank-heimes: ok. I adjusted the whitespaces
(I c&p-ed from the other file, that had the whitespaces)

frank-heimes (commented):
with adjusted whitespaces

frank-heimes (commented):
now with adjusted whitespaces

mwhudson:
@frank-heimes can you join the canonical contributors group on launchpad?

But merging this anyway, thanks.

PR #1064, opened 2021-09-27
Fr 1710 storage

dbungert:


mwhudson:
Seems ok so far :) Didn't look super closely yet.

Makefile
```diff
@@ -68,13 +68,18 @@ flake8:
 	$(PYTHON) -m flake8 $(CHECK_DIRS) --exclude gettext38.py,contextlib38.py
 
 unit: gitdeps
-	python3 -m unittest discover
+	python3 -m pytest --ignore curtin --ignore probert \
+		--ignore subiquity/tests/api
+
+api:
+	@[ "$(shell lsb_release -sr | cut -d. -f1)" -le "18" ] || \
```
mwhudson, line 75: So I don't think we need this any more, we should just stop testing on bionic.

subiquity/common/filesystem/labels.py
```diff
@@ -258,19 +258,36 @@ def for_client(device, *, min_size=0):
 @for_client.register(Disk)
 @for_client.register(Raid)
 def _for_client_disk(disk, *, min_size=0):
+    path = disk.path if getattr(disk, 'path', None) else None
```
mwhudson, line 261: How does this differ from "path = getattr(disk, 'path', None)"? I guess it's different if disk.path is a non-None falsey value but surely that's not what this is about?

subiquity/common/types.py
```diff
@@ -244,20 +244,29 @@ class Bootloader(enum.Enum):
...
 @attr.s(auto_attribs=True)
 class Partition:
-    size: int
-    number: int
-    annotations: List[str]
+    size: Optional[int] = None
+    number: Optional[int] = None
+    preserve: Optional[bool] = None
+    wipe: Optional[str] = None
+    annotations: Optional[List[str]] = []
```
mwhudson, line 251: I guess this is OK in this case but for mutable defaults something like

```
    annotations: List[str] = attr.ib(default=attr.Factory(list))
```

is better. (Also I don't think the `Optional` is appropriate here?)

subiquity/common/types.py
```diff
@@ -244,20 +244,29 @@ class Bootloader(enum.Enum):
...
+    mount: Optional[str] = None
+    format: Optional[str] = None
+    grub_device: Optional[bool] = None
```
mwhudson, line 254: Is grub_device == None different from grub_device == False? Or would `grub_device: bool = False` be better here?

subiquity/common/types.py
```diff
@@ -244,20 +244,29 @@ class Bootloader(enum.Enum):
...
-    ok_for_guided: bool
+    label: Optional[str]
+    type: Optional[str]
+    size: int
+    ok_for_guided: Optional[bool]
+    ptable: Optional[str]
+    preserve: Optional[bool] = None
+    path: Optional[str] = None
+    free_for_partitions: Optional[int] = 0
+    usage_labels: Optional[List[str]] = []
```
mwhudson, line 269: Same comments as before: Optional[int] = 0 looks funny, just int = 0 is fine, use a factory for mutable default.

subiquity/server/controllers/filesystem.py
```diff
@@ -179,6 +181,21 @@ def guided_lvm(self, disk, lvm_options=None):
                 mount="/",
                 ))
 
+    async def guided(self, choice):
```
mwhudson, line 184: There's no reason to think this function will ever need to await is there?

mwhudson (changes requested):
So this looks pretty good, I have a few comments on the type annotations. I think we can lose some testing complexity now we can assume focal or better.

Would it be possible to make the move to pytest in a separate branch?

subiquity/tests/api/test_api.py
```diff
@@ -0,0 +1,521 @@
...
+        params = {}
+        for key in kwargs:
+            params[key] = dumps(kwargs[key])
+        data = dumps(data)
+        info = f'{method} {query}'
+        if params:
+            for i, key in enumerate(params):
+                joiner = '?' if i == 0 else '&'
+                info += f'{joiner}{key}={params[key]}'
+        print(info)
```
mwhudson, line 88: I guess this is only for debugging but surely you don't have to print this by hand? Would printing resp.url below be good enough?

subiquity/tests/api/test_api.py
```diff
@@ -0,0 +1,521 @@
...
+        return run
+    return wrapper
+
+
+def json_print(json_data):
+    print(json.dumps(json_data, indent=4))
+
+
+def loads(data):
+    return json.loads(data) if data else None
```
mwhudson, line 51: Can this check for data == "" rather than just "if data"? If that's the intent. I also kind of dislike the ternary operator form you use here but maybe I just need to get over that.

subiquity/tests/api/test_api.py
```diff
@@ -0,0 +1,521 @@
...
+        self.server = asyncio.create_task(self.proc.communicate())
+
+    async def asyncSetUp(self):
+        if self.need_spawn_server:
+            await self.spawn_server()
+
+        # setup client
+        conn = aiohttp.UnixConnector(path=socket_path)
+        self.session = aiohttp.ClientSession(connector=conn)
+        await self.server_startup()
```
mwhudson, line 127: So I sort of hate using implicit setup for this. Would it be possible to have a helper method that each test can call to set up a server (that would of course arrange for the server to be torn down at the end of the test using addAsyncCleanup)?

subiquity/tests/api/test_api.py
```diff
@@ -0,0 +1,521 @@
...
+#     @unittest.skip("useful for interactive debug only")
+#     async def test_v2_delete_debug(self):
+#         await self.get('/storage/v2')
+
+
+class TestSimple(TestAPI):
+    @timeout(5)
+    async def test_not_bitlocker(self):
+        resp = await self.get('/storage/has_bitlocker')
+        self.assertEqual(0, len(resp))
```
mwhudson, line 152: The problem with implicit set up is that you have to know what it does to understand that this test is correct (not only do you have to know that a server has been implicitly started, you have to know that it uses a config that doesn't have any bitlocker disks!). It's more verbose but I would prefer this test to be written more like:

```
async def test_not_bitlocker(self):
    server = await self.startServerForConfig('examples/simple.json')
    resp = await server.get('/storage/has_bitlocker')
    self.assertEqual(0, len(resp))
```

This would also mean that you could have the "test_not_bitlocker" and "test_yes_bitlocker" tests next to each other.

subiquity/tests/api/test_api.py
```diff
@@ -0,0 +1,521 @@
...
+        self.assertEqual(simple_add, manual_add)
+
+
+class TestWin10(TestAPI):
+    machine_config = 'examples/win10.json'
+
+    @timeout(5)
+    async def test_bitlocker(self):
+        resp = await self.get('/storage/has_bitlocker')
+        self.assertEqual('BitLocker', resp[0]['partitions'][2]['format'])
```
mwhudson, line 236: Uh, OK this test is, let's say, not clearly correct to the first time reader. Let me dump some reading on you! First a talk I gave ages and ages ago and wrote up into a blog post slightly less ages ago:

https://medium.com/@mwhudson/a-patronizing-guide-to-what-makes-a-good-unit-test-fe895951ce35

And another blog post on the same kind of topic: https://thoughtbot.com/blog/mystery-guest

In some ways, this sort of code (and also probert and curtin) are hard to write totally transparent unit tests for, because while we _could_ write a helper function that constructed a machine config with appropriate fields and then test the server responded appropriately, at the end of the day we really care about how the server responds to an actual probert report of a bitlocker partition, not this codes understanding of what such a report looks like.

So I guess what I think this test should do is assert that there is something about the response that indicates bitlocker is in use, but not assert that it's the 2nd partition of the 0th disk, or whatever. Does that make sense?

mwhudson (commented):
But wait there's more! Mostly unsolicited opinions on how to write tests.

Makefile
dbungert: Agree, fixed.

subiquity/common/types.py
dbungert: It is different, actually.  Per J-P, it's easiest for the client if we pass around full Partition structure and not just the few elements that are relevant to an API call.  I'm using None as the 'just use the current value' sentinel for edit_partition, and a 'do what you think is best' indicator for add_partition.

subiquity/common/types.py
dbungert: I'm going to change the Disk type declaration back to what it is on main, as there is no scenario where the client sends a json Disk to the server.  Otherwise I agree and would have adopted that suggestion.

subiquity/server/controllers/filesystem.py
dbungert: Can't come up with a reason, removing the async.

subiquity/tests/api/test_api.py
dbungert: Agree

subiquity/server/controllers/filesystem.py
```diff
@@ -292,6 +297,97 @@ async def has_bitlocker_GET(self) -> List[Disk]:
...
+        raise ValueError(f'Partition {number} on {disk.id} not found')
+
+    async def v2_GET(self) -> StorageResponseV2:
+        disks = self.model._all(type='disk')
+        need_root = not self.model.is_root_mounted()
+        need_boot = self.model.needs_bootloader_partition()
+        return StorageResponseV2(
+                disks=[labels.for_client(d) for d in disks],
+                todos_needs_root_mounted=need_root,
+                todos_needs_bootloader_partition=need_boot,
```
jpnurmi, line 313: I'd be fine with `need_root` and `need_boot` in `StorageResponseV2` too, to avoid so long names. dartfmt insists wrapping at 80 chars, and `todos_needs_bootloader_partition` is almost half of it.

subiquity/tests/api/test_api.py
dbungert: That particular API returns a (possibly empty) list of disks that are known to have 1 or more Bitlockered partitions on them.  I've adjusted it to just do a simple len check.  
I'll give the links some thought in the coming days.

PR #1065, opened 2021-09-28
Serializer: clarify default / optional

dbungert:
Optional means a field can be none.
A field having a default means we don't have to supply it.
Explicitly reject non-real fields.

### DIFF
--- a/subiquity/common/serialize.py
+++ b/subiquity/common/serialize.py
@@ -192,9 +192,10 @@ class Serializer:
             return annotation(*args)
         else:
             args = {}
-            for field in attr.fields(annotation):
+            fields = {field.name: field for field in attr.fields(annotation)}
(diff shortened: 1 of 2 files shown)


subiquity/common/serialize.py
```diff
@@ -192,9 +192,10 @@ def _deserialize_attr(self, annotation, value, metadata, path):
             return annotation(*args)
         else:
             args = {}
-            for field in attr.fields(annotation):
+            fields = {field.name: field for field in attr.fields(annotation)}
+            for key in value.keys():
```
mwhudson, line 196: I tend to write this sort of thing as "for k, v in value.items()" but well, super minor.

mwhudson (approved):
Thanks for spotting my thinko here and adding tests!

PR #1066, opened 2021-09-28
DEENG-38/DEENG-39 - Moving to WSL Configuration and remove old command call

patrick330602:
Completes: 
1. Move all “Commits/Write to file” in the configure controller.
2. Remove ubuntu wsl integration calls and do directly ini file writing in the configure controller.


system_setup/common/conf.py
```diff
@@ -0,0 +1,79 @@
...
+
+import logging
+from configparser import ConfigParser
+
+log = logging.getLogger("system_setup.common.conf")
+
+
+class WSLConfig:
+    def __init__(self, inst_type,  conf_file):
+        self.inst_type = inst_type
```
didrocks: For instance, this is never used, no need to store it

system_setup/common/conf.py
```diff
@@ -0,0 +1,79 @@
...
+        self.config.BasicInterpolcation = None
+        self.config.read(conf_file)
+
+    def update(self, config_section, config_setting, config_value):
+        self.config[config_section][config_setting] = config_value
+        with open(self.conf_file, 'w') as configfile:
+            self.config.write(configfile)
+
+
+class WSLConfigHandler:
```
didrocks: This is neat, but I think you are navigating between generic and non generic code here.

For instance, you want to be generic with a general handler which handles both "wsl" and "ubuntu" configuration. However, the code below
```
        if type_input == "ubuntu":
            return self.ubuntu_conf
        elif type_input == "wsl":
            return self.wsl_conf
        else:
            raise ValueError("Invalid config name. Please check again.")
```
Show that you are indeed restricted to a single set of config.

To avoid too much code (like _select_config, _update) and subsequent file updates (you are updating here each file multiple times in a row), I suggest turning WSLConfigHandler in a serializer and deserializer for wslconf_base and wslconf_ad. It doesn’t need to have any instance.
This means:
- This class has an update() method that takes a `wslconf_base` and `wslconf_ad` parameters directly.
- Then, it splits this up in section and updates both file, each one a single time. (this removes the need for `_select_config` and `_update` as this is the class which knows which property goes to which file. It will simplify also udpate.
dest = "" with a switch statement where all method properties are listed, which then assign to a particular dest.
- Then, this will update in one shot, one for each files, anything loaded on the model.

This should simplify this class in term of metacode, and ensure that only one place will know which keys goes where.

Then, in a separate PR or in the same if you want, when initialize both model, you use this class to load() and return both models.

There are things to think about if a key is the default model value, remove it from the file (don’t write unecessary key. This is what happens with this code)

system_setup/server/controllers/configure.py
```diff
@@ -69,13 +66,88 @@ async def configure(self, *, context):
...
+            dryrun = self.app.opts.dry_run
+            variant = self.app.variant
+            config = WSLConfigHandler(dryrun)
+            if variant == "wsl_setup":
+                wsl_identity = self.model.identity.user
+                if dryrun:
+                    log.debug("mimicking creating user %s",
+                              wsl_identity.username)
+                else:
+                    run_command(["/usr/sbin/useradd", "-m", "-s", "/bin/bash",
```
didrocks, line 78: What happens when the command fails? Can we have something reported to the user os it silent here?

system_setup/server/controllers/configure.py
```diff
@@ -69,13 +66,88 @@ async def configure(self, *, context):
...
+                                 "-p", wsl_identity.password,
+                                 wsl_identity.username])
+                    run_command(["/usr/sbin/usermod", "-a",
+                                 "-c", wsl_identity.realname,
+                                 "-G", self.get_userandgroups(),
+                                 wsl_identity.username])
+            else:
+                wslconf_ad = self.model.wslconfadvanced.wslconfadvanced
+                config.update("WSL.automount.enabled",
```
didrocks, line 87: See my other comment, way too many file updates here, which can be fixed with the requested refactoring (with only one config.update call, passing both models)

system_setup/server/controllers/configure.py
```diff
@@ -69,13 +66,88 @@ async def configure(self, *, context):
...
+
             self.app.update_state(ApplicationState.DONE)
         except Exception:
             kw = {}
             self.app.make_apport_report(
                 ErrorReportKind.INSTALL_FAIL, "configuration failed", **kw)
             raise
 
+    def get_userandgroups(self):
+        usergroups_path = '/usr/share/subiquity/users-and-groups'
```
didrocks, line 135: hardcoding this path is not correct:
- it doesn’t work when running from git, as it doesn’t take the current branch file
- can’t we extract and factor out code from `users-and-groups`?

didrocks (changes requested):
I think we are on the right path. We should refactor and simplify conf.py (and rename the file wslconf.py) as suggested. Happy to clarify anything that is needed!

Also, please add docstring for any methods you add.

system_setup/server/controllers/configure.py
patrick330602: > it doesn’t work when running from git, as it doesn’t take the current branch file

Not sure I get this question correctly, but it do have a function `resource_path('users-and-groups')`; but the problem is that is under WSL dyring the real run, `SUBIQUITY_ROOT` won't be set in the real run. technically saying right now the following lines below already did what you want to do:

https://github.com/canonical/subiquity/blob/2ac6f6d8b7de43343fc10d84636cb64e31fec2a4/system_setup/server/controllers/configure.py#L136-L139

> can’t we extract and factor out code from users-and-groups?

what do you mean by this?

system_setup/server/controllers/configure.py
patrick330602: it will report in info.log that the command exit with code 2

system_setup/common/conf.py
patrick330602: gotcha, I will do a mockup today

system_setup/server/controllers/configure.py
didrocks: No problem, we can set (actually, we already do in the packaging we prepared with jibel `SUBIQUITY_ROOT`. I suggest that if it’s not set, you set it on server startup to a directory relative to the current file, that way, this will work when developping locally!

system_setup/server/controllers/configure.py
didrocks: I don’t think this is enough, as it won’t be user visible. From the user perspective: all is good, I haven’t seen any error.

We should surface the error to the user, I bet @mwhudson or @dbungert can guide you and we can grow an API to reflect to the user any errors both in the TUI and in ubuntu-desktop-installer (@jpnurmi, does this already exists?)

system_setup/common/conf.py
patrick330602: Can you checkout my new implementation for this? Thanks

didrocks (changes requested):
Hey! It’s better than the first draft. The class is now following more what I had in mind. However, there are still some issues like the number of times you write the file and it’s possible to make the whole thing a little bit more generic. Have a look at my comments. Thanks!

system_setup/common/conf.py
didrocks: (see new comments below)

system_setup/common/wsl_utils.py
```diff
@@ -59,3 +61,22 @@ def get_windows_locale():
     except OSError as e:
         log.info(windows_locale_failed_msg + e.strerror)
         return None
+
+
+def get_userandgroups():
```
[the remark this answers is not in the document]
patrick330602: I have simplified the code in the latest commit

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,201 @@
...
+            temp_conf_ref = config_base_ref
+            temp_conf_default = config_base_default
+        elif config_class.__name__ == "WSLConfAdvanced":
+            temp_conf_ref = config_adv_ref
+            temp_conf_default = config_adv_default
+        else:
+            raise TypeError("Invalid type name '{}'.".
+                            format(config_class.__name__))
+        if self.is_dry_run:
+            return
```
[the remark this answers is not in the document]
patrick330602: Added

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,201 @@
...
+                    config_realname = config_settings[config_setting]
+                    config_value = config_class.__dict__[config_realname]
+                    if temp_conf_default[config_type][config_section][
+                            config_setting] == config_value:
+                        self._select_config(config_type). \
+                            drop_if_exists(config_section,
+                                           config_setting)
+                    else:
+                        self._select_config(config_type). \
+                            update(config_section,
```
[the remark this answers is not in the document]
patrick330602: This should now be completely fixed.

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,201 @@
...
+    def __init__(self, conf_file):
+        self.conf_file = conf_file
+
+        self.config = ConfigParser()
+        self.config.BasicInterpolcation = None
+        self.config.read(conf_file)
+
+    def drop_if_exists(self, config_section, config_setting):
+        if config_setting in self.config[config_section]:
+            self.config.remove_option(config_section, config_setting)
```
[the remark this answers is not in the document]
patrick330602: removed

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,201 @@
...
+        },
+        "Motd": {
+            "wslnewsenabled": "true"
+        }
+    }
+}
+
+
+def wsl_config_loader(data, pathname, config_ref, id):
+    if path.exists(pathname):
```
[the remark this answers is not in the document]
patrick330602: done for this

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,201 @@
...
+            "advancedipdetection": "false"
+        },
+        "Motd": {
+            "wslnewsenabled": "true"
+        }
+    }
+}
+
+
+def wsl_config_loader(data, pathname, config_ref, id):
```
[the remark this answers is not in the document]
patrick330602: added docstring to all the functions in wsl_conf.py

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,201 @@
...
+import logging
+from configparser import ConfigParser
+from os import path
+
+log = logging.getLogger("system_setup.common.wsl_conf")
+
+config_base_ref = {
+    "wsl": {
+        "automount": {
+            "root": "custom_path",
```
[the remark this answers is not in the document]
patrick330602: updated and should work in the latest commit

subiquity/common/resources.py
```diff
@@ -24,3 +24,26 @@
 
 def resource_path(relative_path):
     return os.path.join(os.environ.get("SUBIQUITY_ROOT", "."), relative_path)
+
+
+def get_users_and_groups(target, root):
```
didrocks, line 29: As this is impacting a common file between subiquity and system_setup, I suggest we split that in another PR (so, keep the duplicated code in this PR, just keep this commit staged) and we will review and get it reviewed by the subiquity team. What do you think?

subiquity/common/resources.py
```diff
@@ -24,3 +24,26 @@
...
+    from subiquitycore.utils import run_command
+
+    users_and_groups_path = resource_path('users-and-groups')
+    if os.path.exists(users_and_groups_path):
+        groups = open(users_and_groups_path).read().split()
+    else:
+        groups = ['admin']
+    groups.append('sudo')
+
+    command = ['chroot', target, 'getent', 'group']
```
didrocks, line 40: For instance, here, this is something that is not done on subiquity side if I’m correct, so worth a discussion with upstream subiquity. Let’s keep it aside for now.

subiquity/common/types.py
```diff
@@ -356,21 +356,21 @@ class ShutdownMode(enum.Enum):
 
 @attr.s(auto_attribs=True)
 class WSLConfigurationBase:
-    custom_path: str = attr.ib(default='/mnt/')
-    custom_mount_opt: str = ''
-    gen_host: bool = attr.ib(default=True)
-    gen_resolvconf: bool = attr.ib(default=True)
+    automount_root: str = attr.ib(default='/mnt/')
```
didrocks, line 359: I already mentioned it as a nitpick, but you didn’t answer and keep /mnt/ instead of just /mnt.

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,166 @@
...
+                    # drop the section if it become empty
+                    if config[config_section] == {}:
+                        config.remove_section(config_section)
+                else:
+                    config[config_section][config_setting] = config_value
+
+        with open(conf_file + ".new", 'w') as configfile:
+            config.write(configfile)
+
+        if os.isfile(conf_file):
```
didrocks, line 164: Don’t keep the old file around, if there is a crash between rename() and move(), we would end up with no conffile.

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,166 @@
...
+            os.rename(conf_file, conf_file + ".old")
+        os.move(conf_file + ".new", conf_file)
```
didrocks, line 166: This should be os.rename(), no? Why you didn’t use the same python function both times?

Also., os.move doesn’t exit, how did it work for you and in the tests?

system_setup/ui/views/wslconfadvanced.py
```diff
@@ -107,12 +105,15 @@ def __init__(self, controller, configuration_data):
...
             'gui_theme': configuration_data.gui_theme,
             'gui_followwintheme': configuration_data.gui_followwintheme,
-            'legacy_gui': configuration_data.legacy_gui,
-            'legacy_audio': configuration_data.legacy_audio,
-            'adv_ip_detect': configuration_data.adv_ip_detect,
-            'wsl_motd_news': configuration_data.wsl_motd_news,
-            'automount': configuration_data.automount,
-            'mountfstab': configuration_data.mountfstab,
+            'interop_guiintegration': configuration_data
+            .interop_guiintegration,
```
didrocks, line 109: Is this syntax excepted? This is not common python, could we write it with an escape line rather? (same for all others)

didrocks (changes requested):
CI still don’t pass (bionic pass, but WSL is skipped there). Once more, please look at the CI and ensure that the tests pass first before requesting a review.

The code can’t run (see os.move comment below), was this tested by any chance?

system_setup/common/wsl_conf.py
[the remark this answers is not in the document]
didrocks: See comment on new diff

subiquity/common/resources.py
patrick330602: Can you elaborate more on this one, like, should I drop the commit from this PR and start another PR for this fix?

subiquity/common/types.py
patrick330602: The default value for this option by Microsoft is `/mnt/` though

image

subiquity/common/resources.py
patrick330602: This is done in the subiquity side though. It's originally the function of get_target_group() that I merged for modulizing

system_setup/common/wsl_conf.py
patrick330602: this should now be fixed

system_setup/common/wsl_conf.py
patrick330602: This is my mistake here, sorry. This should now be fixed

system_setup/ui/views/wslconfadvanced.py
patrick330602: I have now updated it to a better and more unified style. Also this is one of the three acceptable syntax by flake8

subiquity/common/resources.py
patrick330602: Anyway, I have dropped the commit and apply the function directly in `get_userandgroups()`

didrocks: Yeah, let’s make that another PR once this is merged. Thanks!

subiquity/common/resources.py
didrocks: ack, let’s discuss that in the other PR.

subiquity/common/types.py
didrocks: Ack, we should discuss that with Microsoft, as it’s a nitpick but this isn’t the best practice to put in fstab for instance.

system_setup/ui/views/wslconfadvanced.py
didrocks: Can’t we have something like:
```
'interop_guiintegration': \
     configuration_data.interop_guiintegration,
```
Which would be more aligned with the style used in subiquity?

system_setup/common/wsl_conf.py
didrocks: Ack, I think this meant you didn’t test your changes. Please be more careful next time, local test should pass and automated ones as well. Here none of them passed.

system_setup/client/controllers/wslconfadvanced.py
```diff
@@ -33,21 +33,24 @@ async def make_ui(self):
...
                ['interop_enabled', 'interop_appendwindowspath',
-                'gui_theme', 'gui_followwintheme', 'legacy_gui',
-                'legacy_audio', 'adv_ip_detect',
-                'wsl_motd_news', 'automount', 'mountfstab']):
+                'gui_theme', 'gui_followwintheme', 'interop_guiintegration',
+                'interop_audiointegration', 'interop_advancedipdetection',
+                'motd_wslnewsenabled', 'automount_enabled',
+                'automount_mountfstab']):
             reconfiguration = WSLConfigurationAdvanced(
                 interop_enabled=self.answers['interop_enabled'],
```
didrocks, line 41: Is there a way to make all those assignements, now that there is a one to one mapping easier?
using a **kwargs and passing the dict or something along that idea?
(basically all those lines would become: `reconfiguration = WSLConfigurationAdvanced(self.asnwers)` on the caller side

system_setup/client/controllers/wslconfbase.py
```diff
@@ -16,13 +16,14 @@ async def make_ui(self):
 
     def run_answers(self):
         if all(elem in self.answers for elem in
-               ['custom_path', 'custom_mount_opt',
-                'gen_host', 'gen_resolvconf']):
+               ['automount_root', 'automount_options',
+                'network_generatehosts', 'network_generateresolvconf']):
             configuration = WSLConfigurationBase(
```
didrocks, line 21: ditto, see previous comment.

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # update the config file
+    for config_type in temp_conf_default:
+        config_sections = temp_conf_default[config_type]
+
+        config = ConfigParser()
+        config.BasicInterpolcation = None
+
+        if config_type == "wsl":
+            conf_file = "/etc/wsl.conf"
```
didrocks, line 136: There is some repetiution on the file path now between `default_loader` and `wsl_config_update`. Put it as a constant in the file.

system_setup/common/wsl_conf.py
```diff
@@ -0,0 +1,164 @@
...
+            conf_sec_list = config[conf_sec]
+            for conf_item in conf_sec_list:
+                if conf_item in config_ref[id][conf_sec]:
+                    data[conf_sec.lower()
+                         + "_" + conf_item.lower()] = \
+                             conf_sec_list[conf_item]
+    return data
+
+
+def default_loader(is_advanced):
```
didrocks, line 91: Suggestion: make it a key paramter with default value. That way, the calls would become:
`default_loader(is_advanced=True)` which is more explicit than `default_loader(True)`

system_setup/models/wslconfadvanced.py
```diff
@@ -42,67 +39,22 @@ class WSLConfigurationAdvancedModel(object):
...
     def __init__(self):
         self._wslconfadvanced = None
-        # TODO WSL: Load settings from system
 
-    def apply_settings(self, result, is_dry_run=False):
+    def apply_settings(self, result):
         d = {}
         # TODO: placholder settings; should be dynamically assgined using
         # ubuntu-wsl-integration
         d['interop_enabled'] = result.interop_enabled
```
didrocks, line 47: Same here, can we try passing dynamically result to WSLConfigurationAdvanced() directly? This avoid stuttering

system_setup/server/controllers/configure.py
```diff
@@ -69,6 +66,37 @@ async def configure(self, *, context):
...
+                              wsl_id.username)
+                else:
+                    create_user_act = \
+                        run_command(["/usr/sbin/useradd", "-m", "-s",
+                                     "/bin/bash", "-p",
+                                     wsl_id.password,
+                                     wsl_id.username])
+                    if create_user_act.returncode != 0:
+                        raise Exception("Failed to create user %s"
+                                        % wsl_id.username)
```
didrocks, line 84: Can you add stderr here?

didrocks (changes requested):
Still some changes needed, but there is progress :) Look in particular at removing the duplication of serialization/deserialization and so on…

system_setup/ui/views/wslconfadvanced.py
patrick330602: Unfortunately, that one is the only one I am not able to change since all of the changes violate rules in `flake8`:
image
image
image
image

didrocks: Indeed, we can use the noqa pragma, but anyway, all those use case will be moot if you can remove the stuttering with the dict

system_setup/client/controllers/wslconfadvanced.py
didrocks: Just tested and you can just do:
```
reconfiguration = WSLConfigurationAdvanced(**self.answers)
```

system_setup/ui/views/wslconfadvanced.py
patrick330602: Gotcha

system_setup/server/controllers/configure.py
patrick330602: gotcha

system_setup/models/wslconfadvanced.py
patrick330602: done

system_setup/common/wsl_conf.py
patrick330602: done

system_setup/common/wsl_conf.py
patrick330602: done

system_setup/client/controllers/wslconfadvanced.py
patrick330602: done

system_setup/ui/views/wslconfadvanced.py
```diff
@@ -128,15 +135,24 @@ def __init__(self, controller, configuration_data):
 
     def done(self, result):
         self.controller.done(WSLConfigurationAdvanced(
```
didrocks, line 137: I think the stuttering here can be fixed by building a dict from the self.form values. I suggest you do that in another PR, as this is not related to this one.

didrocks (changes requested):
Ok, apart from my stuttering enhancement which can be done in a separate PR, the requested changes are done.

However, the CI is down and I did run the test both in lxc and locally and there is an error:
```
subiquity/Configure/configure: final system configuration
subiquity/Configure/configure: Invalid type name.
  subiquity/ErrorReporter/1633446579.507226706.install_fail/add_info
  subiquity/ErrorReporter/1633446579.507226706.install_fail/add_info: written to .subiquity/var/crash/1633446579.507226706.install_fail.crash
```

Did you run the tests? Is it passing for you?

system_setup/ui/views/wslconfadvanced.py
patrick330602: Gotcha

patrick330602:
> Ok, apart from my stuttering enhancement which can be done in a separate PR, the requested changes are done.
> 
> However, the CI is down and I did run the test both in lxc and locally and there is an error:
> 
> ```
> subiquity/Configure/configure: final system configuration
> subiquity/Configure/configure: Invalid type name.
>   subiquity/ErrorReporter/1633446579.507226706.install_fail/add_info
>   subiquity/ErrorReporter/1633446579.507226706.install_fail/add_info: written to .subiquity/var/crash/1633446579.507226706.install_fail.crash
> ```
> 
> Did you run the tests? Is it passing for you?

@didrocks yes, both on lxc and on real wsl machine (focal and impish), and both passed (real machine also tested --autoinstall and --answers), no crash log for me, I am not sure what happened....

didrocks:
How are you running the automated tests in lxc/on your machine?

dbungert:
To simulate what the github action does, you should do something like this:
./scripts/test-in-lxd.sh ubuntu-daily:focal "make check"

didrocks:
> To simulate what the github action does, you should do something like this:
./scripts/test-in-lxd.sh ubuntu-daily:focal "make check"

Correct, this is what I run, after looking at the workflow yaml file

patrick330602:
It seems that impish CI is not working, tested in the local impish machine and it is working.

PR #1068, opened 2021-09-29
fixes around logical volume size

mwhudson:
LVM logical volumes can only be multiples of the VG's extent size, which is 4MiB by default.

https://bugs.launchpad.net/subiquity/+bug/1935969

### DIFF
--- a/subiquity/ui/views/filesystem/partition.py
+++ b/subiquity/ui/views/filesystem/partition.py
@@ -57,6 +57,9 @@ from subiquity.ui.mount import (
 log = logging.getLogger('subiquity.ui.filesystem.add_partition')
 
 
+DEFAULT_ALIGNMENT = 1 << 20
+
+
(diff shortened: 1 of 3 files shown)


subiquity/ui/views/filesystem/partition.py
```diff
@@ -567,7 +572,8 @@ def __init__(self, parent, device):
             initial.update(initial_data_for_fs(fs))
         elif not isinstance(device, Disk):
             initial['fstype'] = 'ext4'
-        self.form = PartitionForm(self.model, 0, initial, None, device)
+        self.form = PartitionForm(
+            self.model, 0, initial, None, device, 1 << 20)
```
dbungert, line 576: A constant would be nice, to help ensure it doesn't skew from the value at line 384.

dbungert (approved):

PR #1069, opened 2021-09-29
only do x-override-status: confirm games for first request for a screen

mwhudson:
this uses a rather unsatisfactory way of identifying the first request for a screen.

for https://bugs.launchpad.net/subiquity/+bug/1939176


mwhudson:
I haven't tested this in a vm yet

mwhudson:
(it does seem to fix the bug)

subiquity/client/client.py
```diff
@@ -136,6 +136,8 @@ def __init__(self, opts):
 
         self.conn = aiohttp.UnixConnector(self.opts.socket)
         self.client = make_client_for_conn(API, self.conn, self.resp_hook)
+        self.client1 = make_client_for_conn(
```
dbungert, line 139: Will ubuntu-desktop-installer need to be able to hit client1?  

At some point it will want to support autoinstall, this is part of how interactive-sections works, right?

mwhudson: Yeah this is all around interactive-sections really: basically the client doesn't know when the installer is ready to start curtin so there is a bunch of hair to tell it. Currently udi ignores all this though and just calls .confirm() after submitting the storage config...

dbungert (approved):

mwhudson:
Pushed a much nicer version (that depends on python 3.7 or better, so only works now we're based on core20). I'm not desperate to get this into impish, although it would be nice.

PR #1070, opened 2021-09-29
improve locale support

mwhudson:
This branch does two main things:

1. it installs language support packages in desktop installs
2. it does not configure a locale for a minimal server install


mwhudson:
I haven't tested this in a VM yet...

dbungert (approved):

PR #1072, opened 2021-10-01
snapcraft: core20 base

dbungert:


dbungert:
@mwhudson - you have dissuaded me in the past from adding new wrapper scripts for subiquity-server, but it seems to be required this time.  Passing PYTHONPATH to get the site-packages seems required unless we convert over curtin right now to operate properly with dist-packages, and passing the PYTHONPATH thru the snapcraft environment doesn't seem to be working as expected, but works fine when I set it on the wrapper script.

mwhudson (approved):
As you mentioned, this isn't complete in that we should get curtin deps from the archive too (and possibly the python plugin isn't useful to us when we get that done?) but this works and is definitely a step in the right direction!

PR #1073, opened 2021-10-04
DEENG-41 - Update TUI command line name and remove unneeded options.

patrick330602:
This PR tries to update TUI command line name and remove unneeded options.

The first commit is confirmed not useful one; 
The second commit is the one I am not sure that should be removed; they could be used in debugging.

### DIFF
--- a/system_setup/cmd/tui.py
+++ b/system_setup/cmd/tui.py
@@ -48,28 +48,7 @@ def make_client_args_parser():
     parser.add_argument('--dry-run', action='store_true',
                         dest='dry_run',
                         help='menu-only, do not call installer function')
-    # TODO WSL: remove any uneeded arguments
     parser.add_argument('--socket')
-    parser.add_argument('--serial', action='store_true',
-                        dest='run_on_serial',
-                        help='Run the installer over serial console.')
-    parser.add_argument('--ssh', action='store_true',
-                        dest='ssh',
-                        help='Print ssh login details')
-    parser.add_argument('--ascii', action='store_true',
-                        dest='ascii',
-                        help='Run the installer in ascii mode.')
-    parser.add_argument('--unicode', action='store_false',
-                        dest='ascii',
-                        help='Run the installer in unicode mode.')
-    parser.add_argument('--screens', action='append', dest='screens',
-                        default=[])
-    parser.add_argument('--script', metavar="SCRIPT", action='append',
-                        dest='scripts', default=[],
-                        help=('Execute SCRIPT in a namespace containing view '
-                              'helpers and "ui"'))
-    parser.add_argument('--click', metavar="PAT", action=ClickAction,
-                        help='Synthesize a click on a button matching PAT')
     parser.add_argument('--answers')
     parser.add_argument('--server-pid')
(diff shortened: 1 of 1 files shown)


system_setup/cmd/tui.py
```diff
@@ -48,28 +48,16 @@ def make_client_args_parser():
     parser.add_argument('--dry-run', action='store_true',
                         dest='dry_run',
                         help='menu-only, do not call installer function')
-    # TODO WSL: remove any uneeded arguments
     parser.add_argument('--socket')
     parser.add_argument('--serial', action='store_true',
```
didrocks, line 52: Can you give me some examples on how --serial is used under WSL? (Like a pragmatical step by step guide). I think if this work, we should document and see how this can be useful for users.

system_setup/cmd/tui.py
```diff
@@ -48,28 +48,16 @@ def make_client_args_parser():
...
                         dest='run_on_serial',
                         help='Run the installer over serial console.')
-    parser.add_argument('--ssh', action='store_true',
-                        dest='ssh',
-                        help='Print ssh login details')
     parser.add_argument('--ascii', action='store_true',
```
didrocks, line 55: Same here. (and hilight the benefit/where ascii mode would be required for WSL users)

system_setup/cmd/tui.py
```diff
@@ -48,28 +48,16 @@ def make_client_args_parser():
...
                         dest='ascii',
                         help='Run the installer in ascii mode.')
     parser.add_argument('--unicode', action='store_false',
```
didrocks, line 58: Same here. (and hilight the benefit/where ascii mode would be required for WSL users)

system_setup/cmd/tui.py
```diff
@@ -133,6 +121,10 @@ def main():
     logger.info("Starting System Setup revision {}".format(version))
     logger.info("Arguments passed: {}".format(sys.argv))
 
+    opts.ssh = False
+    opts.screens = None
+    opts.scripts = None
```
didrocks, line 126: Default for ssh is False.
Default for screens and scripts are []. Shouldn’t we use those default values rather than relying on the "falsy" behavior of python?

didrocks (changes requested):

system_setup/cmd/tui.py
patrick330602: That's true...

patrick330602:
After some testing. I am not able to give good examples for `--serial`, `--ascii` and `--unicode` so I am removing all three of them

didrocks (changes requested):
The CI doesn’t pass. This is due to missing arguments in opts.
As you removed and reset the values like `opts.screens = []`, `run_on_serial` and additional parameters should have the same treatment.

patrick330602:
@didrocks This should now be fixed, there is some small issue that came from main too besides my problem

didrocks:
Looks good, test pass despite the lack of initializing opt.ascii/unicode. Merging.

didrocks (approved):

PR #1078, opened 2021-10-05
DEENG-83 - Implement a reboot in shutdown to restart WSL instances to take new configuration in effect.

patrick330602:
This completes "Implement a reboot in shutdown to restart WSL instances to take new configuration in effect."


system_setup/common/wsl_utils.py
```diff
@@ -92,3 +92,33 @@ def get_windows_locale():
     except OSError as e:
         log.info(windows_locale_failed_msg + e.strerror)
         return None
+
+
+def perform_wsl_shutdown(is_restart=False):
+    exec_wsl_map = {
```
didrocks, line 98: I don’t think we should hardcode the exe name here. This is not maintainable. It means that at each new releases, we need to add this element here, rerelease the oobe and so on. Can’t we get from the launcher side the current exe, with current arguments. and reexec this?

I think you tested it, but there is no chance that terminating the process prevents the second part (after the `;` to reexecute?)

Would `wsl.exec -d` bypass the launcher?

didrocks (changes requested):

system_setup/common/wsl_utils.py
patrick330602: > I don’t think we should hardcode the exe name here.

that would be true... I will modify this part

> I think you tested it, but there is no chance that terminating the process prevents the second part (after the ; to reexecute?)
>
> Would wsl.exec -d bypass the launcher?

What do you mean by these two questions? I am not sure I get what you mean... @didrocks

didrocks: > What do you mean by these two questions? I am not sure I get what you mean... @didrocks

* Is there a chance of a race between terminating the WSL instance (so this code) and the restart?
* wsl.exe -d "distro" is still using our launcher, correct?

patrick330602: > Is there a chance of a race between terminating the WSL instance (so this code) and the restart?

No it wont, `;` separates the command and it will always execute first line (here is the `wsl --shutdown`) and then perform the restart

> wsl.exe -d "distro" is still using our launcher, correct?

No, I put this for fallback since not all ubuntu installations on wsl will have the executables; also I just forget the add the logic for the ones that do not have the 'WSL_DISTRO_NAME', those should be fallback directly to `wsl.exe`._

system_setup/common/wsl_utils.py
```diff
@@ -92,3 +94,74 @@ def get_windows_locale():
...
+                    " -ArgumentList '-NoProfile','-NonInteractive',"
+                    "'-Command','wsl.exe','--shutdown',';'")
+    if is_restart:
+        if os.environ.get('WT_PROFILE_ID') is not None:
+            shutdown_cmd = shutdown_cmd + ",'wt.exe'"
+
+        exec_name = ",'wsl.exe'"
+        if os.environ.get('WSL_DISTRO_NAME') is not None:
+            wsl_distro_name = os.environ.get('WSL_DISTRO_NAME')
+            exec_name = exec_name + ",'-d','" + wsl_distro_name + "'"
```
didrocks, line 158: Can you format a string with .format()? That would make things easier to read (same for other string management below)

didrocks (changes requested):
I have a small comment, but otherwise, seems good. However, there is still this part that needs to be discussed:

* How does this work when windows interoperability is disabled? `.exe` file can’t be executed AFAIK.

patrick330602:
> How does this work when windows interoperability is disabled? .exe file can’t be executed AFAIK.

We won't be able to, there is no way we can do as we do not have access to the windows exe currently.

didrocks:
So, this is an issue:
- there is no clear error handling telling that you can’t reboot, it will try… and fail.
- I think we can have another approach, from the launcher itself. WDYT? (A flag that is set somewhere, and that it’s caught by the launcher when the oobe executable. This has to be designed though for the reconfigure mode)

patrick330602:
So I have some update for this one: 
1. set up a file in `/run/subiquity/launcher-state` and read the state from the function, if the command failed, they will pass a status telling launcher should restart or shutdown in the file 
2. set up the user in `/etc/wsl.conf` file so that we do not have to set up the user in the launcher and after executing the command, we can restart/shutdown the ubuntu console directly without passing username in launcher.

didrocks:
Great! Changes are ok, tests pass, let’s merge it.

Do you mind adding an item with a followup PR where you validate the output in dry-run? (there there is a file containing "reboot" and then shutdown/complete when we will have that UI)

PR #1079, opened 2021-10-05
WSL OOBE - modulize the process of getting user groups

patrick330602:
This PR proposes to modulize the process of getting user groups since the WSL OOBE also uses a similar process.


didrocks:
Latest version sounds good to me. Let’s merge it.

PR #1080, opened 2021-10-05
workflows: skip impish

dbungert:
Impish systemd-resolved seems unhappy and it's interfering with the
other test runs.


didrocks:
Everything but impish pass, as expected. The rest of the matrix is now not cancelled and passed, so we can have a look and not ignore CI. Thanks! Ack.

PR #1094, opened 2021-10-12
close the session object after each request to the snapd API

mwhudson:
As reported in https://bugs.launchpad.net/snapd/+bug/1946656,
not-quite idle connections between subiquity and snapd are piling up and
causing issues when trying to restart subiquity. I don't really
understand why, but using a fresh session for each API access makes the
problem go away.


paride:
I built a snap from this branch, injected it into an Impish ISO and I can confirm the fix works.

PR #1100, opened 2021-10-15
WSL OOBE: add `useradd` tests

patrick330602:


didrocks:
I would have prefered the mock/fake path as we suggested, especially as we have a lot of mocking code (not prepared in advanced) in the dry-run path.

However, as dry-run should not be limited to tests, I can see complications with the mock way, so I’m happy to merge it.

PR #1105, opened 2021-10-28
Make subiquity server supporting prefill install files (for WSL System Setup)

CarlosNihelton:
Hi @didrocks and @dbungert! I mentioned you both because of our discussion on #1104 in which I promised another PR with the '--prefill' CLI option added. Feel free to request approval from anybody else if it makes more sense.

This PR stays on domain of WSL `system_setup`. The oly changes outside of that folder is the addition of a few YAML files in the `examples` folder to simulate the pieces of information that WSL launcher may provide the OOBE with.

Thanks to the changes in #1104 we don't need the `system_setup/server/controllers/locale.py` anymore. Also removed one function in wsl_utils which called Powershell on Windows, which couldn't work under certain situations outside of our control.

With the modifications here proposed being accepted and merged, we will complete the cycle to make the experience depicted below. Although demonstrated with Flutter GUI, it works the same way for the TUI.

GIF showing Flutter OOBE with prefill


system_setup/models/system_setup.py
```diff
@@ -61,6 +62,23 @@ def __init__(self, root, hub, install_model_names,
         self.wslconfbase = WSLConfigurationBaseModel()
         self.wslconfadvanced = WSLConfigurationAdvancedModel()
 
+        if prefillInfo:
```
dbungert, line 65: If the controller is handling this we don't also need it in the model, right?

system_setup/models/system_setup.py
```diff
@@ -61,6 +62,23 @@ def __init__(self, root, hub, install_model_names,
...
+            welcome = prefillInfo.get('Welcome', {'lang': None})
+            if welcome is not None and welcome.get('lang') is not None:
+                self.locale.selected_language = welcome['lang']
+                log.debug('Prefill Language: {}'
+                          .format(self.locale.selected_language))
+
+            identity = prefillInfo.get('WSLIdentity', None)
+            if identity:
+                idata = IdentityData()
```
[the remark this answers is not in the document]
dbungert: I think the prefill concept is interesting but I don't have a specific use case for it right now.
Answers is working well for test automation, and autoinstall with or without the interactive-section is a good fit for install automation.
I think if we want prefill we should rework it in that direction at that time.

dbungert (changes requested):

system_setup/models/system_setup.py
CarlosNihelton: That was embarrassing... 😳
That piece of code was supposed to be removed. My fault while experimenting between local branches. Ended up leaving some dead code.

system_setup/server/server.py
```diff
@@ -60,13 +64,32 @@ def __init__(self, opts, block_log_dir):
         self.log_syslog_id = ""
         if is_reconfigure(opts.dry_run):
             self.set_source_variant("wsl_configuration")
+        if self.opts.prefill:
+            with open(self.opts.prefill, 'r') as stream:
```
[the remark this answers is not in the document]
CarlosNihelton: I did a git mistake. The reader logic was supposed to be only in the `SystemSetupServer __init__()` method.

system_setup/cmd/server.py
```diff
@@ -74,6 +79,21 @@ def main():
     logger.info("Starting System Setup server revision {}".format(version))
     logger.info("Arguments passed: {}".format(sys.argv))
 
+    prefillFile = opts.prefill
+    if prefillFile:
+        if os.path.exists(prefillFile):
+            statInfo = os.stat(prefillFile)
+            mode = statInfo.st_mode
+            isRegularFile = (stat.S_ISREG(mode) != 0)
+            if not isRegularFile:
```
dbungert, line 88: https://docs.python.org/3/library/os.path.html#os.path.isfile maybe?

system_setup/cmd/server.py
```diff
@@ -74,6 +79,21 @@ def main():
     logger.info("Starting System Setup server revision {}".format(version))
     logger.info("Arguments passed: {}".format(sys.argv))
 
+    prefillFile = opts.prefill
+    if prefillFile:
```
dbungert, line 83: I'm not going to hold the merge for this as it's more a matter of preference, but is there any reason to scan for the file here and actually parse it later?  This could be in a catch block in system_setup/server/server.py if you want this info for diagnostic purposes.

dbungert (approved):

system_setup/cmd/server.py
CarlosNihelton: Matter of preference, as you said. My reasoning was delaying parsing to the closest point of usage, but doing server side sanity checks the earliest possible to avoid trying to parse an invalid file. Maybe my checks were too shallow, but that was the intent.

system_setup/cmd/server.py
CarlosNihelton: Yeah, it also works fine. Thanks for the advice.

dbungert (approved):

PR #1107, opened 2021-10-29
snapcraft: ssh-import-id lsb_release from live

dbungert:
ssh-import-id in the snap has python loading problems due to trying to
load stuff from the live system instead.  If you mangle that hard enough
your reward is the same result for lsb_release.

In draft until I test this to my liking.  I also now mistrust every single external program we interact with and want to ensure we have tested all such paths.


dbungert:
yea, no ssh-import-id in the bionic live image so I need a new plan

PR #1111, opened 2021-11-03
system_setup: Many fixes for ci

didrocks:
Avoid a crash due to in progress state when summary is displayed.
Fixes on non systemd machines on output printing.
Use correct path in dry-run for socket.

This is part 2 (over 3) on fixing `make integration`.


jibel:
LGTM. Worked and tested together.

jibel (approved):

PR #1112, opened 2021-11-03
Don’t block on running commands in configure

didrocks:
Those commands needs to be async to be able to still answer GET/POST
request while configure is running.
This allows showing up the summary page immediately.

Co-authored-by: Jean-Baptiste Lallement <jean-baptiste@ubuntu.com>


jibel (approved):
Tested on WSL.

PR #1113, opened 2021-11-03
Fix reruning in production system_setup on WSL

didrocks:
On system_setup, we always run the subiquity server. On WSL, ensure we
wipe the server-state file (no systemd fallback) before starting the
server.

Co-authored-by: Jean-Baptiste Lallement <jean-baptiste@ubuntu.com>


jibel (approved):
Tested in WSL.

PR #1116, opened 2021-11-04
System locale manipulation wired into Subiquity.

CarlosNihelton:
Since WSL cannot count on Cloud-Init, Subiquity System Setup Locale controller needs to take all necessary steps to install the locale selected by the user.
This PR achieves that goal by manipulating `/etc/locale.gen` file, running `locale-gen` command and installing language packs recommended by `check-language-support` script.

### DIFF
--- a/system_setup/server/controllers/configure.py
+++ b/system_setup/server/controllers/configure.py
@@ -145,11 +145,15 @@ class ConfigureController(SubiquityController):
                                      apt_pkg.config
                                      .find_dir("Dir::Cache::Archives")[1:])
             os.makedirs(packs_dir, exist_ok=True)
+            try:
             for package in packages:
-                # Downloading instead of installing. Doesn't require root.
+                    # Just write the package uri to a file instead of installing.
                 archive = os.path.join(packs_dir, cache[package].fullname)
-                cmd = ["wget", "-O", archive, cache[package].candidate.uri]
-                await arun_command(cmd, env=env)
+                    with open(archive, "wt") as f:
+                        f.write(cache[package].candidate.uri)
+            except IOError:
+                log.error("Failed to write %s file.", archive)
+                return False        
 
             return True
 
--- a/system_setup/server/controllers/configure.py
+++ b/system_setup/server/controllers/configure.py
@@ -88,13 +88,15 @@ class ConfigureController(SubiquityController):
     def __update_locale_cmd(self, lang) -> List[str]:
         """ Adds mocking cli to update-locale if in dry-run. """
         updateLocCmd = ["update-locale", "LANG={}".format(lang)]
-        if self.app.opts.dry_run:
-            defaultLocDir = os.path.join(self.model.root,
-                                         "etc/default/")
-            os.makedirs(defaultLocDir, exist_ok=True)
-            updateLocCmd += ["--locale-file",
-                             os.path.join(defaultLocDir, "locale"),
-                             "--no-checks"]
+        if not self.app.opts.dry_run:
+            return updateLocCmd
+
+        defaultLocDir = os.path.join(self.model.root,
+                                     "etc/default/")
+        os.makedirs(defaultLocDir, exist_ok=True)
+        updateLocCmd += ["--locale-file",
+                         os.path.join(defaultLocDir, "locale"),
+                         "--no-checks"]
 
         return updateLocCmd
 
@@ -102,7 +104,7 @@ class ConfigureController(SubiquityController):
         """ Last commands to run for locale support. Returns True on success"""
 
         (locGenCmd, ok) = self.__locale_gen_cmd()
-        if ok is False:
+        if not ok:
             log.error("Locale generation failed.")
             return False
 
@@ -216,12 +218,12 @@ class ConfigureController(SubiquityController):
 
         ok = await self._install_check_lang_support_packages(lang, env)
 
-        if ok is False:
+        if not ok:
             log.error("Failed to install recommended language packs.")
             return
 
         ok = await self._activate_locale(lang, env)
-        if ok is False:
+        if not ok:
             log.error("Failed to run locale activation commands.")
             return
 
--- a/system_setup/server/controllers/configure.py
+++ b/system_setup/server/controllers/configure.py
@@ -16,7 +16,7 @@ import os
 import shutil
 import logging
 import re
-from typing import Tuple
+from typing import Tuple, List
 import apt
 import apt_pkg
 
@@ -85,7 +85,7 @@ class ConfigureController(SubiquityController):
             log.error("Failed to modify %s file. %s", cmdFile, err)
             return ("", False)
 
-    def __update_locale_cmd(self, lang) -> list[str]:
+    def __update_locale_cmd(self, lang) -> List[str]:
         """ Adds mocking cli to update-locale if in dry-run. """
         updateLocCmd = ["update-locale", "LANG={}".format(lang)]
         if self.app.opts.dry_run:
--- a/system_setup/server/controllers/configure.py
+++ b/system_setup/server/controllers/configure.py
@@ -53,8 +53,8 @@ class ConfigureController(SubiquityController):
             lineFound = fileContents.find(lang)
             if lineFound == -1:
                 # An unsupported locale coming from our UI is a bug
-                raise AssertionError("Selected language {} not supported."
-                                     .format(lang))
+                log.error("Selected language %s not supported.", lang)
+                return
 
             pattern = r'#+\s*({}.*)'.format(lang)
             commented = re.compile(pattern)
@@ -62,8 +62,11 @@ class ConfigureController(SubiquityController):
             locGenNeedsWrite = n == 1
 
         if locGenNeedsWrite:
-            with open(localeGenPath, "wt") as f:
-                f.write(fileContents)
+            try:
+                with open(localeGenPath, "wt") as f:
+                    f.write(fileContents)
+            except IOError as err:
+                log.error("Failed to modify %s file. %s", localeGenPath, err)
 
     async def _install_check_lang_support_packages(self, lang, env):
         """ Installs any packages recommended by
@@ -90,7 +93,8 @@ class ConfigureController(SubiquityController):
                 packs_dir = os.path.join(self.model.root,
                                          apt_pkg.config
                                          .find_dir("Dir::Cache::Archives")[1:])
-                cmd = ["wget", cache[package].candidate.uri, "-O", packs_dir]
+                archive = os.path.join(packs_dir, cache[package].fullname)
+                cmd = ["wget", "-O", archive, cache[package].candidate.uri]
                 os.makedirs(packs_dir, exist_ok=True)
                 await arun_command(cmd, env=env)
         else:
--- a/system_setup/server/controllers/configure.py
+++ b/system_setup/server/controllers/configure.py
@@ -15,6 +15,9 @@
 import os
 import shutil
 import logging
+import re
+import apt
+import apt_pkg
 
 from subiquity.common.errorreport import ErrorReportKind
 from subiquity.common.types import ApplicationState
@@ -36,6 +39,86 @@ class ConfigureController(SubiquityController):
 
     def start(self):
         self.install_task = self.app.aio_loop.create_task(self.configure())
+
+    def _update_locale_gen(self, localeGenPath, lang):
+        """ Uncomments the line in locale.gen file where lang is found,
+            if found commented. A fully qualified language is expected,
+            since that would have passed thru the Locale
+            controller validation. e.g. en_UK.UTF-8. """
+
+        fileContents: str
+        locGenNeedsWrite = False
+        with open(localeGenPath, "r") as f:
+            fileContents = f.read()
+            lineFound = fileContents.find(lang)
+            if lineFound == -1:
+                # An unsupported locale coming from our UI is a bug
+                raise AssertionError("Selected language {} not supported."
+                                     .format(lang))
+
+            pattern = r'#+\s*({}.*)'.format(lang)
+            commented = re.compile(pattern)
+            (fileContents, n) = commented.subn(r'\1', fileContents, count=1)
+            locGenNeedsWrite = n == 1
+
+        if locGenNeedsWrite:
+            with open(localeGenPath, "wt") as f:
+                f.write(fileContents)
+
+    async def _install_check_lang_support_packages(self, lang, env):
+        """ Installs any packages recommended by
+            check-language-support command. """
+
+        clsCommand = "check-language-support"
+
+        cp = await arun_command([clsCommand, "-l", lang[0:5]], env=env)
+        if cp.returncode != 0:
+            log.error('Command \"%s\" failed with return code %d',
+                      cp.args, cp.returncode)
+            return
+
+        packages = cp.stdout.strip('\n').split(' ')
+        if len(packages) == 0:
+            log.debug("%s didn't recommend any packages. Nothing to do.",
+                      clsCommand)
+            return
+
+        cache = apt.Cache()
+        if self.app.opts.dry_run:
+            for package in packages:
+                # Downloading instead of installing. Doesn't require root.
+                packs_dir = os.path.join(self.model.root,
+                                         apt_pkg.config
+                                         .find_dir("Dir::Cache::Archives")[1:])
+                cmd = ["wget", cache[package].candidate.uri, "-O", packs_dir]
+                os.makedirs(packs_dir, exist_ok=True)
+                await arun_command(cmd, env=env)
+        else:
+            cache.update()
+            cache.open(None)
+            with cache.actiongroup():
+                for package in packages:
+                    cache[package].mark_install()
+                    cache.commit()
+
+    async def _activate_locale(self, lang, env):
+        """ Final commands to run for locale support """
+
+        cmds = [["locale-gen"],
+                ["update-locale", "LANG={}".format(lang)]]
+        if self.app.opts.dry_run:
+            for cmd in cmds:
+                # TODO Figure out about mocks.
+                # It would be good if they offered a -P prefix option,
+                # but they don't.
+                # Chroot'ing is not an option.
+                log.debug('Would run: %s', ' '.join(cmd))
+        else:
+            for cmd in cmds:
+                cp = await arun_command(cmd, env=env)
+                if cp.returncode != 0:
+                    raise SystemError('Command {} failed with return code {}'
+                                      .format(cp.args, cp.returncode))
 
     @with_context(
         description="final system configuration", level="INFO",
@@ -67,6 +150,7 @@ class ConfigureController(SubiquityController):
 
             self.app.update_state(ApplicationState.POST_RUNNING)
 
+            localeGenPath = "/etc/locale.gen"
             dryrun = self.app.opts.dry_run
             variant = self.app.variant
             root_dir = self.model.root
@@ -77,6 +161,7 @@ class ConfigureController(SubiquityController):
                 create_user_base = []
                 assign_grp_base = []
                 usergroups_list = get_users_and_groups()
+                lang = self.model.locale.selected_language
                 if dryrun:
                     log.debug("creating a mock-up env for user %s", username)
                     # creating folders and files for dryrun
@@ -86,6 +171,19 @@ class ConfigureController(SubiquityController):
                     os.makedirs(home_dir, exist_ok=True)
                     pseudo_files = ["passwd", "shadow", "gshadow", "group",
                                     "subgid", "subuid"]
+
+                    # locale
+                    testLocGenPath = os.path.join(etc_dir,
+                                                  os.path
+                                                  .basename(localeGenPath))
+                    shutil.copy(localeGenPath, testLocGenPath)
+                    shutil.copy(localeGenPath,
+                                "{}-".format(testLocGenPath))
+                    # After copies are done, overwrite the original path
+                    # because there is code outside of this 'if'
+                    # depending on 'localeGenPath' being initialized.
+                    localeGenPath = testLocGenPath
+
                     for file in pseudo_files:
                         filepath = os.path.join(etc_dir, file)
                         open(filepath, "a").close()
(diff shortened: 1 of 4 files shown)


CarlosNihelton:
@didrocks , I'd like to have your opinion on this PR before actually submitting it for formal review and merge. I think you'll be able to criticize it towards a higher quality than I'm achieving now. Would you please? 🙄

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+
+    async def _run_locale_support_cmds(self, lang: str):
+        """ Final commands to run when a valid language is POST'ed. """
+        cmds = [["locale-gen"],
+                ["update-locale", "LANG={}".format(lang)]]
+        env = os.environ.copy()
+
+        cp = await arun_command(["check-language-support", "-l", lang[0:5]],
+                                env=env)
+        if cp.returncode == 0:
```
didrocks, line 44: I would rather check != 0 and only return here, to avoid relying on the code below (passing in the else and printing an error)

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+            cmds.append(["apt", "install", "-y"] + packages)
+
+        if self.app.opts.dry_run:
+            for cmd in cmds:
+                log.debug('Would run: ' + ' '.join(cmd))
+        else:
+            for cmd in cmds:
+                cp = await arun_command(cmd, env=env)
+                if cp.returncode:
+                    log.debug('Command \"%s\" failed with return code %d',
```
didrocks, line 55: In the error case, that should rather be .error(), no?

Also, can we return the error to the caller which will handle it?

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+            packages = cp.stdout.strip('\n').split(' ')
+            cmds.append(["apt", "install", "-y"] + packages)
+
+        if self.app.opts.dry_run:
+            for cmd in cmds:
+                log.debug('Would run: ' + ' '.join(cmd))
+        else:
+            for cmd in cmds:
+                cp = await arun_command(cmd, env=env)
+                if cp.returncode:
```
didrocks, line 54: Maybe that’s the Go developer in me, but let’s try to not rely on truthy (!= 0) implicit conversation

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+        """ Final commands to run when a valid language is POST'ed. """
+        cmds = [["locale-gen"],
+                ["update-locale", "LANG={}".format(lang)]]
+        env = os.environ.copy()
+
+        cp = await arun_command(["check-language-support", "-l", lang[0:5]],
+                                env=env)
+        if cp.returncode == 0:
+            packages = cp.stdout.strip('\n').split(' ')
+            cmds.append(["apt", "install", "-y"] + packages)
```
didrocks, line 46: I think we can use python3-apt dep to install packages rather than command line for this

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+
+        if self.app.opts.dry_run:
+            for cmd in cmds:
+                log.debug('Would run: ' + ' '.join(cmd))
```
didrocks, line 50: I think you haven’t figured out mock/test scenarios?

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+        else:
+            for cmd in cmds:
+                cp = await arun_command(cmd, env=env)
+                if cp.returncode:
+                    log.debug('Command \"%s\" failed with return code %d',
+                              cp.args, cp.returncode)
+
+    async def POST(self, data: str):
+        if data == self.autoinstall_default or data == os.environ.get("LANG"):
```
didrocks, line 59: sorry, but you lost me here, what is this about? (Probably a good indication for a needed comment ;))

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+            await super().POST(data)
+            return
+
+        fileContents: str
+        fname = "locale.gen"
+        env = os.environ.copy()
+        if self.app.opts.dry_run:
+            # For debug purposes.
+            fname = ".subiquity/" + fname
+            await arun_command(['cp', '/etc/locale.gen', '.subiquity/'],
```
didrocks, line 69: I think there is already an etc in .subiquity, so you can put it here. Also, hem, there are some method in python to copy a file if this will stay in the end.

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+                               env=env)
+        else:
+            fname = "/etc/" + fname
```
didrocks, line 72: yeah, this is a nitpick but os.path.join?

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+
+        pendingWrite = False
+        with open(fname, "r") as localeGen:
+            # locale.gen is not so big.
+            fileContents = localeGen.read()
+            lineFound = fileContents.find(data)
```
didrocks, line 78: Could we have "data" content being a substring like "pt" for ### pt_PT?
How do you handle as well 
```
# pt_BR ISO-8859-1
# pt_BR.UTF-8 UTF-8
```
As we probably want only the UTF-8 version anyway?

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+            if lineFound == -1:
+                # An unsupported locale coming from our UI is a bug.
+                raise AssertionError("Selected language {} not supported."
+                                     " Rolling back.".format(data))
+
+            commented = "# {}".format(data)
```
didrocks, line 84: Yeah, I’m a little bit scared about the direct matching here for the reason above

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+        env = os.environ.copy()
+        if self.app.opts.dry_run:
+            # For debug purposes.
+            fname = ".subiquity/" + fname
+            await arun_command(['cp', '/etc/locale.gen', '.subiquity/'],
+                               env=env)
+        else:
+            fname = "/etc/" + fname
+
+        pendingWrite = False
```
didrocks, line 74: `needsWrite` maybe?

system_setup/server/controllers/locale.py
```diff
@@ -30,3 +32,65 @@ def start(self):
...
+
+            commented = "# {}".format(data)
+            lineFound = fileContents.find(commented)
+            if lineFound != -1:
+                fileContents = fileContents.replace(commented, data)
+                pendingWrite = True
+
+        if pendingWrite:
+            with open(fname, "wt") as f:
+                f.write(fileContents)
```
didrocks, line 92: what if we can’t write it? (and yes, I know, there are a lot of problems like this in the code), but the exception wouldn’t be posted back to the caller if I’m correct

didrocks (commented):
here we go after this first pass :)

CarlosNihelton:
Thank You for the good comments. One more thing that just occurred me is that installation could be actually moved to the configure controller, like the user creation. That should keep a good user experience. What do you think?

didrocks:
> Thank You for the good comments. One more thing that just occurred me is that installation could be actually moved to the configure controller, like the user creation. That should keep a good user experience. What do you think?

Oh, I haven’t spotted this, thinking this was already the case. Of course, this should be in the configure controller (basically, no change on system before we select "Done" on last screen before recap) :)

system_setup/server/controllers/locale.py
CarlosNihelton: Agreed. Lifted it to its own function to enable early return.

system_setup/server/controllers/locale.py
CarlosNihelton: Changed to an exception. It would really be an exceptional situation running that piece of code as non-root and no --dry-run simultaneosly outside of a dev environment.

system_setup/server/controllers/locale.py
CarlosNihelton: I'll get used to that. I agree that being explicit is always good. It's a bit annoying when you may have False or None, but this one is not the case.

system_setup/server/controllers/locale.py
CarlosNihelton: Sure! Changed into apt.

system_setup/server/controllers/locale.py
CarlosNihelton: Applied os.path.join all around.

system_setup/server/controllers/locale.py
CarlosNihelton: Altered to regex.

system_setup/server/controllers/locale.py
CarlosNihelton: Indeed.

system_setup/server/controllers/locale.py
CarlosNihelton: Agreed.

system_setup/server/controllers/locale.py
CarlosNihelton: I guess you're correct, because I've seen TUI gettng stuck on the failure report screen when server is down due uncaught exceptions. Put into a try-except block and log.error.

system_setup/server/controllers/locale.py
CarlosNihelton: That's still true for locale-gen and update-locale. I wish they had the -P option.

CarlosNihelton:
I think now we can have a second pass.

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +40,90 @@ def __init__(self, app):
...
+
+        fileContents: str
+        locGenNeedsWrite = False
+        with open(localeGenPath, "r") as f:
+            fileContents = f.read()
+            lineFound = fileContents.find(lang)
+            if lineFound == -1:
+                # An unsupported locale coming from our UI is a bug
+                log.error("Selected language %s not supported.", lang)
+                return
```
didrocks, line 57: Maybe we should return false to stop parent on error cases? (same for all error cases below)

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +40,90 @@ def __init__(self, app):
...
+
+            pattern = r'#+\s*({}.*)'.format(lang)
+            commented = re.compile(pattern)
+            (fileContents, n) = commented.subn(r'\1', fileContents, count=1)
+            locGenNeedsWrite = n == 1
```
didrocks, line 62: hem, not a fan :)
Especially that you are only using it once, correct, what about just putting the n == 1 in the if condition?

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +40,90 @@ def __init__(self, app):
...
+
+        if locGenNeedsWrite:
```
didrocks, line 64: you can early return reversing the condition :)

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +40,90 @@ def __init__(self, app):
...
+
+    async def _install_check_lang_support_packages(self, lang, env):
+        """ Installs any packages recommended by
+            check-language-support command. """
+
+        clsCommand = "check-language-support"
+
+        cp = await arun_command([clsCommand, "-l", lang[0:5]], env=env)
+        if cp.returncode != 0:
+            log.error('Command \"%s\" failed with return code %d',
```
didrocks, line 79: hum, do you really need to espace the `"` here? `'` should keep verbatim content

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +40,90 @@ def __init__(self, app):
...
+        if self.app.opts.dry_run:
+            for package in packages:
+                # Downloading instead of installing. Doesn't require root.
+                packs_dir = os.path.join(self.model.root,
+                                         apt_pkg.config
+                                         .find_dir("Dir::Cache::Archives")[1:])
+                archive = os.path.join(packs_dir, cache[package].fullname)
+                cmd = ["wget", "-O", archive, cache[package].candidate.uri]
+                os.makedirs(packs_dir, exist_ok=True)
+                await arun_command(cmd, env=env)
```
didrocks, line 99: and after that you can return

system_setup/server/controllers/configure.py
```diff
@@ -86,6 +175,19 @@ async def configure(self, *, context):
...
+                    testLocGenPath = os.path.join(etc_dir,
+                                                  os.path
+                                                  .basename(localeGenPath))
+                    shutil.copy(localeGenPath, testLocGenPath)
+                    shutil.copy(localeGenPath,
+                                "{}-".format(testLocGenPath))
+                    # After copies are done, overwrite the original path
+                    # because there is code outside of this 'if'
+                    # depending on 'localeGenPath' being initialized.
+                    localeGenPath = testLocGenPath
```
didrocks, line 189: maybe this is a sign that it should be extracted in a function/method of the same file?

didrocks (commented):
And here we go, thanks for pass 2! I think there are still opportunities to get part of the code (outside of the mock question) better, but this is going in the right direction IMHO.

Feel free to tell me if you disagree with anything written here or if things aren’t clear!

system_setup/server/controllers/locale.py
CarlosNihelton: Now i think I figured out!

system_setup/server/controllers/locale.py
CarlosNihelton: + return False.

system_setup/server/controllers/configure.py
CarlosNihelton: Nice tip. Applied across the board.

> And here we go, thanks for pass 2! I think there are still opportunities to get part of the code (outside of the mock question) better, but this is going in the right direction IMHO.
> 
> Feel free to tell me if you disagree with anything written here or if things aren’t clear!

You may not believe, but I really appreciate your critical thinking. It makes me think hard on ways to improve my code. Sometimes I get excited by the result and ended up not looking thoroughly enough to the way the result is achieved.

Thank you for that.

system_setup/server/controllers/configure.py
CarlosNihelton: Opening for read and write is the best way to avoid the need for that variable.

system_setup/server/controllers/configure.py
CarlosNihelton: Redesigned those functions to enable early returns across the board.

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +40,90 @@ def __init__(self, app):
...
+            except IOError as err:
+                log.error("Failed to modify %s file. %s", localeGenPath, err)
+
+    async def _install_check_lang_support_packages(self, lang, env):
+        """ Installs any packages recommended by
+            check-language-support command. """
+
+        clsCommand = "check-language-support"
+
+        cp = await arun_command([clsCommand, "-l", lang[0:5]], env=env)
```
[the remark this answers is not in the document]
CarlosNihelton: There are languages with more than 5 characters anyway, so the best approach I found was actually splitting the language string. I could be, for instance, something like: `nan_TW@latin UTF-8`. I don't think we will ever send the `@latin` part from the UI but checking with the split function won't hurt.
The syntax `clsLang = lang.split('@')[0].split('.')[0].split(' ')[0]` seems strange, but works whether the `lang` string contains dots,`@`, spaces or not.

system_setup/server/controllers/configure.py
CarlosNihelton: Indeed.

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +40,90 @@ def __init__(self, app):
...
+                cmd = ["wget", "-O", archive, cache[package].candidate.uri]
+                os.makedirs(packs_dir, exist_ok=True)
+                await arun_command(cmd, env=env)
+        else:
+            cache.update()
+            cache.open(None)
+            with cache.actiongroup():
+                for package in packages:
+                    cache[package].mark_install()
+                    cache.commit()
```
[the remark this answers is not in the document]
CarlosNihelton: Tab trick. I should have noticed it.

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +40,90 @@ def __init__(self, app):
...
+        cmds = [["locale-gen"],
+                ["update-locale", "LANG={}".format(lang)]]
+        if self.app.opts.dry_run:
+            for cmd in cmds:
+                # TODO Figure out about mocks.
+                # It would be good if they offered a -P prefix option,
+                # but they don't.
+                # Chroot'ing is not an option.
+                log.debug('Would run: %s', ' '.join(cmd))
+        else:
```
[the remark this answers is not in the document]
CarlosNihelton: Agreed.

system_setup/server/controllers/configure.py
CarlosNihelton: Agreed.

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +40,90 @@ def __init__(self, app):
...
+
+        cmds = [["locale-gen"],
+                ["update-locale", "LANG={}".format(lang)]]
+        if self.app.opts.dry_run:
+            for cmd in cmds:
+                # TODO Figure out about mocks.
+                # It would be good if they offered a -P prefix option,
+                # but they don't.
+                # Chroot'ing is not an option.
+                log.debug('Would run: %s', ' '.join(cmd))
```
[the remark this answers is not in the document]
CarlosNihelton: I think I woke up inspired today more than the past week. Found crazy ways to mock up those commands. We'll see how useful those will be. :D

system_setup/server/controllers/configure.py
```diff
@@ -40,89 +41,189 @@ def __init__(self, app):
...
-            since that would have passed thru the Locale
-            controller validation. e.g. en_UK.UTF-8. """
+    def __locale_gen_cmd(self) -> Tuple[str, bool]:
+        """ Returns the locale-gen command path if not in dry-run.
+            Otherwise, copies the locale-gen script, altering the localedef
+            command line to output into a specified directory.
+            Additionally, indicates success by returning True
+            as the second element of the tuple. """
+
+        cmd = "usr/sbin/locale-gen"
```
didrocks, line 51: TBH, I’m not a big fan of it, but I understand why you did this: the fact dry-run can be executed outside of tests only makes mangling the `PATH` env variable more annoying (and you need to ensure you have the local copy somewhere, also…), so I think in the python world, this a good approach (I will show you how this is done generally in Golang, but I dno’t think this should be applied here).

So, it’s well done, I don’t have other counter-proposal, let’s go with this.

system_setup/server/controllers/configure.py
```diff
@@ -40,89 +41,189 @@ def __init__(self, app):
...
+            return (cmdFile, True)
+
+        except IOError as err:
+            log.error("Failed to modify %s file. %s", cmdFile, err)
+            return ("", False)
+
+    def __update_locale_cmd(self, lang) -> list[str]:
+        """ Adds mocking cli to update-locale if in dry-run. """
+        updateLocCmd = ["update-locale", "LANG={}".format(lang)]
+        if self.app.opts.dry_run:
```
didrocks, line 91: can I nitpick asking for an early return here? (feel free to ignore, this is really some personal preference)

system_setup/server/controllers/configure.py
```diff
@@ -40,89 +41,189 @@ def __init__(self, app):
...
+                return False
+        return True
+
+    async def _install_check_lang_support_packages(self, lang, env) -> bool:
         """ Installs any packages recommended by
             check-language-support command. """
 
         clsCommand = "check-language-support"
+        # lang may have more than 5 chars and be separated by dot or space.
+        clsLang = lang.split('@')[0].split('.')[0].split(' ')[0]
```
didrocks, line 125: I would trust you on this one :)

system_setup/server/controllers/configure.py
```diff
@@ -40,89 +41,189 @@ def __init__(self, app):
...
+            return
+
+        ok = await self._install_check_lang_support_packages(lang, env)
+
+        if ok is False:
+            log.error("Failed to install recommended language packs.")
+            return
+
+        ok = await self._activate_locale(lang, env)
+        if ok is False:
```
didrocks, line 224: ah, when I said explicitely comparison, don’t need to check boolean against True or False (as we compare them here)

didrocks (commented):
There were 2 comments on wget missing, and 2 new nitpicks from your changes, but I think we can then move this from Draft to Review, ensure the new tests artifacts are tested and passed, and be done with it!

Well done :)

system_setup/server/controllers/configure.py
CarlosNihelton: Just for the record: :smile:

image

system_setup/server/controllers/configure.py
CarlosNihelton: Alright.

system_setup/server/controllers/configure.py
jibel: maybe a regexp would be more appropriate. Something along the lines:

```
langs = ['nan_tw@latin utf-8', 'en_US', 'C.UTF8' ]

r = re.compile(r'([^.@]+)', re.IGNORECASE)

for l in langs:
      m = r.match(l)
      print(m.groups())
 
('nan_tw',)
('en_US',)
('C',)

```

Plus some error handling in case there is no match of course.

CarlosNihelton:
This PR needs at least one approval. The last editions requested were done: removing wget and the splits in favor of regex on dealing with the selected language. @didrocks, @jibel, Do you guys notice any other issue?

system_setup/server/controllers/configure.py
```diff
@@ -37,6 +41,208 @@ def __init__(self, app):
...
+        localeGenPath = "/etc/locale.gen"
+        if self.app.opts.dry_run is False or self.app.opts.dry_run is None:
+            return localeGenPath
+
+        # For testing purposes.
+        etc_dir = os.path.join(self.model.root, "etc")
+        testLocGenPath = os.path.join(etc_dir,
+                                      os.path.basename(localeGenPath))
+        shutil.copy(localeGenPath, testLocGenPath)
+        shutil.copy(localeGenPath, "{}-".format(testLocGenPath))
```
didrocks, line 224: We generally use `.new` for this ? This is what we are generally using to have atomic changes. (ok, here, it’s contrived because it’s only for tests, but maybe .test?)

didrocks (commented):
2 small nitpicks but then, :+1:

jibel (approved):
LGTM. Thanks!

dbungert (approved):

didrocks:
And I think we are good to go with the latest changes! :+1:

PR #1117, opened 2021-11-04
test/api: make less flaky

dbungert:
This is part of a changeset intended to have the API tests run in
parallel.  I don't consider this ideal - I still see failures when
running tests in parallel - but am providing it now with the intent of
helping people who are seeing api test flakiness even when those tests
are run serially.

* check for socket existence on startup
* lengthen timeouts on shutdown


mwhudson:
Seems to help!

mwhudson (approved):

PR #1120, opened 2021-11-05
kvm-test: assorted updates

mwhudson:
The script is great but I found I needed to make a few changes.

### DIFF
--- a/scripts/kvm-test.py
+++ b/scripts/kvm-test.py
@@ -132,6 +132,8 @@ parser.add_argument('-a', '--autoinstall', default=False,
                     action='store_true', help='use autoinstall')
 parser.add_argument('-b', '--base', default=False, action='store_true',
                     help='use base iso')
+parser.add_argument('--basesnap', default=None, action='store',
+                    help='use slimy-update-snap on this snap')
 parser.add_argument('-B', '--bios', action='store_true', default=False,
                     help='boot in BIOS mode')
 parser.add_argument('-c', '--channel', default=False, action='store',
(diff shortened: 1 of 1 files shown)


scripts/kvm-test.py
```diff
@@ -451,9 +464,8 @@ def install(ctx):
         # appends += ['subiquity-channel=edge']
 
         with mounter(iso, mntdir):
-            appends += grub_get_extra_args(mntdir)
-
             if len(appends) > 0:
+                appends += grub_get_extra_args(mntdir)
```
dbungert, line 468: If we have modified the iso then I think we still need to go looking for the new layer, right?  So we would want this all the time, even if there aren't any other reasons to add to the kernel `-append` list?

mwhudson: livefs-edit edits the grub config to point at the new layer (it should probably edit the initrd but well)

scripts/kvm-test.py
```diff
@@ -243,9 +247,18 @@ def build(ctx):
     # --inject-snap $SUBIQUITY_SNAP_PATH
     if project.startswith('subiquity'):
         if ctx.args.quick:
-            with snap_manager('subiquity_test.snap') as snap:
+            with snap_manager('subiquity_new.snap') as snap:
```
dbungert: It turns out that this 'with' statement makes no sense at all, with either the old or new value.  Even using subiquity_new.snap as the value isn't quite right because the path to that snap file is wrong - that snap file is created in tempdir created by the 'quick' script.  I'm pushing a commit that just removes the with statement and I have no idea why this was ever working.

dbungert (approved):

PR #1121, opened 2021-11-08
Enables preserving existing container instances.

CarlosNihelton:
Hi @dbungert !

I hope you guys find this one useful and nothing breaks GitHub side. I changed the `test-in-lxd` script to preserve existing instances and starting then, which is useful for develop environment. I used that in debugging one issue preventing the server to start on Focal only.

The name of the instance changes from `tester` to the adjective of the Ubuntu Release (extracted from the image name). That enables preserving instances from different releases while doing matrix builds. That makes the dev environment closer to the CI.

Feel free to discard or modify that in case it doesn't fit well for everybody working on Subiquity.

### DIFF
--- a/scripts/test-in-lxd.sh
+++ b/scripts/test-in-lxd.sh
@@ -2,7 +2,7 @@
 set -eux
 IMAGE=$1
 SCRIPT=$2
-TESTER=${IMAGE##*:}
+TESTER=subiquity-${IMAGE##*:}
 
 lxd init --auto
 
--- a/scripts/test-in-lxd.sh
+++ b/scripts/test-in-lxd.sh
@@ -6,7 +6,7 @@ TESTER=${IMAGE##*:}
 
 lxd init --auto
 
-if [ -z "$(lxc list -f csv | grep $TESTER)" ]
+if [ -z "$(lxc list -f csv -c n ^${TESTER}\$)" ]
 then
     lxc launch $IMAGE $TESTER
     lxc config device add $TESTER code disk source=`pwd` path=/subiquity
--- a/scripts/test-in-lxd.sh
+++ b/scripts/test-in-lxd.sh
@@ -2,34 +2,40 @@
 set -eux
 IMAGE=$1
 SCRIPT=$2
+TESTER=${IMAGE##*:}
 
 lxd init --auto
 
-lxc launch $IMAGE tester
-lxc config device add tester code disk source=`pwd` path=/subiquity
+if [ -z "$(lxc list -f csv | grep $TESTER)" ]
+then
+    lxc launch $IMAGE $TESTER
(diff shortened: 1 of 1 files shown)


scripts/test-in-lxd.sh
```diff
@@ -2,34 +2,40 @@
 set -eux
 IMAGE=$1
 SCRIPT=$2
+TESTER=${IMAGE##*:}
 
 lxd init --auto
 
-lxc launch $IMAGE tester
-lxc config device add tester code disk source=`pwd` path=/subiquity
+if [ -z "$(lxc list -f csv | grep $TESTER)" ]
```
dbungert, line 9: `grep $TESTER` might catch the wrong name, depending on the name.
What about something like:
```suggestion
if lxc list -f csv | grep -qv "^${TESTER},"
```

dbungert (changes requested):

scripts/test-in-lxd.sh
jibel: You can altogether avoid the grep and catch the right instance with something like

if [ -z "$(lxc list -f csv -c n ^${TESTER}$)" ]; then

(probably the $ at the end of the regex must be escaped)

CarlosNihelton: Thank you guys for the suggestions. I'll go with @jibel, avoiding grep. It works on bash with and without escaping the last $.

scripts/test-in-lxd.sh
```diff
@@ -2,34 +2,40 @@
 set -eux
 IMAGE=$1
 SCRIPT=$2
+TESTER=${IMAGE##*:}
```
dbungert, line 5: This is looking great.
However, I think I want to rename the test image so it is less likely to conflict with existing containers.
What do you think about this?
```suggestion
TESTER=subiquity-${IMAGE##*:}
```

CarlosNihelton: Sure it's safer than `TESTER=tester-${IMAGE##*:}`. I'll push the change in a bit.

dbungert (approved):

scripts/test-in-lxd.sh
```diff
@@ -2,34 +2,40 @@
 set -eux
 IMAGE=$1
 SCRIPT=$2
+TESTER=subiquity-${IMAGE##*:}
 
 lxd init --auto
 
-lxc launch $IMAGE tester
-lxc config device add tester code disk source=`pwd` path=/subiquity
+if [ -z "$(lxc list -f csv -c n ^${TESTER}\$)" ]
```
mwhudson, line 9: This is fine, but if it was me I'd do something like "if lxc query /1.0/containers/bl >/dev/null 2>&1; then" to avoid the regexp. But I'll merge this anyway :)

mwhudson (approved):

PR #1122, opened 2021-11-09
implement the bulk of subiquity-configure-apt in the server process

mwhudson:
This uses a kind of crazy approach where the hook script we get curtin
to run makes an API call to the server process to do the actual
configuration. I can't work out if this is genius or insane or both but
anyway it is only a step on to the way to doing the apt configuration
before we start curtin so it won't be around for long.


subiquity/server/controllers/install.py
```diff
@@ -147,14 +155,11 @@ async def install(self, *, context):
...
 
-            config_location = self.write_config()
-
             if os.path.exists(self.model.target):
                 await self.unmount_target(
                     context=context, target=self.model.target)
 
-            await self.curtin_install(
-                context=context, config_location=config_location)
+            await self.curtin_install(context=context)
```
dbungert, line 162: looks like this needs a config_location arg

dbungert (changes requested):
I tested in kvm and found the item in the comment.  Unknown if there are further issues but otherwise things looks good.

subiquity/server/controllers/install.py
mwhudson: Of course when I test, then make a "trivial" refactoring, I break stuff. Fix pushed (not tested yet though)

dbungert (approved):

PR #1124, opened 2021-11-11
SystemSetupModel now acks the variant at start.

CarlosNihelton:
Without this change, `wsl_configuration` variant is not recognized without client_variant.POST. The server is expected to start in that variant if the environment is properly set.
Issue and suggestions of fixes originally reported by @jpnurmi.

### DIFF
--- a/system_setup/server/server.py
+++ b/system_setup/server/server.py
@@ -78,8 +78,10 @@ class SystemSetupServer(SubiquityServer):
         root = '/'
         if self.opts.dry_run:
             root = os.path.abspath('.subiquity')
-        return SystemSetupModel(root, self.hub, INSTALL_MODEL_NAMES,
-                                POSTINSTALL_MODEL_NAMES)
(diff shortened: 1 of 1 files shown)


mwhudson (commented):
This seems fine, if a bit around the houses. Why not make it an argument to SystemSetupModel itself? Do you still need the call to set_source_variant in __init__?

mwhudson (approved):

CarlosNihelton:
> This seems fine, if a bit around the houses. Why not make it an argument to SystemSetupModel itself? Do you still need the call to set_source_variant in **init**?

Thanks, @mwhudson !

Regarding the `set_source_variant` in the server init, we just need to have `self.variant` initialized,. which is all that function is actually doing. It could be replaced by a direct initialization, but I chose for the less effort. The same reason why not passing the variant to the model `__init__` method as an argument. Less effort: touch less files.

jibel (approved):
+1. Thanks!

PR #1125, opened 2021-11-15
aptdeps: +snapcraft

dbungert:


mwhudson:
Hm. What's the motivation for this? It's perfectly reasonable to have snapcraft installed via snap, I think?

dbungert:
Attempting to make the dev setup a bit easier.  Since snapcraft is ultimately a snap maybe we just point to that.  I'll drop this for now.

PR #1127, opened 2021-11-17
check-language-support: fix empty pkg list

dbungert:
The check-language-support on my machine was providing a single empty
newline as output, and the length check isn't catching that as the
resulting packages array has a value of `['']`.
Drop empty values from the list to avoid a `KeyError: ''` exception.


mwhudson (approved):

didrocks (approved):
Indeed, this is a valid one. Thanks for picking and fixing it! :+1:

PR #1128, opened 2021-11-17
Complementary to 1127 on integration side

CarlosNihelton:
Hi, @dbungert !

This fix avoids integration tests to fail even if `check-language-support` returns an empty list of packages, which would be the case when all packages needed by one specific configuration of locale are already installed.

This complements #1127 .

### DIFF
--- a/system_setup/server/controllers/configure.py
+++ b/system_setup/server/controllers/configure.py
@@ -116,10 +116,10 @@ class ConfigureController(SubiquityController):
 
         return True
 
-    async def __recommended_packages_to_install(self, lang, env) -> List[str]:
+    async def __recommended_language_packs(self, lang, env) -> List[str]:
         """ Return a list of package names recommended by
          check-language-support (or a fake list if in dryrun).
-         List returned can be empty, but never None.
+         List returned can be empty on success. None for failure.
         """
         clsCommand = "check-language-support"
         # lang code may be separated by @, dot or spaces.
@@ -128,12 +128,12 @@ class ConfigureController(SubiquityController):
         matches = pattern.match(lang)
         if matches is None:
             log.error("Failed to match expected language format: %s", lang)
-            return []
+            return None
 
         langCodes = matches.groups()
         if len(langCodes) != 1:
             log.error("Only one match was expected for: %s", lang)
-            return []
+            return None
 
         clsLang = langCodes[0]
         # Running that command doesn't require root.
@@ -141,7 +141,7 @@ class ConfigureController(SubiquityController):
         if cp.returncode != 0:
             log.error('Command "%s" failed with return code %d',
                       cp.args, cp.returncode)
-            return []
+            return None
 
         packages = cp.stdout.strip().split(' ')
         # We will always have language-pack-{clsLang}-base in dryrun.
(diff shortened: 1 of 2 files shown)


system_setup/server/controllers/configure.py
```diff
@@ -143,29 +143,35 @@ async def _install_check_lang_support_packages(self, lang, env) -> bool:
...
             packs_dir = os.path.join(self.model.root,
                                      apt_pkg.config
                                      .find_dir("Dir::Cache::Archives")[1:])
             os.makedirs(packs_dir, exist_ok=True)
             try:
+                if len(packages) == 0:
+                    message = "{} didn't recommend any packages." \
+                              " Nothing to do.".format(clsCommand)
+                    log.debug(message)
+                    msgFile = os.path.join(packs_dir, "msgFile")
```
dbungert, line 157: Thanks for looking into this.

Note that dry run is used outside of the integration tests - in particular ubuntu-desktop-installer uses it heavily for devel purposes.

While I appreciate that this should solve the problem, I wonder if we should consider another route.
What if we added a mock `check-language-support` to the integration tests that has output you control?  This could also mean that you could do a more calculated test - not just check that the cache is present or absent, but check for specific files to be present.  That would also let you test the no-op case.  What do you think?

system_setup/server/controllers/configure.py
```diff
@@ -143,29 +143,35 @@ async def _install_check_lang_support_packages(self, lang, env) -> bool:
             return False
 
         packages = [pkg for pkg in cp.stdout.strip().split(' ') if pkg]
-        if len(packages) == 0:
```
dbungert, line 146: Note that the removal of this means that the non-dry-run case does touch the cache with an open/commit set of calls.  I doubt that's a problem but wanted to confirm that this is intentional or at least benign.

dbungert (changes requested):

system_setup/server/controllers/configure.py
didrocks: I think @dbungert's strategy is the best one. Wdyt @CarlosNihelton?

CarlosNihelton: Sure! I'll take care of that. Thanks for the recommendations.

system_setup/server/controllers/configure.py
CarlosNihelton: Thanks for pointing that out.
That could lead to unexpected effects if there were other packages marked to install previously by other routines. Whether that's feasible or not under the context of system_setup I couldn't figure out, honestly. But I made sure the cache is only touched if the packages list is not empty. Safer.

system_setup/server/controllers/configure.py
```diff
@@ -127,29 +128,39 @@ async def _install_check_lang_support_packages(self, lang, env) -> bool:
...
+            return []
 
-        packages = [pkg for pkg in cp.stdout.strip().split(' ') if pkg]
-        if len(packages) == 0:
-            log.debug("%s didn't recommend any packages. Nothing to do.",
-                      clsCommand)
-            return True
+        packages = cp.stdout.strip().split(' ')
+        # We will always have language-pack-{clsLang}-base in dryrun.
+        if len(packages) == 0 and self.app.opts.dryrun:
```
dbungert, line 148: I like where this is going.  However, if c-l-s fails, we won't hit the packages == [] case, it will actually return an empty list.

dbungert (changes requested):

system_setup/server/controllers/configure.py
CarlosNihelton: You're right! Good catch! C-L-S could fail under live-run and be interpreted as recommending nothing, which is not a failure. `None` to the rescue.

dbungert (approved):

PR #1129, opened 2021-11-19
snapcraft: remove python links from snap bin dir

dbungert:
The python links in the bin dir of the snap were resolving to system
python instead of the python binary found at usr/bin.  This appears to
be the root cause of why python scripts would fail to find their own
resources, such as ssh-import-id and lsb_release.


mwhudson (commented):
Argh wtf. Is it snapcraft creating these links? Can it not do that for classic snaps? Anyway this is clearly an improvement, so approved (& I'll merge given the day of the week)

mwhudson (approved):

PR #1131, opened 2021-11-26
offer to install drivers with ubuntu-drivers if any are available

mwhudson:
This is mostly a request for comments, we probably need some ubuntu-drivers work to make it really shiny (https://bugs.launchpad.net/ubuntu/+source/ubuntu-drivers-common/+bug/1952147). But it does work!


subiquity/server/controllers/drivers.py
```diff
@@ -0,0 +1,97 @@
...
+log = logging.getLogger('subiquity.server.controllers.drivers')
+
+
+class DriversController(SubiquityController):
+
+    endpoint = API.drivers
+
+    autoinstall_key = model_name = "drivers"
+    autoinstall_schema = {
+        'type': 'boolean',
```
dbungert, line 37: I'm imaging a world where we want some flexibility here, like we want wifi things but not gpu ones.    If that arises, what would we need to change in autoinstall to adapt to that scenario?

subiquity/server/controllers/drivers.py
```diff
@@ -0,0 +1,97 @@
...
+                return
+        async with apt.overlay() as d:
+            try:
+                await self.app.command_runner.run(
+                    ['chroot', d, 'sh', '-c', "command -v ubuntu-drivers"])
+            except subprocess.CalledProcessError:
+                self.has_drivers = False
+                await self.configured()
+                return
+            # XXX need to check ubuntu-drivers is available here!
```
dbungert, line 82: Just adding a comment as a reminder to come back to this one, as this looks like a good problem to handle.

autoinstall-schema.json
```diff
@@ -361,6 +361,9 @@
                 "additionalProperties": false
             }
         },
+        "drivers": {
+            "type": "boolean"
```
ogayot, line 365: I think it's not immediately obvious what `true` or `false` would mean for `drivers`. Would you consider either of these?
```json
"install_drivers" : { "type": "boolean" }
```
```json
"drivers": {
    "type": "object",
    "properties": {
        "install": {
            "type": "boolean"
        }
    }
}
```

subiquity/models/drivers.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+
+log = logging.getLogger('subiquity.models.drivers')
+
+
+class DriversModel(object):
```
ogayot, line 21: You can drop the explicit inheritance from object but it's up to you :)

subiquity/common/apidef.py
```diff
@@ -290,6 +290,10 @@ class edit_partition:
                 def POST(data: Payload[ModifyPartitionV2]) \
                     -> StorageResponseV2: ...
 
+    class drivers:
+        def GET(wait: bool = False) -> Optional[bool]: ...
```
ogayot, line 294: I think we would benefit from injecting the value of `install` / `do_install` into the response, so that we can save/restore the status of the checkbox "install the drivers?" after unchecking it and hitting back.

Actually, I was thinking about something like:

```python
class drivers:
    def GET() -> bool:
        """ Return the value of do_install (aka. install drivers if available) """
    def POST(install: bool) -> None:
        """ Set the value of do_install """

    class available:
        def GET(wait: bool = False) -> Optional[bool]:
            """ Return whether drivers are available for installation. None means retry later.
            In the future, it could also be a list of drivers available.
            """ 
```

What is your opinion on this?

subiquity/server/controllers/drivers.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2018 Canonical, Ltd.
```
ogayot, line 1: minor: year to update :) 

But since it's 2022 now ; I'm not sure it's relevant.

subiquity/server/controllers/drivers.py
ogayot: Seems like the change was finally implemented so I think what's left to do is to drop the XXX comment :)

subiquity/ui/views/drivers.py
```diff
@@ -0,0 +1,106 @@
...
+
+        connect_signal(
+            self.form, 'submit',
+            lambda result: self.done(result.install.value))
+        connect_signal(self.form, 'cancel', self.cancel)
+
+        self._w = self.form.as_screen(excerpt=_(excerpt))
+
+    def done(self, result):
+        log.debug(f"User input: {result}")
```
ogayot, line 102: You can use lazy interpolation if you want:
```python
log.debug("User input: %s", result)

File: subiquity/server/apt.py [hunk 8]
```diff
@@ -52,7 +54,9 @@ async def mount(self, device, mountpoint, options=None, type=None):
             ['mount'] + opts + [device, mountpoint])
         self._mounts.append(mountpoint)
 
-    async def unmount(self, mountpoint):
+    async def unmount(self, mountpoint, remove=True):
+        if remove:
```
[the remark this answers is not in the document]
mwhudson: I can't remember what this api change is about now :) I don't think it's around success or failure though.

autoinstall-schema.json
mwhudson: Yeah that would probably be better (and more future proof)

subiquity/common/apidef.py
mwhudson: Yes certainly the users choice should be sticky. Seems a bit silly to require two api calls but well. Details.

subiquity/models/drivers.py
mwhudson: Heh was obviously having a flashback to 20 year old Python for some reason :)

subiquity/server/apt.py
```diff
@@ -119,35 +124,48 @@ async def configure(self, context):
...
-        await self.mount('/cdrom', f'{for_install}/cdrom', options='bind')
+        os.mkdir(f'{self.for_install}/cdrom')
+        await self.mount('/cdrom', f'{self.for_install}/cdrom', options='bind')
 
         if self.app.base_model.network.has_network:
             os.rename(
-                f'{for_install}/etc/apt/sources.list',
-                f'{for_install}/etc/apt/sources.list.d/original.list')
+                f'{self.for_install}/etc/apt/sources.list',
+                f'{self.for_install}/etc/apt/sources.list.d/original.list')
```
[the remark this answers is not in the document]
mwhudson: We certainly end up copying it around a lot. I think I didn't realize curtin backed up the entire file -- that seems unnecessary. Subiquity isn't really "backing up" the file here, it gets copied back later.

subiquity/server/controllers/drivers.py
mwhudson: Well, I guess we'd need ubuntu-drivers to support that concept to start with... I don't think it does now.

subiquity/server/controllers/drivers.py
mwhudson: Yeah I certainly wrote code to handle this properly.

subiquity/common/apidef.py
ogayot: Having both information in a single endpoint is fine as well IMO.

ogayot:
@mwhudson should we close this now?

mwhudson:
Yes I think so

PR #1132, opened 2021-11-30
filesystem: allow creation of 4th msdos partition

dbungert:


dbungert:
@mwhudson - is this valid for s390x?

mwhudson:
Uh can we hit this code path?

mwhudson:
Oh right, I think we can only hit this on fdasd systems where I think this is not valid -- but we should ask @frank-heimes ...

dbungert:
It came up from https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1952545.  The user forced the MBR partition table with external partitioning and subiquity is willing to create MBR partitions in that state.

mwhudson:
Oh right, IIRC I meant to treat an empty partition table the same as _no_ partition table, so that you would end up creating GPT partitions in this scenario. Would be easy to get that wrong though...

subiquity/common/filesystem/actions.py
```diff
@@ -218,9 +218,13 @@ def _can_partition_device(device):
...
-    # only support 3 partitions. As and when we support editing
-    # partition msdos tables we'll need to be more clever here.
-    if device.ptable in ['vtoc', 'msdos'] and len(device._partitions) >= 3:
+    # only support 3 partitions.
+    if device.ptable == 'vtoc' and len(device._partitions) >= 3:
+        return False
+    # Existing msdos partition tables will be used though, so offer at least
+    # the 4 partitions there.  When we support editing msdos partition tables
+    # we'll need to be more clever here.
+    if device.ptable == 'msdos' and len(device._partitions) >= 4:
```
xnox, line 227: does this need to handle extended msdos partitions too?

dbungert: Eventually yes.  My immediate goal was to start separating what s390x needs out from general MBR partition support on other arches.

mwhudson:
Everything about this code is going to change as part of supporting editing partitions, so maybe we can leave it for now? We can merge this or something like it if we are going to release again before that is ready?

dbungert:
OK, let's revisit with partition editing.

PR #1134, opened 2021-12-01
Compare the snaps versions to avoid suggesting a downgrade

ogayot:
Hello,

In specific scenarios, subiquity was suggesting to upgrade to a version that is older to the one currently running ; as in the following example [1]

_Version 21.01.2 of the installer is now available (21.06.01 is currently running)._

### Rationale
- To check if an update is available, subiquity queries the snapd daemon through its HTTP API
- The response sent back by snapd seems to 1:1 correspond to what is returned by `$ snap refresh --list`. I’d assume it’s using the same API call internally
- If snapd response contains a result with the “subiquity” title, then it is considered an new available version ; and the update available screen will effectively be displayed. subiquity does not do any filtering based e.g. on the version name returned.

### Implementation
I implemented basic filtering of the snap versions to avoid suggesting something that would result in a downgrade.
We parse the version number of the currently running version and compare it to the version of the snap returned by snapd.
The following formats are supported:

- `{major}.{minor}.{patch}`
- `{major}.{minor}.{patch}+git{build_id}.{commit_id}`

If a different version format is read, the parsing would fail. Consequently, the version returned by snapd would be suggested as an upgrade to the user. Therefore, the current behavior would not change in this scenario.

Thanks,
Olivier

[1] https://bugs.launchpad.net/subiquity/+bug/1936190

### DIFF
--- a/subiquity/common/snap.py
+++ b/subiquity/common/snap.py
@@ -1,4 +1,6 @@
 from dataclasses import dataclass
+from typing import Optional
+import re
 
 
 class SnapVersionParsingError(Exception):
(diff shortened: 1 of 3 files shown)


dbungert:
@mwhudson - do we ever do version numbers like x.y.z-rc1?

dbungert (approved):
Looks great, thanks!
Have a look at the minor complaint from the linter.

mwhudson:
> @mwhudson - do we ever do version numbers like x.y.z-rc1?

Not so far, and I think we should avoid doing so.

dbungert:
An interesting outcome with this PR is that if we inject a snap into the ISO, then no matter what we aren't going to offer the updated version, which arguably is the behavior we want.  Thoughts @mwhudson ?

mwhudson:
I don't think this is a change in behaviour -- snapd will never show an unasserted snap as having an update available anyway. If you inject a snap + assertion, that's no different from how the image is made (I hope!)

PR #1137, opened 2021-12-06
move apt configurer to mirror controller

mwhudson:
And some related refactoring.

Feel free to review this quite hard!


mwhudson:
FWIW https://github.com/mwhudson/subiquity/tree/mirror-check-in-ui is the next part along this road which adds checking of the provided mirror in the UI.

subiquity/server/apt.py
```diff
@@ -50,58 +110,45 @@ async def mount(self, device, mountpoint, options=None, type=None):
...
-        #    'configured' overlay.
-
-        self.configured = self.tdir()
-
-        config_upper = await self.setup_overlay(self.source, self.configured)
+
+        def lowerdir_for(lower):
+            if isinstance(lower, str):
+                return lower
+            if isinstance(lower, Mountpoint):
```
dbungert, line 131: We could simplify this sequence a little by adding a method that the `_MountBase`s implement

dbungert (approved):

subiquity/server/apt.py
mwhudson: I'm not sure quite what you meant here, but I refactored this a little anyway and in the process noticed a bug, so thanks for the suggestion :)

subiquity/server/controllers/mirror.py
```diff
@@ -81,6 +90,10 @@ def on_geoip(self):
             self.model.set_country(self.app.geoip.countrycode)
         self.cc_event.set()
 
+    def on_source(self):
+        if self.configured_once:
+            self._apply_apt_config_task.start_sync()
+
```
ogayot, line 96: Hi @mwhudson,

Sorry for bringing this PR back from the archives.
While doing some debugging, I noticed that the `_apply_apt_config` method is called twice (maybe more) during an installation. It seems to be causing at least /some/ problems.
I think it is because the `self.configured_once` variable is always `True`. It is initialized to `True` and then is assigned to `True` again in the `configure` method.

Changing the initialization value to `False` would seemingly fix the  logic but if you remember the 

Thanks,
Olivier

mwhudson: Hi, yes the initialization of configured_once is clearly wrong :( and always results in _apply_apt_config being cancelled at least once.

The reason this causes problems rather than just wastes a bit of resources is because I thought that cancelling asyncio.Task that was waiting on a subprocess killed the subprocess -- but this just isn't true. I talked to @dbungert about this yesterday and I think we're going to go for a solution where we run the apt-config bits as a transient systemd unit with PrivateMounts=true -- that way we can just get systemd to shoot it in the head and not have to worry about child processes or bind mounts lingering.

PR #1138, opened 2021-12-06
filesystem: add disk model and vendor name

jpnurmi:
Read ID_MODEL/VENDOR_ENC for disks to allow the GUI installer to show
pretty disk names without having to call libudev through FFI.

As a bonus, this also fixes disk names to match the machine config in
dry-run mode.

Related issues
- FR-1710
- canonical/ubuntu-desktop-installer#415

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -569,7 +569,6 @@ class Disk(_Device):
     wwn = attr.ib(default=None)
     multipath = attr.ib(default=None)
     path = attr.ib(default=None)
-    model = attr.ib(default=None)
     wipe = attr.ib(default=None)
     preserve = attr.ib(default=False)
     name = attr.ib(default="")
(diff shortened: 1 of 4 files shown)


subiquity/models/filesystem.py
```diff
@@ -569,7 +569,6 @@ class Disk(_Device):
     wwn = attr.ib(default=None)
     multipath = attr.ib(default=None)
     path = attr.ib(default=None)
-    model = attr.ib(default=None)
```
dbungert, line 572: Why is this line being removed?

jpnurmi: I replaced this with a model property because I didn't see this being set anywhere. The only reference I could find is below but I think it was always None.

https://github.com/canonical/subiquity/blob/bd4e4785a9bc4f44ceb7423bac09502719819971/subiquity/models/filesystem.py#L601

mwhudson (approved):

PR #1139, opened 2021-12-07
storage/v2: add os-prober data

dbungert:
On the Partition objects, add a 'os' field.
This contains os-prober results for that partition.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -31,7 +31,7 @@ from curtin.util import human2bytes
 
 from probert.storage import StorageInfo
 
-from subiquity.common.types import Bootloader
+from subiquity.common.types import Bootloader, OsProber
 
 
 log = logging.getLogger('subiquity.models.filesystem')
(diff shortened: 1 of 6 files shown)


subiquity/models/filesystem.py
```diff
@@ -708,6 +708,17 @@ def ok_for_raid(self):
             return False
         return True
 
+    @property
+    def path(self):
+        return self.device.path + str(self.number)
```
mwhudson, line 713: unfortunately this is too simple, it's probably ok for desktop but it won't work for RAIDs (a partition of /dev/md0 is something like /dev/md0p1) or multipath disks (where it all gets complicated). I'm probably happy with doing this this way purely for the purpose of os-prober look ups but not in an API that looks generally useful :)

It's all a bit frustrating of course because we know exactly the DEVPATH that corresponds to this partition object when making the partition object in curtin.storage_config but I don't think we attach that in any reliable way. Maybe we should... (but probably not for this branch!)

mwhudson (changes requested):

subiquity/models/filesystem.py
dbungert: See the fixup commit for my modification, and yes having curtin output the path sounds useful!

mwhudson (approved):

mwhudson:
Maaybe add a comment saying the path calculation is deliberately simple?

PR #1140, opened 2021-12-08
Add initial integration of Ubuntu Advantage in Subiquity

ogayot:
Hello,

This PR introduces a new view, model and controllers pair to enable Ubuntu Advantage from Subiquity.
For now, we expect the user to input an alphanumeric token and rely on cloud-init to enable the features associated with the UA subscription.

This implementation comes with a number of limitations:

- inability to validate the token before first-boot
- inability to check what features are associated with the Ubuntu Advantage subscription, before first-boot
- the user must input the token directly: there is no mechanism to deduce the token based on email address and password.

The new screen right after the identity screen. It is only shown if we are running a LTS release.

Thanks,
Olivier


dbungert:
Something in the CI tests went wrong and they were running for 4+ hours.  Please take a peek.  Consider running locally
```
sudo ./scripts/test-in-lxd.sh ubuntu-daily:focal "make check"
```

mwhudson:
Hi, this looks like a good start. To get the integration tests to work you'll need to add "answers" support to the new client controller and answers to all of the examples/answers*.yaml files (this last bit is not the greatest design). I'm not sure why they are hanging forever though -- they are supposed to time out rather than doing that :(

I do wonder if we want to show this screen when installing a non-LTS?

ogayot:
Updated to fix the failing integration tests.
I had to squash the two commits together because, for some reason, I didn't manage to have the integration tests succeed with the new screen implemented but not shown.

ogayot:
> I do wonder if we want to show this screen when installing a non-LTS?

Good point, do we know if non-LTS releases can benefit from some of the components that can be enabled through UA?

ogayot:
Updated to show screen only when running on LTS version.
In dry-run mode, we always consider that we are on LTS -> we load `examples/lsb-release-focal` instead of `/etc/lsb-release`.

dbungert (approved):

dbungert (changes requested):
This is all great!

One change however - we'd like to hide the screen for now.  We'd like to be able to show the screen once we have the ability to validate the token.

ogayot:
Thanks @Dan!
while implementing the change that you requested, I noticed that raising `Skip` from the client prevented the installation from finishing. 
The reason was that the server-side controller still needs its method `.configured()` to be called to signal that the model has finished configuring. Thanks to @mwhudson for pointing me to the right direction :)

To fix this issue, I created a separate `POST` resource (called skip) to be called by the client when it wants to skip the screen. As a result, to effectively skip the screen, the client must do:

```
# Make sure the server-side controller calls .configured()
await self.endpoint.skip.POST()
raise Skip(...)
```

subiquity/client/controllers/ubuntu_advantage.py
```diff
@@ -31,6 +35,14 @@ class UbuntuAdvantageController(SubiquityTuiController):
 
     async def make_ui(self) -> UbuntuAdvantageView:
         """ Generate the UI, based on the data provided by the model. """
+        path_lsb_release: Optional[str] = None
+        if self.app.opts.dry_run:
+            # In dry-run mode, always pretend to be on LTS
+            path_lsb_release = "examples/lsb-release-focal"
+        if "LTS" not in lsb_release(path_lsb_release)["description"]:
```
dbungert, line 47: This PR is getting a little long and gathering some additional items, so I want you to do something in a follow-on PR.
We call lsb_release in six spots, and I'd like them to have consistent answers (even in dry-run mode).
So I'd like to add an argument to lsb_release to tell it to use dry-run mode and then that can trigger the usage of the example data.  Then we can tell all lsb_release usage about dryrun.

dbungert (approved):

PR #1141, opened 2021-12-08
apt: get settings from mirror + proxy

dbungert:


mwhudson (approved):
AAAA yes good spot. It's kinda noticeable how we're moving from a two step -- install, postinstall -- installation process to something much more fine grained -- apt, curtin, cloud-init config, packages, cleanup -- and I wonder if we should embrace that a little more.

dbungert:
I absolutely think we should embrace it.  We're able to provide better results by being more granular.

PR #1145, opened 2021-12-14
snapcraft: add os-prober

dbungert:


mwhudson:
We discussed this in a back channel a bit but could we consider patching os-prober to support something like $OS_PROBER_PLUGIN_DIR before doing this? I'm not completely against the sed but I'm not a fan of it either :)

mwhudson (approved):
Thanks. Any chance of making a mp on salsa and linking that in a comment?

dbungert:
https://salsa.debian.org/installer-team/os-prober/-/merge_requests/13

PR #1148, opened 2021-12-16
[SystemSetup] Moving pieces of configuration from Subiquity to WSL Launcher

CarlosNihelton:
Hi there!

With this pull request I aim to remove the default user configuration from SystemSetup. That piece is being moved to the C++ launcher (Windows side). This results in removal of some code and sections of golden files. With that we also make the way Subiquity SystemSetup should communicate the required actions to the launcher a bit more strict. That required changing SystemSetup Configure and Shutdown controllers.


system_setup/server/controllers/configure.py
```diff
@@ -343,6 +344,16 @@ async def configure(self, *, context):
                     raise Exception("Failed to create user %s: %s"
                                     % (username, create_user_proc.stderr))
                 log.debug("created user %s", username)
+                with open(os.path.join(etc_dir, "passwd")) as f:
+                    for line in f:
+                        if username not in line:
```
dbungert, line 349: This could trigger some false matches
* the username being searched for could be a subset of other usernames
* the username being searched for could appear in other fields

Something like this maybe?
```
tokens = line.split(":")
if username == tokens[0]:
    if len(tokens) < 3:
        raise Exception('Invalid passwd entry')
    uid = int(tokens[2])
    break
```

dbungert: Also in general this function is getting a bit long, you may want to break it up in smaller chunks.

dbungert: More thoughts
* uid is never assigned outside of the loop, so a failure to find the username will raise a NameError
* is it valid to look for the root user?  Maybe it's best to use a None as the uid initializer instead of zero.

system_setup/server/controllers/shutdown.py
```diff
@@ -50,18 +50,24 @@ async def _wait_install(self):
     @with_context(description='mode={self.mode.name}')
     def shutdown(self, context):
         self.shuttingdown_event.set()
-        launcher_status = "complete"
+        comments = ["# This file was auto generated by subiquity server.",
```
dbungert, line 53: Nitpick: random users may not understand the server portion - or subiquity in general.  Maybe just call it system-setup?

dbungert (changes requested):

system_setup/server/controllers/configure.py
CarlosNihelton: Thank you, @dbungert, for the good review.

I'll go with your suggestion on how to avoid false positives, but I'll be stricter, since `/etc/passwd` entries must have 7 and only 7 fields.

```
/etc/passwd contains one line for each user account, with seven fields delimited by colons (“:”). These fields are:
```

While root is a valid user it would be a waste of CPU to set it as default user. WSL defaults to root if we don't have a default UID set. That's why we check for None and 0 in the Shutdown controller before adding the defaultUid line:

```python
        if default_uid is not None and default_uid != 0:
            launcher_status += ["defaultUid={}".format(default_uid)]
```

So being None or 0 would be the same. I'll go with None per your suggestion.

system_setup/server/controllers/shutdown.py
CarlosNihelton: That's fine. I'll go with your suggestion.

dbungert (approved):

PR #1151, opened 2021-12-17
Validate UA token & list available services

ogayot:
Hello,

This PR adds a mechanism to check the supplied token and list the available services associated with the subscription.

We rely on the following command to check the token and retrieve the list of services associated. The support for this command is a work in progress.

```bash
$ ubuntu-advantage enable --simulate-with-token "${token}" --format json
```

In case of error, the “$ ubuntu-advantage” command will exit with a non-zero status. When that happens, we will ask the user if he wants to go back or continue anyway. Unfortunately, in case of an invalid token, the “$ ubuntu-advantage” command also exits with a non-zero status. Therefore, we cannot easily differentiate between a network/internal error and an invalid token provided.

### Screen captures:
valid token:
https://user-images.githubusercontent.com/4038023/146580276-9b7964ce-d3dd-4a73-83fb-7e0fe7b8bff8.png

expired token:
https://user-images.githubusercontent.com/4038023/146579696-d986320d-91e6-4323-b503-3fbd46593995.png

invalid token (only shown in dry-run mode currently): 
https://user-images.githubusercontent.com/4038023/146579530-b9c4b2ff-a04d-4222-86e2-f28a1a4b19bd.png

internal / network error:
https://user-images.githubusercontent.com/4038023/146580380-6c4954ff-784f-4280-81ee-e2137bd69631.png

To test the flow in dry run mode, the following rules have been put in place:
* a token starting with "x" will be considered expired
* a token starting with "i" will be considered invalid
* a token starting with "f" will cause an internal error / network error to be reported
* any other token will be considered valid

Thanks,
Olivier

### DIFF
--- a/subiquity/client/controllers/ubuntu_advantage.py
+++ b/subiquity/client/controllers/ubuntu_advantage.py
@@ -18,7 +18,18 @@
 import logging
 from typing import Optional
 
+from subiquitycore.async_helpers import schedule_task
+
 from subiquity.client.controller import SubiquityTuiController
+from subiquity.common.ubuntu_advantage import (
+    InvalidUATokenError,
+    ExpiredUATokenError,
+    CheckSubscriptionError,
+    UAInterface,
+    UAInterfaceStrategy,
+    MockedUAInterfaceStrategy,
+    UAClientUAInterfaceStrategy,
+)
 from subiquity.common.types import UbuntuAdvantageInfo
 from subiquity.ui.views.ubuntu_advantage import UbuntuAdvantageView
 
@@ -32,6 +43,16 @@ class UbuntuAdvantageController(SubiquityTuiController):
     """ Client-side controller for Ubuntu Advantage configuration. """
 
     endpoint_name = "ubuntu_advantage"
+
+    def __init__(self, app):
+        """ Initializer for client-side UA controller. """
+        strategy: UAInterfaceStrategy
+        if app.opts.dry_run:
+            strategy = MockedUAInterfaceStrategy()
+        else:
+            strategy = UAClientUAInterfaceStrategy()
+        self.ua_interface = UAInterface(strategy)
+        super().__init__(app)
 
     async def make_ui(self) -> UbuntuAdvantageView:
         """ Generate the UI, based on the data provided by the model. """
@@ -49,6 +70,31 @@ class UbuntuAdvantageController(SubiquityTuiController):
     def run_answers(self) -> None:
         if "token" in self.answers:
             self.done(self.answers["token"])
+
+    def check_token(self, token: str):
+        """ Asynchronously check the token passed as an argument. """
+        async def inner() -> None:
+            try:
+                svcs = await self.ua_interface.get_avail_services(token=token)
+            except InvalidUATokenError:
+                if isinstance(self.ui.body, UbuntuAdvantageView):
+                    self.ui.body.show_invalid_token()
+            except ExpiredUATokenError:
+                if isinstance(self.ui.body, UbuntuAdvantageView):
+                    self.ui.body.show_expired_token()
+            except CheckSubscriptionError:
+                if isinstance(self.ui.body, UbuntuAdvantageView):
+                    self.ui.body.show_unknown_error()
+            else:
+                if isinstance(self.ui.body, UbuntuAdvantageView):
+                    self.ui.body.show_available_services(svcs)
+
+        self._check_task = schedule_task(inner())
+
+    def cancel_check_token(self) -> None:
+        """ Cancel the asynchronous token check (if started). """
+        if self._check_task is not None:
+            self._check_task.cancel()
 
     def cancel(self) -> None:
         self.app.prev_screen()
--- a/subiquity/client/controllers/ubuntu_advantage.py
+++ b/subiquity/client/controllers/ubuntu_advantage.py
@@ -35,11 +35,6 @@ class UbuntuAdvantageController(SubiquityTuiController):
 
     async def make_ui(self) -> UbuntuAdvantageView:
         """ Generate the UI, based on the data provided by the model. """
-
-        # TODO remove these two lines to enable this screen
-        await self.endpoint.skip.POST()
-        raise Skip("Hiding the screen until we can validate the token.")
-
         path_lsb_release: Optional[str] = None
         if self.app.opts.dry_run:
             # In dry-run mode, always pretend to be on LTS
--- a/subiquity/common/tests/test_ubuntu_advantage.py
+++ b/subiquity/common/tests/test_ubuntu_advantage.py
@@ -1,0 +1,132 @@
+# Copyright 2021 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from subprocess import CalledProcessError, CompletedProcess
+import unittest
+from unittest.mock import patch
+
+from subiquity.common.ubuntu_advantage import (
+    InvalidUATokenError,
+    ExpiredUATokenError,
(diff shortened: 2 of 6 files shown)


subiquity/client/controllers/ubuntu_advantage.py
```diff
@@ -33,13 +44,18 @@ class UbuntuAdvantageController(SubiquityTuiController):
...
+            strategy = UAClientUAInterfaceStrategy()
+        self.ua_interface = UAInterface(strategy)
+        super().__init__(app)
+
     async def make_ui(self) -> UbuntuAdvantageView:
         """ Generate the UI, based on the data provided by the model. """
-
-        # TODO remove these two lines to enable this screen
-        await self.endpoint.skip.POST()
-        raise Skip("Hiding the screen until we can validate the token.")
```
dbungert, line 41: I think this means that on a non-LTS, the user would see
1. screen is presented
2. user inputs 30 character token
3. subiquity informs the user that the release is not supported

Is that accurate?

subiquity/common/tests/test_ubuntu_advantage.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+from subiquitycore.tests.util import run_coro
+
+
+class TestMockedUAInterfaceStrategy(unittest.TestCase):
+    def setUp(self):
+        self.strategy = MockedUAInterfaceStrategy(scale_factor=1_000_000)
+
+    def test_query_info_invalid(self):
+        # Tokens starting with "i" in dry-run mode cause the token to be
```
dbungert, line 36: are the i/f/x sentinel values valid starting token characters?
If so I'd like to pick another strategy for dryrun.

subiquity/common/ubuntu_advantage.py
```diff
@@ -0,0 +1,151 @@
...
+    async def query_info(self, token: str) -> dict:
+        """ Return the subscription info associated with the supplied
+        UA token. No actual query is done to the UA servers in this
+        implementation. Instead, we create a response based on the following
+        rules:
+        * Tokens starting with "x" will be considered expired.
+        * Tokens starting with "i" will be considered invalid.
+        * Tokens starting with "f" will generate an internal error.
+        """
+        await asyncio.sleep(1 / self.scale_factor)
```
dbungert, line 81: We want to leverage the environment variable for consistency with scripts/runtests.sh.  See also https://github.com/canonical/subiquity/blob/1aac580dfd6e70aee6ea1004bf64c03816591f83/subiquitycore/core.py#L69-L70

dbungert: Do you have a feel for how long the validation should normally take?  If significantly longer than 1 second, that would be nice to use before dividing by the timescale.

subiquity/common/ubuntu_advantage.py
```diff
@@ -0,0 +1,151 @@
...
+        return await self.strategy.query_info(token)
+
+    async def get_avail_services(self, token: str) -> list:
+        """ Return a list of available services for the subscription
+        associated with the token provided.
+        """
+        info = await self.get_subscription(token)
+
+        expiration = dt.fromisoformat(info["expires"])
+        if expiration.timestamp() <= dt.utcnow().timestamp():
```
dbungert, line 144: Sometimes units have extremely incorrect system times.  Can we run a test with a valid token and system time that is wrong enough to trigger the expiration?  I'd like to know if the u-a client enforces this already.

dbungert (changes requested):

subiquity/common/tests/test_ubuntu_advantage.py
ogayot: I haven't gotten an "official" answer on this yet but I believe they would indeed be valid starting token characters.
Any suggestion for a different strategy would be welcome. Do you think picking non alphanumerical characters would be enough?

subiquity/common/ubuntu_advantage.py
ogayot: Thanks, I'll make sure to leverage on `SUBIQUITY_REPLAY_TIMESCALE` when creating the `MockedUAInterfaceStrategy` here : https://github.com/canonical/subiquity/pull/1151/files/4fffca685e1047cec6ac620d9ad4a7f37527715f#diff-6363a69b28f40d7cbb460b917d2f53ac172366824645fc7e8a39b024b907a771R51

In normal conditions, the validation seems to be almost instantaneous. But it can take a while is there is some network misconfiguration.
The reason why I added a `sleep` here is so that the "Loading ..." screen has the opportunity to be visible in dry-mode for more than a few milliseconds. (so that we can test what happens it the user clicks on "cancel").

If we leave it to one second without using the timescale, it significantly slows down the unit tests - although we could also use some mocking to avoid the sleep.

What do you think is the best way?

subiquity/common/ubuntu_advantage.py
ogayot: I tried to perform this test but unfortunately, changing the system date leads to TLS certificate validity exception when running `ua attach`. We might need to disable TLS verification in ubuntu-advantage-client or to generate a token that expired just a few days ago.

subiquity/common/tests/test_ubuntu_advantage.py
dbungert: OK.  I think the risk is low of a collision.  I accept this strategy.

subiquity/common/ubuntu_advantage.py
dbungert: I think we're in agreement.

subiquity/common/ubuntu_advantage.py
dbungert: Give this one some more thought.  I lean toward removing this explicit expiration check.

ogayot: I played around with ua attach some more and it seems that the "attach" operation would succeed even if the token is expired. 

I think we have two options:
* removing everything related to token expiration from Subiquity altogether
* changing the workflow to allow the user to go through even if we think the token is expired. That means changing the screen from:
```
Token expired
The Ubuntu Advantage token that you provided has expired.
Please use a different token.
[ Close ]
```
to something like
```
Token expired
The Ubuntu Advantage token that you provided seems to have expired. 
Do you want to go back or continue anyway?
[ Back ] [ Continue anyway ]
```

Do you have a preference?

dbungert: We're due to have a UX discussion on the matter anyhow, so we can discuss as part of that.
Let's go with status-quo until then.

dbungert (approved):
LGTM but take a peek at the merge conflict.

subiquity/client/controllers/ubuntu_advantage.py
```diff
@@ -51,6 +68,31 @@ def run_answers(self) -> None:
         if "token" in self.answers:
             self.done(self.answers["token"])
 
+    def check_token(self, token: str):
+        """ Asynchronously check the token passed as an argument. """
+        async def inner() -> None:
+            try:
+                svcs = await self.ua_interface.get_avail_services(token=token)
```
mwhudson, line 75: Did you consider adding an API for this and doing it in the server? Because I don't think we want to reimplement this logic again in the desktop client.

subiquity/ui/views/ubuntu_advantage.py
```diff
@@ -89,11 +140,107 @@ def on_cancel(_: UbuntuAdvantageForm):
...
+        self.parent = parent
+
+        ok = ok_btn(label=_("OK"), on_press=self.ok)
+
+        title = _("Available Services")
+        header = _("List of services that will be enabled through your "
+                   "Ubuntu Advantage subscription:")
+
+        widgets: List[Widget] = [Text(header)]
+        widgets.extend([Text(f"* {svc['description']}") for svc in services])
```
mwhudson, line 214: I think this dialog should have a blank line between the header and services and between the services and the button. And the button should be focused by default.

Actually, it could probably also have some text along the lines of "you can enable these services using the 'ua' command line tool" between the list and the button.

mwhudson: I think the first two comments apply to the other dialogs too.

mwhudson (commented):
Hi, thanks so much for working on this! I have a few comments but they are pretty minor.

subiquity/ui/views/ubuntu_advantage.py
ogayot: Thanks!
I created https://github.com/canonical/subiquity/pull/1157 to address this comment.

subiquity/client/controllers/ubuntu_advantage.py
ogayot: Thanks! I've carded this since it would indeed be beneficial but requires some more work than mere UI adjustments :)

PR #1152, opened 2022-01-04
Make sure test-this-branch.sh exits if distro-info is not installed

ogayot:
Although the script is running with `set -e`, having two distinct invocations of a subshell in the same instruction masks failures in the first subshell invocation. It is similar in essence to what the `pipefail` option controls.

As a consequence, the following instruction does not fail if `distro-info` is not installed:

```bash
  isoname=$(distro-info -d)-live-server-$(dpkg --print-architecture).iso
```

And therefore, we end up with something like:

```bash
  isoname=-live-server-amd64.iso
```

Fixed by first assigning the value of `$(distro-info -d)` to a variable.


dbungert (approved):
Nice catch, and some unfortunate behavior for 'set -e'.

PR #1153, opened 2022-01-04
Add support for ubuntu-advantage with autoinstall

ogayot:
Hello,

This PR adds support for Ubuntu Advantage token when running Subiquity with `--autoinstall`.
Also added a dummy token in the integration tests.

Thanks,
Olivier

refs FR-1652

### DIFF
--- a/autoinstall-schema.json
+++ b/autoinstall-schema.json
@@ -260,6 +260,14 @@
                 }
             ]
         },
+        "ubuntu-advantage": {
+            "type": "object",
+            "properties": {
+                "token": {
+                    "type": "string"
+                }
+            }
+        },
         "proxy": {
             "type": [
                 "string",
--- a/autoinstall-schema.json
+++ b/autoinstall-schema.json
@@ -264,7 +264,11 @@
             "type": "object",
             "properties": {
                 "token": {
-                    "type": "string"
+                    "type": "string",
+                    "minLength": 24,
+                    "maxLength": 30,
+                    "pattern": "^C[1-9A-HJ-NP-Za-km-z]+$",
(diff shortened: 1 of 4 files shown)


autoinstall-schema.json
```diff
@@ -260,6 +260,14 @@
                 }
             ]
         },
+        "ubuntu-advantage": {
+            "type": "object",
+            "properties": {
+                "token": {
+                    "type": "string"
```
dbungert, line 267: I'm considering if we want to do crude token checks at the schema level.
For example, 30 chars, upper/lower/digits only.
If the token changes format we'd need to update but I'm OK with that.
What do you think?

ogayot: Now that the format is defined, I was thinking at using:
```json
{
    "type": "string",
    "minLength": 24,
    "maxLength": 30,
    "pattern": "^C[1-9A-HJ-NP-Za-km-z]{23,29}$
}
```

but maybe it's overkill.

dbungert: I don't think that check is overkill, except maybe the {23,29} is redundant with min/maxLength. Let's also give this a brief comment explaining the format at a basic level - the constant at the front and a pointer to https://pkg.go.dev/github.com/btcsuite/btcutil/base58#CheckEncode would be good

dbungert (approved):

dbungert (changes requested):
sorry, meant to click request changes for the schema enforcement changes + requested comment

autoinstall-schema.json
ogayot: Done, thanks!
Added the pattern without the redundant check.
Also added the brief explanation in the "description" field of the schema.

dbungert (approved):

PR #1154, opened 2022-01-05
apt: fix disable_components w/ autoinstall

dbungert:
* The mixture of storage for disable_components in and out of the config
  dict was confusing and caused bugs, depending on the flow
* disable_components now solely lives in the dict, like other items

### DIFF
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -165,11 +165,8 @@ class MetaController:
     async def free_only_POST(self, enable: bool) -> None:
         self.free_only = enable
         to_disable = {'restricted', 'multiverse'}
-        if enable:
-            # enabling free only mode means disabling components
-            self.app.base_model.mirror.disable_components |= to_disable
-        else:
-            self.app.base_model.mirror.disable_components -= to_disable
+        # enabling free only mode means disabling components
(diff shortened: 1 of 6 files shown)


subiquity/server/server.py
```diff
@@ -165,11 +165,8 @@ async def free_only_GET(self) -> bool:
     async def free_only_POST(self, enable: bool) -> None:
         self.free_only = enable
         to_disable = {'restricted', 'multiverse'}
-        if enable:
-            # enabling free only mode means disabling components
-            self.app.base_model.mirror.disable_components |= to_disable
-        else:
-            self.app.base_model.mirror.disable_components -= to_disable
+        # enabling free only mode means disabling components
+        self.app.base_model.mirror.disable_components(to_disable, enable)
```
ogayot, line 169: I don't know if it is a valid use case, but are free-only and disable_components mutually exclusive options?
If we configure `disable_components=["multiverse"]` explicitly, shouldn't
```
POST("/a/free_only", data={"enable", False})
```
be a no-op rather than making `disable_components` empty?

ogayot: Thanks. That's fair. I suggest to keep it the way it is then.

ogayot (approved):

PR #1155, opened 2022-01-06
Fix language packs detection in integration tests

ogayot:
Hello,

While looking at the logs from the integration tests, I observed some weird behavior:
```bash
      ++ ls .subiquity/var/cache/apt/archives/
      + '[' -z 'language-pack-en:amd64
      wamerican:amd64
      wbritish:amd64 | grep ' ']'
```

this comes from this instruction:

```bash
            if [ -z "$( ls .subiquity/var/cache/apt/archives/) | grep $lang" ] ; then
```

One issue is that the `grep` is not being executed. It is used as a string literal and therefore causes the `-z` check to always be false. I rewrote the test and dropped the use of the subshell since we don't need it.

The other issue is that the lang variable is empty. I figured out it's empty because etc/default/locale contains:

```bash
LANG=en_US
```

and we expect it to contain:

```bash
LANG="en_US"
```

I changed the pattern matching to not expect the quotes.

Thanks,
Olivier

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -90,7 +90,7 @@ validate () {
                 exit 1
             fi
             lang="$(grep -Eo 'LANG="([^.@ _]+)' .subiquity/etc/default/locale | cut -d \" -f 2)"
-            if [ -z "$( ls .subiquity/var/cache/apt/archives/) | grep $lang" ] ; then
+            if ! ls .subiquity/var/cache/apt/archives/ | grep --fixed-strings --quiet -- "$lang"; then
                 echo "expected $lang language packs in directory var/cache/apt/archives/"
                 exit 1
             fi
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -89,7 +89,8 @@ validate () {
                 echo "user not assigned with the expected group sudo"
                 exit 1
             fi
-            lang="$(grep -Eo 'LANG="([^.@ _]+)' .subiquity/etc/default/locale | cut -d \" -f 2)"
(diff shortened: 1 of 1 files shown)


scripts/runtests.sh
```diff
@@ -89,8 +89,8 @@ validate () {
                 echo "user not assigned with the expected group sudo"
                 exit 1
             fi
-            lang="$(grep -Eo 'LANG="([^.@ _]+)' .subiquity/etc/default/locale | cut -d \" -f 2)"
-            if [ -z "$( ls .subiquity/var/cache/apt/archives/) | grep $lang" ] ; then
+            lang="$(grep -Eo 'LANG=([^.@ _]+)' .subiquity/etc/default/locale | cut -d= -f 2-)"
```
CarlosNihelton, line 92: I'd keep the quote removal in the end. I'm not quite sure under what circumstances `update-locale` quotes the values it writes to the `/etc/default/locale` file. In my system file it has quotes, under the integration tests it seems not to add them. I'd rather follow your approach with the following addition:
```suggestion
            lang="$(grep -Eo 'LANG=([^.@ _]+)' .subiquity/etc/default/locale | cut -d= -f 2- | cut -d\" -f 2-)"
```

In case of absent quotes that wouldn't fail anyway, so we should be good to go. Other than that, LGTM.

CarlosNihelton (approved):
Just a comment for your appreciation.

Thanks for spotting that.

scripts/runtests.sh
ogayot: Good point, I didn't realize `cut` would print the string if no delimiter is found.
Thanks!

dbungert (approved):

PR #1156, opened 2022-01-06
Add support for APT preferences through autoinstall in Subiquity

ogayot:
Hello,

This PR adds partial support to Subiquity for apt preferences (aka. pinning / priorities) when using autoinstall.

The autoinstall schema for "apt" now supports the "preferences" configuration. Each preference element should contain the properties described below, that 1:1 map with the keywords from apt_preferences(5):

```
  * package      <-> Package:
  * pin          <-> Pin:
  * pin-priority <-> Pin-Priority:
```

These preferences are forwarded to curtin through subiquity-curtin-apt.conf. Another PR will need to be done in curtin to have full support for this feature.

### DIFF
--- a/subiquity/server/controllers/mirror.py
+++ b/subiquity/server/controllers/mirror.py
@@ -50,6 +50,28 @@ class MirrorController(SubiquityController):
                     'enum': ['universe', 'multiverse', 'restricted',
                              'contrib', 'non-free']
                 }
+            },
+            "preferences": {
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "properties": {
+                        "package": {
+                            "type": "string",
+                        },
+                        "pin": {
+                            "type": "string",
+                        },
+                        "pin-priority": {
+                            "type": "integer",
+                        },
+                    },
+                    "required": [
+                        "package",
+                        "pin",
+                        "pin-priority",
+                    ],
+                }
             }
         }
     }
--- a/autoinstall-schema.json
+++ b/autoinstall-schema.json
@@ -306,6 +306,28 @@
                             "non-free"
                         ]
                     }
+                },
+                "preferences": {
+                    "type": "array",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "package": {
+                                "type": "string"
+                            },
+                            "pin": {
+                                "type": "string"
+                            },
+                            "pin-priority": {
+                                "type": "integer"
+                            }
+                        },
+                        "required": [
+                            "package",
+                            "pin",
+                            "pin-priority"
+                        ]
+                    }
                 }
             }
         },
--- a/examples/autoinstall.yaml
+++ b/examples/autoinstall.yaml
@@ -22,6 +22,9 @@ apt:
   disable_components:
     - non-free
     - restricted
+  preferences:
+    - {package: "python3-*", pin: "origin *ubuntu.com*", pin-priority: 200}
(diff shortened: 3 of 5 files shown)


subiquity/server/controllers/mirror.py
```diff
@@ -50,6 +50,28 @@ class MirrorController(SubiquityController):
...
+            "preferences": {
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "properties": {
+                        "package": {
+                            "type": "string",
+                        },
+                        "pin": {
+                            "type": "string",
```
dbungert, line 63: 1. What are thoughts on the Pin value?  
If desired, we could be more specific and limit Pin to just having lines that start with origin, release, version (are there others?)
(That's a question, not a demand)

2. The apt_preferences manpage documents Pin like so
```
Pin: origin "ftp.de.debian.org"
```
Do things get weird in the yaml if part of it is quoted?

ogayot: Thanks for the feedback!

> 1. What are thoughts on the Pin value?
>    If desired, we could be more specific and limit Pin to just having lines that start with origin, release, version (are there others?)
>    (That's a question, not a demand)

I have been hesitant about that very thing. At first I wanted to establish stricter rules but I realized that the driver for this use-case is to avoid the current workaround (which consists of using `late_commands`).

When using `late_commands`, there is no safety net to guard against non-existent keywords or invalid stanzas. So I have assumed that using stricter rules would not bring that much more to the table. But I would be happy to change that if you think it would be better.

> 2. The apt_preferences manpage documents Pin like so
> 
> ```
> Pin: origin "ftp.de.debian.org"
> ```
> 
> Do things get weird in the yaml if part of it is quoted?

It should be fine, APT behaves the same if the origin field is quoted or unquoted:
https://salsa.debian.org/apt-team/apt/-/blob/main/apt-pkg/versionmatch.cc#L124
```cpp
   if (Type == Origin)
   {
      if (Data[0] == '"' && Data.length() >= 2 && Data.end()[-1] == '"')
	 OrSite = Data.substr(1, Data.length() - 2);
      else
	 OrSite = Data;
      return;
   }   
```
as for YAML itself, these values are equivalent (as far as my tests seem to show):
```yaml
pin: origin "ftp.de.debian.org"
pin: 'origin "ftp.de.debian.org"'
pin: "origin \"ftp.de.debian.org\""
```
However, I can foresee the character `"\n"` as being problematic when de-serializing the data into stanzas. E.g.:
```yaml
preferences:
  {package: dummy, pin: "some\nvalue", pin-priority: -1}
```
->
```
Package: dummy
Pin: some
value
Pin-Prioriy: -1
```
Maybe it's worth rejecting this character from pin, package and pin-priority. To be fair, there might be other characters that could make things go wrong. But since the stanzas from APT are more user-friendly than machine-friendly, it's difficult to come up with something perfect IMO.

Thanks,
Olivier

dbungert (approved):

dbungert:
I think I want to land this after we have the matching curtin change.

dbungert:
So for this one we have the curtin changes needed, right?
Next we need to update the curtin commit hash in the snapcraft file to point to a newer version.

ogayot:
> So for this one we have the curtin changes needed, right? Next we need to update the curtin commit hash in the snapcraft file to point to a newer version.

That's right, thanks!
I have added a commit to change the version of curtin being picked up when building the snap.

PR #1158, opened 2022-01-12
Embed ubuntu-advantage-tools into Subiquity snap and use it

ogayot:
We now embed the package `ubuntu-advantage-tools` in the Subiquity snap.
The goal is to make sure that u-a-c implements the `status --simulate-with-token` option, that might not be present in old ISOs.

We now have two different installations of ubuntu-advantage-tools (one from the ISO and one from the snap).
When invoking `ubuntu-advantage`, we need to make sure we use the Python installation from the snap rather than from base ISO. Similarly, we also need to make sure we use right `ubuntu-advantage` entry point (the one from the snap).

NOTE: the version of `ubuntu-advantage` that we place in the snap is currently too old. The version is picked up from `focal-updates`. We need to wait until 27.5~20.04.1 lands in `focal-updates` to get `--simulate-with-token` to work.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -114,6 +114,7 @@ parts:
       - python3-requests-unixsocket
       - python3-requests
       - python3-pyudev
+      - ubuntu-advantage-tools
     stage:
       - "*"
       - -bin/python
--- a/subiquity/client/controllers/ubuntu_advantage.py
+++ b/subiquity/client/controllers/ubuntu_advantage.py
@@ -16,6 +16,7 @@
 """
 
 import logging
+import os
 
 from subiquitycore.async_helpers import schedule_task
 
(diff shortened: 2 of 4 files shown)


bin/run-ua-from-snap
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+set -u
+
+export PYTHONPATH="$SNAP"/usr/lib/python3/dist-packages
+"$SNAP"/usr/bin/ubuntu-advantage "$@"
```
dbungert, line 7: I think in this scenario it's going to use the system python instead of the one in the Subiquity snap.
We have been setting the PYTHON env variable, which can be done in the snapcraft file.
See this example:
https://github.com/canonical/subiquity/blob/main/snapcraft.yaml#L14

If you do that, do you need the wrapper script?  I though it picked up dist-packages automatically and that it was the site packages causing the problem.

ogayot: Thanks for the feedback.
You are right, the following construct picks up the right version of Python and `ubuntu-advantage`:

```bash
$PYTHON "${SNAP}"/usr/bin/ubuntu-advantage --version
```
I guess I can get rid of the wrapper script.

ogayot: Done! I'm not forcing the use of the Python installation from the snap.

dbungert (approved):

PR #1161, opened 2022-01-13
Fix two bug reports from ubuntu-desktop-installer

dbungert:
* apt_configurer cleanup is async, so needs an await
* ovmnt.lowers is a str, so fix list concatenation

### DIFF
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -15,6 +15,7 @@
 
 import datetime
 import functools
+import logging
 import os
 import shutil
 import tempfile
@@ -28,6 +29,8 @@ from subiquitycore.lsb_release import lsb_release
 from subiquitycore.utils import arun_command
 
 from subiquity.server.curtin import run_curtin_command
+
+log = logging.getLogger('subiquity.server.apt')
 
 
 class _MountBase:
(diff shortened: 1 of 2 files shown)


subiquity/server/apt.py
```diff
@@ -75,7 +78,7 @@ def _lowerdir_for_mnt(mnt):
 def _lowerdir_for_ovmnt(ovmnt):
     # One cannot indefinitely stack overlayfses so construct an
     # explicit list of the layers of the overlayfs.
-    return lowerdir_for(ovmnt.lowers + [ovmnt.upperdir])
+    return lowerdir_for([ovmnt.lowers] + [ovmnt.upperdir])
```
mwhudson, line 81: [ovmnt.lowers] + [ovmnt.upperdir] can be written [ovmnt.lowers, ovmnt.upperdir] :)

It took some me some work to convince myself this was correct. I really should have written unit tests for this :/

mwhudson (approved):
Thanks for cleaning up my mess!

subiquity/server/apt.py
dbungert: lol - Thanks!

PR #1162, opened 2022-01-14
Fix some typing issues reported by mypy, including potential crash when comparing snap versions

ogayot:
I ran mypy over the source tree and had a look at some of the most obvious issues reported.

One of the most concerning was this one:
```
subiquity/common/snap.py:84: error: Unsupported operand types for > ("int" and "None")
subiquity/common/snap.py:84: error: Unsupported operand types for < ("int" and "None")
subiquity/common/snap.py:84: error: Unsupported left operand type for > ("None")
subiquity/common/snap.py:84: note: Both left and right operands are union
```
There is a part of a false-positive here. mypy is not able to infer that, but the operands in question can't be a mix of`int` and `None` based on the logic that happens before.
However, there is one scenario that leads to both operands of the comparison being `None`. And this scenario occurs when comparing a snap version to itself  and when there is no git_build_id specified:

```python
SnapVersion.from_string("1.0.0") > SnapVersion.from_string("1.0.0")
TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
```

Fixed by first converting the build ID to a number before making the comparison. Also added a unit test to cover this scenario.

Also added some type hinting in base classes of controllers to silence some redundant errors.

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -21,6 +21,7 @@ import os
 import signal
 import sys
 import traceback
+from typing import Optional
 
 import aiohttp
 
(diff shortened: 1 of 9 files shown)


dbungert (approved):

mwhudson (approved):
Is there a reason this didn't get merged?

dbungert:
Nope!

PR #1167, opened 2022-01-19
snaplist: handle cancel error paths

dbungert:


subiquity/client/controllers/snaplist.py
```diff
@@ -63,7 +65,11 @@ def cancel(self, sender=None):
         self.app.prev_screen()
 
     async def get_list_wait(self):
-        return await self.endpoint.GET(wait=True)
+        try:
+            return await self.endpoint.GET(wait=True)
+        except aiohttp.client_exceptions.ServerDisconnectedError as sde:
+            log.debug('failed to get snaplist', sde)
+            return SnapListResponse(status=SnapCheckState.FAILED)
```
mwhudson, line 72: I don't think this is the correct thing really, we should retry on cancellation in the server.

subiquity/ui/views/snaplist.py
```diff
@@ -463,10 +463,13 @@ def make_main_screen(self, data):
...
-        self.controller.done(sorted(
-            self.selections_by_name.values(),
-            key=lambda s: s.name))
+        if getattr(self, "selections_by_name", None):
+            log.debug("snaps to install %s", self.selections_by_name)
+            self.controller.done(sorted(
+                self.selections_by_name.values(),
+                key=lambda s: s.name))
+        else:
+            self.controller.done([])
```
mwhudson, line 472: Um. I don't understand the flow that gets to this path, but probably just setting `self.selections_by_name` to {} in `__init__` is easier?

mwhudson (changes requested):

subiquity/ui/views/snaplist.py
dbungert: Yes, that's much more sensible

mwhudson (approved):
Nice. Did SingleInstanceTask make sense? :slightly_smiling_face: (I'm not completely sure it does to me any more :slightly_frowning_face:)

dbungert:
I need to get a better understanding of how the Propagate Errors bit is working, but otherwise it's fine.  Maybe we think about the name or some comments.

PR #1168, opened 2022-01-19
Add --use-fuse switch to kvm-test --install so it can run as non-root

ogayot:
Passing `--use-fuse` to `scripts/kvm-test.py --install` allows to run as non-root.

It requires installation of the package fuseiso so the switch is disabled by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,8 @@ aptdeps:
 		libnl-route-3-dev python3-attr python3-distutils-extra python3-requests \
 		python3-requests-unixsocket python3-jsonschema python3-apport \
 		python3-bson xorriso isolinux python3-aiohttp cloud-init ssh-import-id \
-		curl jq build-essential python3-pytest python3-async-timeout language-selector-common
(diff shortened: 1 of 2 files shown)


dbungert (changes requested):
I love the idea so much I want it to be the default and not an option.
Would you add fuse / fuseiso to the packages installed in `make aptdeps`?

mwhudson:
Would you be able to do something similar for https://github.com/mwhudson/livefs-editor? :)

ogayot:
> I love the idea so much I want it to be the default and not an option. Would you add fuse / fuseiso to the packages installed in `make aptdeps`?

sure thing :)

> Would you be able to do something similar for https://github.com/mwhudson/livefs-editor? :)

I can give it a try !

dbungert (approved):

dbungert:
Great idea, thanks!  It bugged me to need sudo in this script and I now have hope that we can make it not need root at all.

PR #1169, opened 2022-01-20
initialize MirrorController.configure_once to False

mwhudson:
oops :/


ogayot (approved):
Thanks!

ogayot:
With this patch, I can still observe two distinct invocations of `curtin apt-config` during an installation but at least they both succeed now ; so this is already better.

PR #1170, opened 2022-01-20
Fix hanging when network is the last postinst model configured

ogayot:
When running an autoinstallation with only “network” as an interactive section, the installation hanged after the following steps:
```
  finalizing installation
    running 'curtin hook'
      curtin command hook
  executing late commands
```
where in normal conditions, the next step is:
```
  final system configuration
```
Also, querying the meta/status endpoint at this point shows that the app is in POST_WAIT state.

Rationale:

* the network model is declared as both an “install” model and a “postinstall” model (it is the only one we have AFAIK)
* when calling `.configured()` for a given model, we only trigger /at max/ one event. Either:
  * `install_event` (if the model is an “install” model - and it is the last one configured)
  * `postinstall_event` (if the model is a “postinstall” model but not an “install” model - and it is the last one configured)
  * no event (as in most cases)

* So when we call `.configured()` for the network model, we /can/ trigger the `install_event` but can never trigger the `postinstall event`.
Therefore Subiquity, will wait forever until something triggers the `postinstall_event`.

Fixed by accepting to trigger the `install_event` and the `postinstall_event` in a single call to `.configured()`.

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -18,6 +18,7 @@ from collections import OrderedDict
 import functools
 import logging
 import os
+from typing import Set
 import uuid
 import yaml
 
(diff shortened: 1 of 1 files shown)


dbungert (commented):
LGTM but I'd like to wait for @mwhudson's feedback on network being in both categories.

ogayot:
> LGTM but I'd like to wait for @mwhudson's feedback on network being in both categories.

For reference, network was added as a postinstall model in the following PR:
https://github.com/canonical/subiquity/pull/1052

mwhudson (approved):
Good spot! I've had it in the back of my mind to refactor this state machine into something more reusable and then we could write unit tests for this sort of thing...

PR #1173, opened 2022-01-24
snap: fix os-prober related non-amd64 build failures

dbungert:
snap builds for non-amd64 are failing on the current patch-based plan,
as the list of included files are different on different arches.

* revert to previous logic
* continue to advertise os-prober as a snap binary
* fix file selection for files to sed


mwhudson (approved):
Baaaaah. OK.

PR #1174, opened 2022-01-25
Don't list UA services that are not entitled

ogayot:
Hello,

So far, for a valid contract token, we would list all the UA services that are marked "available".

After discussing with @alnvdl, it turns out that we also want to make sure these services are marked "entitled":

_available_ means it is available for the current machine type (e.g., fips may be available for a x86_64 machine running focal). This is irrespective of the contract; whereas
_entitled_ means that your contract gives you access to the service

Instead of only checking if a given UA service is available, we now also check if it is entitled.

Also fixed the unit test that expected "entitled" to be a boolean field ; but is actually a "yes" or a "no" string.

### DIFF
--- a/subiquity/common/ubuntu_advantage.py
+++ b/subiquity/common/ubuntu_advantage.py
@@ -147,9 +147,9 @@ class UAInterface:
         """ Return a dictionary containing the subscription information. """
         return await self.strategy.query_info(token)
 
-    async def get_avail_services(self, token: str) -> List[dict]:
-        """ Return a list of available services for the subscription
-        associated with the token provided.
+    async def get_activable_services(self, token: str) -> List[dict]:
+        """ Return a list of activable services (i.e. services that are
+        entitled to the subscription and available on the current hardware).
         """
         info = await self.get_subscription(token)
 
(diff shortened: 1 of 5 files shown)


subiquity/common/ubuntu_advantage.py
```diff
@@ -158,7 +158,7 @@ async def get_avail_services(self, token: str) -> List[dict]:
             raise ExpiredUATokenError(token, expires=info["expires"])
 
         def is_avail_service(service: dict) -> bool:
```
dbungert, line 160: 1) Now that we check for both available && entitled, we probably should adjust the name of the methods to not use the term 'available' - `get_display_services` maybe?  Is there a term we can borrow from u-a for something that is both available and entitled?  Does entitled imply available?
2) The commit message is good material and I'd like to include it as a direct code comment.

dbungert (changes requested):

subiquity/common/ubuntu_advantage.py
ogayot: I agree. The name seems confusing now.
I guess an adequate term would be "enableable" ; which is surprisingly difficult to write. Are you happy with the term "activable" ?
I will include the content of the commit message as soon as we agree on the wording :)

dbungert: Sure, let's do activable.  Thanks!

dbungert (approved):

PR #1175, opened 2022-01-25
api test: run in parallel

dbungert:
By splitting up the output directory instead of just writing to .subiquity,
we can run api tests in parallel.  About 7x speedup on my machine over running
them serially.

### DIFF
--- a/subiquity/tests/api/test_api.py
+++ b/subiquity/tests/api/test_api.py
@@ -102,14 +102,16 @@ class Server(Client):
         except aiohttp.client_exceptions.ServerDisconnectedError:
             return
 
-    async def spawn(self, socket_path, machine_config, bootloader='uefi'):
+    async def spawn(self, output_base, socket, machine_config,
+                    bootloader='uefi'):
         env = os.environ.copy()
         env['SUBIQUITY_REPLAY_TIMESCALE'] = '100'
-        cmd = 'python3 -m subiquity.cmd.server --dry-run' \
-              + ' --bootloader ' + bootloader \
-              + ' --socket ' + socket_path \
-              + ' --machine-config ' + machine_config
-        cmd = cmd.split(' ')
+        cmd = ['python3', '-m', 'subiquity.cmd.server',
+               '--dry-run',
+               '--bootloader', bootloader,
+               '--socket', socket,
+               '--output-base', output_base,
+               '--machine-config', machine_config]
         self.proc = await astart_command(cmd, env=env)
 
     async def close(self):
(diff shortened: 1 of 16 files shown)


subiquitycore/controllers/network.py
```diff
@@ -134,7 +134,7 @@ def __init__(self, app):
         super().__init__(app)
         self.apply_config_task = SingleInstanceTask(self._apply_config)
         if self.opts.dry_run:
-            self.root = os.path.abspath(".subiquity")
+            self.root = self.opts.output_base
```
ogayot, line 137: are we fine with `.subiquity` (i.e. the default value for `--output-base`) no longer being an absolute path now?
I don't know the reason why we resolved it into an absolute path before ; just wondering.

subiquity/tests/api/test_api.py
```diff
@@ -102,14 +102,15 @@ async def server_shutdown(self, immediate=True):
         except aiohttp.client_exceptions.ServerDisconnectedError:
             return
 
-    async def spawn(self, socket_path, machine_config, bootloader='uefi'):
+    async def spawn(self, tempdir, socket, machine_config, bootloader='uefi'):
```
ogayot, line 105: Maybe `output_base` is better than `tempdir` in this context? I don't know

subiquity/tests/api/test_api.py
```diff
@@ -102,14 +102,15 @@ async def server_shutdown(self, immediate=True):
...
-              + ' --bootloader ' + bootloader \
-              + ' --socket ' + socket_path \
-              + ' --machine-config ' + machine_config
-        cmd = cmd.split(' ')
+        cmd = ['python3', '-m', 'subiquity.cmd.server',
+               '--dry-run',
+               '--bootloader', bootloader,
+               '--socket', socket,
+               '--output-base', tempdir,
+               '--machine-config', machine_config]
```
ogayot, line 113: This looks much cleaner without the string operations + split! Thanks

ogayot (approved):
Good idea!

subiquity/tests/api/test_api.py
dbungert: I like that better, will do

subiquitycore/controllers/network.py
dbungert: Safer to keep it.

mwhudson:
Thanks for doing this. BTW, I think if we make output_base '/' in a non-dry-run build we can probably get rid of a lot of dry-run checks here and there...

PR #1177, opened 2022-01-28
serializer: fix and test for KeyError '$type'

dbungert:


mwhudson:
Oh hm, I'd failed to understand why this change is needed and now I'm a bit less sure it's a good idea. In general I'd expect a client to deserialize and reserialize the objects it gets from the API so that would drop the $type but I guess the API test doesn't do that. Would it be super painful to .pop('$type') where it's needed in the api tests? Not 100% opposed to this if it would be painful. I should try running the api tests in your other branch with this bit backed out I guess :)

dbungert:
Use case:
resp = get('/storage/v2')
select Partition from response
because Partition is in a list of Partitions and Gaps, Partition contains '$type' field
post('/storage/v2/delete_partition', partition_to_delete)

If the caller has to remove the $type field, that feels really strange - why would subiquity send a Partition with a $type field then not accept it?
If the fix is not applied and the caller doesn't remove the $type field, we get an exception of the form `KeyError '$type'`
I'd rather have the deserializer be permissive about accepting that field, since the serializer is the one that adds it.
What do you think?

mwhudson (approved):
I'm still not super happy about this but I guess the thing that is really bothering me is how I represented Union in the serialization. It seems cattrs does things automagically if the two attr classes have different sets of field names: https://cattrs.readthedocs.io/en/latest/unions.html. Ah well. Let's merge this :)

PR #1178, opened 2022-02-01
Expose endpoint to get list of UA activable services

ogayot:
The list of activable UA services was only retrieved from the client-side of Subiquity (using `ubuntu-advantage-tools`). Therefore, the desktop installer would need to reimplement the same logic should they need access to the list of services ; which is inconvenient.

We now expose a new endpoint in the API that takes the token as a parameter and returns a status (+ a list of services if the token is valid and not expired).

```
  /a/ubuntu_advantage/check_token?token="XXXX"
```

This new endpoint is a read-only GET endpoint. It is not designed as a replacement for the POST to `/a/ubuntu_advantage` that the client must (still) do to pass the token to the model.

We now use this new endpoint internally in Subiquity so that the retrieval of the activable services is done on the server-side only.


dbungert (approved):

subiquity/client/controllers/ubuntu_advantage.py
```diff
@@ -44,21 +37,6 @@ class UbuntuAdvantageController(SubiquityTuiController):
...
-            executable = (
-                os.environ["PYTHON"],
-                os.path.join(os.environ["SNAP"], "usr/bin/ubuntu-advantage"),
-            )
-            strategy = UAClientUAInterfaceStrategy(executable=executable)
-        self.ua_interface = UAInterface(strategy)
-        super().__init__(app)
-
     async def make_ui(self) -> UbuntuAdvantageView:
         """ Generate the UI, based on the data provided by the model. """
```
mwhudson, line 41: github won't let me comment on the relevant code, but the decision on whether or not to skip this page should probably be more server-side too (fine for a follow up branch!)

subiquity/client/controllers/ubuntu_advantage.py
```diff
@@ -77,21 +55,19 @@ def run_answers(self) -> None:
...
-            except CheckSubscriptionError:
+            elif answer.status == TokenStatus.UNKNOWN_ERROR:
                 if isinstance(self.ui.body, UbuntuAdvantageView):
                     self.ui.body.show_unknown_error()
             else:
                 if isinstance(self.ui.body, UbuntuAdvantageView):
-                    self.ui.body.show_activable_services(svcs)
+                    self.ui.body.show_activable_services(answer.services)
 
         self._check_task = schedule_task(inner())
```
mwhudson, line 73: again github won't let me comment in the right place, but below here is a check "self._check_task is not None" but self._check_task is never initialized to None.

subiquity/server/controllers/ubuntu_advantage.py
```diff
@@ -15,9 +15,23 @@
...
+import os
 
 from subiquity.common.apidef import API
-from subiquity.common.types import UbuntuAdvantageInfo
+from subiquity.common.types import (
+    UbuntuAdvantageInfo,
+    UbuntuAdvantageCheckTokenAnswer,
+    UbuntuAdvantageCheckTokenStatus,
+)
+from subiquity.common.ubuntu_advantage import (
```
mwhudson, line 26: Pedantry, but this module can move to subiquity.server now I think?

subiquity/common/types.py
```diff
@@ -394,6 +394,20 @@ class UbuntuAdvantageInfo:
...
+    INVALID_TOKEN = enum.auto()
+    EXPIRED_TOKEN = enum.auto()
+    UNKNOWN_ERROR = enum.auto()
+
+
+@attr.s(auto_attribs=True)
+class UbuntuAdvantageCheckTokenAnswer:
+    status: UbuntuAdvantageCheckTokenStatus
+
+    services: Optional[List[dict]]
```
mwhudson, line 408: Is 'dict' all we want to say about the contents here? I guess it's just want comes out of the ua binary so perhaps it's not worth saying overmuch. But afaict the client only looks for the description, so maaybe this could just be a list of descriptions?

Not going to insist on any change here.

mwhudson (approved):
Thanks! It speaks well to your initial design that this is pretty non-disruptive. Some comments but nothing major.

subiquity/common/types.py
ogayot: I have been originally hesitant about this one ; namely because I didn't know what information would be valuable for the user. The description makes sense because we display it but I wouldn't be surprised if the name would also be useful as an identifier e.g. maybe at some point we want to do:

```python
is_livepatch_activable = any([service["name"] == "livepatch" for service in services])
```

I don't know if it makes sense to keep the other keys around:
* entitled -> at this point we know it's true
* auto_enabled -> TBH, I don't know what this is ; I should probably check
* available -> at this point we know it's true

I'd be happy to change but I'm not sure what direction to take yet.

subiquity/server/controllers/ubuntu_advantage.py
ogayot: True!
Maybe it would make sense to keep the exception types (e.g., `InvalidUAToeknError`, `ExpiredUATokenError`, `CheckSubscriptionError`) under `common` so they can be reused on the client side (which we're not doing ATM) but we can come back to it later and move them to common if needed. I'll move everything under `server` for now.

subiquity/common/types.py
ogayot: Using a `typing.TypedDict` would probably make sense.
```python
from typing import TypedDict
class ServiceDict(TypedDict):
    name: str
    description: str
```
And then we'd need to assert the presence of these two keys (if they're the ones we want to include) after calling `u-a-c`

subiquity/client/controllers/ubuntu_advantage.py
ogayot: Thanks! I'll go for the follow-up branch :)

I think it would make sense to have a generic way to skip a page based on the server controller's decision. Maybe by marking the controller as non-interactive if we can?

ogayot:
* Moved `subiquity/common/ubuntu_advantage.py` and `subiquity/common/tests/test_ubuntu_advantage.py` under `subiquity/server/`.
* Initialized `_check_token_task` in the initializer of the UA client-side controller.

subiquity/common/types.py
mwhudson: Is there a difference between TypedDict and an attr.s class for this sort of thing really? But that doesn't support extra keys either. One approach would be to do some filtering in check_token_GET to only the keys that the api defines I guess. Or just leave this as is for now :)

Speaking generally I don't want us to wast time thinking too hard about "what might be useful in the future" for api stuff. We have very tight bindings between the clients and servers and that's fine -- if needs change, the API changes :)

mwhudson (approved):

subiquity/client/controllers/ubuntu_advantage.py
mwhudson: Yeah, overriding interactive() is the cleanest and most generic way to make the client skip a page. Maybe we should support raising a particular exception from a GET method as well to trigger the same behaviour, that would be a fairly easy change in the middleware.

subiquity/common/types.py
ogayot: I'll go with the `attr.s` + filtering. I think it makes the most sense. Thanks

ogayot: I decided to filter the keys of the services soon after calling `u-a-c` (i.e. just after filtering out the services that are not activable). Because some unit tests were expecting the "entitled" & "available" keys to still be around, I had to rework the unit tests a little.

ogayot:
Made the new GET endpoint expect the token in the HTTP body - rather than the query string - to avoid ending up in the access log.

PR #1179, opened 2022-02-03
Server-side implementation of third-party drivers

ogayot:
This is the server side implementation of the third-party drivers installation.

This PR is a rework of https://github.com/canonical/subiquity/pull/1131 that has been around for a while.

It only contains the server-side implementation of the third-party drivers installation.
Currently, an autoinstall section controls if the third-party drivers get installed.

```yaml
drivers:
    install: true
```
or (the default)
```yaml
drivers:
    install: false
```
Example output from a GET call to `/a/drivers`:
```json
{
    "install": true,
    "drivers": ["nvidia-driver-470"]
}
```
An empty array for `drivers` means that no drivers are available. However, this information (i.e. whether drivers are available or not) is only available after reaching some point in the installation process. `drivers` will be set to `null` until the information is available:
```json
{
    "install": true,
    "drivers": null
}
```
I have yet to test this on hardware that has a nvidia card or something reported by `ubuntu-drivers list --recommended`.


subiquity/server/runner.py
```diff
@@ -24,22 +24,27 @@ class LoggedCommandRunner:
     def __init__(self, ident):
         self.ident = ident
 
-    async def start(self, cmd):
-        proc = await astart_command([
-            'systemd-cat', '--level-prefix=false', '--identifier='+self.ident,
-            ] + cmd)
+    async def start(self, cmd, *, capture=False):
```
mwhudson, line 27: So I know I talked to @dbungert about this but I can't remember if I talked to you about this (or if Dan talked to you about this) but I have this scheme I can't work out if it's hair-brained or genius in this area. A problem we can have -- and possibly this is the issue you are seeing with mounts, although I'm not going to claim certainty in this -- is that curtin creates mounts when it runs (to bind mount /dev into the target when running a chrooted process, for example) and if these are either still running or killed too unceremoniously, these mounts can cause problems for other parts of the code.

One way to make this problem just go away would be to use systemd-run to run these commands in a private mount namespace (something like `systemd-run --unit $blah -p PrivateMounts=true python3 -m curtin in-target ....`). Then the mounts that this process creates just don't affect the rest of the system and if we need to terminate it we can just use systemctl stop to make the whole mess go away as if it had never existed.

The output from such a service goes to the journal by default but we can always read that afterwards if we need to look at the output.

One problem this has is that the services run like this have a clean environment -- for most cases we'd probably want some way to propagate the server's environment into the subprocess (I don't really want to try to play the game of deciding which values matter -- very certainly PATH and PYTHONPATH do but which others?). We could do this with a zillion -E flags to systemd-run or by dumping things out to a file and using -p EnvironmentFile=whatever.

What do you think of this idea? Neat or just too complicated?

mwhudson:
This looks pretty good. I'm sorry you're having overlay troubles :( let me know if you want me to help with those.

ogayot:
Rebased on top of main to fix the failure with hirsute.
Also fixed mis-merge leading to mixed up overlay options when mounting.

It seems to solve at least most of the problems I had with overlays.

subiquity/server/runner.py
ogayot: I don't remember discussing this so I'd guess we haven't talked about it.

First impression, this is neat.

A couple years ago, I had a somewhat similar scenario where I was starting a process that would create new routes, IP addresses and whatnot during its lifetime. I banged my head trying to patch it to make sure everything would be cleaned when the process would exit (sometimes unexpectedly)

Turned out that the only reliable way I found was to make that process run in isolation (i.e., in a different network namespace)

I guess this is analogous to what we have here ; except maybe that we have (full) control over curtin.

IMO the downside is that we might be slower to notice issues in curtin when everything gets cleaned up automatically.

ogayot:
> This looks pretty good. I'm sorry you're having overlay troubles :( let me know if you want me to help with those.

Well.. turns out that this is mostly the result of a mis-merge on my end. Sometimes apt-config still fails but I haven't yet figured out what makes it fail (i.e., unfortunately we don't always capture `stderr` ; which I think is a mistake).

I will keep you posted when I reproduce the issue :)

subiquity/server/runner.py
mwhudson: I'm not sure I would regard curtin not cleaning up after itself when interrupted at arbitrary points as an especially serious bug. Getting that sort of thing 100% right is just a lot of work and we have a mechanism for doing it cleanly, I think we should probably do that.

mwhudson:
> Well.. turns out that this is mostly the result of a mis-merge on my end. Sometimes apt-config still fails but I haven't yet figured out what makes it fail (i.e., unfortunately we don't always capture `stderr` ; which I think is a mistake).

If we ran things using systemd-run, stderr would end up in the journal by default :)

subiquity/server/runner.py
ogayot: Addressed in https://github.com/canonical/subiquity/pull/1185

subiquity/server/controllers/install.py
```diff
@@ -151,14 +155,16 @@ async def install(self, *, context):
 
             self.app.update_state(ApplicationState.RUNNING)
 
-            for_install_path = await self.configure_apt(context=context)
+            self.for_install_path = await self.configure_apt(context=context)
```
dbungert, line 158: any reason to store this on self?

subiquity/server/controllers/drivers.py
```diff
@@ -0,0 +1,103 @@
...
+                self.drivers = []
+            else:
+                self.drivers = await self.ubuntu_drivers.list_drivers(
+                    root_dir=d.mountpoint,
+                    context=context)
+        log.debug("Available drivers to install: %s", self.drivers)
+        if not self.drivers:
+            await self.configured()
+        else:
+            # TODO Remove this once we have the GUI controller.
```
dbungert, line 92: Are we expecting to have a GUI element for the Drivers step for Server?
cc @mwhudson

subiquity/server/server.py
```diff
@@ -200,6 +200,7 @@ def get_installer_password_from_cloudinit_log():
     })
 
 POSTINSTALL_MODEL_NAMES = ModelNames({
+    "drivers",
```
dbungert, line 203: We could choose to make the drivers controller one that does not have to be configured, which would simplify some of the force-configuration in the controller, but that depends on if the Server TUI will have the drivers UI element or not.

dbungert (approved):
None of the feedback blocks merge, but some of it we might revisit in a later PR.

subiquity/server/controllers/install.py
ogayot: Not anymore, thanks! It was a residual change from before the `AptConfigurer.overlay` context manager.
I'll remove it.

subiquity/server/controllers/drivers.py
ogayot: The plan was to have a page show up if at least one driver is available for installation ; but skipped if no drivers to install were found. Maybe I shouldn't have kept it for a follow up PR.

Here's a screenshot:
third-party-drivers

subiquity/server/server.py
ogayot: Thanks! As mentioned in another comment, we're going to have a UI component for server (skipped if no drivers are found) so I don't think we can go for the that construct. 
Having said that, I'll keep the option in mind. I think it can be useful for future components to come :)

PR #1180, opened 2022-02-03
workflow: -hirsute

dbungert:


dbungert:
The CI workflow failed on Hirsute on https://github.com/canonical/subiquity/pull/1179 and I don't think we should bother figuring out why.

mwhudson (approved):

PR #1181, opened 2022-02-08
Lp 1960162

dbungert:


mwhudson:
I think this looks basically ok. Can we change answers/guided-lvm.yaml to request encryption with a password of "passw0rd" so the grepping scripts/runtests.sh does will guard against regression?

subiquity/common/apidef.py
```diff
@@ -230,7 +230,7 @@ class guided:
             def GET(wait: bool = False) -> GuidedStorageResponse:
                 pass
 
-            def POST(choice: Optional[GuidedChoice]) \
+            def POST(data: Payload[GuidedChoice]) \
```
ogayot, line 233: Are we sure that the other API consumers (e.g. desktop installer) never pass `None` / `null` to this endpoint?
Otherwise, we might need to keep `Payload[Optional[GuidedChoice]]` and the check for `choice is not None` on the server side.

ogayot (approved):
Looks good to me, thanks!
Just one question related to not accepting `None` / `null` anymore in `/storage/guided` POST endpoint.

subiquity/common/types.py
```diff
@@ -391,7 +391,7 @@ class TimeZoneInfo:
 
 @attr.s(auto_attribs=True)
 class UbuntuAdvantageInfo:
-    token: str
+    token: str = attr.ib(default='', repr=False)
```
ogayot, line 394: can we avoid the default value?

dbungert: Yes, I wasn't trying to change behavior on that point.

PR #1184, opened 2022-02-15
error: don't actually send dry run error reports

dbungert:


mwhudson:
I dunno, sending crash reports to staging does let more of the code actually be exercised (and validates that the error reports are not horrendously malformed, i guess). What motivates this?

dbungert:
OK, good news that it goes to staging and doesn't pollute real reports.  Thanks!

mwhudson:
Ah yes, that would be bad :)

PR #1185, opened 2022-02-15
Add option to run commands with private mounts and use it for curtin in-target

ogayot:
Hello,

Based on a discussion in a different PR [1], @mwhudson mentioned the idea of executing curtin in-target commands using `systemd-run` to benefit from private mounts.

The `LoggedCommandRunner` now has the option to start a command with private mounts enabled - see `mount_namespaces(7)`.

To use this feature, we now rely on `systemd-run` instead of `systemd-cat`. The main benefit is that it should allow us not to worry about mounts refusing to be un-mounted after executing a curtin in-target command.

On the downside, `systemd-run` requires root privileges (unless `--user` is available) so we can't use it in dry-run mode. For dry-run mode, we still need to rely on `systemd-cat`.
**EDIT**: we now use `systemd-run` at all times and pass the `--user` switch when running as non-root.

Another downside is that the commands passed to `systemd-run` are run in a clean environment. Therefore, we must explicitly set the environment variables that we want to pass to the commands to execute.

I came up with a short list of environment variables that we need but we might need to add more later.

We now use private mounts when executing in-target commands through curtin.

I was able to run a full installation and the system booted properly after. However, I noticed a few (potentially insignificant?) errors from `systemd-udev` in the journal that I would like to mention:

```
Feb 15 20:09:16 ubuntu-server systemd-udevd[2395]: vda2: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/vda2' failed with exit code 1.
Feb 15 20:09:16 ubuntu-server systemd-udevd[2571]: vda1: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/vda1' failed with exit code 1.
Feb 15 20:09:16 ubuntu-server systemd-udevd[2661]: vda3: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/vda3' failed with exit code 1.
Feb 15 20:09:16 ubuntu-server systemd-udevd[2395]: vda3: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/vda3' failed with exit code 1.
Feb 15 20:09:16 ubuntu-server systemd-udevd[2571]: vda1: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/vda1' failed with exit code 1.
Feb 15 20:09:16 ubuntu-server systemd-udevd[2661]: vda2: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/vda2' failed with exit code 1.
Feb 15 20:09:17 ubuntu-server systemd-udevd[2395]: vda3: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/vda3' failed with exit code 1.
Feb 15 20:09:17 ubuntu-server systemd-udevd[2571]: vda1: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/vda1' failed with exit code 1.
Feb 15 20:09:17 ubuntu-server systemd-udevd[2661]: vda2: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/vda2' failed with exit code 1.
Feb 15 20:09:17 ubuntu-server systemd-udevd[2571]: dm-0: Process '/usr/bin/unshare -m /usr/bin/snap auto-import --mount=/dev/dm-0' failed with exit code 1.
```

Thanks,

[1] https://github.com/canonical/subiquity/pull/1179#discussion_r799104385

### DIFF
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -148,13 +148,14 @@ class AptConfigurer:
         if type is not None:
             opts.extend(['-t', type])
         await self.app.command_runner.run(
-            ['mount'] + opts + [device, mountpoint])
+            ['mount'] + opts + [device, mountpoint], private_mounts=False)
         m = Mountpoint(mountpoint=mountpoint)
         self._mounts.append(m)
         return m
 
     async def unmount(self, mountpoint: str):
-        await self.app.command_runner.run(['umount', mountpoint])
+        await self.app.command_runner.run(['umount', mountpoint],
+                                          private_mounts=False)
 
     async def setup_overlay(self, lowers: List[Lower]) -> OverlayMountpoint:
         tdir = self.tdir()
@@ -191,7 +192,7 @@ class AptConfigurer:
 
         await run_curtin_command(
             self.app, context, 'apt-config', '-t', self.configured_tree.p(),
-            config=config_location)
+            config=config_location, private_mounts=True)
 
     async def configure_for_install(self, context):
         assert self.configured_tree is not None
@@ -220,7 +221,7 @@ class AptConfigurer:
 
         await run_curtin_command(
             self.app, context, "in-target", "-t", self.install_tree.p(),
-            "--", "apt-get", "update")
+            "--", "apt-get", "update", private_mounts=True)
 
         return self.install_tree.p()
 
@@ -247,7 +248,7 @@ class AptConfigurer:
         if self.app.base_model.network.has_network:
             await run_curtin_command(
                 self.app, context, "in-target", "-t", target_mnt.p(),
-                "--", "apt-get", "update")
+                "--", "apt-get", "update", private_mounts=True)
         else:
             await _restore_dir('var/lib/apt/lists')
 
@@ -256,7 +257,7 @@ class AptConfigurer:
         if self.app.base_model.network.has_network:
             await run_curtin_command(
                 self.app, context, "in-target", "-t", target_mnt.p(),
-                "--", "apt-get", "update")
+                "--", "apt-get", "update", private_mounts=True)
 
 
 class DryRunAptConfigurer(AptConfigurer):
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import contextlib
 import functools
 import logging
 import os
@@ -153,9 +154,12 @@ class AptConfigurer:
         self._mounts.append(m)
         return m
 
-    async def unmount(self, mountpoint: str):
-        await self.app.command_runner.run(['umount', mountpoint],
-                                          private_mounts=False)
+    async def unmount(self, mountpoint: Mountpoint, remove=True):
+        if remove:
+            self._mounts.remove(mountpoint)
+        await self.app.command_runner.run(
+                ['umount', mountpoint.mountpoint],
+                private_mounts=False)
 
     async def setup_overlay(self, lowers: List[Lower]) -> OverlayMountpoint:
         tdir = self.tdir()
@@ -225,9 +229,27 @@ class AptConfigurer:
 
         return self.install_tree.p()
 
+    @contextlib.asynccontextmanager
+    async def overlay(self):
+        overlay = await self.setup_overlay([
+                self.install_tree.upperdir,
+                self.configured_tree.upperdir,
+                self.source
+            ])
+        try:
+            yield overlay
+        finally:
+            # TODO self.unmount expects a Mountpoint object. Unfortunately, the
+            # one we created in setup_overlay was discarded and replaced by an
+            # OverlayMountPoint object instead. Here we re-create a new
+            # Mountpoint object and (thanks to attr.s) make sure that it
+            # compares equal to the one we discarded earlier.
+            # But really, there should be better ways to handle this.
+            await self.unmount(Mountpoint(mountpoint=overlay.mountpoint))
+
     async def cleanup(self):
         for m in reversed(self._mounts):
-            await self.unmount(m.mountpoint)
+            await self.unmount(m, remove=False)
         for d in self._tdirs:
             shutil.rmtree(d)
 
@@ -240,7 +262,9 @@ class AptConfigurer:
                 'cp', '-aT', self.configured_tree.p(dir), target_mnt.p(dir),
                 ])
 
-        await self.unmount(target_mnt.p('cdrom'))
+        await self.unmount(
+                Mountpoint(mountpoint=target_mnt.p('cdrom')),
+                remove=False)
         os.rmdir(target_mnt.p('cdrom'))
 
         await _restore_dir('etc/apt')
@@ -254,13 +278,12 @@ class AptConfigurer:
 
         await self.cleanup()
 
-        if self.app.base_model.network.has_network:
-            await run_curtin_command(
-                self.app, context, "in-target", "-t", target_mnt.p(),
-                "--", "apt-get", "update", private_mounts=True)
-
 
 class DryRunAptConfigurer(AptConfigurer):
+
+    async def unmount(self, mountpoint: Mountpoint, remove=True):
+        if remove:
+            self._mounts.remove(mountpoint)
 
     async def setup_overlay(self, lowers: List[Lower]) -> OverlayMountpoint:
         # XXX This implementation expects that:
@@ -286,8 +309,12 @@ class DryRunAptConfigurer(AptConfigurer):
             mountpoint=target,
             upperdir=None)
 
+    @contextlib.asynccontextmanager
+    async def overlay(self):
+        yield await self.setup_overlay(self.install_tree.mountpoint)
+
     async def deconfigure(self, context, target):
-        return
+        await self.cleanup()
 
 
 def get_apt_configurer(app, source: str):
--- a/subiquity/server/runner.py
+++ b/subiquity/server/runner.py
@@ -14,7 +14,10 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import asyncio
+from contextlib import suppress
+import os
 import subprocess
+from typing import List
 
 from subiquitycore.utils import astart_command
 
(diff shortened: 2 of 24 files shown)


subiquity/server/runner.py
```diff
@@ -49,16 +76,36 @@ def __init__(self, ident, delay):
         super().__init__(ident)
         self.delay = delay
 
-    async def start(self, cmd):
+    def _forge_systemd_cmd(self, cmd: List[str], private_mounts: bool) \
+            -> List[str]:
+        # We would like to use systemd-run here but unfortunately it requires
+        # root privileges.
+        # Using systemd-run --user would be an option but it not available
+        # everywhere ; so we fallback to using systemd-cat.
```
mwhudson, line 84: where is it not available, out of curiosity?

subiquity/server/tests/test_runner.py
```diff
@@ -0,0 +1,129 @@
...
+            "systemd-run",
+            "--wait", "--same-dir",
+            "--property", "SyslogIdentifier=my-identifier",
+            "--property", "PrivateMounts=yes",
+            "--setenv", "PATH=/snap/subiquity/x1/bin",
+            "--setenv", "PYTHONPATH=/usr/lib/python3.8/site-packages",
+            "--setenv", "PYTHON=/snap/subiquity/x1/usr/bin/python3.8",
+            "--setenv", "TARGET_MOUNT_POINT=/target",
+            "--setenv", "SNAP=/snap/subiquity/x1",
+            "/bin/ls", "/root",
```
mwhudson, line 57: I'm not sure asserting the order of the --setenv's is a great test but well. Perhaps there should be a test that other variables are not carried over?

subiquity/server/runner.py
```diff
@@ -23,23 +26,47 @@ class LoggedCommandRunner:
...
-            ] + cmd)
-        proc.args = cmd
+    def _forge_systemd_cmd(self, cmd: List[str], private_mounts: bool) \
+            -> List[str]:
+        """ Return the supplied command prefixed with the systemd-run stuff.
+        """
+        prefix = [
+            "systemd-run",
+            "--wait", "--same-dir",
+            "--property", f"SyslogIdentifier={self.ident}",
```
mwhudson, line 43: I think for ubuntu-drivers we're going to want the output of a command run like this, but I think we can do that by giving the unit created by systemd-run a unique name and then reading the journal afterwards.

mwhudson (approved):
Well this looks pretty good to me! I'd like Dan to look at it too though.

subiquity/server/runner.py
ogayot: That's right. We can use something like:
```bash
unit_name="subiquity-run-$(uuidgen)"
systemd-run --unit "$unit_name" ...
```
 Another way to do it would be to read from `stderr` when running `systemd-run` since it prints out the unit name that it generates from us:
```
systemd-run ls
Running as unit: run-rdb7cd9bb2f824b2f8fd7df2625a14ea3.service
```
Also, the first approach I considered was to use `systemd-run --pipe` which would pipe `stdout` and `stderr`back to the caller.
```systemd-run --pipe ls /root
Running as unit: run-u683.service
bin   cdrom  etc   lib	  lib64   lost+found  mnt  proc  run   snap  sys  usr
boot  dev    home  lib32  libx32  media       opt  root  sbin  srv   tmp  var
```
However with `--pipe`, the output does not go into the journal. I still think it's a good option.

Any preference?

subiquity/server/runner.py
ogayot: "not available" might not be accurate. My first implementation was using the `--user` switch all the time and it did not work in the installer because it runs as root:
```bash
# systemd-run --user /bin/ls
Failed to connect to bus: $DBUS_SESSION_BUS_ADDRESS and $XDG_RUNTIME_DIR not defined (consider using --machine=<user>@.host --user to connect to bus of other user)
```
We can't say "use `--user` when in dry-run mode and not when in normal mode" because sometimes we run integration tests (i.e. dry-run) as root.

Maybe the "if not root, then use `--user` would work" but I don't know ; it feels brittle. What do you think?

ogayot:
I ran the following command locally and got an exception related to C.UTF-8 not being available. I think it's not related to this PR:
```bash
$ scripts/test-in-lxd.sh ubuntu-daily:impish "make check"
```

```python
2022-02-16 10:03:10,608 ERROR root:39 finish: subiquity/Configure/configure: FAIL: [Errno 2] No such file or directory: '/usr/lib/locale/C.UTF-8/'                                                                                                            
2022-02-16 10:03:10,609 INFO root:39 start: subiquity/ErrorReporter/1645005790.604914188.install_fail/add_info: 
2022-02-16 10:03:10,611 ERROR subiquity.server.server:414 top level error
  File "/root/subiquity/subiquity/server/controllers/cmdlist.py", line 104, in _run
    await Install.install_task
  File "/root/subiquity/subiquitycore/context.py", line 148, in decorated_async
    return await meth(self, **kw)
  File "/root/subiquity/system_setup/server/controllers/configure.py", line 385, in configure
    await self.apply_locale(lang)
  File "/root/subiquity/system_setup/server/controllers/configure.py", line 276, in apply_locale
    ok = await self._activate_locale(lang, env)
  File "/root/subiquity/system_setup/server/controllers/configure.py", line 104, in _activate_locale
    (locGenCmd, ok) = self.__locale_gen_cmd()
  File "/root/subiquity/system_setup/server/controllers/configure.py", line 61, in __locale_gen_cmd
    shutil.copytree("/usr/lib/locale/C.UTF-8/", outDir,
  File "/usr/lib/python3.9/shutil.py", line 564, in copytree
    with os.scandir(src) as itr:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/lib/locale/C.UTF-8/'
2022-02-16 10:03:10,612 ERROR subiquity.server.server:414 top level error
  File "/root/subiquity/system_setup/server/controllers/shutdown.py", line 46, in _wait_install
    await self.app.controllers.Install.install_task
  File "/root/subiquity/subiquity/server/controllers/cmdlist.py", line 104, in _run
    await Install.install_task
  File "/root/subiquity/subiquitycore/context.py", line 148, in decorated_async
    return await meth(self, **kw)
  File "/root/subiquity/system_setup/server/controllers/configure.py", line 385, in configure
    await self.apply_locale(lang)
  File "/root/subiquity/system_setup/server/controllers/configure.py", line 276, in apply_locale
    ok = await self._activate_locale(lang, env)
  File "/root/subiquity/system_setup/server/controllers/configure.py", line 104, in _activate_locale
    (locGenCmd, ok) = self.__locale_gen_cmd()
  File "/root/subiquity/system_setup/server/controllers/configure.py", line 61, in __locale_gen_cmd
    shutil.copytree("/usr/lib/locale/C.UTF-8/", outDir,
  File "/usr/lib/python3.9/shutil.py", line 564, in copytree
    with os.scandir(src) as itr:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/lib/locale/C.UTF-8/'
```

subiquity/server/tests/test_runner.py
ogayot: Updated. I added a variable called `DUMMY` in the mocked environment and made sure it's not in the resulting command.
Regarding the order of `--setenv` commands, I agree... I tried to split the resulting command into three distinct parts so we can do something more intelligent with the `setenv`s:
* the prefix:
```python
["systemd-run", "--wait", "--same-dir",
    "--property", "SyslogIdentifier=my-identifier",
    "--property", "PrivateMounts=yes"]
```
* the suffix (which is the supplied command prefixed with `"--"`:
```python
["--", "/bin/ls", "/root"]
```
* and what's leftover: 
```python
["--setenv", "PATH=/snap/subiquity/x1/bin",
    "--setenv", "PYTHONPATH=/usr/lib/python3.8/site-packages",
    "--setenv", "PYTHON=/snap/subiquity/x1/usr/bin/python3.8",
    "--setenv", "TARGET_MOUNT_POINT=/target",
    "--setenv", "SNAP=/snap/subiquity/x1"]
```
The implementation was based on crazy list slicing and felt way too complex for unit tests ...

subiquity/server/apt.py
```diff
@@ -182,7 +183,7 @@ async def apply_apt_config(self, context):
 
         await run_curtin_command(
             self.app, context, 'apt-config', '-t', self.configured_tree.p(),
-            config=config_location)
+            config=config_location, private_mounts=False)
```
mwhudson, line 186: Why is this private_mounts=False? I *think* this is one of the ones we really want to be isolated, isn't it?

subiquity/server/runner.py
mwhudson: I think "if not root, then use --user" might actually make sense, even if it is a little strange. The more we can exercise stuff like this in dry-run mode, the better.

subiquity/server/tests/test_runner.py
mwhudson: OK let's leave this like this then!

subiquity/server/runner.py
mwhudson: I didn't know about --pipe, that's probably easier indeed.

Having everything end up in the journal would be useful in some ways but probably not worth the complication.

mwhudson (commented):
This still looks basically fine to me. I would still like @dbungert to take a look :)

subiquity/server/apt.py
ogayot: Yes, you are right.
In the first place I set `private_mounts=True` almost everywhere and obviously it didn't work because some calls are expected to leave some mounts available for later (e.g. /target).
So I ended-up do it only for the `curtin in-target` commands. But obviously, `curtin apt-config` should use private mounts too. Thanks!

subiquity/server/runner.py
ogayot: There we go!

subiquity/server/apt.py
ogayot: Updated!

dbungert (approved):
LGTM.  I will specifically ask that we put this through some ISO testing before merge though.

ogayot:
> LGTM. I will specifically ask that we put this through some ISO testing before merge though.

Thanks! Yes, this change can have a significant impact. Would tests using an injected snap be enough in your opinion? Maybe let's discuss what kind of tests we can do.

dbungert:
> > LGTM. I will specifically ask that we put this through some ISO testing before merge though.
> 
> Thanks! Yes, this change can have a significant impact. Would tests using an injected snap be enough in your opinion? Maybe let's discuss what kind of tests we can do.

Yes, just a bit of kvm-test or equivalent should be fine.

ogayot:
Took me a few attemps to get the merge done properly but it is working as expected now.
Verified on multiple installs using kvm. Logs and installation look sane.

mwhudson:
Let's do it then!

PR #1186, opened 2022-02-16
filesystem: add disk scaling for /boot and efi

dbungert:
Add a scaling system to adjust the size of /boot and efi partitions in
response to the size of the disk.

Caveat:  scaling for efi is not smart enough to know if /boot is needed or not.

### DIFF
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -15,6 +15,8 @@
 
 import logging
 
+import attr
+
 from subiquity.common.filesystem import boot, gaps
 from subiquity.common.types import Bootloader
 from subiquity.models.filesystem import (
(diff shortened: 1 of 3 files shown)


subiquity/common/filesystem/manipulator.py
```diff
@@ -25,9 +27,52 @@
...
+
+
+def scale_partitions(all_factors, disk_size):
+    """for the list of scale factors, provide list of scaled partition size.
+    Assumes at most one scale factor with maximum==-1, and
+    disk_size is at least as big as the sum of all partition minimums.
+    The scale factor with maximum==-1 is given all remaining disk space."""
+    ret = []
+    for cur in all_factors:
+        sum_priorities = sum([factor.priority for factor in all_factors])
```
mwhudson, line 63: This line can be outside the loop?

mwhudson (approved):
Looks nice! Thanks. One trivial nit.

subiquity/common/filesystem/manipulator.py
dbungert: Done, thanks.

PR #1187, opened 2022-02-17
Fixes missing directory for system_setup integration test

CarlosNihelton:
By providing alternative directories to find LC_* locale definitions.

libc-bin package was recently updated to include the locale files for C.UTF-8 in `/usr/lib/locale/C.utf8` , breaking the system_setup tests. With this PR alternative directory paths are provided and tested for existence (and error checking as well) before attempting to copy the directory tree.

### DIFF
--- a/system_setup/server/controllers/configure.py
+++ b/system_setup/server/controllers/configure.py
@@ -58,8 +58,14 @@ class ConfigureController(SubiquityController):
         cmdFile = os.path.join(self.model.root, cmd)
         shutil.copy(os.path.join("/", cmd), cmdFile)
         # Supply LC_* definition files to avoid complains from localedef.
-        shutil.copytree("/usr/lib/locale/C.UTF-8/", outDir,
-                        dirs_exist_ok=True)
+        candidateSourceDirs = ["/usr/lib/locale/C.UTF-8/",
+                               "/usr/lib/locale/C.utf8/"]
+        sourceDirs = [d for d in candidateSourceDirs if os.path.exists(d)]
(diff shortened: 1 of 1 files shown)


system_setup/server/controllers/configure.py
```diff
@@ -58,8 +58,14 @@ def __locale_gen_cmd(self) -> Tuple[str, bool]:
         cmdFile = os.path.join(self.model.root, cmd)
         shutil.copy(os.path.join("/", cmd), cmdFile)
         # Supply LC_* definition files to avoid complains from localedef.
-        shutil.copytree("/usr/lib/locale/C.UTF-8/", outDir,
-                        dirs_exist_ok=True)
+        candidateSourceDirs = ["/usr/lib/locale/C.UTF-8/",
+                               "/usr/lib/locale/C.utf8/"]
+        sourceDirs = [d for d in candidateSourceDirs if os.path.exists(d)]
+        if len(sourceDirs) == 0 or sourceDirs is None:
```
mwhudson, line 64: The "or sourceDirs is None" part should go, not only is there no way for sourceDirs to be None, len(sourceDirs) would have crashed before this is evaluated if was.

mwhudson (approved):

system_setup/server/controllers/configure.py
CarlosNihelton: Got it. Removed the unnecessary check.

dbungert (approved):

dbungert:
Thanks!

PR #1188, opened 2022-02-21
Add type hints to overlayfs helpers and add some unit tests.

ogayot:
Hello,

After struggling quite a bit with issues related to how we use overlays, I decided to do some improvements (hopefully?) to make it easier to work with overlays:
* added type hints where relevant to make it easier to understand what is what in the different contexts. Sometimes we expect `str` objects, sometimes we expect `Mountpoints` objects, sometimes expect `OverlayMountpoints` objects, or a combination of all the above ; with usually the name for the variables irrespective of the type. The recent refactoring of APT configuration also changed the type returned by the overlay-related functions (i.e., from the upperdir to the mountpoint) so it felt challenging to figure out what was what. I hope that the type hints make it a little simpler now.
* added some unit tests for the functions translating different objects (i.e., `str`, `Mountpoint`, `OverlayMountpoint` or a list of all of the above) into an options string that can be passed to `mount -t overlay`.
* changed the signature of `AptConfigurer.setup_overlay(self, lowers)` a little bit:
Before this change, we could pass either a sequence of `Lower` or a single `Lower` directly:
```python
# These three forms were passing type-checking before
setup_overlay("/lower1")
setup_overlay(["/lower1"])
setup_overlay(["/lower1", "/lower2"])
```
We now force the user to pass a _sequence_ of `Lower`'s (where `Lower` is either a `str`, `Mountpoint` or `OverlayMountpoint`).
```python
# Now these two forms only pass type-checking.
setup_overlay(["/lower1"])
setup_overlay(["/lower1", "/lower2"])
```
* made `DryRunAptConfigurer.setup_overlay(self, source)` consistent with `AptConfigurer.setup_overlay(self, lowers)`.
* made use of `attr.define` (i.e., a newer API for `attr.s`) for `Mountpoint` and `OverlayMountpoint` so that using one of these object in a function from the `logging` module prints out some useful information (e.g., list of lowers, upper, mountpoint); while keeping the signature identical. This is very useful for debugging! **EDIT** finally used `attr.s` instead since `attr.define` is not available on focal.

Thanks,
Olivier

### DIFF
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -20,6 +20,8 @@ import shutil
 import tempfile
 from typing import List, Optional, Union
 
+import attr
+
 from curtin.config import merge_config
 
 from subiquitycore.file_util import write_file, generate_config_yaml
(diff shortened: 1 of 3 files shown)


mwhudson (approved):
LGTM thanks

dbungert (approved):
Thanks for doing this.

PR #1189, opened 2022-02-22
Revert probert again to remove os-prober usage

dbungert:
os-prober is leaving around invalid devices and we don't know what to do
about it yet, but we do know that we don't require os-prober for
20.04.4.  Back this out.  LP: #1961640 LP: #1961628


mwhudson (approved):
Sigh. Who knew that os-prober would be such a pain!

PR #1190, opened 2022-02-22
snapcraft.yaml: remove bson from python-packages

mwhudson:
The one installed by pip no longer works with our error report code.


mwhudson:
confirmed that this helps

dbungert (approved):

dbungert:
nice

PR #1193, opened 2022-02-28
os-prober: installer newer version

dbungert:
The version of os-prober in focal has some device mapper code that we don't
want.  So instead pick up a version including the v1.75 changes.  Also pick up
my branch that allows setting env variables to find the os-prober dependent
scripts and utility program.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -197,11 +197,12 @@ parts:
 
   os-prober:
     plugin: nil
-    stage-packages: [os-prober]
-    override-stage: |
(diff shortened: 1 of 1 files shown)


mwhudson (approved):
Oh that's not too ugly at all. I assume you've tested this works :)

dbungert:
I have tested this, but I was not able to reproduce the device mapper problem from LP: #1961628 , so while this continues to run os-prober fine we'll need some help making certain that the device mapper problem is fixed.  I will ping the relevant testers.

PR #1194, opened 2022-03-01
Fix POST to drivers not using payload wrapper for install

ogayot:
In the following patch, I changed the signature of the POST request handler for `/drivers` so that we expect the a JSON body instead of query string parameters:

3c9e5ca98fee453c37660c961de26ebe8e7dcaa8  Use body for parameters of POST /drivers

I forgot to update the function calling `self.POST` explicitly (i.e. without making an actual HTTP request). Therefore, it prevents the `drivers` component from being marked configured when drivers are found - subsequently making Subiquity hang on "executing late commands".

### DIFF
--- a/subiquity/server/controllers/drivers.py
+++ b/subiquity/server/controllers/drivers.py
@@ -90,7 +90,9 @@ class DriversController(SubiquityController):
             await self.configured()
         else:
             # TODO Remove this once we have the GUI controller.
-            await self.POST(install=True)
+            await self.POST(data=DriversPayload(
+                install=True,
+                ))
 
     async def GET(self, wait: bool = False) -> DriversResponse:
         if wait:


mwhudson (approved):
fine I guess and clearly more correct! is there a reason for DriversPayload existing though?

ogayot:
> fine I guess and clearly more correct! is there a reason for DriversPayload existing though?

When @dbungert started to work on hiding sensitive information from the access log files, I figured it might be worth being more consistent and pass POST parameters in the HTTP body rather than in query strings parameters.

Do you have an opinion on this?

mwhudson:
Payload[bool] would work just as well for that I think? But anyway, no, I don't really have an opinion :)

ogayot:
> Payload[bool] would work just as well for that I think? But anyway, no, I don't really have an opinion :)

Yes, but then it drops the ability to "name" the parameters. In my opinion:

```
POST /drivers
{"install": true}
```
makes more sense than:
```
POST /drivers
true
```

mwhudson:
> Having said that, I'm not too happy about defining a new type each time we add a new POST endpoint :)

Yeah these single-use datatypes are a bit annoying. Not sure if there's really a better answer though. Anyway, enough talking about a simple change, let's merge this.

PR #1195, opened 2022-03-01
generated bond configuration (active-backup) is invalid

maristeslk:
mii-monitor-interval default value is Zero. 
Mode Active-backup depends on mii-monitor-interval to set the correct value, which is usually 100.

### DIFF
--- a/subiquitycore/models/network.py
+++ b/subiquitycore/models/network.py
@@ -105,6 +105,8 @@ class BondConfig:
         params = {
             'mode': self.mode,
             }
+        if mode == "active-backup":
+            params['mii-monitor-interval'] = 100
         if mode in BondParameters.supports_xmit_hash_policy:
             params['transmit-hash-policy'] = self.xmit_hash_policy
         if mode in BondParameters.supports_lacp_rate:


dbungert:
@slyon thoughts?

@maristeslk - Thanks for your contribution.  Getting a better default set sounds like a nice fix.  Before I can merge this I will need to ask that you sign the Contributor License Agreement.  https://ubuntu.com/legal/contributors  Thanks!

slyon:
I agree that having the `mii-monitor-interval` set to 100 would be the common case for detecting link failures of the underlying physical interfaces of the bond. I did not find any definitive documentation about it being necessary, other than https://wiki.linuxfoundation.org/networking/bonding#bonding_driver_options (or the kernel's `bonding.txt` file) which does not really match the given use case, though.

@maristeslk can you provide a link to any documentation to back up that statement?

maristeslk:
> I agree that having the `mii-monitor-interval` set to 100 would be the common case for detecting link failures of the underlying physical interfaces of the bond. I did not find any definitive documentation about it being necessary, other than https://wiki.linuxfoundation.org/networking/bonding#bonding_driver_options (or the kernel's `bonding.txt` file) which does not really match the given use case, though.
> 
> @maristeslk can you provide a link to any documentation to back up that statement?

Sorry it took so long to reply.
If miimon is set to 0, I tested it and found that the backup port is not automatically switched after the primary port shutdown.
It's true that the docs don't explicitly state that the case where value is 0 means disable, but we can check the parameter description of arp_interval to say that 0 is indeed disable.
arp_interval and miimon are themselves parameters for the same purpose.
https://www.kernel.org/doc/Documentation/networking/bonding.txt
image
image

PR #1196, opened 2022-03-02
Add client-side controller and view to install third-party drivers on server installer

ogayot:
We now have a GUI component to control whether we want to install available third-party drivers.
In normal conditions, the page shows a checkbox "Install the available drivers" alongside the list of drivers that are available. However, if no drivers are available, we will skip the page upon reaching it.

Now, if by the time we reach the page, we haven't been able to find the list of drivers available yet, then a progress bar will be displayed instead.
Then we will either display the list of drivers or show "no drivers are available". 

Hitting the "continue" button while the progress bar ls displayed will not install drivers if they are available.

To test the feature using dry-run:
```bash
# Test with a nvidia driver available
SUBIQUITY_DEBUG=has-drivers make dry-run

# Test with no drivers available
make dry-run
```

looking
no-applicable
install-drivers

### DIFF
--- a/examples/answers-swap.yaml
+++ b/examples/answers-swap.yaml
@@ -39,5 +39,5 @@ SnapList:
       classic: false
 InstallProgress:
   reboot: yes
-
-
+Drivers:
+  install: no
--- a/subiquity/client/controllers/drivers.py
+++ b/subiquity/client/controllers/drivers.py
@@ -15,6 +15,7 @@
 
 import asyncio
 import logging
+from typing import List
 
 from subiquitycore.tuicontroller import Skip
 
@@ -31,17 +32,14 @@ class DriversController(SubiquityTuiController):
 
     async def make_ui(self) -> DriversView:
         response: DriversResponse = await self.endpoint.GET()
-        if response.drivers is not None and not response.drivers:
+        if not response.drivers and response.drivers is not None:
             raise Skip
-        if response.drivers is None:
-            return DriversView(self, None, response.install)
-        else:
-            return DriversView(self, bool(response.drivers), response.install)
+        return DriversView(self, response.drivers, response.install)
 
-    async def _wait_drivers(self) -> bool:
+    async def _wait_drivers(self) -> List[str]:
         response: DriversResponse = await self.endpoint.GET(wait=True)
         assert response.drivers is not None
-        return bool(response.drivers)
+        return response.drivers
 
     async def run_answers(self):
         if 'install' not in self.answers:
--- a/subiquity/client/controllers/drivers.py
+++ b/subiquity/client/controllers/drivers.py
@@ -1,0 +1,79 @@
+# Copyright 2021 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import asyncio
+import logging
+
+from subiquitycore.tuicontroller import Skip
+
+from subiquity.common.types import DriversPayload, DriversResponse
+from subiquity.client.controller import SubiquityTuiController
+from subiquity.ui.views.drivers import DriversView
+
+log = logging.getLogger('subiquity.client.controllers.drivers')
+
+
+class DriversController(SubiquityTuiController):
+
+    endpoint_name = 'drivers'
+
+    async def make_ui(self) -> DriversView:
+        response: DriversResponse = await self.endpoint.GET()
+        if response.drivers is not None and not response.drivers:
+            raise Skip
+        if response.drivers is None:
+            return DriversView(self, None, response.install)
+        else:
+            return DriversView(self, bool(response.drivers), response.install)
+
+    async def _wait_drivers(self) -> bool:
+        response: DriversResponse = await self.endpoint.GET(wait=True)
+        assert response.drivers is not None
+        return bool(response.drivers)
+
+    async def run_answers(self):
+        if 'install' not in self.answers:
+            return
+
+        import urwid
+        from subiquitycore.testing.view_helpers import (
+            click,
+            find_button_matching,
+            find_with_pred,
+            )
+
+        def text_finder(txt):
+            def pred(w):
+                return isinstance(w, urwid.Text) and txt in w.text
+            return pred
+
+        view = self.app.ui.body
+        while find_with_pred(view, text_finder('Looking for')):
+            await asyncio.sleep(0.2)
+        if find_with_pred(view, text_finder('No applicable')):
+            click(find_button_matching(view, "Continue"))
+            return
+
+        view.form.install.value = self.answers['install']
+
+        click(view.form.done_btn)
+
+    def cancel(self):
+        self.app.prev_screen()
+
+    def done(self, install):
+        log.debug("DriversController.done next_screen install=%s", install)
+        self.app.next_screen(
+                self.endpoint.POST(DriversPayload(install=install)))
--- a/subiquity/client/controllers/drivers.py
+++ b/subiquity/client/controllers/drivers.py
@@ -62,10 +62,10 @@ class DriversController(SubiquityTuiController):
 
         click(view.form.done_btn.base_widget)
 
-    def cancel(self):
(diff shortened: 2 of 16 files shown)


subiquity/ui/views/drivers.py
```diff
@@ -0,0 +1,124 @@
...
+            self.spinner,
+            ]
+        btn = ok_btn(_("Continue"), on_press=lambda sender: self.done(False))
+        self._w = screen(rows, [btn])
+        asyncio.create_task(self._wait(install))
+
+    async def _wait(self, install: bool) -> None:
+        """ Wait until the "list" of drivers is available and change the view
+        accordingly. """
+        drivers = await self.controller._wait_drivers()
```
dbungert, line 82: We recently merged a [fix](https://github.com/canonical/subiquity/pull/1192) for an issue where the task to check if a subiquity update was available was causing answers automated install hangs.

I wonder if we need to do that for every task that is implemented on the server as a network check - that we should not leave a task around.  Once we have made the decision to continue, we should cancel that task.  We also probably want to wrap the task in a `schedule_task` call.

What do you think?

subiquity/ui/views/drivers.py
```diff
@@ -0,0 +1,124 @@
...
+    available drivers or not. """
+
+    cancel_label = _("Back")
+
+    install = BooleanField(_("Install the drivers"))
+
+
+class DriversView(BaseView):
+
+    title = _("Third-party drivers.")
```
dbungert, line 52: The other titles don't have a period
```suggestion
    title = _("Third-party drivers")
```

subiquity/client/controllers/drivers.py
```diff
@@ -0,0 +1,53 @@
...
+        return DriversView(self, response.drivers, response.install)
+
+    async def _wait_drivers(self) -> List[str]:
+        response: DriversResponse = await self.endpoint.GET(wait=True)
+        assert response.drivers is not None
+        return response.drivers
+
+    async def run_answers(self) -> None:
+        if 'install' in self.answers:
+            self.done(self.answers["install"])
```
mwhudson, line 45: I realise this might be my code but in general run_answers should operate the UI a bit more. So this should be something more like self.view.form.install.checked=True; view_helpers.click(self.view.form.ok_btn)

subiquity/ui/views/drivers.py
mwhudson: I think this is good hygiene in general

examples/answers-swap.yaml
```diff
@@ -39,5 +39,5 @@ SnapList:
       classic: false
 InstallProgress:
   reboot: yes
-
-
+Drivers:
+  install: no
```
mwhudson, line 43: It would be nice if we didn't have to do this to every file every time we add a screen wouldn't it... (I don't really have any ideas here). Remember we'll need to do this to the ubuntu-testcases repo too!

mwhudson (approved):
Just some minor comments, looks very good overall!

subiquity/ui/views/drivers.py
ogayot: The problem I have with cancelling the task is that upon hitting "back" from the later screen, the user is not going to have a second chance of getting the relevant information shown.

I haven't dug enough into the root-cause of https://github.com/canonical/subiquity/pull/1192 to understand if other screens can be affected or if it's just because the "Refresh" screen is shown at two possible positions.

Any thoughts?

examples/answers-swap.yaml
ogayot: Indeed, it would be!
One way that I can think of would be to generate the `examples/answers-*.yaml` files based on:
* a single baseline `.yaml` file (where we have "default" values for each item)
* other `.yaml` files that override specific items.
Any thoughts?

I must admit I didn't know about the ubuntu-test-cases / live-server repo's existence :eyes: 
Lesson learned, I guess!

subiquity/client/controllers/drivers.py
ogayot: Updated. It took me forever to figure out that:
```diff
-click(view.form.done_btn)
+click(view.form.done_btn.base_widget)
```
is necessary so signals get emitted properly.

Also, I didn't keep the text finding stuff because it breaks when the content is translated. Instead I stored a status variable in the view.

subiquity/ui/views/drivers.py
ogayot: I implemented this change.

The task is now cancelled on POST request to /drivers. Subsequent GET requests (if any) having wait=true will restart the task.

Unfortunately, if any task/coroutine is `await`'ing on the `_drivers_task` when it gets cancelled, an `asyncio.CancelledError` exception is raised and the server exits silently. This is not specific to the third-party drivers. I was able to reproduce this behavior in any other controller.

In the normal scenario:
* the client sends a `GET /drivers` query without `wait=true` -> the server returns immediately without any info about the drivers
* the client sends another `GET /drivers` query with `wait=true` and the server "blocks" until it can respond
* while it waits for an answer, the client can send a `POST /drivers` query to skip the third-party drivers page. This will effectively make the server raise a `asyncio.CancelledError` exception in the GET handler.

Therefore, we have to catch this exception in the GET handler. And what to return to the client when a task cancelled is another problem to solve.
I figured it would be good not to respond to the client with a normal body with "fake" data in it.
I implemented a mechanism that - upon raising a `HTTPConflict` exception from the server - raises a `ServerTaskCancelled` on the client side. The client can then handle this kind of exception using various means.

subiquity/client/controllers/drivers.py
mwhudson: Oh right yes, I guess we should keep this translation neutral!

subiquity/ui/views/drivers.py
mwhudson: Hm. One of the areas where asyncio is objectively terrible is cancellation, unfortunately. One annoying thing is that if there is an expectation that a task is going to be cancelled, you have to be careful to think if any await in it should have an asyncio.shield() wrapped around it. I think we're OK here on this.

I guess for the single client case, in some sense the GET(wait=True) task should be cancelled when the user clicks continue. But there could be other clients in the same case (even if that's not very likely), and clearly having the server silently exit is bad! (I'm actually surprised by that -- exceptions that reach the top level are supposed to be handled by SubiquityServer._handle_exception which puts things into an error report and sets an error state but not exit. Do you know what's going on there? Is something handling CancelledError specially?). So I think this solution is fine.

ogayot: Thanks mwhudson! What you said about the exceptions being handled by `_handle_exception` made complete sense so I went to look again at what was happening.

Turns out I misinterpreted my observations. Here's what actually happens when the server raises a `CancelledError` in a `GET` handler. 
TL;DR The server does not exit. The client actually dies and as the result sends a SIGINT to the server before exiting (at least in dry-run mode)

1. The server catches the `CancelledError` exception.
2. As the result of the exception, the server closes the socket associated with the GET query without sending a response.
3. This makes the client receive a `EPIPE` error (or something equivalent for UNIX domain sockets) that is transformed by `aiohttp` into a `aiohttp.client_exceptions.ServerDisconnectedError`.
4. The client has no handler setup for `ServerDisconnectedError` so it dies
5. Before dying, the client kills the server (as it usually does):

_generating crash report
report saved to .subiquity/var/crash/1647245854.068176031.ui.crash
**killing server 25055**_


```
report saved to .subiquity/var/crash/1647245854.068176031.ui.crash
killing server 25055
[...]
  File "/home/olivier/dev/canonical/subiquity/subiquity/ui/views/drivers.py", line 92, in _wait
    response: DriversResponse = await self.endpoint.GET(wait=True)
  File "/home/olivier/dev/canonical/subiquity/subiquity/common/api/client.py", line 45, in impl
    async with make_request(
[...]
  File "/usr/lib/python3/dist-packages/aiohttp/streams.py", line 604, in read
    await self._waiter
aiohttp.client_exceptions.ServerDisconnectedError: Server disconnected
```

thanks!

ogayot: As discussed with @dbungert, I reverted this change and will open a follow-up PR with it.

subiquity/client/controllers/drivers.py
```diff
@@ -0,0 +1,69 @@
...
+        if 'install' not in self.answers:
+            return
+
+        from subiquitycore.testing.view_helpers import (
+            click,
+        )
+
+        view = self.app.ui.body
+        while view.status == DriversViewStatus.WAITING:
+            await asyncio.sleep(0.2)
```
dbungert, line 54: I'm prepared to let this one go, but if status never moves on for some reason this can hang test automation.  Think about it but I don't consider it merge blocking.

ogayot: Thanks!
Well, if the status never moves, then there's a bug in our implementation that we need to address. So I would not mind having at least _some_ test break so we are aware of the issue.
Would a 5-10 minutes timeout help?

dbungert: 10 minutes sounds nice.  Long enough that we would not ever expect to hit it during proper operation.

PR #1198, opened 2022-03-04
Allow calling installdeps.sh from anywhere

jpnurmi:
It's nice for the desktop installer contributors to be able to run
"./path/to/subiquity/scripts/installdeps.sh" without changing dir.

### DIFF
--- a/scripts/installdeps.sh
+++ b/scripts/installdeps.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 set -eux
+cd "$(dirname ${BASH_SOURCE:0})/.."
+
 apt-get update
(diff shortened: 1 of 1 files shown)


scripts/installdeps.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 set -eux
+pushd "$(dirname ${BASH_SOURCE:0})/.."
+trap popd EXIT
```
dbungert, line 4: Great idea!
Since subshells cannot change the parent, we can skip pushd/popd and just cd to the required directory.

dbungert (changes requested):

dbungert (approved):

PR #1199, opened 2022-03-04
aptdeps: cloud-init, cleanup

dbungert:
* document cloud-init as an apt-dep (it was only half there)
* centralize this info so we don't have to update 2 spots


mwhudson (approved):
Thanks for getting rid of that annoying duplication!

ogayot (approved):

PR #1200, opened 2022-03-04
track start position of partitions

dbungert:
* adjust some routines to accommodate the possibility that we may be editing partitions
* standardize on the term 'start' to indicate the starting position of the partition or gap
* standardize on a sort order for partitions based on number, falling back to offset
* several routines updated to expect a Gap object

### DIFF
--- a/subiquity/common/filesystem/actions.py
+++ b/subiquity/common/filesystem/actions.py
@@ -66,10 +66,10 @@ class DeviceAction(enum.Enum):
         r = _checkers[self](device)
         if isinstance(r, bool):
             return r, None
-        elif isinstance(r, str):
-            return False, r
+        elif r is None:
+            return False, _("Unknown reason")
         else:
-            return r
+            return False, str(r)
 
 
 @functools.singledispatch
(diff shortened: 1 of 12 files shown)


dbungert:
Wait, curtin wants the 'offset' term, will change to that

mwhudson (changes requested):
Hmmm I left it a bit late in the day to have enough brain to review this. Maybe that's not entirely a bad thing... Even though the diff is not that large it sure feels like there is a lot going on. Is it possible to break it up a bit more? Even trivial stuff like moving get_efi_size to top level functions would be fine, I promise to review that sort of thing very quickly :)

Also the offsets for new partitions are a bit bogus currently. I don't know if this really matters until we represent multiple gaps and generate v2 configs but playing with the UI creating and deleting and resizing partitions gave me a curtin config like this for the partitions of one disk:

```
  - size: 953M  number: 1 offset: 0
  - size: 4165M number: 2 offset: 11193M
  - size: 5120M number: 3 offset: 5118M
```

offset: 0 isn't right for the ESP, and that offset for the middle disk is considerable past the end of the disk...

mwhudson:
Oh and I'm not sure we should be attached to the branch history at all :)

dbungert:
I think everything interesting is brought over, except some of the action handling bit.

PR #1201, opened 2022-03-07
Fix not awaited ubuntu-drivers command existence check

ogayot:
The `ensure_cmd_exists()` coroutine from ubuntu_drivers was never awaited. Therefore, the check would always silently succeed. Taking a look in `server-output` would show the warning though:

```
RuntimeWarning: coroutine 'UbuntuDriversHasDriversInterface.ensure_cmd_exists' was never awaited
  self.ubuntu_drivers.ensure_cmd_exists(d.mountpoint)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

Also, the implementation of `ensure_cmd_exists()` when `SUBIQUITY_DEBUG=run-drivers` is set was failing because we use a shell builtin as an executable. This was not noticed since `ensure_cmd_exists()` was not executed and `arun_command` does not check for errors by default.

### DIFF
--- a/subiquity/server/controllers/drivers.py
+++ b/subiquity/server/controllers/drivers.py
@@ -78,7 +78,7 @@ class DriversController(SubiquityController):
         async with apt.overlay() as d:
             try:
                 # Make sure ubuntu-drivers is available.
-                self.ubuntu_drivers.ensure_cmd_exists(d.mountpoint)
+                await self.ubuntu_drivers.ensure_cmd_exists(d.mountpoint)
             except CommandNotFoundError:
                 self.drivers = []
             else:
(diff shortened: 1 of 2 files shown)


mwhudson (approved):

mwhudson:
I wonder if there is any way we can make this sort of thing poke us in the eye more. Maybe the integration tests should grep server-output for warnings. Maybe we can even run them with PYTHONTRACEMALLOC set?

dbungert (approved):

ogayot:
> I wonder if there is any way we can make this sort of thing poke us in the eye more. Maybe the integration tests should grep server-output for warnings. Maybe we can even run them with PYTHONTRACEMALLOC set?

I like this idea a lot! I will open a follow up PR

ogayot:
> > I wonder if there is any way we can make this sort of thing poke us in the eye more. Maybe the integration tests should grep server-output for warnings. Maybe we can even run them with PYTHONTRACEMALLOC set?
> 
> I like this idea a lot! I will open a follow up PR

https://github.com/canonical/subiquity/pull/1203

PR #1202, opened 2022-03-08
Force ubuntu-advantage screen show in jammy installer until it is marked LTS

ogayot:
Because we decided to hide the ubuntu-advantage screen on non-LTS versions, we need to wait until jammy is marked LTS (i.e., until the `"LTS"` string is present in `/etc/lsb-release`) to test the screen in new installs.

Since we need to do more testing now, I've white-listed jammy so that installs show the ubuntu-advantage screen even though it's not marked LTS yet.
One thing to keep in mind though is that u-a-t returns `"available": false` for all services since Jammy is not yet marked LTS in the back-end. Therefore, the list of services shown is empty:

services

Also, while testing with the ISO, I encountered a crash when the token expiration date came with the "Z" suffix instead of the "+00:00" UTC offset. This is specified by RFC3339 but not currently implemented by Python `datetime.date.fromisoformat`. I worked around this limitation by replacing any "Z" character with "+00:00".

### DIFF
--- a/subiquity/client/controllers/ubuntu_advantage.py
+++ b/subiquity/client/controllers/ubuntu_advantage.py
@@ -49,7 +49,11 @@ class UbuntuAdvantageController(SubiquityTuiController):
         """ Generate the UI, based on the data provided by the model. """
 
         dry_run: bool = self.app.opts.dry_run
-        if "LTS" not in lsb_release(dry_run=dry_run)["description"]:
+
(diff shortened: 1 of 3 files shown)


gene1wood:
I filled [this bug](https://bugs.launchpad.net/subiquity/+bug/1970631) on the isoformat issue before finding this PR, sorry about that.

PR #1204, opened 2022-03-08
keyboard: run setxkbmap for the desktop installer

jpnurmi:
This helps to fix https://github.com/canonical/ubuntu-desktop-installer/issues/250. Related discussion at https://github.com/canonical/ubuntu-desktop-installer/pull/154#issuecomment-878660593.

### DIFF
--- a/subiquity/server/controllers/keyboard.py
+++ b/subiquity/server/controllers/keyboard.py
@@ -16,6 +16,7 @@
 import logging
 from typing import Dict, Optional
 import os
+import shutil
 
 import attr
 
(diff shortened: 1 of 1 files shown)


subiquity/server/controllers/keyboard.py
```diff
@@ -202,6 +202,11 @@ async def set_keyboard(self):
             ['setupcon', '--save', '--force', '--keyboard-only'],
             [resource_path('bin/subiquity-loadkeys')],
             ]
+        if self.app.base_model.source.current.variant == 'desktop':
```
dbungert, line 205: Instead of an explicit check on desktop, I think the key detail is if setxkbmap is around (or not).
What do you think about a shutil.which('setxkbmap') instead, and adding to the list if that is the case?

mwhudson: Yes I was going to say the same thing :)

jpnurmi: Nice, then I don't even need to touch the test :+1:

mwhudson (approved):

dbungert (approved):

PR #1206, opened 2022-03-10
filesystem: add partition alignment table

dbungert:
Part 2 of breaking https://github.com/canonical/subiquity/pull/1200 into chunks.


mwhudson (approved):
I guess pedantically we should change the implementation of parts_and_gaps to use this but as that's going to get entirely replaced with one that does use it fairly soon, that would be makework. I guess PartitionStretchy could use this?

It might make sense for a  VG to define alignment_data too, although only the alignment member would really make any sense (there is apparently a limit on the number of lvs you can create, but it's 2047!). In fact this might be a good idea because subiquity assumes all VGs use the default chunk size which isn't necessarily the case, and it would be nice (eventually, not in this branch) to have probert read the actual chunk size somehow.

PR #1209, opened 2022-03-10
riscv: extend block probing timeout

dbungert:


mwhudson (approved):
I guess my only question is: are we completely sure that platform.machine() returns that value on the architecture Ubuntu calls 'riscv64'?

dbungert:
The test I did on that check was to follow https://wiki.qemu.org/Documentation/Platforms/RISCV - and at least on that rawhide image linked there, this was the resulting value.  Also, it appears to match the magic strings in the kernel.

PR #1217, opened 2022-03-14
Rename UbuntuAdvantage -> UbuntuPro

ogayot:
Nothing too crazy - just replacing UbuntuAdvantage with UbuntuPro wherever it makes sense.

There are still occurrences of "UbuntuAdvantage" for:
* modules & classes that interface with `ubuntu-advantage-tools`.
* integration with cloud-init (which expects the `ubuntu_advantage` key)

We also keep the old "ubuntu-advantage" autoinstall key as a deprecated alias so that we ensure backward-compatibility.

~~Making it a draft so it does not get merged until I open a relevant merge proposal in ubuntu-test-cases.~~
EDIT: https://code.launchpad.net/~ogayot/ubuntu-test-cases/ubuntu-advantage-pro/+merge/416881

ubuntu-pro

Thanks,
Olivier

### DIFF
--- a/subiquity/server/controllers/ubuntu_pro.py
+++ b/subiquity/server/controllers/ubuntu_pro.py
@@ -19,13 +19,13 @@ import os
 
 from subiquity.common.apidef import API
 from subiquity.common.types import (
-    UbuntuAdvantageInfo,
-    UbuntuAdvantageCheckTokenAnswer,
-    UbuntuAdvantageCheckTokenStatus,
+    UbuntuProInfo,
+    UbuntuProCheckTokenAnswer,
+    UbuntuProCheckTokenStatus,
 )
 from subiquity.server.ubuntu_advantage import (
-    InvalidUATokenError,
-    ExpiredUATokenError,
+    InvalidTokenError,
+    ExpiredTokenError,
     CheckSubscriptionError,
     UAInterface,
     UAInterfaceStrategy,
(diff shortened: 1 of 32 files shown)


subiquity/server/controllers/ubuntu_pro.py
```diff
@@ -34,20 +34,20 @@
...
-class UbuntuAdvantageController(SubiquityController):
-    """ Represent the server-side Ubuntu Advantage controller. """
+class UbuntuProController(SubiquityController):
+    """ Represent the server-side Ubuntu Pro controller. """
 
-    endpoint = API.ubuntu_advantage
+    endpoint = API.ubuntu_pro
 
-    model_name = "ubuntu_advantage"
-    autoinstall_key = "ubuntu-advantage"
```
dbungert, line 50: We have shipped subiquity where ubuntu-advantage was a valid autoinstall key, and fixing autoinstall errors is kind of hard for end users.
I would like to have a light compatibility layer that is willing to read in something under the ubuntu-advantage name and populates the value under ubuntu-pro.

dbungert (commented):
see comment about autoinstall

subiquity/server/controllers/ubuntu_pro.py
ogayot: Good point, thanks!

dbungert (approved):

ogayot:
@jpnurmi this PR has API-breaking changes:

HTTP `/ubuntu_advantage`   ->  ` /ubuntu_pro`

PR #1218, opened 2022-03-14
console-conf: fix crash on network info

dbungert:
If you run dryrun for console-conf, go to the network page, go to an
interface, then info, a crash of the form
    TypeError: object str can't be used in 'await' expression
can be seen.  The signature for the core version of get_info_for_netdev
is not async, but a non-async method returning str.

So skip the loading dialog, which wants the awaitable.

### DIFF
--- a/subiquity/server/controllers/network.py
+++ b/subiquity/server/controllers/network.py
@@ -405,4 +405,4 @@ class NetworkController(BaseNetworkController, SubiquityController):
         self.delete_link(dev_name)
 
     async def info_GET(self, dev_name: str) -> str:
-        return self.get_info_for_netdev(dev_name)
+        return await self.get_info_for_netdev(dev_name)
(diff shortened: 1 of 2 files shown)


mwhudson (changes requested):
This breaks info in the subiquity case though (it's the same view in both apps). I think the core get_info_for_netdev needs to be made async.

mwhudson (commented):
Looks like we need this as well:

```
diff --git a/subiquity/server/controllers/network.py b/subiquity/server/controllers/network.py
index 48c432e6c..6a3d3196c 100644
--- a/subiquity/server/controllers/network.py
+++ b/subiquity/server/controllers/network.py
@@ -405,4 +405,4 @@ class NetworkController(BaseNetworkController, SubiquityController):
         self.delete_link(dev_name)
 
     async def info_GET(self, dev_name: str) -> str:
-        return self.get_info_for_netdev(dev_name)
+        return await self.get_info_for_netdev(dev_name)
```

mwhudson (approved):

PR #1219, opened 2022-03-14
Offsets and stretchy

dbungert:


mwhudson:
Going in the right direction I think. You need to pass the gap to the handler in PartitionStretchy.done next?

subiquity/ui/views/filesystem/partition.py
```diff
@@ -405,6 +406,9 @@ def __init__(self, parent, disk, partition=None):
                 lvm_names.remove(partition.name)
         else:
             initial['fstype'] = 'ext4'
+            if gap:
+                initial['size'] = humanize_size(self.gap.size)
```
mwhudson, line 410: This is a change in behaviour I think? size used to be empty by default

subiquity/common/filesystem/manipulator.py
```diff
@@ -197,9 +197,10 @@ def reformat(self, disk):
             self.delete_partition(p, True)
         self.clear(disk)
 
-    def partition_disk_handler(self, disk, partition, spec):
-        log.debug('partition_disk_handler: %s %s %s', disk, partition, spec)
-        log.debug('disk.freespace: {}'.format(gaps.largest_gap_size(disk)))
+    def partition_disk_handler(self, disk, spec, *, partition=None, gap=None):
+        log.debug('partition_disk_handler: %s %s %s %s',
+                  disk, spec, partition, gap)
+        log.debug('disk.freespace: %s', disk.available_for_partitions)
```
mwhudson, line 203: Not quite sure why you'd change this back to disk.available_for_partitions, I thought we wanted to kill that? Maybe just drop the log message I guess...

subiquity/common/filesystem/manipulator.py
```diff
@@ -231,14 +232,16 @@ def partition_disk_handler(self, disk, partition, spec):
...
         log.debug("Successfully added partition")
 
-    def logical_volume_handler(self, vg, lv, spec):
+    def logical_volume_handler(self, vg, spec, *, partition, gap):
+        # keep the partition name for compat with PartitionStretchy.handler
+        lv = partition
+
         log.debug('logical_volume_handler: %s %s %s', vg, lv, spec)
-        log.debug('vg.freespace: {}'.format(gaps.largest_gap_size(vg)))
+        log.debug('vg.freespace: %s', vg.available_for_partitions)
```
mwhudson, line 244: Ditto

subiquity/ui/views/filesystem/partition.py
```diff
@@ -405,6 +406,9 @@ def __init__(self, parent, disk, partition=None):
                 lvm_names.remove(partition.name)
         else:
             initial['fstype'] = 'ext4'
+            if gap:
```
mwhudson, line 409: I think that if partition is None, gap cannot be None here? If not I'm getting confused

mwhudson (approved):
A few comments, but overall :+1:

subiquity/common/filesystem/manipulator.py
dbungert: Yes, the available value was the only reasonable alternative.  Dropping the log statement(s).

subiquity/ui/views/filesystem/partition.py
dbungert: Dropped

subiquity/ui/views/filesystem/partition.py
dbungert: The intention was that a gap or a partition would be supplied, but I seem to have forgot that with this check.  I have backed this out and used a more stringent up-front check.

subiquity/common/filesystem/gaps.py
```diff
@@ -165,3 +165,15 @@ def largest_gap_size(device):
...
+def gap_after_partition(partition):
+    pg = parts_and_gaps(partition.device)
+    part_idx = pg.index(partition)
+    try:
+        gap = pg[part_idx + 1]
+    except IndexError:
+        return None
+    if isinstance(gap, Gap):
+        return gap
+    return None
```
mwhudson, line 179: So I don't know if this is the branch to do it, but I don't think this is quite the right logic or -- more importantly -- interface. I know we've talked about this issue of "sliding" batches of partitions around when edits are made. 

In this branch, if you take an empty 10G disk, add two 1G partitions and then edit the first, there is no gap after it and so the max size is 1G. In main, the max size is 9G and that feels more right. So I think the interface here needs to be "get me the movable partitions after this one and then any trailing gap". And then PartitionStretchy can call this API and only look at the gap bit, and the code that handles the edits can adjust the offsets of the movable partitions.

I hesitate to pile more onto this branch but actually it might not be so bad. The hardest part might be coming up with a name for the API...

mwhudson (changes requested):
I think there's one more bit we should get done in this branch. But I'm happy for you to try to talk me out of it.

subiquity/common/filesystem/gaps.py
dbungert: OK, I'm ejecting what is currently called gap_after_partition from this branch and we'll sort it out.  I can live with max_size being wrong until then.

mwhudson (approved):
Thanks.

PR #1220, opened 2022-03-15
better logic for where to put the bios_grub partition

mwhudson:
Probably too complicated, but also probably v2 compatible!


dbungert:
So far so good

mwhudson:
Mostly subsumed by #1232 (that code isn't v2-partitioning at all and this is, but it makes sense to add v2-partitioning-ness after that has been merged)

PR #1222, opened 2022-03-15
Optional v2 storage

mwhudson:
Add a way to turn on more interesting gaps behaviour on the command line, and enable deleting preserved partitions in that mode. I think this approach has legs, it lets us break everything in the new world while not worrying about breaking existing behaviour. With a finished version of [dbungert:offsets-and-stretchy](https://github.com/dbungert/subiquity/tree/offsets-and-stretchy) merged in it might not even result in completely mad behaviour!

### DIFF
--- a/subiquity/cmd/server.py
+++ b/subiquity/cmd/server.py
@@ -71,6 +71,8 @@ def make_server_args_parser():
         '--output-base', action='store', dest='output_base',
         default='.subiquity',
         help='in dryrun, control basedir of files')
+    parser.add_argument(
+        '--storage-version', action='store', type=int, default=1)
     return parser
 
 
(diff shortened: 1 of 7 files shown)


dbungert:
So far so good.  I really like the feature flag argument.

mwhudson:
This is now just the absolute minimal version.

mwhudson:
Scratch that, I put the more interesting parts_and_gaps back because that's really fundamental to everything else.

dbungert (approved):
LGTM.
As an additional test, I experimented with the API tests set for storage version 2.  All tests pass except for test_v2_flow, which also passes if I update it to not assume a certain order for objects in the partitions array and lookup partitions by their number.

PR #1224, opened 2022-03-15
probert: update for python 3.10 fix (lp-1964571)

dbungert:


mwhudson (approved):
Do we want to merge this to server/jammy too? Or create a server/jammy branch for probert and cherry pick things into that?

dbungert:
My hunch for probert is that 100% of the things that land there in the next monthish are stuff we want for Jammy, so I lean towards leaving it alone and just pulling from main.
If we find ourselves landing some feature we can do a branch on probert, and if that day comes it shouldn't require touching subiquity at all since we pick up probert by a fixed commit (which could be on any branch).
What do you think?

dbungert:
therefore I would want to merge this PR to the subiquity server/jammy branch.

mwhudson:
:+1: -- my only hesitation is: does the probert referenced by the server/jammy branch invoke os-prober?

dbungert:
It does and we don't need it to.  It would be safer to back out the os-prober call (I saw an os-prober looking failure in one of Heinrich's logs)
New proposal:
1) Merge this as is
2) create probert stable branch
3) Change the probert stable branch to not include a call to os-prober
4) merge to server/jammy that resulting probert commit

mwhudson:
That all sounds good to me!

PR #1226, opened 2022-03-16
Systemd experimental support

didrocks:
Grow a new API for enabling/disabling systemd experimental support. We do this by chaining a specific command= in the boot section. We preserve the user customized command= content when systemd experimental support is not enabled.

### DIFF
--- a/system_setup/common/wsl_conf.py
+++ b/system_setup/common/wsl_conf.py
@@ -60,15 +60,16 @@ conf_type_to_file = {
 }
 
 
-def wsl_config_loader(data, config_ref, id):
+def wsl_config_loader(config_ref, id):
     """
     Loads the configuration from the given file type,
     section and reference config.
 
-    :param data: dict, the data to load into
-    :param pathname: string, the path to the file to load
+    :param config_ref: dict, default entries
     :param id: string, the name of the section to load
+    return the data loaded
     """
+    data = {}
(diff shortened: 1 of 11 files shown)


system_setup/common/wsl_conf.py
```diff
@@ -74,9 +80,20 @@ def wsl_config_loader(data, config_ref, id):
             conf_sec_list = config[conf_sec]
             for conf_item in conf_sec_list:
                 if conf_item in config_ref[id][conf_sec]:
+                    # Handle systemd experimental feature (string -> bool)
+                    if (
+                            conf_sec.lower() == "boot" and
+                            conf_item.lower() == "command"):
```
CarlosNihelton, line 86: Nitpick: I'm not quite sure why this line break after the `if(` clause.

CarlosNihelton (approved):
Great job! I don't see any issues. As before, it only affects `system_setup`, so we should be fine. Just one minor nitpick, but feel free to disregard if you don't agree with.

system_setup/common/wsl_conf.py
didrocks: That was for the linter to not complain, let’s see what other says, I dno’t really like it either (but 80 characters, avoid alignements errors…)

dbungert (approved):

PR #1227, opened 2022-03-16
Allow to use HTTP 409 for tasks cancelled on server end

ogayot:
_This was originally implemented as part of https://github.com/canonical/subiquity/pull/1196 to avoid potential UI freezes when tasks are not cancelled._ 

We now allow the server controller to raise a HTTPConflict exception inHTTP request handlers.

This will indicate that we cancelled a task that we were waiting on to satisfy a HTTP request.

The problem that we are trying to solve is this:
 * the client sometimes runs an HTTP query that is blocking on the server side (most of the time it has the wait=True parameter). The server will most likely await on a coroutine before returning an HTTP response.
 * in the meantime the client can send another HTTP query that will cancel the task we are waiting on. E.g., skip the check for updates.
 * the response to the original HTTP request will not be able to contain the information that the clients expects ; because the task to retrieve the information was cancelled.
 * the client should have a simple way to be informed that the server wasn't able to complete the action requested.

Example of usage:

```python
# Client side
async def _wait_drivers(self) -> List[str]:
     try:
         response: DriversResponse = await self.endpoint.GET(wait=True)
     except ServerTaskCancelled:
         log.warning("_wait_drivers: task interrupted on server end")
         return []
     return response.drivers

# Server side
async def GET(self, wait: bool = False) -> DriversResponse:
    if wait:
        try:
            await self._drivers_task
        except asyncio.CancelledError:
            raise HTTPConflict(text="Drivers listing task was interrupted.")
        return DriversResponse(install=self.model.do_install, drivers=self.drivers)

```

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -69,6 +69,11 @@ class Abort(Exception):
         self.error_report_ref = error_report_ref
 
 
+class ServerTaskCancelled(Exception):
+    """ Exception to raise when the server cancels a task that we are waiting
+    on through an HTTP query. """
+
+
 DEBUG_SHELL_INTRO = _("""\
 Installer shell session activated.
 
@@ -221,7 +226,9 @@ class SubiquityClient(TuiApplication):
             raise Abort(ref)
         try:
             response.raise_for_status()
-        except aiohttp.ClientError:
+        except aiohttp.ClientError as e:
+            if isinstance(e, aiohttp.ClientResponseError) and e.status == 409:
+                raise ServerTaskCancelled()
             report = self.error_reporter.make_apport_report(
                 ErrorReportKind.SERVER_REQUEST_FAIL,
                 "request to {}".format(response.url.path))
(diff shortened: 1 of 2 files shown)


mwhudson:
Do we still want to do this? It's been open for a while.

ogayot:
> Do we still want to do this? It's been open for a while.

AFAIK, there hasn't been any discussion about this change.
Right now there isn't a specific need for it anymore but I'd like to keep it as an option should the need such mechanism at some point.

I can cancel the PR or mark it draft if that helps.

mwhudson:
I'm not sure about using HTTP 409 for this, I think that might be a bit confusing. We could just define our own exception type to raise in the server and communicate the fact that it's a cancellation via the x-override-status header?

Did you think about just catching CanceledError in the generic server code?

ogayot:
>  I'm not sure about using HTTP 409 for this, I think that might be a bit confusing. We could just define our own exception type to raise in the server and communicate the fact that it's a cancellation via the x-override-status header?

Yes, using x-status would work I guess. That's a good alternative.

> Did you think about just catching CanceledError in the generic server code?

I did think about it at the time but was afraid we would swallow a bunch of irrelevant exceptions. Having a little more experience with cancellation now, I might try it out. Thanks.

PR #1233, opened 2022-03-17
FilesystemModel._actions_from_config: catch config is None

xypron:
None is not iterable. If _actions_from_config is called with config is
None, an exception "TypeError: 'NoneType' object is not iterable"
is raised. Let's just log this situation. (LP: #1965169)

Signed-off-by: Heinrich Schuchardt <heinrich.schuchardt@canonical.com>

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1157,6 +1157,9 @@ class FilesystemModel(object):
         byid = {}
         objs = []
         exclusions = set()
+        if config is None:
+            log.debug("config is None")
+            return []
         for action in config:
             if is_probe_data and action['type'] == 'mount':
                 if not action['path'].startswith(self.target):


xypron:
The problem occurred if the target device was already partitioned by a previous installer run.

ogayot:
I don't think this is the right way to fix the issue but would like somebody more familiar with the FS controller to confirm. From what I understand from the logs ; when the crash occurs, we went through the following path:

```python
        if choice is not None:
            coro = self.endpoint.guided.POST(choice)  # <-- choice was not None when the crash occurred
        else:
            coro = self.endpoint.GET()
        status = await self.app.wait_with_progress(coro)
        # XXX When the occurred, we had:
        # status.status = "PROBING"
        # status.bootloader = None
        self.model = FilesystemModel(status.bootloader)
        self.model.load_server_data(status)
```
Therefore, we constructed a `FilesystemModel` with `None` as the bootloader and attempted to parse data that is incomplete.
I think we should rather wait until the probing information is finished (unless maybe it should not be probing anymore at this point). Maybe we can implement something like this:
```python
coro = self.endpoint.guided.POST(choice, wait=True)
coro = self.endpoint.GET(wait=True)
```
but I guess this is more likely an action on my end.

ogayot (changes requested):
I investigated this further and added the details at https://bugs.launchpad.net/subiquity/+bug/1962205

I tested your fix in dry-run mode after applying
```diff
     async def guided_POST(self, data: GuidedChoice) -> StorageResponse:
         log.debug(data)
+        self._udev_event()
         self.guided(data)
         return await self.GET()
```
and although it allows us to get to the next screen, all the relevant information are missing. See captures below:

[Expected]
with-info
[With the fix applied and udev event being simulated]
with-fix

xypron:
I understand that this patch may not be the complete solution. But it is good enough to stop a crash. If you can avoid calling the function with None in a later patch, that will simply make these extra three lines dead code which does no harm.

mwhudson:
#1234 is a better fix I think.

PR #1234, opened 2022-03-18
never return a PROBING response from guided_POST

mwhudson:


dbungert (approved):

mwhudson:
This should be backported to server/jammy I guess...

dbungert:
Merged to jammy branch in #1253

PR #1236, opened 2022-03-18
integration: use output-base=$tmpdir

dbungert:


didrocks:
LGTM, and this is a needed change to avoid writing in current subiquity directory.

didrocks (approved):

ogayot (approved):
Thanks for doing this!

You can remove the variable assignment at line 4 since we don't use it anymore.
```sh
testschema=.subiquity/test-autoinstall-schema.json
```

dbungert:
> Just a note: I would just echo the `tmpdir` variable so that, in case of failure, I would know which directory is the current one (multiple runs with failures would keep the directories on disk)

I meant for this to happen, but the current implementation uses a trap function on ERR, which doesn't run in every case that we would want it to.  trap on EXIT seems more what we want.

PR #1241, opened 2022-03-21
probert: use os-prober only if asked

dbungert:
* add command line arg for using os-prober
* use probe_types to pass message along

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -185,7 +185,7 @@ parts:
       - libnl-route-3-dev
     source: https://github.com/canonical/probert.git
     source-type: git
-    source-commit: e9da1c5958917c6ee39a90230ffc7497c0b8fa68
+    source-commit: 66ca2f19f421de55ed3f6b5a7c31aa68d4b04ca5
     requirements: [requirements.txt]
     stage:
       - "*"
--- a/subiquity/cmd/server.py
+++ b/subiquity/cmd/server.py
@@ -73,6 +73,8 @@ def make_server_args_parser():
         help='in dryrun, control basedir of files')
     parser.add_argument(
         '--storage-version', action='store', type=int, default=1)
+    parser.add_argument(
+        '--use-os-prober', action='store_true', default=False)
     return parser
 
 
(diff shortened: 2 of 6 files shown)


dbungert:
@jpnurmi - when this merges, ubuntu-desktop-installer will want https://github.com/canonical/ubuntu-desktop-installer/pull/749

subiquity/cmd/server.py
```diff
@@ -73,6 +73,8 @@ def make_server_args_parser():
         help='in dryrun, control basedir of files')
     parser.add_argument(
         '--storage-version', action='store', type=int, default=1)
+    parser.add_argument(
+        '--use-os-prober', action='store', type=bool, default=False)
```
ogayot, line 77: I don't think this does what you want, or does it?
`action='store'` indicates that the next argument must be consumed and converted into a boolean, e.g.:

```python
server.py --use-os-prober foo     # bool("foo") => True
server.py --use-os-prober ""      # bool("") => False
server.py --use-os-prober False   # bool("False") => True
server.py --use-os-prober True    # bool("True") => True
server.py --use-os-prober         # -> will fail / missing argument
```

I think this is what you want instead:
```python
parser.add_argument(
    '--use-os-prober', action='store_true')
```

ogayot (changes requested):

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2019 Canonical, Ltd.
```
ogayot, line 1: 2022 :D

subiquity/cmd/server.py
dbungert: Yes, I wanted store_true.

ogayot (approved):

snapcraft.yaml
```diff
@@ -183,9 +183,9 @@ parts:
       - libnl-3-dev
       - libnl-genl-3-dev
       - libnl-route-3-dev
-    source: https://github.com/canonical/probert.git
+    source: https://github.com/dbungert/probert.git
```
mwhudson, line 186: Not this bit :)

mwhudson (approved):

PR #1243, opened 2022-03-22
filesystem: add_partition requires offset

dbungert:


mwhudson (approved):
Nice. The "offset=0" ones make me twitch a bit but I guess it's not important.

PR #1244, opened 2022-03-22
Removes apt python library from system_setup

CarlosNihelton:
It's been quite painful trying to work with apt python library in the WSL snap environment so far, so today we decided to drop it in favor of `subprocess.run` with the environment adjusted to ensure we can run system `apt` instead of adding it to the snap.

This pull request only affects the scope of `system_setup` and adjusts `apply_locale()` and `create_user()` functions to perform their duties using system binaries.

Besides the subiquity CI tests, this code has been tested in WSL and behaved as expected.

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -100,10 +100,16 @@ validate () {
             fi
             # Extract value of the LANG variable from etc/default/locale (with or without quotes)
             lang="$(grep -Eo 'LANG=([^.@ _]+)' $tmpdir/etc/default/locale | cut -d= -f 2- | cut -d\" -f 2-)"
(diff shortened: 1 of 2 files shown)


dbungert (approved):

dbungert:
LGTM, @didrocks feel free to merge when you're ready

didrocks (approved):
I’m not a big fan of using hardcoded paths for useradd and others outside of the snap, but that would do for now to unblock us. Can we revisit it later on? I think the issue may be due to PATH not containing what you need. However, here, we would like to use the snap version for adding users IMHO.

Otherwise, good, this strips the environment where it’s needed as we discussed, let’s merge.

PR #1245, opened 2022-03-23
storage/v2: add_partition takes a gap arg

dbungert:


dbungert:
@jpnurmi - API change on storage/v2/add_partition, gap object is required.   Send a gap object acquired from the partition list returned by a previous step.
```
data = {
    'disk_id': disk_id,
    'gap': gap,
    'partition': {
        'format': 'ext3',
        'mount': '/',
    }
}
add_resp = await inst.post('/storage/v2/add_partition', data)
```

subiquity/server/controllers/filesystem.py
```diff
@@ -351,7 +352,7 @@ async def v2_add_boot_partition_POST(self, disk_id: str) \
         self.add_boot_disk(disk)
         return await self.v2_GET()
 
-    async def v2_add_partition_POST(self, data: ModifyPartitionV2) \
+    async def v2_add_partition_POST(self, data: AddPartitionV2) \
```
mwhudson, line 355: Er. Shouldn't this function actually use data.gap now?

mwhudson (approved):
One quibble which may or may not be trivial.

subiquity/server/controllers/filesystem.py
dbungert: :facepalm: yes

subiquity/server/controllers/filesystem.py
```diff
@@ -351,29 +352,26 @@ async def v2_add_boot_partition_POST(self, disk_id: str) \
...
             -> StorageResponseV2:
         log.debug(data)
         if data.partition.format is None:
             raise ValueError('add_partition must supply format')
         if data.partition.boot is not None:
             raise ValueError('add_partition does not support changing boot')
         disk = self.model._one(id=data.disk_id)
-        gap = gaps.largest_gap(disk)
-        if gap.size == 0:
```
mwhudson, line 363: Huh it's not like it matters now but this is buggy I think -- gaps.largest_gap returns None if there is no gap.

mwhudson (approved):

PR #1246, opened 2022-03-24
Faster curtin replay tests

dbungert:
Allow the TIMESCALE value to flow to the curtin command runner.


ogayot (approved):
Oops! I guess this explains why integration tests felt so slow these days...
Thanks for cleaning up my mess!

dbungert:
No problem - reading the diff on the commit that started it (0ca56056ca58647c8541ad7964304e3e3d46d08d) this was not at all clear!

PR #1247, opened 2022-03-24
lables: handle gaps in storage_version=2

dbungert:


subiquity/models/filesystem.py
```diff
@@ -35,7 +35,10 @@
 
 log = logging.getLogger('subiquity.models.filesystem')
 
-MiB = 1024 * 1024
+
+kiB = 1 << 10
+MiB = 1 << 20
+GiB = 1 << 30
```
mwhudson, line 41: I mean, fine, but not uses anywhere in this branch?

mwhudson (commented):
Fine I guess but can we just remove the field instead? The desktop client can compute the amount of space used / available if it wants to for some reason? What does @jpnurmi think?

jpnurmi:
Right, we don't need it anymore now that we have the gaps.

subiquity/models/filesystem.py
dbungert: GiB was used in the tests.  I wanted to have them predefined to make it easier for later uset.

dbungert:
OK great, I'll just drop free_for_partitions then instead.  I prefer that.

dbungert:
Replaced by #1248

PR #1249, opened 2022-03-30
Gap after partition

dbungert:


dbungert:
#1250 is a better version of this

PR #1250, opened 2022-04-01
keep partition offsets up to date as they are deleted and resized

mwhudson:
Also correct the max_size in the partition stretchy.

### DIFF
--- a/subiquity/common/filesystem/gaps.py
+++ b/subiquity/common/filesystem/gaps.py
@@ -169,13 +169,19 @@ def largest_gap_size(device):
 
 
 def movable_trailing_partitions_and_gap(partition):
-    pg = parts_and_gaps(partition.device)
-    part_idx = pg.index(partition)
+    pgs = parts_and_gaps(partition.device)
+    part_idx = pgs.index(partition)
     trailing_partitions = []
-    for idx in range(part_idx + 1, len(pg)):
-        p = pg[idx]
-        if isinstance(p, Partition):
-            trailing_partitions.append(p)
+    in_extended = partition.flag == "logical"
+    for idx in range(part_idx + 1, len(pgs)):
+        pg = pgs[idx]
+        if isinstance(pg, Partition):
+            if in_extended and pg.flag != "logical":
+                break
+            trailing_partitions.append(pg)
         else:
-            return (trailing_partitions, p)
+            if pg.in_extended == in_extended:
+                return (trailing_partitions, pg)
+            else:
+                return (trailing_partitions, None)
     return (trailing_partitions, None)
--- a/subiquity/common/filesystem/gaps.py
+++ b/subiquity/common/filesystem/gaps.py
@@ -173,8 +173,7 @@ def movable_trailing_partitions_and_gap_size(partition):
     part_idx = pgs.index(partition)
     trailing_partitions = []
     in_extended = partition.flag == "logical"
-    for idx in range(part_idx + 1, len(pgs)):
-        pg = pgs[idx]
+    for pg in pgs[part_idx + 1:]:
         if isinstance(pg, Partition):
             if pg.preserve:
                 break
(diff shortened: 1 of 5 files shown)


subiquity/common/filesystem/gaps.py
```diff
@@ -165,3 +166,24 @@ def largest_gap_size(device):
...
     return 0
+
+
+def movable_trailing_partitions_and_gap_size(partition):
+    pgs = parts_and_gaps(partition.device)
+    part_idx = pgs.index(partition)
+    trailing_partitions = []
+    in_extended = partition.flag == "logical"
+    for idx in range(part_idx + 1, len(pgs)):
+        pg = pgs[idx]
```
ogayot, line 177: Would something like the below make sense? It seems to me that you don't really need to store the index.
```python
for pg in pgs[part_idx + 1:]:
```
or
```python
from itertools import islice
for pg in islice(pgs, part_idx + 1, None):
```

subiquity/common/filesystem/tests/test_gaps.py
```diff
@@ -165,3 +167,153 @@ def test_half_extended_half_logical(self):
...
+            min_end_offset=10, primary_part_limit=10))
+        # 0----10---20---30---40---50---60---70---80---90---100
+        # #####[ p1                                   ]#####
+        m, d = make_model_and_disk(size=100)
+        p = make_partition(m, d, offset=10, size=80)
+        self.assertEqual(
+            ([], 0),
+            gaps.movable_trailing_partitions_and_gap_size(p))
+
+    def test_immediately_trailing_gap(self):
```
dbungert, line 193: These two tests in a row strike me as a bit odd.  In this test, p2 is not preserved, so because there is a gap before it, we can return the small gap as the answer.  In `test_one_trailing_movable_partition_and_gap`, we instead show the gap after p2, which only makes sense if we slide p2 to the end, but we could do that here also right?

I'm willing to accept 'follow-on branch' as an answer.

subiquity/common/filesystem/tests/test_gaps.py
```diff
@@ -165,3 +167,153 @@ def test_half_extended_half_logical(self):
...
+        self.use_alignment_data(PartitionAlignmentData(
+            part_align=10, min_gap_size=1, min_start_offset=10,
+            min_end_offset=10, primary_part_limit=10))
+        # 0----10---20---30---40---50---60---70---80---90---100
+        # #####[ p1      ]         [ p2 ]              #####
```
dbungert, line 198: I love the ASCII art diagrams!
Nitpick: there is a slight inconsistency in the placement of the end bracket and the intended end position of the partition in the test, compared to some of the others such as `test_one_trailing_movable_partition_and_gap`.  This nitpick may or may not be related to me squinting at off-by-one errors in the curtin resize tests.

dbungert (approved):

subiquity/common/filesystem/gaps.py
mwhudson: Oh yeah, totally.

subiquity/common/filesystem/tests/test_gaps.py
mwhudson: Ah hah yes you are correct but

 1) I don't particularly want to think about the fact that in this case you might end up having to slide partitions by varying amounts
 2) fortunately I don't think there is or will be a way to create the test_immediately_trailing_gap situation, right? With this branch there's no way to create a hole without preserved partitions being around. So perhaps the correct response is to delete that test.

subiquity/common/filesystem/tests/test_gaps.py
mwhudson: > I love the ASCII art diagrams!

Me too! I wondered about implementing a parser for them so they could actually be the input to the test...

> Nitpick: there is a slight inconsistency in the placement of the end bracket and the intended end position of the partition in the test, compared to some of the others such as `test_one_trailing_movable_partition_and_gap`. This nitpick may or may not be related to me squinting at off-by-one errors in the curtin resize tests.

Even more of a reason to delete the test!

PR #1251, opened 2022-04-01
Subiquity doesn't work with remote img as source

p53:


subiquity/models/source.py
```diff
@@ -89,14 +89,19 @@ def load_from_file(self, fp):
 
     def get_source(self):
         path = os.path.join(self._dir, self.current.path)
+        scheme = self.current.type
+
+        if self.current.path.startswith("http"):
```
dbungert, line 94: This would misbehave if the local file name really starts with http.  This is arguably a poor choice for a file name but I can see it happening if someone caches a local copy of the file and uses the URI as the input for generation of the name.  So I would like a deeper inspection of the path value to check if it looks like a URI or not.

dbungert (changes requested):
Hi @p53, thanks for the proposal.  Before we can merge this we will need to ask that you sign the Contributor License Agreement. https://ubuntu.com/legal/contributors - Thanks!

On this specific proposal, see the comment, plus I would like to see a test covering this case so that we don't regress it.

PR #1255, opened 2022-04-05
autoinstall: fix interaction with early-commands 

dbungert:
Early commands is currently documented as follows:
> The autoinstall config is available at /autoinstall.yaml
> (irrespective of how it was provided) and the file will be re-read
> after the early-commands have run to allow them to alter the config
> if necessary.

The previous change to let cloud take precedence over iso location broke
this functionality.  Fix that by copying to the iso location the file of
choice.

Continuing from #1237


subiquitycore/file_util.py
```diff
@@ -73,3 +74,16 @@ def generate_config_yaml(filename, content, **kwargs):
         now = datetime.datetime.utcnow()
         tf.write(f'# Autogenerated by Subiquity: {now} UTC\n')
         tf.write(yaml.dump(content))
+
+
+def copy_file_if_exists(source, target):
+    """If source exists, copy to destination.  Ignore error that dest may be a
+    duplicate.  Create destination parent dirs as needed."""
+    if source is None or not os.path.exists(source):
```
ogayot, line 82: What would you think of marking `source` as a `str` and dropping the check for `source is None`?
I understand that it means making the call to `copy_file_if_exists` conditional (i.e., based on whether `self.autoinstall is None`) but I don't know about silently ignoring `None` values in this function.

ogayot (approved):
Would a symlink be an option rather than copying the contents of the file?

dbungert:
> Would a symlink be an option rather than copying the contents of the file?

Maybe.  I'm learning about the behavior of autoinstall during a refresh.  It would be strange if early-commands ran twice and double-modified the file, but double-running early commands is also probably bad for many other reasons.

subiquitycore/file_util.py
dbungert: I'll do this if we don't end up with the symlink

ogayot: Alright, choice is up to you!

mwhudson:
> > Would a symlink be an option rather than copying the contents of the file?

I worry a little about early-commands editing the file in a way that does not respect symlinks. E.g. one that does "cp /cdrom/my-autoinstall.yaml /autoinstall.yaml" or similar.

> Maybe. I'm learning about the behavior of autoinstall during a refresh. It would be strange if early-commands ran twice and double-modified the file, but double-running early commands is also probably bad for many other reasons.

early-commands should definitely only run once!

subiquity/server/tests/test_server.py
```diff
@@ -40,55 +42,73 @@ def setUp(self):
...
-            self.assertEqual(expected, fp.read())
+        self.assertIsNone(self.server.select_autoinstall())
 
     def test_bogus_autoinstall_argument(self):
         self.server.opts.autoinstall = self.path('nonexistant.yaml')
         with self.assertRaises(Exception):
-            self.server.select_autoinstall_location()
+            self.server.select_autoinstall()
+
+    def test_early_commands_changes_autoinstall(self):
```
mwhudson, line 94: Could you add a test or modify this test to simulate the autoinstall config being supplied via cloud-init config?

subiquity/server/server.py
```diff
@@ -539,7 +541,7 @@ async def wait_for_cloudinit(self):
                 "cloud-init status: %r, assumed disabled",
                 status_txt)
 
-    def select_autoinstall_location(self):
+    def select_autoinstall(self):
```
mwhudson, line 544: I'm not sure this method name is quite appropriate any more. But I don't have any better ideas.

mwhudson (approved):
I'd like one more test please, as I thought you'd missed a case but on close reading don't think you have.

subiquitycore/file_util.py
dbungert: Doing the non-symlink implementation, so I have done the above.

subiquity/server/server.py
dbungert: I sat here for a few minutes pondering a better name and I have nothing.  If we think of one later I'm open to changing it.

subiquity/server/tests/test_server.py
dbungert: Good point, it's the entire reason for the PR.  Done!

PR #1256, opened 2022-04-05
Enables passing the autoinstall CLI to the server.

CarlosNihelton:
The `system_setup` server was prepared to receive this CLI option, but the TUI client was not, thus in live-run that option had no effect.


mwhudson (approved):
Seems fine. Have you seen the way the equivalent code in subiquity forwards all unknown arguments to the server? Would that make this sort of thing less error prone?

PR #1257, opened 2022-04-06
Some more code cleanup

ogayot:
Changes included:
* added/updated type hints where type used was ambiguous/wrong
* avoided redefinition of types that would mask the former symbol
* ~~introduced `ProcessWithArgs` to avoid adding member variables dynamically (type checkers don't like it)~~
* fixed typo in `addresses` variable name
* renamed `arun_command` variable that was interfering the `arun_command` function when using `ctags -R`
* accepted `Sequence[str]` (aka. `List[str]` or `Tuple[str]`) for commands instead of just `List[str]` in `utils.py`. ubuntu_pro.py was already passing a `Tuple[str]` to one of these functions so mypy was not happy about it.
* helped type checkers figure out that `returncode` cannot be `None` anymore after calling `.communicate()`

When using `mypy`, it lowers the number of errors by 21.

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -21,7 +21,7 @@ import os
 import signal
 import sys
 import traceback
-from typing import Optional
+from typing import Dict, List, Optional
 
 import aiohttp
 
@@ -118,7 +118,7 @@ class SubiquityClient(TuiApplication):
         "Progress",
         ]
 
-    variant_to_controllers = {}
+    variant_to_controllers: Dict[str, List[str]] = {}
 
     def __init__(self, opts):
         if is_linux_tty():
--- a/subiquity/common/filesystem/boot.py
+++ b/subiquity/common/filesystem/boot.py
@@ -15,7 +15,7 @@
 
 import abc
 import functools
-from typing import Optional
+from typing import Any, Optional
 
 import attr
 
@@ -125,7 +125,7 @@ class SetAttrPlan(MakeBootDevicePlan):
 
     device: object
     attr: str
-    val: str
+    val: Any
 
     def apply(self, manipulator):
         setattr(self.device, self.attr, self.val)
--- a/subiquity/models/network.py
+++ b/subiquity/models/network.py
@@ -15,12 +15,12 @@
 
 import logging
 
-from subiquitycore.models.network import NetworkModel
+from subiquitycore.models.network import NetworkModel as CoreNetworkModel
 
 log = logging.getLogger('subiquity.models.network')
 
 
-class NetworkModel(NetworkModel):
+class NetworkModel(CoreNetworkModel):
 
     def __init__(self):
         super().__init__("subiquity")
--- a/subiquity/server/controllers/reporting.py
+++ b/subiquity/server/controllers/reporting.py
@@ -25,14 +25,12 @@ from curtin.reporter.events import (
     report_start_event,
     status,
     )
-from curtin.reporter.handlers import (
-    LogHandler,
-    )
+from curtin.reporter.handlers import LogHandler as CurtinLogHandler
 
 from subiquity.server.controller import NonInteractiveController
 
 
-class LogHandler(LogHandler):
+class LogHandler(CurtinLogHandler):
     def publish_event(self, event):
         level = getattr(logging, event.level)
         logger = logging.getLogger('')
--- a/subiquity/server/runner.py
+++ b/subiquity/server/runner.py
@@ -78,6 +78,8 @@ class LoggedCommandRunner:
     async def wait(self, proc: asyncio.subprocess.Process) \
             -> subprocess.CompletedProcess:
         stdout, stderr = await proc.communicate()
+        # .communicate() forces returncode to be set to a value
+        assert(proc.returncode is not None)
         if proc.returncode != 0:
             raise subprocess.CalledProcessError(proc.returncode, proc.args)
         else:
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -141,7 +141,7 @@ class MetaController:
         return self.app.variant
 
     async def ssh_info_GET(self) -> Optional[LiveSessionSSHInfo]:
-        ips = []
+        ips: List[str] = []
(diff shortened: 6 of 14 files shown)


subiquity/server/runner.py
```diff
@@ -17,11 +17,23 @@
 from contextlib import suppress
 import os
 import subprocess
-from typing import List, Optional
+from typing import Any, List, Optional
 
 from subiquitycore.utils import astart_command
 
 
+class ProcessWithArgs:
```
dbungert, line 25: This is hard to look at to convince myself, with code review only, that it is not introducing a logic change.
 Especially as there is some typing handling already in place.  Do you feel like existing tests cover the users of ProcessWithArgs well?

subiquity/server/server.py
```diff
@@ -141,7 +141,7 @@ async def client_variant_GET(self) -> str:
         return self.app.variant
 
     async def ssh_info_GET(self) -> Optional[LiveSessionSSHInfo]:
-        ips = []
+        ips: List[str] = []
```
dbungert, line 144: It looks like all users of actual_global_ip_addresses want them in strs anyhow, so maybe that should just return them as a list of strs.  Suggest looking into it in a follow-on PR.

dbungert (changes requested):

subiquity/client/client.py
```diff
@@ -118,7 +118,7 @@ def make_ui(self):
         "Progress",
         ]
 
-    variant_to_controllers = {}
+    variant_to_controllers: Dict[str, List[str]] = {}
```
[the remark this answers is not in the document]
ogayot: That's not my intention, no :)

I used the following definition from `system_setup/client/client.py` to specify these hints:

```
    controllers = [
        "Welcome",
        "WSLIdentity",
        "WSLConfigurationBase",
        "Summary",
    ]


    variant_to_controllers = {
        "wsl_setup": controllers,                                                                                              
        "wsl_configuration": [
            "WSLConfigurationBase",
            "WSLConfigurationAdvanced",
            "Summary",
        ]
    }
```

 `variant_to_controllers` seems to be used only by `system_setup`.

subiquity/server/server.py
ogayot: Yes, that seems correct.
I quite like passing around `ip_address` objects rather than strings though. The API is quite nice and I can imagine that at some point we will have a use-case where we need to do things like the following:

```python
if addr.version == 4:
    print(addr)
else:
    print(f"[{addr}]")
```
We don't have this kind of use-case today so I'm down to change it if you like `str` objects better.

subiquity/server/runner.py
ogayot: Yeah, I kind of agree with you... It felt like a good option when I wrote it but now I wonder if that was really a good thing. Maybe I should drop this patch and think again how to improve this.

The types hints in place currently are sort of broken for this use case ; because we rely on a member variable that we add dynamically, but isn't present in we just instantiate a `asyncio.subprocess.Process`. This is the error I'm trying to address:
```
subiquity/server/runner.py:75: error: "Process" has no attribute "args"
subiquity/server/runner.py:82: error: "Process" has no attribute "args"
subiquity/server/runner.py:85: error: "Process" has no attribute "args"
```

subiquity/server/server.py
dbungert: Yea, I can see the use case for the objects.  Let's leave it.

subiquity/server/runner.py
dbungert: If we have a test, I'm OK with it.
If we don't have a test, I'd like to see one or pull this hunk out and bring it back later with a test.

subiquity/common/filesystem/boot.py
```diff
@@ -125,7 +125,7 @@ class SetAttrPlan(MakeBootDevicePlan):
 
     device: object
     attr: str
-    val: str
+    val: Union[str, bool]
```
mwhudson, line 128: Really I think Any is sort of more appropriate here. (I guess it has to be something that can be serialized as yaml and /probably/ `Union[str,bool,int]` would cover anything this is ever likely to be but...)

subiquity/models/network.py
```diff
@@ -15,12 +15,12 @@
 
 import logging
 
-from subiquitycore.models.network import NetworkModel
+import subiquitycore.models.network
```
mwhudson, line 18: I think I would prefer "from subiquitycore.models.network import NetworkModel as CoreNetworkModel" here

subiquity/server/controllers/reporting.py
```diff
@@ -25,14 +25,12 @@
     report_start_event,
     status,
     )
-from curtin.reporter.handlers import (
-    LogHandler,
-    )
+import curtin.reporter.handlers
```
mwhudson, line 28: same as before

subiquity/server/runner.py
mwhudson: So this touches on something I've been a bit concerned about anyway: curtin.py has an API where you call start_curtin_command and get a Command object, which then has methods like .wait() and so on. But runner.py has an API where you get a runner and call start() to get a, well, thing, that you then pass back to other runner methods. I wonder if runner.py having an API more like curtin.py would be better. Viewed one way, this is taking what you've done here and going a bit further with it (and will less attribute forwarding please).

If you decide this is a good idea, it would be good for another branch I think.

mwhudson (changes requested):
A few quibbles but nothing major. I think I agree with Dan that the ProcessWithArgs stuff should be a separate branch though.

subiquity/server/runner.py
ogayot: Okay I'm dropping this patch for now. Thank you both for your input!

subiquity/common/filesystem/boot.py
ogayot: Yes, that's more sensible.

mwhudson (approved):

dbungert (approved):

PR #1258, opened 2022-04-07
Ensures snap_dir='/' if not better specified.

CarlosNihelton:
UDI sets the SNAP env var to '.' for development purposes, affecting the behavior of `system_setup`, which server is started by the GUI.

See: https://github.com/canonical/ubuntu-desktop-installer/commit/9eb6f04

It is unlikely that under test or production that env var will ever by just '.'. On the other hand in dry-run we want this controller to interpret it as '/' if not properly set, thus discarding the '.' directory.


didrocks:
Ternary expression! Ok :p Let’s still get it :) It makes sense for testing purposes.

didrocks (approved):

PR #1259, opened 2022-04-07
Wsl remove systemd

jibel:
Removed systemd from the UI. However it can still be enabled manually from wsl.conf.


didrocks:
Changes only impacts the UI until systemd support is ready. Merging then, keeping the API backend.

didrocks (approved):

PR #1263, opened 2022-04-08
filesystem: going back to guided also does reset

dbungert:
Per LP: #1968160, with 2 or more disks, go to guided storage config, hit
done.  At file system summary, hit back, and choose the other disk.
While this screen does say so, one might not notice that the first disk
is still setup to be formatted.

Instead, when going back to guided storage, a reset is also done.

### DIFF
--- a/subiquity/client/controllers/filesystem.py
+++ b/subiquity/client/controllers/filesystem.py
@@ -242,16 +242,17 @@ class FilesystemController(SubiquityTuiController, FilesystemManipulator):
     def guided(self):
         self.app.aio_loop.create_task(self._guided())
 
-    def reset(self):
+    def reset(self, refresh_view):
         log.info("Resetting Filesystem model")
         self.app.ui.block_input = True
-        self.app.aio_loop.create_task(self._reset())
+        self.app.aio_loop.create_task(self._reset(refresh_view))
 
-    async def _reset(self):
+    async def _reset(self, refresh_view):
         status = await self.endpoint.reset.POST()
         self.app.ui.block_input = False
         self.model.load_server_data(status)
-        self.ui.set_body(FilesystemView(self.model, self))
+        if refresh_view:
+            self.ui.set_body(FilesystemView(self.model, self))
 
     def cancel(self):
         self.app.prev_screen()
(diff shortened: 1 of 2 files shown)


mwhudson:
I wonder if we should do the reset before applying the new guided config, i.e. if you go back and then select custom should you still have whatever you did before? wdyt?

ogayot:
What would you think about doing the following? I think it might be better for the user's experience:

* adding a reset button in the Guided storage configuration screen - and keep it disabled/gray-ed out at the beginning
* after the user moves on to the next screen and chooses to hit "Back":
  * enabling the reset button and disabling all the other options
  * if the user wants to change his decision, he must first hit the "reset" button. Otherwise, he can move forward again and keep the settings he made.

Probably too much work to integrate it this PR. But do you think it would be worth it as a follow-up PR?

before-decision
after-decision

ogayot (commented):
It would be good to give a little heads up to the user so that he know that modifications are lost upon hitting back. 

Maybe changing the text on the Back button to "Cancel and Back" or something would be enough? What do you think?

dbungert:
@mwhudson - on second thought, isn't that already what the confirmation screen is after choosing guided but before saying 'Continue' to the 'Confirm destructive action' dialog?   Is there an interesting difference between customizing the guided storage changes on the confirm screen versus using the back button to select 'Custom storage layout'? 

I'm starting to think the answer to this discussion is 'leave it alone and ask UX'

mwhudson:
> The above means that the annotations list on disk 1 is lies - we'll show annotations that are implied from the original guided storage choice, then promptly ignore said actions and instead show in the confirmation screen annotations as if the user choose disk 2 in the first place.

Well we could base the annotations on the original config (which we keep around). It would be a bit messy to implement I guess.

> I'm starting to think the answer to this discussion is 'leave it alone and ask UX'

Back when Claudio was the UX guy we dealt with, we discussed some issues around this and how it was less than perfect. We never really came up with a design we were both happy with though.

I think the problem @xypron ran into is real and this branch is better than the status quo.

dbungert:
@elioqoshi - I'd like your opinion on the above

mwhudson (approved):
Let's do it

PR #1265, opened 2022-04-11
boot: further increase suggested sizing: min 1.7G

dbungert:


mwhudson (approved):
This seems very large for server somehow but it's important to align with ubiquity so let's do it.

PR #1266, opened 2022-04-12
Fix storage screen sometimes not refreshing after slow probing

ogayot:
_Not the greatest nor the worst way to fix https://bugs.launchpad.net/subiquity/+bug/1968161_

We now implement an _optional_ indirection mechanism for `make_ui()` functions and use it in the filesystem controller to work around the screen not refreshing issue.

Hopefully, the commit messages should explain in details what's going on and what the patches bring to the table.

### DIFF
--- a/subiquitycore/tui.py
+++ b/subiquitycore/tui.py
@@ -234,8 +234,6 @@ class TuiApplication(Application):
             except Exception:
                 self.controllers.index = cur_index
                 raise
-            else:
-                return
 
     async def move_screen(self, increment, coro):
         view_or_callable = await self.wait_with_progress(
--- a/subiquitycore/tui.py
+++ b/subiquitycore/tui.py
@@ -18,6 +18,7 @@ import inspect
 import logging
 import os
 import signal
+from typing import Callable, Optional, Union
 
 import urwid
 
@@ -34,6 +35,7 @@ from subiquitycore.tuicontroller import Skip
 from subiquitycore.ui.utils import LoadingDialog
 from subiquitycore.ui.frame import SubiquityCoreUI
 from subiquitycore.utils import astart_command
+from subiquitycore.view import BaseView
 
 log = logging.getLogger('subiquitycore.tui')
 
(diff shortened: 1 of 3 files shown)


ogayot:
Rebased on top of main to resolve conflicts.

subiquitycore/tui.py
```diff
@@ -205,7 +208,8 @@ def hide_load():
     async def wait_with_progress(self, awaitable):
         return await self._wait_with_indication(awaitable, self.show_progress)
 
-    async def _move_screen(self, increment, coro):
+    async def _move_screen(self, increment, coro) \
+            -> Optional[Union[BaseView, Callable[[], BaseView]]]:
```
dbungert, line 212: This function finishes with a bare `return`, which of course means `return None` but is upsetting to my C-laden eyeballs.   Would you make that explicit?

subiquitycore/tui.py
```diff
@@ -234,9 +238,13 @@ async def _move_screen(self, increment, coro):
                 return
 
     async def move_screen(self, increment, coro):
-        view = await self.wait_with_progress(
+        view_or_callable = await self.wait_with_progress(
             self._move_screen(increment, coro))
-        if view is not None:
+        if view_or_callable is not None:
+            try:
```
dbungert, line 244: In client.py we check with `callable`.  I'm sure either is fine but I would like consistency.   If there is no other criteria, we could choose the faster one.

dbungert (commented):
Thanks for this, it's nice to see progress on some of the screen change glitches.

subiquitycore/tui.py
ogayot: Oh yes, good catch!

Actually this `else` block is unreachable code now ; so I will just get rid of it.
One can't reach the `else` block if there is an unconditional `return` statement in the `try` block.

This `else` block became dead code with this patch: a7bcc7faf081054dc2ec20871c44e952335c50b0

subiquitycore/tui.py
ogayot: No problem. I will go with `callable()` in both places then.
Using the `try ... except` in the other function would result in the callable actually being called needlessly just for the sake of testing if it is callable. I'd rather avoid that.

subiquitycore/tui.py
dbungert: even better!

dbungert (approved):

PR #1267, opened 2022-04-12
Enable resize

dbungert:
Allow resize by way of storage/v2/edit_partition

### DIFF
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -43,7 +43,7 @@ class FilesystemManipulator:
         self.model.remove_mount(mount)
 
     def create_filesystem(self, volume, spec):
-        if spec['fstype'] is None:
+        if spec.get('fstype') is None:
             # prep partitions are always wiped (and never have a filesystem)
             if getattr(volume, 'flag', None) != 'prep':
                 volume.wipe = None
@@ -61,9 +61,9 @@ class FilesystemManipulator:
                 volume.flag = "swap"
             elif volume.flag == "swap":
                 volume.flag = ""
-        if spec['fstype'] == "swap":
+        if spec.get('fstype') == "swap":
             self.model.add_mount(fs, "")
-        if spec['fstype'] is None and spec['use_swap']:
+        if spec.get('fstype') is None and spec.get('use_swap'):
             self.model.add_mount(fs, "")
         self.create_mount(fs, spec)
         return fs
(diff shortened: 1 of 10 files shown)


dbungert:
@jpnurmi - API change but existing usage should not notice any difference - when merged, resize is done by using edit_partition and specifying a different size on the partition object.  Requires storage version 2.  We will want to be able to show the acceptable lower bound of the FS size, which is not yet exposed on the API, so for now one merely needs to choose a known correct value.

dbungert:
Moving back to draft, will mark ready when I'm happy with the state of curtin resize code.

subiquity/common/filesystem/manipulator.py
```diff
@@ -165,19 +167,29 @@ def reformat(self, disk, ptable=None):
...
                 trailing, gap_size = \
                      gaps.movable_trailing_partitions_and_gap_size(partition)
                 new_size = align_up(spec['size'])
                 size_change = new_size - partition.size
                 if size_change > gap_size:
                     raise Exception("partition size too large")
+                if not self.can_resize_partition(partition):
+                    raise Exception("partition cannot support resize")
                 partition.size = new_size
+                partition.resize = True
```
mwhudson, line 192: I guess it's not a problem exactly but I feel like it might be less confusing to only set this on a preserve=True partition.

subiquity/tests/api/test_api.py
```diff
@@ -847,6 +850,68 @@ async def test_use_free_space_after_existing(self):
             }
             self.assertDictSubset(e4, p4)
 
+    @timeout()
+    async def test_resize(self):
+        # load config, edit size, use that for server
+        with open('examples/ubuntu-and-free-space.json', 'r') as fp:
+            data = json.load(fp)
+        # expand to use the rest of the disk
+        data['storage']['blockdev']['/dev/sda3']['attrs']['size'] = 40899706880
```
mwhudson, line 859: This number is a mystery guest. Can you read the disk size out of the blockdev data for /dev/sda or something?

mwhudson (approved):
Just a couple of comments but mostly very nice!

subiquity/common/filesystem/manipulator.py
dbungert: Good point, that matches the curtin assumptions but isn't a showstopper as I have set up curtin to be flexible about non-resizes masquerading as resizes (such as a partition reformat).  I will think about it for future branches.

subiquity/tests/api/test_api.py
dbungert: Fixed in the rebase.

PR #1269, opened 2022-04-13
storage/v2: permit delete with storage-version=2

dbungert:
On preserved disks, we can now allow partition deletes without requiring
a wholesale format of the disk, if using storage-verison=2.


ogayot (approved):
LGTM

PR #1272, opened 2022-04-14
drivers: prevent client crash if GET /drivers is closed from client side

ogayot:
When a HTTP client sends a query but closes the socket before an answer is received,  aiohttp signals it on the server end by raising an `asyncio.CancelledError` in the associated query handler.

By default, when a task is cancelled with asyncio, the task(s) that it is currently awaiting on are cancelled as well.

The GET handler for `/drivers?wait=true` awaits on the "list drivers" task. Therefore, if the GET handler gets cancelled, so will be the "list drivers" task.

When that happens, any subsequent call to `GET /drivers?wait=true` will make the server raise an `asyncio.CancelledError` because the "list drivers" task has already been cancelled. This results in:

 * the socket being closed from the server end
 * an `aiohttp.client_exceptions.ServerDisconnectedError` exception raised on the client end. This type of exception is unhandled and makes the client crash.

Fixed by preventing the "list drivers" task from being cancelled when the `GET /drivers` query handler gets cancelled.

https://bugs.launchpad.net/subiquity/+bug/1968729

### DIFF
--- a/subiquity/server/controllers/drivers.py
+++ b/subiquity/server/controllers/drivers.py
@@ -91,7 +91,7 @@ class DriversController(SubiquityController):
 
     async def GET(self, wait: bool = False) -> DriversResponse:
         if wait:
-            await self._drivers_task
+            await asyncio.shield(self._drivers_task)
         return DriversResponse(install=self.model.do_install,
                                drivers=self.drivers)
 
--- a/subiquity/tests/api/test_api.py
+++ b/subiquity/tests/api/test_api.py
@@ -920,18 +920,18 @@ class TestCancel(TestAPI):
     async def test_cancel_drivers(self):
         with patch.dict(os.environ, {'SUBIQUITY_DEBUG': 'has-drivers'}):
             async with start_server('examples/simple.json') as inst:
+                # /drivers?wait=true is expected to block until APT is
+                # configured.
+                # Let's make sure we cancel it.
+                with self.assertRaises(asyncio.TimeoutError):
+                    await asyncio.wait_for(inst.get('/drivers', wait=True),
+                                           0.1)
                 names = ['locale', 'keyboard', 'source', 'network', 'proxy',
                          'mirror', 'storage']
                 await inst.post('/meta/mark_configured', endpoint_names=names)
                 await inst.get('/meta/status', cur='WAITING')
                 await inst.post('/meta/confirm', tty='/dev/tty1')
                 await inst.get('/meta/status', cur='NEEDS_CONFIRMATION')
-
-                try:
-                    await asyncio.wait_for(inst.get('/drivers', wait=True),
-                                           0.001)
-                except asyncio.exceptions.TimeoutError:
-                    pass
 
                 # should not raise ServerDisconnectedError
                 resp = await inst.get('/drivers', wait=True)
--- a/subiquity/tests/api/test_api.py
+++ b/subiquity/tests/api/test_api.py
@@ -8,6 +8,7 @@ import json
 import os
 import tempfile
 import unittest
+from unittest.mock import patch
 from urllib.parse import unquote
 
 from subiquitycore.utils import astart_command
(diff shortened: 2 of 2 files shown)


dbungert (commented):
I took the liberty of writing an API test to help understand just what was going on here.  The test as-is depends on a timeout that is probably not robust (is the dryrun path for drivers effectively instant?), so I would appreciate some thought on how to improve it (sleep() calls on the drivers dryrun path?).  I'm good to merge this once we decide on a test, but we should also open a card to analyze this category of problem - API endpoint waits on task, which might be cancelled and called again.

ogayot:
> I took the liberty of writing an API test to help understand just what was going on here.

Thanks @dbungert !
This is much better with a test indeed.

> The test as-is depends on a timeout that is probably not robust (is the dryrun path for drivers effectively instant?), so I would appreciate some thought on how to improve it (sleep() calls on the drivers dryrun path?). I'm good to merge this once we decide on a test,

In dry-run mode, yes, the `_list_drivers` task is instant (unless `SUBIQUITY_DEBUG=run-drivers` but this is not reliable for automated tests).
However, the `_list_drivers` task only runs after APT has finished configuring.
Before that, we can reliably expect the `GET /drivers?wait=true` to block.

Therefore, I guess we can do:

```python
# Cancel a call to /drivers?wait=true.
with self.assertRaises(asyncio.TimeoutError):
    await asyncio.wait_for(inst.get('/drivers', wait=True), 0.1)
# Mark controllers configured so drivers listing can operate
names = ['locale', 'keyboard', 'source', 'network', 'proxy', 'mirror', 'storage']
await inst.post('/meta/mark_configured', endpoint_names=names)
await inst.get('/meta/status', cur='WAITING')
await inst.post('/meta/confirm', tty='/dev/tty1')
await inst.get('/meta/status', cur='NEEDS_CONFIRMATION')
#  should not raise ServerDisconnectedError
resp = await inst.get('/drivers', wait=True)
self.assertEqual(['nvidia-driver-470-server'], resp['drivers']) 
```
> but we should also open a card to analyze this category of problem - API endpoint waits on task, which might be cancelled and called again.

I'm not sure exactly what you mean here. Is it
1. endpoint waits on task - task gets cancelled - task gets rescheduled? or
2. endpoint waits on task - endpoint gets cancelled and cancels the task by mistake - endpoint gets called again?

(this PR addresses an instance of the latter (2.))

dbungert:
> I'm not sure exactly what you mean here. Is it
> 
> 1. endpoint waits on task - task gets cancelled - task gets rescheduled? or
> 2. endpoint waits on task - endpoint gets cancelled and cancels the task by mistake - endpoint gets called again?

Right, item 2 was what I was referencing, but checking correctness of item 1 at the same time sounds prudent.

ogayot:
> Right, item 2 was what I was referencing, but checking correctness of item 1 at the same time sounds prudent.

I originally opened [LP 1968729](https://bugs.launchpad.net/subiquity/+bug/1968729) to track this category of problem rather than the occurrence on third-party drivers. But I can change that if needed.

dbungert:
OK cool, that LP works for me then to track the general problem so long as we keep it open after this PR.

ogayot:
> > The test as-is depends on a timeout that is probably not robust (is the dryrun path for drivers effectively instant?), so I would appreciate some thought on how to improve it (sleep() calls on the drivers dryrun path?). I'm good to merge this once we decide on a test,
> 
> In dry-run mode, yes, the `_list_drivers` task is instant (unless `SUBIQUITY_DEBUG=run-drivers` but this is not reliable for automated tests). However, the `_list_drivers` task only runs after APT has finished configuring. Before that, we can reliably expect the `GET /drivers?wait=true` to block.
> 
> Therefore, I guess we can do:
> 
> ```python
> # Cancel a call to /drivers?wait=true.
> with self.assertRaises(asyncio.TimeoutError):
>     await asyncio.wait_for(inst.get('/drivers', wait=True), 0.1)
> # Mark controllers configured so drivers listing can operate
> names = ['locale', 'keyboard', 'source', 'network', 'proxy', 'mirror', 'storage']
> await inst.post('/meta/mark_configured', endpoint_names=names)
> await inst.get('/meta/status', cur='WAITING')
> await inst.post('/meta/confirm', tty='/dev/tty1')
> await inst.get('/meta/status', cur='NEEDS_CONFIRMATION')
> #  should not raise ServerDisconnectedError
> resp = await inst.get('/drivers', wait=True)
> self.assertEqual(['nvidia-driver-470-server'], resp['drivers']) 
> ```

@dbungert would you be okay with this approach?

dbungert:
> @dbungert would you be okay with this approach?

Yes, that works for me.

dbungert (approved):

PR #1274, opened 2022-04-15
Fix guided direct

dbungert:
A scenario that should happen nearly always for guided direct, when we
add_boot_disk from partition_disk_handler we must update the gap or the
target partition is at the wrong offset.

(This bug was introduced with the offsets&gap work and is not preset on Jammy)

### DIFF
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -196,12 +196,15 @@ class FilesystemManipulator:
             self.add_boot_disk(disk)
 
             # adjust downward the partition size (if necessary) to accommodate
-            # bios/grub partition
-            if spec['size'] > gaps.largest_gap_size(disk):
+            # bios/grub partition.  It's OK and useful to assign a new gap:
+            # 1) with len(partitions()) == 0 there could only have been 1 gap
+            # 2) having just done add_boot_disk(), the gap is no longer valid.
+            gap = gaps.largest_gap(disk)
+            if spec['size'] > gap.size:
                 log.debug(
                     "Adjusting request down from %s to %s",
-                    spec['size'], gaps.largest_gap_size(disk))
-                spec['size'] = gaps.largest_gap_size(disk)
+                    spec['size'], gap.size)
+                spec['size'] = gap.size
 
         self.create_partition(disk, gap, spec)
 
--- a/subiquity/server/controllers/tests/test_filesystem.py
+++ b/subiquity/server/controllers/tests/test_filesystem.py
@@ -19,6 +19,11 @@ from subiquity.server.controllers.filesystem import FilesystemController
 
 from subiquitycore.tests.util import run_coro
 from subiquitycore.tests.mocks import make_app
+from subiquity.common.types import Bootloader
+from subiquity.models.tests.test_filesystem import (
+    make_disk,
+    make_model,
+    )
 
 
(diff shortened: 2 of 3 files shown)


subiquity/common/filesystem/manipulator.py
```diff
@@ -197,11 +197,12 @@ def partition_disk_handler(self, disk, spec, *, partition=None, gap=None):
 
             # adjust downward the partition size (if necessary) to accommodate
             # bios/grub partition
-            if spec['size'] > gaps.largest_gap_size(disk):
+            gap = gaps.largest_gap(disk)
```
mwhudson, line 200: I don't know if this is overthinking, but is it worth adding a comment that in this case it's safe to ignore the passed-in gap because there can only be one?

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -44,3 +49,54 @@ def test_probe_defaults_and_os(self):
         self.app.opts.use_os_prober = True
         run_coro(self.fsc._probe_once(context=None, restricted=False))
         self.app.prober.get_storage.assert_called_with({'defaults', 'os'})
+
+
+class TestGuided(TestCase):
+    def _guided_direct(self, bootloader, ptable):
+        self.app = make_app()
```
mwhudson, line 56: Wait, doesn't guided always get you a gpt partition table?

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -44,3 +49,54 @@ def test_probe_defaults_and_os(self):
...
+        self.controller.model._probe_data = {'blockdev': {}}
+        self.d1 = make_disk(self.controller.model, ptable=ptable)
+        self.controller.guided_direct(self.d1)
+
+    def test_guided_direct_UEFI_GPT(self):
+        self._guided_direct(Bootloader.UEFI, 'gpt')
+        [d1p1, d1p2] = self.d1.partitions()
+        self.assertEqual('/boot/efi', d1p1.mount)
+        self.assertEqual('/', d1p2.mount)
+        self.assertEqual(d1p1.size + d1p1.offset, d1p2.offset)
```
mwhudson, line 69: I've wondered for a while if a helper that checks a disk's partitions for sanity would be good, like here we could use a helper that asserts a disk is full of partitions with no overlaps and no gaps. But it doesn't have to be part of this branch :)

mwhudson (approved):
Looks good, a couple of questions.

subiquity/common/filesystem/manipulator.py
dbungert: I like this, I had to stare at the code for a bit to convince myself that assigning a new gap was ok.  Will do.

subiquity/server/controllers/tests/test_filesystem.py
dbungert: Yes.  I was working on things related to msdos tables after this and squished that commit into here.  So for now this code is unit-test only.

subiquity/server/controllers/tests/test_filesystem.py
dbungert: That does sound helpful.

mwhudson (approved):

dbungert:
While I'm not going to attempt it now, it occurs to me that it might be interesting for partition operations that take a gap to return a gap object that represents the space remaining.

PR #1275, opened 2022-04-15
v2/reformat: accept ptable argument

dbungert:
Move storage/v2/reformat_disk to body instead of query args.
Add optional ptable arg, which can be used to format to a msdos table.

### DIFF
--- a/subiquity/common/filesystem/tests/test_manipulator.py
+++ b/subiquity/common/filesystem/tests/test_manipulator.py
@@ -503,3 +503,28 @@ class TestFilesystemManipulator(unittest.TestCase):
         disk = make_disk(manipulator.model, preserve=True, size=2002*MiB)
         self._test_add_boot_full_resizes_larger(
             manipulator, disk, sizes.PREP_GRUB_SIZE_BYTES)
+
+
+class TestReformat(unittest.TestCase):
+    def setUp(self):
+        self.manipulator = make_manipulator()
+
+    def test_reformat_default(self):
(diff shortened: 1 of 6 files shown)


subiquity/common/filesystem/tests/test_manipulator.py
```diff
@@ -503,3 +503,28 @@ def test_add_boot_PREP_full_resizes_larger(self):
...
+
+    def test_reformat_default(self):
+        disk = make_disk(self.manipulator.model, ptable=None)
+        self.manipulator.reformat(disk)
+        self.assertEqual(None, disk.ptable)
+
+    def test_reformat_keep_current(self):
+        disk = make_disk(self.manipulator.model, ptable='msdos')
+        self.manipulator.reformat(disk)
+        self.assertEqual('msdos', disk.ptable)
```
mwhudson, line 520: I realise this looks like the current behaviour but is this actually what we want? In my mind we always reformat to GPT but maybe that's not what happens...

mwhudson (approved):

subiquity/common/filesystem/tests/test_manipulator.py
mwhudson: Yeah that makes sense.

dbungert:
@jpnurmi - change to reformat_disk, allows creation of msdos partition tables, arguments moved to body.  Recommend continuing to use gpt partition tables for now as we're working through bugs in support for msdos style tables.

PR #1277, opened 2022-04-20
Sets with the prefilled language before starting controllers

CarlosNihelton:
TUI client was able to GET() the language before it was set with the prefilled option in production. This attempts to ensure the earliest initialization possible.


didrocks:
ack, that makes sense, nicely spotted!

didrocks (approved):

PR #1279, opened 2022-04-20
Revert "Move to aiohttp for GeoIP requests"

dbungert:
This reverts commit ebdf1db63619424d06f7597f0c358e72cae4eedc.

Jammy merge of #1278
LP: #1969393


mwhudson (approved):
At this point in the cycle this seems like the right thing to do

dbungert:
Merging after discussion with @vorlonofportland

PR #1281, opened 2022-04-21
Fix WSL crash on autoinstall due attempts to log to systemd journal

CarlosNihelton:
EarlyController was still logging to systemd journal for WSL.


didrocks (approved):
As tested together, it works and tests still pass.

PR #1282, opened 2022-04-22
tests: consistently run Subiquity commands

ogayot:
The use of `sh -c "$command"` to run the subiquity commands was not necessary and it makes it easy to forget to quote arguments here and there.
Instead, we can execute commands directly with, which saves us the shell invocation + potential trouble where quoting properly is not straightforward.

Also, I took the opportunity to reorder the arguments passed to Subiquity in a consistent order across the different invocations that we have:
1. Arguments used by the client & the server (e.g., `--dry-run`, `--output-base`) 
2. Arguments used by the client only
3. Arguments used by the server only

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -155,13 +155,13 @@ for answers in examples/answers*.yaml; do
         # The --foreground is important to avoid subiquity getting SIGTTOU-ed.
         LANG=C.UTF-8 timeout --foreground 60 \
             python3 -m subiquity.cmd.tui < "$tty" \
-            --bootloader uefi \
-            --answers "$answers" \
             --dry-run \
-            --snaps-from-examples \
-            --machine-config "$config" \
             --output-base "$tmpdir" \
-            "${opts[@]}"
+            --answers "$answers" \
+            "${opts[@]}" \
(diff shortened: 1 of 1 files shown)


scripts/runtests.sh
```diff
@@ -166,18 +174,26 @@ for answers in examples/answers*.yaml; do
                 reconf_settings="true"
                 validate_subtype="answers-reconf"
             fi
-            timeout --foreground 60 sh -c "DRYRUN_RECONFIG=$reconf_settings LANG=C.UTF-8 python3 -m system_setup.cmd.tui --answers $answers --output-base $tmpdir --dry-run " < $tty
+            DRYRUN_RECONFIG="$reconf_settings" LANG=C.UTF-8 timeout --foreground 60 \
+                python3 -m system_setup.cmd.tui < "$tty" \
```
dbungert, line 178: Just when you think you understand shell syntax.

dbungert (approved):
Thanks for work on this.  That file has gotten a little ugly.

mwhudson:
Thanks for this (how much more until we rewrite this file in Python?). BTW we might be able to drop the --foreground in the --timeout calls now, not sure.

PR #1283, opened 2022-04-22
gitignore: .tags, .mypy

dbungert:


.gitignore
```diff
@@ -5,6 +5,9 @@ __pycache__/
 # C extensions
 *.so
 
+# ctags
+.tags
```
ogayot, line 9: I'm using ctags from `universal-ctags` and the default filename is `tags` (without the leading `.`)
Add it as well maybe?

```suggestion
.tags
tags
```

ogayot (approved):

PR #1292, opened 2022-05-04
Add visual indication for starred snap developers

ogayot:
This PR now shows a visual indication for starred developers / publishers.
It also, changes the alternative character (i.e., used in basic mode) for verified publishers from a `+` sign to `**`.
As for the green color used for verified developers in basic mode. we dropped it and used default colors instead. There are no other place where we display colors in basic mode.

~~I'm relying on a hack (based non-breakable spaces) to display two stars instead of a single check-mark in basic mode. I hate it but don't have any better idea ATM. Substituting a single char with two chars does not seem to be an option. urwid computes the maximum number of columns needed based on the original number of characters.
Any better idea would be most welcome.~~

I'm adding some screen shots. Currently, the ✪ sign is not easily noticeable in the list of snaps (probably because of the white font). Adding a space before it would probably look visually better, but it would also add a space before the star in basic mode.

~~Marking this as a draft until we get confirmation that:~~
~~* 2 stars = ✓ = verified ~~
~~* 1 star = ✪ = starred~~
.. and not the other way around.~~
EDIT: confirmed!

basic-list
basic-microk8s
basic-ralight
rich-list
rich-microk8s
rich-ralight


ogayot:
@elioqoshi would you mind having a look and giving us some feedback? Thanks!

mwhudson:
Oh I had forgotten just how fun a hack subiquity's ascii mode was!

You'll need to add \N{circled white star} to the font we use on the linux console. I hope font/README.md makes sense!

mwhudson:
BTW, the hack seems ok ish to me. I certainly don't have any better ideas.

ogayot:
Changed the hack to use two checkmarks (each replaced by a single star in basic mode).
In rich mode, we mask one of the check-marks by using the same font and background (we lean on urwid display attributes to do that). Thanks @mwhudson for the suggestion!

elioqoshi:
@ogayot I am not the best person for this, though @AnnSoMuller worked on this IIRC.

ogayot:
I included a glyph representing the ✪ character in our subiquity font.
Unfortunately, I didn't find a proper 8-points-wide glyph so I had to chop off a 9-wide one. This does not look great...

mwhudson:
I think the code looks fine, although there should probably be a comment /somewhere/ (not sure where) about the visibility hacks?

Maybe we can get design to think about how a starred developer should be presented in a 8x16 (or whatever it is) bitmap -- it guess it doesn't have to be a bitmap version of ✪.

Oh and you'll need to check the generated psf file in -- we can't build the font at snap build time (afaik! would be nice to be wrong about this)

AnnSoMuller:
Hi there :)
I had the confirmation that it is:
2 stars = ✓ = verified
1 star = ✪ = starred
as written in the [spec doc](https://docs.google.com/document/d/1Q7ChjpcHyVifo30-QwUC3eAPJFwqRP6r-otJqrrh6xA/edit#heading=h.epvc134qwols)
That was "changed" as verified is "stronger" than starred

Then we agreed also to use orange as the colour for ✪ instead of white when we use green for ✓
with-space
Thanks!

ogayot:
Updated screenshots after design reviewing with @AnnSoMuller.
✪ signs are now displayed with orange font, making them easier to notice.

I updated the .psf font file as well. I don't have clear understanding of do and don'ts when building snaps to know what prevents us from building the font automatically.
If it's because we need to enable deb-src, maybe we can work around this issue by using pull-lp-source instead of apt source? Otherwise, I think a git submodule pointing to a given revision of https://git.launchpad.net/ubuntu/+source/console-setup would do fine. @mwhudson any thoughts ?

mwhudson (approved):
> I think the code looks fine, although there should probably be a comment /somewhere/ (not sure where) about the visibility hacks?

This is still true :)

I think the character you made is fine, we can update it when design get back to us.

If you want to change things that sounds great but not in this branch please. I would sort of prefer a part that had the console-source import branch as a source over a submodule but I'm not that fussed really.

ogayot:
> > I think the code looks fine, although there should probably be a comment /somewhere/ (not sure where) about the visibility hacks?
> 
> This is still true :) 
> 
Oops! I forgot to push the changes I did in that regard. Should be sorted now, thanks!
> If you want to change things that sounds great but not in this branch please. I would sort of prefer a part that had the console-source import branch as a source over a submodule but I'm not that fussed really.

Yep, not planning on doing such a change now in this branch. I will give it some more thoughts and hopefully will change it before we need to introduce another glyph :)

mwhudson (approved):

PR #1294, opened 2022-05-06
server/jammy: merge two fixes

dbungert:
Updated curtin for proxy fix
Update probert for raid fix


ogayot (approved):
Thanks!

dbungert:
Verified the raid fix against a VM configured for RAID 1 that was known to produce the bug.

PR #1296, opened 2022-05-10
Server-side username validation

CarlosNihelton:
Hello!

This PR is backed by [the spec](https://docs.google.com/document/d/1Luulbn0DR2xxxmVO6KlxGDIUieqCg2G5f7fbrmlQfes/preview) I curculated a few days ago on the New Desktop Installer and motivated by a bug report of a crash in WSL installer due the GUI not being able to prevent the user to submit system reserved usernames. Instead of duplicating the logic that exists in TUI, with this PR I aim to move it to the server, so both clients can benefit from it. Also, being in the server, allows easier and more consistent extensibility. For instance: today the list of reserved names used for validation is quite old and doesn't contain popular names such as lxd or systemd-oom. If at some point we decide that those names should be added, there will be only once central place to change to make that hit all platforms where we install Ubuntu (thinking of the day when Subiquity will be omnipresent).


didrocks (approved):
Those changes looks good and backed up with multiple tests. Thanks for doing this! Also the logic system setup side is more straightforward.

I’ll let subiquity upstream reviewing it too.

subiquity/client/controllers/identity.py
```diff
@@ -48,3 +49,10 @@ def done(self, identity_data):
             "IdentityController.done next_screen user_spec=%s",
             identity_data)
         self.app.next_screen(self.endpoint.POST(identity_data))
+
+    async def _validate_username(self, username):
```
ogayot, line 53: I would prefer not to store the result of the request in this object.
When the user is typing the username, there will likely be multiple `_validate_username` coroutines running concurrently. So it would not be clear what actual username `self.username_validation` corresponds to at at given time.

ogayot (commented):
Thanks for doing this. I have a few comments but nothing major.

subiquity/server/controllers/identity.py
```diff
@@ -77,4 +112,23 @@ async def GET(self) -> IdentityData:
...
         await self.configured()
+
+    async def validate_username_GET(self, username: str) -> UsernameValidation:
+        if username in self._existing_usernames:
+            return UsernameValidation.ALREADY_IN_USE
+
+        if username in self._system_reserved_names:
+            return UsernameValidation.SYSTEM_RESERVED
+
+        if not re.match(USERNAME_REGEX, username):
```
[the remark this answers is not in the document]
CarlosNihelton: Indeed. Added the proposed test case and it proved the need for the `fullmatch`.

subiquity/server/controllers/identity.py
```diff
@@ -16,20 +16,43 @@
...
 
 log = logging.getLogger('subiquity.server.controllers.identity')
 
+USERNAME_MAXLEN = 32
+USERNAME_REGEX = r'[a-z_][a-z0-9_-]*'
+
+
+def _reserved_names_from_file(path: str) -> Set[str]:
+    names = set()
+    if os.path.exists(path):
```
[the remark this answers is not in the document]
CarlosNihelton: Used a try-block as suggested.

subiquity/server/controllers/identity.py
```diff
@@ -77,4 +112,23 @@ async def GET(self) -> IdentityData:
 
     async def POST(self, data: IdentityData):
         self.model.add_user(data)
+        if await self.validate_username_GET(data.username) != \
+                UsernameValidation.OK:
+            log.error("Username <%s> is invalid and should not be submitted.",
+                      data.username)
         await self.configured()
```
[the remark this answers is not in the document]
CarlosNihelton: Indeed. Changed to raise an exception. As you correctly deduced, having POST'ed an invalid username configures UI bug.

subiquity/server/controllers/identity.py
```diff
@@ -16,20 +16,43 @@
...
+
+    return names
+
 
 class IdentityController(SubiquityController):
 
     endpoint = API.identity
 
+    _system_reserved_names = set()
+    _existing_usernames = set()
```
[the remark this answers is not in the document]
CarlosNihelton: Good catch. Removed.

subiquity/client/controllers/identity.py
CarlosNihelton: Here I struggled the most. The way I found was to store the latest async task in the Form, as part of its state. Since for sure a couple of coroutines will be spawn due user editting the text field, at each new change the previous task is cancelled. When the `username_validation_state` method is called, we return the result of the last task launched, or `UsernameValidation.OK` by default. Handling this task object so explicitly uglyfied the API test code a little bit, IHMO, but I think we made the point that the validation result is deterministically retrieved.

subiquity/server/controllers/identity.py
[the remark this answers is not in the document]
ogayot: Thanks! I see we still do the `self.model.add_user`, don't we want to make it conditional as well?

subiquity/client/controllers/identity.py
ogayot: Thanks! This looks much better IMO :)

subiquity/server/controllers/identity.py
```diff
@@ -43,6 +66,18 @@ class IdentityController(SubiquityController):
...
+    # TODO: Find THE way to initialize application_reserved_names.
+    def __init__(self, app):
+        super().__init__(app)
+        core_reserved_path = resource_path("reserved-usernames")
+        self._system_reserved_names = \
+            _reserved_names_from_file(core_reserved_path)
+        self._system_reserved_names.add('root')
+
+        self._existing_usernames = {
+            u.pw_name for u in pwd.getpwall()
```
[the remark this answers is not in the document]
ogayot: I'd like a second opinion on this. Maybe @mwhudson or @dbungert ?
Although it seems better than using etc/passwd from the installer,  I'm afraid the target won't necessarily be populated with etc/passwd by the time we hit this screen. Also, we'd need to confirm if the flutter installer (and flavors) would install a different etc/passwd maybe?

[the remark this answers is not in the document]
CarlosNihelton: That should not be a problem if the `reserved-usernames` protects against any interesting username any of the flavors would ship. Yet, I understand it's quite hard to ensure that in a future-proof way. We may consider creating a "protocol" to allow flavors to extend the `reserved-usernames` by shipping a version of it somewhere inside the installer filesystem Subiquity would load names from and even report a specific validation code for that. If that direction is desired, I'd prefer doing it in a next PR.

subiquity/server/controllers/identity.py
```diff
@@ -16,15 +16,68 @@
...
+    names = set()
+    try:
+        with open(path, "r") as f:
+            for line in f:
+                line = line.strip()
+                if not line or line.startswith('#'):
+                    continue
+                names.add(line.split()[0])
+
+    except FileNotFoundError:
```
dbungert, line 46: I'm concerned about only logging these failures.  Failure to process the reserved-names file to me represents an installer bug and I think we should allow those Exceptions to bubble up to the surface.

subiquity/server/controllers/identity.py
```diff
@@ -16,15 +16,68 @@
...
+        log.error("Could not find %s", path)
+
+    except OSError:
```
dbungert, line 49: Same

subiquity/server/controllers/identity.py
```diff
@@ -16,15 +16,68 @@
...
+        log.error("Failed to process %s", path)
+
+    except Exception:
```
dbungert, line 52: Catching raw `Exception` in general is probably not what we want, at least without re-raise.  In this specific case you're welcome to do so for logging purposes, but please re-raise it.  We use `Exception` in many places in the Subiquity codebase as a fatal error that triggers a backtrace.

dbungert (changes requested):

subiquity/server/controllers/identity.py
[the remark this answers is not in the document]
dbungert: Reading the runtime /etc/passwd file is going to be a race with the earlier installation steps.  Taking one's time in the installer with the target on NVME and this file might actually be ready, but in other cases that file probably won't be there.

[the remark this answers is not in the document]
CarlosNihelton: Thank you for the considerations, guys. My current understanding is that there is not a reliable and portable way to find the list of usernames considered existing from the point of view of the server. So I'm defaulting it to just `{'root'}` and using it as a customisation point for variants. In WSL we can safely read the target's passwd database, so there I allow initializing the `existing_usernames` field from the target's passwd file.

subiquity/server/controllers/identity.py
```diff
@@ -43,6 +96,22 @@ class IdentityController(SubiquityController):
         'additionalProperties': False,
         }
 
+    # TODO: Find THE way to initialize application_reserved_names.
```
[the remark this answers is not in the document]
CarlosNihelton: My bad. That was a leftover that survived across the rebases. Removed.

subiquity/server/controllers/identity.py
```diff
@@ -16,15 +16,68 @@
...
+        log.error("Failed to process %s", path)
+
+    except Exception:
+        log.error("Unexpected error happened when attempted to read %s",
+                  path)
+
+    return names
+
+
+def _existing_user_names(path: str) -> Set[str]:
```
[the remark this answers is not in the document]
CarlosNihelton: You're absolutely right. Removed all of those with the assumption that if that file is missing we have a serious bug in the installer or the target. Either way, it's outside of the scope of the server to do anything else with that error. Let the exception propagate.

subiquity/client/controllers/identity.py
```diff
@@ -25,6 +25,7 @@
 class IdentityController(SubiquityTuiController):
 
     endpoint_name = 'identity'
+    username_validation = UsernameValidation.OK
```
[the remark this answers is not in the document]
CarlosNihelton: Not anymore. Removed.

subiquity/server/controllers/identity.py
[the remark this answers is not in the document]
dbungert: Thanks!

dbungert (approved):
LGTM.  Let's allow Olivier one more chance to review.

subiquity/ui/views/identity.py
```diff
@@ -95,6 +99,23 @@ def __init__(self, reserved_usernames, initial):
...
+        if self.validation_task is not None:
+            self.validation_task.cancel()
+
+        self.validation_task = \
+            schedule_task(self.controller.validate_username(value))
+
+    def username_validation_state(self):
+        task = self.validation_task
+        if task is None or not task.done():
+            return UsernameValidation.OK
```
mwhudson, line 115: Oh boy, asynchronous validation! I certainly wasn't thinking about that when I wrote the forms code.

While this is probably fine in practice, it is theoretically racy to assume that if validation hasn't completed it's OK because you can enter invalid data, leave the field before validation completes and then hit done, which will crash because the validation in POST fails.

Fixing this properly probably involves some delicate hacking of the generic form code -- if the user moves out of a form with validation incomplete, we could do something like freeze the ui for 0.1s and if the validation still has not completed show a spinner and update the validation status when the result comes in. Sounds fiddly though and definitely not something for Friday afternoon!

In this case I think it's probably OK though because the server side of the validation method is not actually asynchronous and won't force you to do this "properly" :)

ogayot: I know you're not a fan of using multiple HTTP requests but I think we can do something like:

when the OK/Done button is pressed, don't submit the form just yet.
* send an asynchronous validation request (through GET).
* display a spinner if needed
* on success, submit the token that was validated (through POST).

To reduce the number of requests, we can also make the POST call return a boolean and make sure we check it on the client-side.

IMO, I don't think we should rely too much on the client-side validation to decide if the application crashes or not :)

mwhudson: Well, we should send the validation request when the focus moves out of the field (or when a key is pressed, maybe) rather than when OK/Done is pressed but otherwise that makes sense as a plan I think.

mwhudson: It might be worth spending a few minutes thinking about how to handle validation being asynchronous in the generic code. I guess an "if iscoroutine()" somewhere around https://github.com/canonical/subiquity/blob/main/subiquitycore/ui/form.py#L241 would be the start... but I expect that might be a bit painful.

CarlosNihelton: > Well, we should send the validation request when the focus moves out of the field (or when a key is pressed, maybe) rather than when OK/Done is pressed but otherwise that makes sense as a plan I think.

That's exactly what I'm doing on the Flutter client. I could not find a suitable signal to connect in the TUI form editor, though.

mwhudson: Ah there's no signal but you can define a lost_focus method on a widget and it will get called when the widget loses focus. But the validate_foo method is called by such a hook (https://github.com/canonical/subiquity/blob/main/subiquitycore/ui/form.py#L97) -- so you can start the request there?

CarlosNihelton: Ok. I built on top of your suggestion of moving the async calls for when we leave focus. I did https://github.com/canonical/subiquity/pull/1296/commits/df8b157ae891fc441b33ed96a9d0a305eed6da52 as the good-enough approach, but decided to take out the async related logic into a specific mixin class, that can be combined with other editors in https://github.com/canonical/subiquity/pull/1296/commits/e6a6f7a92b2c02eae037e8e580a384940160ec68. Shall we decide to validate hostname in future, there it will be a building block. 

Either of the commits should be fine. I'll leave both for your appreciation. If you decide that the latest is overkill, I'll drop it. Otherwise, I'll be happy to hear from you on the current design.

subiquity/ui/views/tests/test_identity.py
```diff
@@ -41,6 +66,27 @@ def test_done_enabled_when_valid(self):
         view_helpers.enter_data(view.form, valid_data)
         self.assertTrue(view.form.done_btn.enabled)
 
+    async def test_username_validation_system_reserved(self):
```
mwhudson, line 69: This doesn't do anything unless you change the class to inherit from unittest.IsolatedAsyncioTestCase

subiquity/ui/views/tests/test_identity.py
```diff
@@ -41,6 +66,27 @@ def test_done_enabled_when_valid(self):
...
+        view = self.make_view()
+        view.controller.validate_username.return_value = \
+            UsernameValidation.SYSTEM_RESERVED
+        view_helpers.enter_data(view.form, system_reserved)
+        await asyncio.wait(view.form.validation_task)
```
mwhudson, line 74: whereupon we find that wait takes an iterable. so this would be more correct if it were await asyncio.wait({view.form.validation_task}) but it would be even better if it was just await view.form.validation_task

subiquity/ui/views/tests/test_identity.py
```diff
@@ -41,6 +66,27 @@ def test_done_enabled_when_valid(self):
...
+        self.assertFalse(view.form.done_btn.enabled)
+
+    async def test_username_validation_in_use(self):
+        view = self.make_view()
+        view.controller.validate_username.return_value = \
+            UsernameValidation.ALREADY_IN_USE
+        view_helpers.enter_data(view.form, already_taken)
+        await asyncio.wait(view.form.validation_task)
```
mwhudson, line 82: ditto

subiquity/ui/views/identity.py
mwhudson: I think the current approach is OK, and can be made more generic as and when we need it. Some test changes required though!

mwhudson (changes requested):
Just the test stuff I think now

subiquity/ui/views/tests/test_identity.py
CarlosNihelton: Indeed, but just inheriting my test class from IsolatedAsyncioTestCase breaks TestMessageHub. I'm fighting for a while with that break without getting a clue of what's going on. Every call to `run_coro`  inside tests_pubsub.py throws `RuntimeError: There is no current event loop in thread 'MainThread'`. I'm not quite sure why testing the identity view disturbs MessageHub tests.

subiquity/ui/views/tests/test_identity.py
```diff
@@ -42,7 +42,21 @@
...
-class IdentityViewTests(unittest.TestCase):
+def tearDownModule() -> None:
+    # Set empty loop policy, so that subsequent get_event_loop() returns a new
+    # loop. If there is no running event loop set, that function will return
+    # the result of `get_event_loop_policy().get_event_loop()` call. If there
+    # is a policy there must be a running loop. IsolatedAsyncioTestCase
+    # closes the loop during tear down, though. It doesn't touch the policy.
+    # By having it as None, it autoinits and the next tests run smoothly.
+    # Another approach would be set a new event_loop for the current policy on
+    # test fixture tearDown, as pytest-ayncio does.
```
mwhudson, line 53: Typo, "pytest-asyncio"

subiquity/ui/views/tests/test_identity.py
mwhudson: Hah! I had no idea about all of that.

mwhudson (approved):

ogayot (approved):

PR #1299, opened 2022-05-24
filesystem: fix udev infinite loop

dbungert:
If the probert probe triggers a udev event, we can loop forever.
Fixed by not restarting an active probe, which can lose information if
it appears mid-probe but that's better than an infinite loop.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -29,6 +29,7 @@ from subiquitycore.async_helpers import (
     run_in_thread,
     schedule_task,
     SingleInstanceTask,
+    TaskAlreadyRunningError,
     )
 from subiquitycore.context import with_context
 from subiquitycore.utils import (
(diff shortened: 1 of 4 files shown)


subiquitycore/async_helpers.py
```diff
@@ -69,6 +71,10 @@ async def start(self, *args, **kw):
         return self.task
 
     def start_sync(self, *args, **kw):
+        if not self.cancel_restart:
+            if self.task is not None and not self.task.done():
+                # start() expects an awaitable
+                return asyncio.sleep(0)
```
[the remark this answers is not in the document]
dbungert: I like that.  I now have clarifying logs covering which scenario happened - might be useful.  Also makes it clearer if the task is never started for some reason.

ogayot (approved):
LGTM, thanks!

subiquity/server/controllers/filesystem.py
```diff
@@ -507,22 +505,17 @@ def stop_listening_udev(self):
         loop.remove_reader(self._monitor.fileno())
 
     def _udev_event(self):
-        cp = run_command(['udevadm', 'settle', '-t', '0'])
-        if cp.returncode != 0:
-            log.debug("waiting 0.1 to let udev event queue settle")
-            self.stop_listening_udev()
-            loop = asyncio.get_event_loop()
-            loop.call_later(0.1, self.start_listening_udev)
-            return
```
mwhudson, line 516: I'm not sure why you're removing this?

mwhudson (approved):
I don't like this but I don't have any good ideas for what to do better :(

subiquity/server/controllers/filesystem.py
dbungert: It made sense when I was turning off the udev listener, but less so now.  Restored.

PR #1300, opened 2022-05-24
debug shell: use a clean environment

dbungert:
Per LP: #1975629, environment variables used to run Subiquity are leaked
into the debug environment, and in the case of things like PYTHONPATH
can harm commands run in that shell.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -31,8 +31,15 @@ apps:
     daemon: simple
     restart-condition: always
     environment:
+      # Save original values of environment variables, we want to restore them
+      # for the debug shell (LP: #1975629).
+      PYTHONPATH_ORIG: $PYTHONPATH
+      PATH_ORIG: $PATH
+      PYTHONIOENCODING_ORIG: $PYTHONIOENCODING
       PYTHONIOENCODING: utf-8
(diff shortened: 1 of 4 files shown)


dbungert:
debating not using 'env' for this

ogayot:
For some reason, it's always been a pain to invoke a shell with a clean-ish environment... :shrug:
Would it make sense to keep the HOME variable as well?

ogayot (approved):
LGTM

mwhudson (commented):
Er hm. What does $PATH end up being in this shell?

dbungert:
> Er hm. What does $PATH end up being in this shell?

LS_COLORS has a real value, I truncated it for readability.
```
# env
PWD=/
LS_COLORS=rs=0:di=01;34:ln=01;...
LESSCLOSE=/usr/bin/lesspipe %s %s
TERM=linux
LESSOPEN=| /usr/bin/lesspipe %s
SHLVL=1
LC_ALL=C
_=/usr/bin/env
# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
This value is a packaging-enforced default.
https://git.launchpad.net/ubuntu/+source/bash/tree/debian/tests/builtin-path.sh

mwhudson (approved):

PR #1301, opened 2022-05-24
TUI Help/About message customisation point for variants

CarlosNihelton:
WSL installer TUI currently crashes if user clicks on Help menu due missing network parameter in the app basemodel. When fixing that we noticed that the Help/About message required some customisation. This PR aims to provide the customisation point for variants and provide the fix for the WSL case.
This aims to close https://github.com/ubuntu/WSL/issues/202

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -120,13 +120,13 @@ class SubiquityClient(TuiApplication):
 
     variant_to_controllers: Dict[str, List[str]] = {}
 
-    def __init__(self, opts):
+    def __init__(self, opts, about_msg=None):
         if is_linux_tty():
             self.input_filter = KeyCodesFilter()
         else:
             self.input_filter = DummyKeycodesFilter()
 
-        self.help_menu = HelpMenu(self)
+        self.help_menu = HelpMenu(self, about_msg)
         super().__init__(opts)
         self.interactive = None
         self.server_updated = None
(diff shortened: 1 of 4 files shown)


subiquity/ui/views/help.py
```diff
@@ -394,8 +394,9 @@ def _close(self, sender):
 
 class HelpMenu(PopUpLauncher):
 
-    def __init__(self, app):
+    def __init__(self, app, about_msg=None):
```
dbungert, line 397: This arg isn't used, correct?

CarlosNihelton: Indeed. It could be, though, but I avoided changing the SubiquityClient init code.

mwhudson: Please feel free to modify the subiquity code if it makes sense to do so!

subiquity/client/client.py
```diff
@@ -120,13 +120,13 @@ def make_ui(self):
 
     variant_to_controllers: Dict[str, List[str]] = {}
 
-    def __init__(self, opts):
+    def __init__(self, opts, about_msg=None):
```
CarlosNihelton, line 123: This was what I originally wanted, but gave up in the middle of the process and you guys encouraged me doing 😄 
Thanks for that.

dbungert (approved):
LGTM

mwhudson (approved):
This looks fine, feel free to merge.

It does make me a bit sad again that system_setup inherits so much from the subiquity package -- really we should move the generic client / server bits out of subiquity to somewhere else and leave subiquity to have the installer specific parts. But well, that's a lot of refactoring. Maybe one day.

PR #1302, opened 2022-05-24
filesystem: report estimated partition size minimums

dbungert:


subiquity/common/types.py
```diff
@@ -268,6 +268,7 @@ class Partition:
     boot: Optional[bool] = None
     os: Optional[OsProber] = None
     offset: Optional[int] = None
+    estimated_min_size: Optional[int] = -1
```
ogayot, line 271: If I understand correctly, this should never be None but I might have overlooked something

dbungert: The Partition object is a bit different as it has two major use cases
* Originating from Subiquity, in which case we're going to give all the fields reasonable values.
* Originating from a client, which is only going to fill out a few fields.

While we could ask clients to send a slightly different object with just the required fields, doing it this way allows the client to get a Partition from a previous call, modify it, then send it back.
Also, the serializer objects if there are extra fields.

ogayot: Thanks for the details! I understand the use-cases better.
Would None be a valid value when the Partition objects originates from Subiquity? or when it originates from the client?
I was expecting the value -1 to be set by the deserializer if `estimated_min_size` is absent from the request, is that incorrect?

mwhudson (commented):
I think you should add or adapt an API test to exercise this?

dbungert:
> I think you should add or adapt an API test to exercise this?

Done

subiquity/models/filesystem.py
```diff
@@ -702,6 +702,16 @@ def boot(self):
...
 
+    @property
+    def estimated_min_size(self):
+        fs_data = self._m._probe_data.get('filesystem', {}).get(self._path())
+        if fs_data is None:
+            return -1
+        val = fs_data.get('ESTIMATED_MIN_SIZE')
+        if val == 0:
+            return self.device.alignment_data().part_align
+        return val
```
mwhudson, line 713: I don't know how likely this is to be a problem in practice but this should round the probert supplied value to part_align, and also round any value smaller than min_gap_size to min_gap_size (I'm now getting side-tracked by the design of PartitionAlignmentData -- how like is it that part_align and min_gap_size will be different? Why is it called min_gap_size and not min_part_size?)

mwhudson (changes requested):
Sorry, I found something else to quibble over.

subiquity/models/filesystem.py
mwhudson: Er yes, I guess min(align_up(ESTIMATED_MIN_SIZE), actual_size) or something.

mwhudson (approved):

subiquity/models/filesystem.py
```diff
@@ -702,6 +702,18 @@ def boot(self):
...
+    def estimated_min_size(self):
+        fs_data = self._m._probe_data.get('filesystem', {}).get(self._path())
+        if fs_data is None:
+            return -1
+        val = fs_data.get('ESTIMATED_MIN_SIZE', -1)
+        if val == 0:
+            return self.device.alignment_data().part_align
+        if val == -1:
+            return -1
+        return align_up(val)
```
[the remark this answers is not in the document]
dbungert: ty

PR #1304, opened 2022-06-01
take storage version from kernel command line

mwhudson:


subiquity/cmd/server.py
```diff
@@ -26,6 +27,24 @@
     )
 
 
+class CommandLineParams:
```
ogayot, line 30: I'd appreciate to have a `__str__` or `__repr__` defined so we can log the content of a CommandLineParams object without going through `.__dict__`
Would leaning on `attr.s` or `dataclass` make sense?

ogayot (approved):
LGTM, just one suggestion.

subiquity/cmd/server.py
ogayot: Thanks!

PR #1306, opened 2022-06-06
Enables TCPSite for WSL

CarlosNihelton:
Per aiohttp documentation, the `host` parameter of the TCPSite sets the interface on which the server will listen on. Listening on 127.0.0.1, as you know, means only clients running on the same host can reach that server. Interestingly, with WSL2, clients running on Windows can make requests to 127.0.0.1 which crosses the vEthernet adapter and reach the VM as if it was the same host. This behavior was tested on Windows 10 and 11. Thus by simply locking to the loopback interface can give us some level of security and keep the solution simple.

This PR enables the component in charge of starting the server to choose which port to bind to. This is a key feature to enable migrating the OOBE GUI to a Windows application.

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -245,11 +245,32 @@ if [ "${RELEASE%.*}" -ge 20 ]; then
         echo "Timeout reached before Subiquity TCP socket started listening"
         exit 1
     fi
-    curl "http://localhost:$port/meta/status"
-    curl_ec=$?
+    loopback_failed=0
+    unallowed_failed=0
+    # Assert that only loopback interface is accepted.
+    interfaces=($(ip --json link show up | jq -r '.[]["ifname"] | select ( . != null )'))
+    for if in ${interfaces[@]}; do
+        curl_ec=0
+        timeout 10s curl "http://localhost:$port/meta/status" --interface $if || curl_ec=$?
+        # Loopback should exit 0
+        if [ $if = "lo" ]; then
+            if [ $curl_ec -ne 0 ]; then
+                loopback_failed=1
+            fi
+        # Other interfaces shoud not
+        else
+            if [ $curl_ec -eq 0 ]; then
+                unallowed_failed=1
+            fi
+        fi
+    done
     kill $subiquity_pid
-    if [ $curl_ec != 0 ]; then
-        echo "GET Request to meta/status failed with code: $curl_ec"
+    if [ $loopback_failed -ne 0 ]; then
+        echo "Loopback was expected to connect"
+        exit 1
+    fi
+    if [ $unallowed_failed -ne 0 ]; then
+        echo "Only the loopback interface should be allowed."
         exit 1
     fi
 
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -250,19 +250,23 @@ if [ "${RELEASE%.*}" -ge 20 ]; then
     # Assert that only loopback interface is accepted.
     interfaces=($(ip --json link show up | jq -r '.[]["ifname"] | select ( . != null )'))
     for if in ${interfaces[@]}; do
-        curl_ec=0
-        timeout 10s curl "http://localhost:$port/meta/status" --interface $if || curl_ec=$?
-        # Loopback should exit 0
-        if [ $if = "lo" ]; then
-            if [ $curl_ec -ne 0 ]; then
-                loopback_failed=1
+        for ipv in 4 6; do
+            curl_ec=0
+            timeout 10s \
+                curl -$ipv "http://localhost:$port/meta/status" --interface $if \
+                || curl_ec=$?
+            # Loopback should exit 0 on IPv4
+            if [ $if = "lo" ]; then
+                if [ $curl_ec -ne 0 -a $ipv -eq 4 ]; then
+                    loopback_failed=1
+                fi
+            # Everything else should not.
+            else
+                if [ $curl_ec -eq 0 ]; then
+                    unallowed_failed=1
+                fi
             fi
-        # Other interfaces shoud not
-        else
-            if [ $curl_ec -eq 0 ]; then
-                unallowed_failed=1
-            fi
-        fi
+        done
     done
     kill $subiquity_pid
     if [ $loopback_failed -ne 0 ]; then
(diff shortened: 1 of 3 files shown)


system_setup/server/server.py
```diff
@@ -88,3 +92,13 @@ def make_model(self):
     async def wait_for_cloudinit(self):
         self.cloud_init_ok = True
         return
+
+    async def start_site(self, runner: web.AppRunner):
+        port = self.opts.tcp_port
+        if port is None:
+            return await super().start_site(runner)
+
+        # Subiquity runs with root privileges. We don't wont outsiders to
```
didrocks, line 101: s/wont/want/

didrocks: + I would clarify this is why we only listen on localhost.

system_setup/server/server.py
```diff
@@ -88,3 +92,13 @@ def make_model(self):
     async def wait_for_cloudinit(self):
         self.cloud_init_ok = True
         return
+
+    async def start_site(self, runner: web.AppRunner):
+        port = self.opts.tcp_port
+        if port is None:
+            return await super().start_site(runner)
```
didrocks, line 99: Shouldn’t we still call parent start_site() in case we pass a tcp port?

didrocks (commented):
Thanks for the PR! This looks mostly good and simple :) I have a couple of comments, but overall, this looks like what we need :)

system_setup/server/server.py
CarlosNihelton: Thanks for noticing that typo.

system_setup/server/server.py
CarlosNihelton: `super().start_site()` handles the default behavior (creating a Unix socket site and changing the socket permissions). When variants override this method to create other types of sites, its their responsibility to start it. Internally, SubiquityServer will await this method in its `start_api_server()` method, as you can see here:

https://github.com/canonical/subiquity/blob/a9ac5a5fb28168aa328e2f9453b1f4768fd4e54c/subiquity/server/server.py#L494-L505

This is a little different from the original intent (which was make a site factory), it changed after some conversation I had with Dan Bungert. The spec contains the updated proposal, which was already merged in Subiquity.

CarlosNihelton:
> Thanks for the PR! This looks mostly good and simple :) I have a couple of comments, but overall, this looks like what we need :)

Thank you for the suggested test. I was really unconfident about this idea, but it worked quite well!

scripts/runtests.sh
```diff
@@ -231,6 +231,29 @@ grep -q 'finish: subiquity/Install/install/postinstall/run_unattended_upgrades:
...
+    subiquity_pid=$!
+    next_time=3
+    until [ $next_time -eq 0 ] || [ ! -z "$(ss -Hlt sport = $port)" ]; do
+        sleep $(( next_time-- ))
+    done
+    if [ $next_time -eq 0 ]; then
+        echo "Timeout reached before Subiquity TCP socket started listening"
+        exit 1
+    fi
+    curl "http://localhost:$port/meta/status"
```
dbungert, line 248: I won't require it for this PR, but a nice-to-have to consider - an integration test that connects on a non-loopback interface should never work.  As mentioned previously, I wish to be extremely certain that we don't listen on a public-facing interface.

(Once, when doing a windows server install, it was behaving weirdly, and I made the determination that the server had been compromised before I had even completed the install.  The internet can be scary.)

system_setup/server/server.py
dbungert: Hi @didrocks, yes this implementation is consistent with my request to Carlos to ensure that the unix socket creation happens in the same place as the chmod of that socket, which was implemented in #1305.  An earlier plan had them split and the `chmod` was behind a new `if` check, and I didn't want even a little risk that that check could fail to run `chmod`, as Subiquity for Server would be fine but ubuntu-desktop-install would fail.

So +1 to having system-setup override the `start_site()` behavior.

dbungert (approved):

scripts/runtests.sh
CarlosNihelton: Hey, @dbungert , do you think it would be worthy to add a middleware in system_setup to check each if  `request.remote == "127.0.0.1" `  and raise `HttpForbidden` if not?  Or is it too silly?

dbungert: I do not think that would be too silly.

CarlosNihelton: I'm asking because I don't know those attacks could be, but if they were like a reverse proxy, checking for request.remote would be misleading, since the requests reach the server coming from the proxy. In some sense the virtual adapters connecting Windows and Linux on WSL remind me of those reverse proxies.

scripts/runtests.sh
```diff
@@ -231,6 +231,50 @@ grep -q 'finish: subiquity/Install/install/postinstall/run_unattended_upgrades:
 
 # The OOBE doesn't exist in WSL < 20.04
 if [ "${RELEASE%.*}" -ge 20 ]; then
+    # Test TCP connectivity (system_setup only)
+    clean
+    port=50321
+    LANG=C.UTF-8 timeout --foreground 60 \
```
dbungert, line 237: If you do much more of this, consider converting to something more like what subiquity/tests/api/test_api.py does.  That file is overdue to restructure to split the tests a bit (and so is runtests.sh).  My first API test was similar, forking the server process and using `curl` to communicate back with it.

dbungert (approved):

scripts/runtests.sh
```diff
@@ -231,6 +231,50 @@ grep -q 'finish: subiquity/Install/install/postinstall/run_unattended_upgrades:
...
+        echo "Timeout reached before Subiquity TCP socket started listening"
+        exit 1
+    fi
+    loopback_failed=0
+    unallowed_failed=0
+    # Assert that only loopback interface is accepted.
+    interfaces=($(netstat -i | awk "NR>2 {print}" | cut -d" " -f1))
+    for if in ${interfaces[@]}; do
+        curl_ec=0
+        timeout 10s curl "http://localhost:$port/meta/status" --interface $if || curl_ec=$?
```
ogayot, line 254: I'd suggest to try over IPv4 and IPv6. Sometimes, there are surprises :)
```
$ curl -4
$ curl -6
```

scripts/runtests.sh
CarlosNihelton: I loved this idea. Thank you!

didrocks (approved):

PR #1307, opened 2022-06-08
Check child forms status when disabling/enabling Done button of a form

ogayot:
When the validation of a field fails in a form, we disable the Done button. With child forms, however, it did not work because they have their own set of hidden buttons ; that are not the ones the users interacts with.

This patch makes parent forms recurse on the child forms when checking if any field is in error. Also, when a child form undertakes validation, it now propagates to the parent, so that the done button can be immediately enabled/disabled.

Having a validation error in a child form that is not currently enabled should not prevent the user from moving forward, so we disable recursion for child forms that are disabled.

--

The above change was needed as part of the Ubuntu Pro changes where we make use of child forms for manual token or Ubuntu One email. Since Guided Storage also makes use of SubForms, I took some time to check how that screen would be affected. I'm quite happy about the change.

Before, the Done button would never be disabled in the GuidedStorage screen, even in case of invalid LUKS passphrase or some other error. The Done button is now disabled if a validation error occurs ; unless it's in a disabled child form.

luks-enabled-passwords-do-not-match
luks-disabled-passwords-do-not-match

### DIFF
--- a/subiquitycore/ui/form.py
+++ b/subiquitycore/ui/form.py
@@ -111,27 +111,6 @@ class _Validator(WidgetWrap):
         self.field.validate()
 
 
-class FormField(abc.ABC):
-
-    next_index = 0
-    takes_default_style = True
-    caption_first = True
-
-    def __init__(self, caption=None, help=None):
-        self.caption = caption
-        self.help = help
-        self.index = FormField.next_index
-        FormField.next_index += 1
-
-    @abc.abstractmethod
-    def _make_widget(self, form):
-        pass
-
-    def bind(self, form):
-        widget = self._make_widget(form)
-        return BoundFormField(self, form, widget)
-
-
 class WantsToKnowFormField(object):
     """A marker class."""
     def set_bound_form_field(self, bff):
@@ -161,16 +140,8 @@ class BoundFormField(object):
             widget.set_bound_form_field(self)
 
     def is_in_error(self) -> bool:
-        """ Tells whether this field is in error. We will also check if any of
-        the enabled subform reports an error.
-        """
-        if self.in_error:
-            return True
-
-        if not self._enabled or not isinstance(self.widget, SubFormWidget):
-            return False
-
-        return self.widget.form.has_validation_error()
+        """ Tells whether this field is in error. """
+        return self.in_error
 
     def _build_table(self):
         widget = self.widget
@@ -308,6 +279,42 @@ class BoundFormField(object):
         self._enabled = val
         for row in self._rows:
             row.enabled = val
+
+
+class BoundSubFormField(BoundFormField):
+    def is_in_error(self):
+        """ Tells whether this field is in error. We will also check if the
+        subform (if enabled) reports an error.
+        """
+        if super().is_in_error():
+            return True
+
+        if not self._enabled:
+            return False
+
+        return self.widget.form.has_validation_error()
+
+
+class FormField(abc.ABC):
+
+    next_index = 0
+    takes_default_style = True
+    caption_first = True
+    bound_field_class = BoundFormField
+
+    def __init__(self, caption=None, help=None):
+        self.caption = caption
+        self.help = help
+        self.index = FormField.next_index
+        FormField.next_index += 1
+
+    @abc.abstractmethod
+    def _make_widget(self, form):
+        pass
+
+    def bind(self, form):
+        widget = self._make_widget(form)
+        return self.bound_field_class(self, form, widget)
 
 
 def simple_field(widget_maker):
@@ -517,7 +524,7 @@ class Form(object, metaclass=MetaForm):
             narrow_rows=narrow_rows)
 
     def has_validation_error(self) -> bool:
-        """ Tells if any field (or subformfield) is in error. """
+        """ Tells if any field is in error. """
         return any(map(lambda f: f.is_in_error(), self._fields))
 
     def validated(self):
@@ -554,6 +561,7 @@ class SubFormWidget(WidgetWrap):
 class SubFormField(FormField):
 
     takes_default_style = False
+    bound_field_class = BoundSubFormField
 
     def __init__(self, form_cls, caption=None, help=None):
         super().__init__(caption=caption, help=help)
--- a/subiquitycore/ui/form.py
+++ b/subiquitycore/ui/form.py
@@ -160,6 +160,18 @@ class BoundFormField(object):
         if isinstance(widget, WantsToKnowFormField):
             widget.set_bound_form_field(self)
 
+    def is_in_error(self) -> bool:
+        """ Tells whether this field is in error. We will also check if any of
+        the enabled subform reports an error.
+        """
+        if self.in_error:
+            return True
+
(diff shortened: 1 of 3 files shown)


subiquitycore/ui/form.py
```diff
@@ -504,13 +517,13 @@ def as_screen(self, focus_buttons=True, excerpt=None, narrow_rows=False):
...
+        return any(map(lambda f: f.is_in_error(recurse=recurse), self._fields))
+
     def validated(self):
-        in_error = False
-        for f in self._fields:
-            if f.in_error:
-                in_error = True
-                break
-        if in_error:
+        if self.has_validation_error(recurse=True):
```
dbungert, line 526: Just curious - can you tell me about the use case of recurse=False?

dbungert (approved):

subiquitycore/ui/form.py
```diff
@@ -160,6 +160,19 @@ def __init__(self, field, form, widget):
...
 
+    def is_in_error(self, recurse: bool = True) -> bool:
+        """ Tells whether this field is in error. If recurse is True, we will
+        also check if any of the enabled subform reports an error.
+        """
+        if self.in_error:
+            return True
+
+        if not recurse or not self._enabled \
+                or not isinstance(self.widget, SubFormWidget):
```
mwhudson, line 171: This check feels wrong. It feels like maybe it would be better to somehow allow the FormField to define what "in error" means for a given situation. This could be a method on FormField that takes a BoundFormField, or it could be overriding bind() in SubFormField to return a subclass of BoundFormField that overrides is_in_error().

subiquitycore/ui/form.py
mwhudson: Yeah looks like this argument is no longer needed?

mwhudson (commented):
Thanks for the clear description of the problem. I have a counter suggestion for how to fix it though. I was also going to say you should add a test, but apparently I never wrote any tests for the form code at all. Sorry about that!

subiquitycore/ui/form.py
ogayot: Yes, we can get rid of it now. It's pretty useful in the debugger to figure out why a form would enable/disable the Done button. But yeah, not needed in the codebase. I'll remove it.

ogayot:
> I was also going to say you should add a test, but apparently I never wrote any tests for the form code at all. Sorry about that!

I've added tests that relevant tests for the use case, which assert that:
* `Form.has_validation_error()` returns the expected result (with and without child forms - optionally disabled)
* Calling `Form.validated()` enables/disables the `Done` button as expected
* Calling `SubForm.validated()` propagates to the parent (and therefore enable/disable the `Done` button in the parent form)

I didn't have the courage to write tests for a wider range of the existing form stuff though.

subiquitycore/ui/form.py
```diff
@@ -160,6 +139,10 @@ def __init__(self, field, form, widget):
         if isinstance(widget, WantsToKnowFormField):
             widget.set_bound_form_field(self)
 
+    def is_in_error(self) -> bool:
+        """ Tells whether this field is in error. """
```
mwhudson, line 143: I guess maybe this should check self._enabled too? Not sure.

subiquitycore/ui/form.py
```diff
@@ -504,13 +523,12 @@ def as_screen(self, focus_buttons=True, excerpt=None, narrow_rows=False):
             focus_buttons=focus_buttons, excerpt=excerpt,
             narrow_rows=narrow_rows)
 
+    def has_validation_error(self) -> bool:
+        """ Tells if any field is in error. """
+        return any(map(lambda f: f.is_in_error(), self._fields))
+
     def validated(self):
```
mwhudson, line 530: I'm not sure I have any better ideas but I'm not sure that "validated" is a very useful name for this method. Can you think of a better one? :)

subiquitycore/ui/tests/test_form.py
```diff
@@ -0,0 +1,151 @@
...
+
+from subiquitycore.ui.form import (
+    Form,
+    StringField,
+    SubForm,
+    SubFormField,
+    )
+
+
+class TestForm(TestCase):
```
mwhudson, line 26: Thanks for this!

mwhudson (approved):
Thanks for the changes, I like this a lot now.

It's funny reading the forms code again, I haven't really looked at it for a long time. One thing that's kind of a pain is tying the enabled/disabled state of form elements to the state of other form elements. If you have any ideas for how to do that more declaratively I am all ears :)

subiquitycore/ui/form.py
ogayot: Adding this test crossed my mind multiple times.
I think generally speaking we never want errors in disabled fields to be considered meaningful. Or at least I can't think of any scenario where we would want that. So it makes sense to me to return False is the element is disabled.

The main reason why I didn't add this test is because I was afraid it could break something. For some reason, this test was not in place before (disabled fields did affect the done button) and I couldn't figure out if it was on purpose. Probably not?

subiquitycore/ui/tests/test_form.py
ogayot: welcome!

subiquitycore/ui/form.py
ogayot: Agreed. How do you feel about the following?

```python
def on_field_validated(self) -> None:
    """ Method to call when a field is validated. """
    self.refresh_done_button()
    
def refresh_done_button(self) -> None:
    """ Look for any error in the fields and enable/disable the Done button accordingly. """
    # here goes the body of `validated`
```

fields would call `on_field_validated` after undertaking validation.

ogayot:
> It's funny reading the forms code again, I haven't really looked at it for a long time. One thing that's kind of a pain is tying the enabled/disabled state of form elements to the state of other form elements. If you have any ideas for how to do that more declaratively I am all ears :)

I don't have any better ideas at the moment, but I'm sure I will have the opportunity to look at that more closely in the near future!

subiquitycore/ui/form.py
mwhudson: Yeah maybe something for another branch. I have a hard time believing we want an invalid disabled field to disable the done button.

mwhudson:
> > It's funny reading the forms code again, I haven't really looked at it for a long time. One thing that's kind of a pain is tying the enabled/disabled state of form elements to the state of other form elements. If you have any ideas for how to do that more declaratively I am all ears :)
> 
> I don't have any better ideas at the moment, but I'm sure I will have the opportunity to look at that more closely in the near future!

I wonder if react or something like this has any patterns to follow here. Or maybe even Django by now!

subiquitycore/ui/form.py
ogayot: I did implement this change but I'm not too happy with it at the moment ; so I'm not pushing it yet.
The change in subiquity/ui/views/filesystem/guided.py proves that there are other scenarios where we'd want to refresh the done button (e.g., toggling a radio button); not just when a field undertakes validation.
I want to give it more thoughts and I will go with a follow up PR.

PR #1309, opened 2022-06-10
gap: several utilities

dbungert:
Add gap.split() and gaps.at_offset().

### DIFF
--- a/subiquity/common/filesystem/gaps.py
+++ b/subiquity/common/filesystem/gaps.py
@@ -205,3 +205,11 @@ def movable_trailing_partitions_and_gap_size(partition):
             else:
                 return (trailing_partitions, 0)
     return (trailing_partitions, 0)
+
+
+def at_offset(device, offset):
+    for pg in parts_and_gaps(device):
+        if isinstance(pg, Gap):
+            if pg.offset == offset:
+                return pg
+    return None
--- a/subiquity/common/filesystem/gaps.py
+++ b/subiquity/common/filesystem/gaps.py
@@ -28,7 +28,8 @@ from subiquity.models.filesystem import (
     )
 
 
-@attr.s(auto_attribs=True)
+# should also set on_setattr=None with attrs 20.1.0
+@attr.s(auto_attribs=True, frozen=True)
 class Gap:
     device: object
     offset: int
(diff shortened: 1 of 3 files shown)


subiquity/common/filesystem/gaps.py
```diff
@@ -41,6 +41,26 @@ class Gap:
     def id(self):
         return 'gap-' + self.device.id
 
+    def copy(self):
```
mwhudson, line 44: A method like this clearly suggests that Gaps should be immutable. Do we want to consider using the attrs support for this? https://www.attrs.org/en/stable/examples.html#immutability

subiquity/common/filesystem/tests/test_gaps.py
```diff
@@ -31,26 +32,74 @@
...
+        self.assertEqual(gap.offset, new_gaps[0].offset)
+        self.assertEqual(gap.offset + size, new_gaps[1].offset)
+
+
+class TestAtOffset(unittest.TestCase):
+    def test_zero(self):
+        self.assertIsNone(gaps.at_offset(make_disk(), 0))
+
+    def test_one_mib(self):
+        self.assertIsNotNone(gaps.at_offset(make_disk(), MiB))
```
mwhudson, line 76: The MiB is kind of a mystery guest here. d=make_disk(); g=largest_gap(d); assertEqual(gaps.at_offset(d, g.offset), g) seems like a better test.

subiquity/common/filesystem/gaps.py
```diff
@@ -41,6 +41,26 @@ class Gap:
...
+        return Gap(device=self.device,
+                   offset=self.offset,
+                   size=self.size,
+                   in_extended=self.in_extended)
+
+    def split(self, size):
```
mwhudson, line 50: Should this care about alignment issues at all? Or is that the callers responsibility? (I say probably the callers responsibility?)

mwhudson (approved):
This is all fine as is, I have a few leading questions which you can address or ignore as you see fit.

subiquity/common/filesystem/gaps.py
dbungert: Good question.  Since the `gaps.find_disk_gaps` functions are alignment aware, probably other gaps operations should be also.

The `align_up` in `FilesystemModel.add_partition()` concerns me, I think the right place is for that to happen a bit earlier, and maybe gaps help with that.

Will consider for a future branch.

subiquity/common/filesystem/gaps.py
dbungert: Yes, that is reasonable, and with that `copy()` is not as helpful, so I ended up dropping `copy()`.

subiquity/common/filesystem/tests/test_gaps.py
dbungert: Fixed, thanks.
I left the zero one, there will always be partition junk there.

mwhudson (approved):
Nice

PR #1310, opened 2022-06-13
console-conf: fix crash on network info

mwhudson:
If you run dryrun for console-conf, go to the network page, go to an
interface, then info, a crash of the form
TypeError: object str can't be used in 'await' expression
can be seen. The signature for the core version of get_info_for_netdev
is not async, but a non-async method returning str.

So mark the core version of the function async.

Co-authored-by: Michael Hudson-Doyle <michael.hudson@canonical.com>
(cherry picked from commit 4b1277ae2ddfd3660af64f474b8c60cfe79810e1)

### DIFF
--- a/subiquity/server/controllers/network.py
+++ b/subiquity/server/controllers/network.py
@@ -405,4 +405,4 @@ class NetworkController(BaseNetworkController, SubiquityController):
         self.delete_link(dev_name)
 
     async def info_GET(self, dev_name: str) -> str:
-        return self.get_info_for_netdev(dev_name)
+        return await self.get_info_for_netdev(dev_name)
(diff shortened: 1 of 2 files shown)


mwhudson:
For https://bugs.launchpad.net/subiquity/+bug/1978339. The core22 snap is following this branch but specifies a commit so that will need to be updated too.

ogayot (approved):

ogayot:
Thanks! I was under the impression that the next release would be based on main though?

dbungert:
> Thanks! I was under the impression that the next release would be based on main though?

It will be, but this branch is still available for maintenance things. So this still helps.
But yes, the builder that creates snaps targetting the Beta channel needs to look at a new branch.

dbungert (approved):

PR #1315, opened 2022-06-15
ubuntu-pro: don't assume an overlay is present in show_* functions

ogayot:
ubuntu-pro: don't assume an overlay is present in show_* functions
    
show_* functions from the UbuntuProView object always assumed that the loading overlay was present. When called, they would therefore attempt to close the overlay. This results in a crash if no overlay is present.
    
With the current design, it is fine because we never call the show_* functions without an overlay being present.

However, with new design of Ubuntu Pro, sometimes the show_* functions will be called without an overlay.
    
We now delegate the responsibility of removing the overlay by using a callback system.

### DIFF
--- a/subiquity/ui/views/ubuntu_pro.py
+++ b/subiquity/ui/views/ubuntu_pro.py
@@ -25,7 +25,10 @@ from urwid import (
     Widget,
     )
 
-from subiquity.common.types import UbuntuProService
+from subiquity.common.types import (
+    UbuntuProCheckTokenStatus,
+    UbuntuProService,
+    )
 from subiquitycore.view import BaseView
 from subiquitycore.ui.buttons import (
     back_btn,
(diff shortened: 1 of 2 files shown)


dbungert (commented):
It sounds to me that remove_overlay() should be robust against the possibility that there might not actually be an overlay.  It seems like it tries to.  What sort of problem happens there?

ogayot:
> Do you mind fixing remove_overlay first? I'd like to do so while we have a good test case handy. It could be in this PR or another one.

Took me a while to fully understand the semantics of `remove_overlay` but I think I got it right now.
`StretchyOverlay` (from subiquitycore.ui.stretchy) and `Overlay` (from urwid) both have a `bottom_w` attribute but these classes don't inherit from one another.
BTW, the crash with `AttributeError` would occur only if the `stretchy` parameter of `remove_overlay` is None.
I created a different PR to fix this: https://github.com/canonical/subiquity/pull/1324

subiquity/ui/views/ubuntu_pro.py
```diff
@@ -146,11 +149,27 @@ def done(self, form: UbuntuProForm) -> None:
         asynchronously check if the token is valid. """
         token: str = form.token.value
         if token:
+            def on_success(services: List[UbuntuProService]) -> None:
+                self.remove_overlay()
```
dbungert, line 153: What value of not_found_ok is desired?  Is the default appropriate?

dbungert (approved):

PR #1316, opened 2022-06-15
filesystem: several size related tweaks

dbungert:
* send gaps of the intended partition size to create_partition
* create_partition now uses the gap size instead of spec size
* partition size scaling functions for boot & uefi expect an available space value instead of a disk (which it would only use to get the total disk size)

### DIFF
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -220,6 +220,7 @@ class FilesystemManipulator:
                     spec['size'], gap.size)
                 spec['size'] = gap.size
 
+        gap = gap.split(spec['size'])[0]
         self.create_partition(disk, gap, spec)
 
         log.debug("Successfully added partition")
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -79,7 +79,7 @@ class FilesystemManipulator:
 
     def create_partition(self, device, gap, spec, **kw):
         part = self.model.add_partition(
-            device, size=spec["size"], offset=gap.offset, **kw)
+            device, size=gap.size, offset=gap.offset, **kw)
         self.create_filesystem(part, spec)
         return part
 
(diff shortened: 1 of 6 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -376,7 +371,8 @@ async def v2_add_partition_POST(self, data: AddPartitionV2) \
             'mount': data.partition.mount,
         }
 
-        self.create_partition(disk, data.gap, spec, wipe='superblock')
+        gap = gaps.at_offset(disk, data.gap.offset).split(requested_size)[0]
```
mwhudson, line 374: This translation between types.Gap and gaps.Gap is, hmm. A thing :) I guess it's needed but it feels a bit silly somehow.

subiquity/common/filesystem/manipulator.py
```diff
@@ -220,6 +220,7 @@ def partition_disk_handler(self, disk, spec, *, partition=None, gap=None):
                     spec['size'], gap.size)
                 spec['size'] = gap.size
 
+        gap = gap.split(spec['size'])[0]
```
mwhudson, line 223: I wonder how many callers of split are going to care about the remainder gap? I guess we'll find out...

mwhudson (approved):
Love the small PRs!

subiquity/common/filesystem/manipulator.py
dbungert: :smile: 
The remainder actually gets used with the core commits for use_gap.  It works extremely well for creating multiple partitions in one gap, which is why I started `gap.split`.

mwhudson: Ah right, that makes sense!

subiquity/server/controllers/filesystem.py
dbungert: Yes, the same-sounding types between what is provided by common.types and what filesystem actually uses is a bit unfortunate.  If we can come up with an answer for clarity there in general it will also probably help this case.

PR #1317, opened 2022-06-15
deps: fix missing developer dep python3-mock for apt-deps make target

blackboxsw:
Fix missing python3-mock dep which allows folks to run unittests on a new jammy container

# Steps to reproduce
lxc launch ubuntu-daily:jammy dev-j
lxc exec dev-j bash
git clone https://github.com/canonical/subiquity.git
apt install make
make aptdeps
make unit

# Desired behavior
....
======================= 436 passed, 68 warnings in 4.88s =======================

# failure:
```
_________ ERROR collecting subiquity/tests/test_timezonecontroller.py __________
ImportError while importing test module '/root/subiquity/subiquity/tests/test_timezonecontroller.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
subiquity/tests/test_timezonecontroller.py:16: in <module>
    import mock
E   ModuleNotFoundError: No module named 'mock'
```

### DIFF
--- a/subiquity/tests/test_timezonecontroller.py
+++ b/subiquity/tests/test_timezonecontroller.py
@@ -13,7 +13,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import mock
+from unittest import mock
 
 from subiquity.common.types import TimeZoneInfo
 from subiquity.models.timezone import TimeZoneModel


dbungert (commented):
Thanks for doing this, Chad!

While this is a valid solution, I think we should solve this differently.
That one failing test file should be fine with the python-provided mock from unittest.  If we import from there, we can skip adding another dependency from apt.

I'd like to fix test_timezonecontroller.py instead, what do you think?

blackboxsw:
> 

  Totally agreed @dbungert I should have looked generically at what you folks use throughout the general test framework instead of reacting quickly to the first trace I saw. +1 I rebased and dropped the original commit. Using stdlib's unittest for this level of mock handling makes sense.

ogayot (approved):

dbungert (approved):

PR #1318, opened 2022-06-15
ui/stretchy: do not assign to underscore

dbungert:
Underscore is used with the gettext integration, so remove this
otherwise-harmless assignment that is intended as disposal of an unused
variable.


mwhudson (approved):
I wonder if we can coerce
"make lint" into complaining about this?

ogayot:
I'm curious, what are the consequences of assigning to _ in this context? is it breaking something?
We have other existing assignments too:

```
console_conf/controllers/identity.py:        key, _, value = line.partition("=")
subiquitycore/lsb_release.py:        k, _, v = tok.partition("=")
```

plus many uses of _ as a iterator variable.

mwhudson:
_ is how we do translations.

ogayot:
> _ is how we do translations.

Right, but - does it break translations when we use `_` for other purposes - is what I meant?
If it does, I think we want to remove all other uses of `_` across the whole tree, not just this one, don't we?

And if we decide to search and replace, I recommend replacing `_` by  `__` for throwaway variables. This is probably safer since:
```python
x, _ = iterable
```
and 
```python
x = iterable[0]
```
are only equivalent when iterable has exactly two elements to unpack.

dbungert:
> Right, but - does it break translations when we use `_` for other purposes ?

In this case, no.

I realized yesterday that I have been just assigning to `_` recently but wasn't thinking of the consequences.  I think it's a bad habit and I want to break that habit.

So like Michael hinted above, I want to set up a lint and have us stop assigning to `_` in general.

> We have other existing assignments too
> plus many uses of _ as a iterator variable.

I missed those in my search then - thanks for identifying them!  They are next on the list then - will you send a PR or shall I?

dbungert:
> I recommend replacing _ by __ for throwaway variables.

I think I'm going to ask that we not do that, that's probably OKish in the terminal font but I think it might be hard to spot in code review in other places.

PR #1319, opened 2022-06-16
cloud-config: schema drop deprecated dict user.groups value type

blackboxsw:
Align with latest cloud-init schema avoid using depracated
value types for users. groups definitions in cloud-config.

Also fix get_users_and_groups to return a list instead of
dict as all call sites expected it to return a list.

### DIFF
--- a/subiquity/common/resources.py
+++ b/subiquity/common/resources.py
@@ -18,6 +18,7 @@
 
 import logging
 import os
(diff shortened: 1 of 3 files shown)


mwhudson (approved):
Thanks

dbungert (approved):

PR #1320, opened 2022-06-16
unittests: fix group check

dbungert:
The recently changed group list check is a bit fragile to the exact
groups available on the system.  Mock out the list of actually existing
groups.


blackboxsw (approved):
+1 makes complete sense. Good fix, thank you. LGTM!

mwhudson (approved):

PR #1321, opened 2022-06-16
[system_setup] Avoid integration test failure due server exited

CarlosNihelton:
`timeout 60` might be too short for someone with lots of network interfaces.
Since we grab the PID and run the server in the background, the timeout is not critically necessary.

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -234,8 +234,7 @@ if [ "${RELEASE%.*}" -ge 20 ]; then
     # Test TCP connectivity (system_setup only)
     clean
     port=50321
-    LANG=C.UTF-8 timeout --foreground 60 \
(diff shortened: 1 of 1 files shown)


scripts/runtests.sh
```diff
@@ -234,8 +234,7 @@ if [ "${RELEASE%.*}" -ge 20 ]; then
     # Test TCP connectivity (system_setup only)
     clean
     port=50321
-    LANG=C.UTF-8 timeout --foreground 60 \
-        python3 -m system_setup.cmd.server --dry-run --tcp-port=$port &
+    LANG=C.UTF-8 python3 -m system_setup.cmd.server --dry-run --tcp-port=$port &
```
dbungert, line 237: If we take the next `exit 1`, this process will still be running.

dbungert (changes requested):

dbungert (approved):

PR #1323, opened 2022-06-16
filesystem: autoinstall support use_gap mode

dbungert:
Marked draft as this needs a bit more testing before I would be comfortable merging.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -133,28 +133,26 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
                 "autoinstall config did not create needed bootloader "
                 "partition")
 
-    def guided_direct(self, disk):
-        self.reformat(disk)
-        gap = gaps.largest_gap(disk)
-        spec = {
-            "size": gap.size,
-            "fstype": "ext4",
-            "mount": "/",
-            }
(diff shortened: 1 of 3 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -461,19 +460,36 @@ async def _probe(self, *, context=None):
...
+        self.validate_layout_mode(mode)
+
+        if mode == 'reformat_disk':
+            match = layout.get("match", {'size': 'largest'})
+            disk = self.model.disk_for_match(self.model.all_disks(), match)
+            if not disk:
+                raise Exception("autoinstall cannot configure storage "
+                                "- no disk found large enough for install")
+        elif mode == 'use_gap':
+            gap = self.largest_gap(self.model.all_disks())
```
mwhudson, line 475: This should filter the disks through "can_be_boot(d, with_reformatting=False)" I think. In general the device selection here should probably follow FilesystemController.guided_GET more closely but well.  autoinstall into the biggest gap on a VROC subvolume isn't something that has come up as a feature request yet.

mwhudson (approved):
Just one quibble!

dbungert:
Gave this some VM testing, found some bugs, decided to send fixes in new PRs as the core is fine.

PR #1324, opened 2022-06-20
ui: avoid crashing when removing overlay that does not exist

ogayot:
`BaseView.remove_overlay()` would crash with `AttributeError` if no overlay was found. We now add a `not_found_ok` parameter (defaulting to ~~True~~ False) that makes the function silently return if the overlay could not be found.

~~Passing `not_found_ok=False` and~~ catching `OverlayNotFoundError` can be helpful in some scenarios to do something different if no overlay was found.

### DIFF
--- a/subiquity/ui/views/ssh.py
+++ b/subiquity/ui/views/ssh.py
@@ -297,5 +297,7 @@ class SSHView(BaseView):
             ConfirmSSHKeys(self, ssh_data, identities))
 
     def fetching_ssh_keys_failed(self, msg, stderr):
-        self.remove_overlay()
+        # FIXME in answers-based runs, the overlay does not exist so we pass
+        # not_found_ok=True.
+        self.remove_overlay(not_found_ok=True)
         self.show_stretchy_overlay(SomethingFailed(self, msg, stderr))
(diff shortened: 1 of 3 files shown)


mwhudson (approved):
It would be nice if we could avoid bugs where we call remove_overlay when we shouldn't but this is pragmatic I guess!

dbungert (approved):
Thanks for looking into this.

ogayot:
> It would be nice if we could avoid bugs where we call remove_overlay when we shouldn't but this is pragmatic I guess!

Agreed! That's what https://github.com/canonical/subiquity/pull/1315 was about, really. But I agreed to also relax `remove_overlay()` a bit. Maybe the default value should change from `True` to `False` is we want to catch more errors?

ogayot:
I noticed this PR addresses a failure in integration tests when we attempt to fetch an SSH key but fail to contact github/launchpad. The SSH client controller does not open an overlay but it attempts to close an overlay in case of failure.

Although in this case, the PR "solves" the issue, I'm afraid we're also going to mask other failures that would need to be addressed differently.

ogayot:
I've resurrected this PR since the CI broke on the SSH key import thingy. https://github.com/canonical/subiquity/actions/runs/3637249399/jobs/6138051033

I changed the default value of `not_found_ok` to False to avoid masking potential bugs. We can explicitly pass `not_found_ok=True` if we know the code is buggy ; just like I did for SSH

dbungert (approved):

PR #1327, opened 2022-06-21
filesystem: non-overlapping partition numbering

dbungert:
A basic implementation of partition numbering so that they do not overlap.
Uses the previously-discussed strategy of number in order of creation,
which does have some interesting implications for the implicitly created boot partition.

I have a local variant of number-by-offset which we can switch to if we decide this isn't great.

This makes no attempts to address the implications of msdos partition table numbering, so we'll need to do something similar there in a logical-aware way.


subiquity/models/filesystem.py
```diff
@@ -689,10 +689,16 @@ def serialize_number(self):
 
     @property
     def _number(self):
-        if self.preserve:
```
mwhudson, line 692: This seems a strange place to do this, so on the first access to a partitions _number attribute, all the partitions of the disk get assigned? Why not do it as the partitions are created? If there is a reason for this it should go into a comment :)

mwhudson (commented):
Seems OK, one question though.

subiquity/tests/api/test_api.py
```diff
@@ -374,8 +372,9 @@ async def test_v2_add_boot_partition(self):
 
             await inst.post('/storage/v2/reset')
 
-            # these manual steps are expected to be equivalent to just adding
-            # the single partition and getting the automatic boot partition
+            # these manual steps are expected to be mostly equivalent to just
+            # adding the single partition and getting the automatic boot
+            # partition
```
mwhudson, line 377: It is a bit strange that the boot partition ends up being partition 2. I think we could fix this if we wanted by having v2_add_partition_POST explicitly call add_boot_disk if it needs to, the way partition_disk_handler does. But I guess that may make the gap passed to v2_add_partition_POST no longer valid so blah. Maybe we see if anyone notices :)

mwhudson (approved):
I'm not sure the partition numbering we've settled on for the spec and this implementation is the most elegant but it should work at least and we can fix it for elegance if anyone other than me notices :-)

subiquity/models/tests/test_filesystem.py
```diff
@@ -177,7 +177,11 @@ def make_partition(model, device=None, *, preserve=False, size=None,
     partition = Partition(m=model, device=device, size=size, offset=offset,
                           preserve=preserve, **kw)
     if preserve:
-        partition.number = len(device._partitions)
+        number = kw.get('number')
+        if number is not None:
+            partition.number = number
+        else:
+            partition.number = len(device._partitions)
```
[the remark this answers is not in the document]
dbungert: Nope!

PR #1331, opened 2022-06-22
ubuntu-pro: update screens to match design mockups

ogayot:
Hello,

This PR updates the different Ubuntu Pro screens to match the new mockups.
I deliberately did not include any of the related magic-attach related screens ; because I'm still working on that part.
The presence of the radio button in the screen asking to input the contract token can look stupid because there is no other radio button.
When magic-attach comes in, the user will either choose to manually input the token or to supply a Ubuntu One email address ; so the radio button will make sense when we get there. I could try to hide the radio button for now but the mechanism we have to manage subforms is a bit of an adventure ; so I'd prefer to keep it.

I've uploaded screen captures at https://people.canonical.com/~ogayot/subiquity-captures

Besides purely visual changes, the PR also includes:
* format checking for a contract token
* a way to prevent the user from supplying a second time a token which was already considered expired or invalid
* a way to submit the token asynchronously if check_token is successful
* the use of the new auto-enabled flag for Ubuntu Pro services

Thanks,
Olivier

### DIFF
--- a/subiquity/client/controllers/ubuntu_pro.py
+++ b/subiquity/client/controllers/ubuntu_pro.py
@@ -70,6 +70,7 @@ class UbuntuProController(SubiquityTuiController):
         async def inner() -> None:
             answer = await self.endpoint.check_token.GET(token)
             if answer.status == TokenStatus.VALID_TOKEN:
+                await self.endpoint.POST(UbuntuProInfo(token=token))
                 on_success(answer.subscription.services)
             else:
                 on_failure(answer.status)
@@ -85,6 +86,12 @@ class UbuntuProController(SubiquityTuiController):
         self.app.prev_screen()
 
     def done(self, token: str) -> None:
+        """ Submit the token and move on to the next screen. """
         self.app.next_screen(
             self.endpoint.POST(UbuntuProInfo(token=token))
         )
+
+    def next_screen(self) -> None:
+        """ Move on to the next screen. Assume the token should not be
+        submitted (or has already been submitted). """
+        self.app.next_screen()
--- a/subiquity/client/controllers/ubuntu_pro.py
+++ b/subiquity/client/controllers/ubuntu_pro.py
@@ -18,6 +18,8 @@ import asyncio
 import logging
 from typing import Callable, List, Optional
 
+from urwid import Widget
+
 from subiquitycore.async_helpers import schedule_task
 
 from subiquity.client.controller import SubiquityTuiController
@@ -26,7 +28,12 @@ from subiquity.common.types import (
     UbuntuProCheckTokenStatus as TokenStatus,
     UbuntuProService,
     )
-from subiquity.ui.views.ubuntu_pro import UbuntuProView
+from subiquity.ui.views.ubuntu_pro import (
+    UbuntuProView,
+    UpgradeYesNoForm,
+    UpgradeModeForm,
+    TokenAddedWidget,
+    )
 
 from subiquitycore.lsb_release import lsb_release
 from subiquitycore.tuicontroller import Skip
@@ -58,9 +65,66 @@ class UbuntuProController(SubiquityTuiController):
         ubuntu_pro_info = await self.endpoint.GET()
         return UbuntuProView(self, ubuntu_pro_info.token)
 
-    def run_answers(self) -> None:
-        if "token" in self.answers:
-            self.done(self.answers["token"])
+    async def run_answers(self) -> None:
+        """ Interact with the UI to go through the pre-attach process if
+        requested. """
+        if "token" not in self.answers:
+            return
+
+        from subiquitycore.testing.view_helpers import (
+            click,
+            enter_data,
+            find_button_matching,
+            find_with_pred,
+            keypress,
+        )
+
+        view = self.app.ui.body
+
+        def run_yes_no_screen(skip: bool) -> None:
+            if skip:
+                radio = view.upgrade_yes_no_form.skip.widget
+            else:
+                radio = view.upgrade_yes_no_form.upgrade.widget
+
+            keypress(radio, key="enter")
+            click(find_button_matching(view, UpgradeYesNoForm.ok_label))
+
+        def run_token_screen(token: str) -> None:
+            keypress(view.upgrade_mode_form.with_contract_token.widget,
+                     key="enter")
+            data = {"with_contract_token_subform": {"token": token}}
+            # TODO: add this point, it would be good to trigger the validation
+            # code for the token field.
+            enter_data(view.upgrade_mode_form, data)
+            click(find_button_matching(view, UpgradeModeForm.ok_label))
+
+        async def run_token_added_overlay() -> None:
+            def is_token_added_overlay(widget: Widget) -> bool:
+                try:
+                    if widget._text == f" {TokenAddedWidget.title} ":
+                        return True
+                except AttributeError:
+                    return False
+
+            # Wait until the "Token added successfully" overlay is shown.
+            while not find_with_pred(view, is_token_added_overlay):
+                await asyncio.sleep(.2)
+
+            click(find_button_matching(view, TokenAddedWidget.done_label))
+
+        def run_services_screen() -> None:
+            click(find_button_matching(view._w,
+                                       UbuntuProView.services_done_label))
+
+        if not self.answers["token"]:
+            run_yes_no_screen(skip=True)
+            return
+
+        run_yes_no_screen(skip=False)
+        run_token_screen(self.answers["token"])
+        await run_token_added_overlay()
+        run_services_screen()
 
     def check_token(self, token: str,
                     on_success: Callable[[List[UbuntuProService]], None],
--- a/subiquity/ui/views/ubuntu_pro.py
+++ b/subiquity/ui/views/ubuntu_pro.py
@@ -308,6 +308,81 @@ class UbuntuProView(BaseView):
                 excerpt=excerpt,
                 focus_buttons=True)
 
+    def services_screen(self, services) -> Widget:
+        """
+        +---------------------------------------------------------+
+        | List of your enabled services:                          |
+        |                                                         |
+        |   * ...                                                 |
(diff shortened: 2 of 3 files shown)


subiquity/client/controllers/ubuntu_pro.py
```diff
@@ -85,6 +86,12 @@ def cancel(self) -> None:
         self.app.prev_screen()
 
     def done(self, token: str) -> None:
+        """ Submit the token and move on to the next screen. """
         self.app.next_screen(
             self.endpoint.POST(UbuntuProInfo(token=token))
         )
+
+    def next_screen(self) -> None:
```
dbungert, line 94: This is a bit unusual as the client controllers go.  Can you elaborate?  Did this used to have an extra step before app.next_screen()?

subiquity/ui/views/ubuntu_pro.py
```diff
@@ -206,47 +461,207 @@ def show_unknown_error(self) -> None:
...
+
+        ok = ok_btn(label=_("Continue"), on_press=lambda unused: self.close())
+
+        title = _("About Ubuntu Pro")
+        header = _("Ubuntu Pro subscription gives you access to multiple"
+                   " security & compliance features such as:")
+
+        services = [
+            _("Security patching for over 30.000 packages, with a focus on"
+              " High and Critical CVEs (extended from 2.500)"),
```
dbungert, line 587: The indent here between bullet points when we need a second line for a single bullet point doesn't quite match mockups.   I appreciate that this is probably non-trivial.  Thoughts?

subiquity/ui/views/ubuntu_pro.py
```diff
@@ -81,23 +82,94 @@ def valid_char(self, ch: str) -> bool:
...
+    ok_label = _("Continue")
+    group: List[RadioButtonField] = []
 
-    token = UATokenField(_("Ubuntu Pro token:"), help=ua_help)
+    upgrade = RadioButtonField(
+            group, _("Upgrade to Ubuntu Pro"),
+            help=NO_HELP)
+    skip = RadioButtonField(
+            group, _("Do this later"),
+            help=_("You can always enable Ubuntu Pro later via the"
```
dbungert, line 168: Check the indent on this line.  It's supposed to be down another line, and over.

subiquity/ui/views/ubuntu_pro.py
```diff
@@ -125,76 +197,259 @@ def cancel(self, sender) -> None:
...
+        |                        [ Continue ]                     |
+        |                        [ Back     ]                     |
+        +---------------------------------------------------------+
+        """
+
+        excerpt = _("To upgrade to Ubuntu Pro, you can enter your token"
+                    " manually.")
+
+        how_to_register_btn = menu_btn(
+                _("How to Register"),
```
dbungert, line 257: Lowercase 'r' :)

subiquity/ui/views/ubuntu_pro.py
```diff
@@ -125,76 +197,259 @@ def cancel(self, sender) -> None:
...
+        |                       [ Back     ]                      |
+        +---------------------------------------------------------+
+        """
+        auto_enabled = [svc for svc in services if svc.auto_enabled]
+        not_auto_enabled = [svc for svc in services if not svc.auto_enabled]
+
+        svc_rows: List[Widget] = []
+
+        if auto_enabled:
+            svc_rows.append(Text(_("List of your enabled services:")))
```
dbungert, line 336: check the indent here also

dbungert (changes requested):
If the screens are done-ish we should get some answers files going.
I've got a few items listed here, if they need a new PR let's keep the comments open until those are all responded to.

subiquity/ui/views/ubuntu_pro.py
ogayot: I found a solution to achieve that using a `Columns` construct. If the bullet point + text with indent is something we need for other screens going forward, I can move it somewhere under `subiquitycore.ui`.

ogayot: Screenshot updated.

subiquity/client/controllers/ubuntu_pro.py
ogayot: So, the reason why I added this is because I wanted to submit the contract token to the server _before_ showing the "Token added successfully" popup (and also before showing the screen with the list of services)

so we:
1. submit the token through POST /ubuntu_pro/ (this marks the controller configured)
2. display the "token added successfully" popup
3. display the screen with the list of services

if we implemented step 1. with a simple call to `controller.done()` (as most sets of screens do), we would directly move on to the SSH screen ; without giving us a chance to do step 2. and 3.

in the previous implementation, we would wait until the user clicks OK in the list of services popup to submit the token:

ok

But this popup had no "Back" button back then.
Now, we want the settings to be saved even if the users decides to go back ; so that they don't start over with an empty token.

back

Let me know if that does not make sense. I can show examples.

subiquity/ui/views/ubuntu_pro.py
```diff
@@ -206,47 +461,207 @@ def show_unknown_error(self) -> None:
...
+
+        title = _("How to Register")
+        header = _("You can register for a free Ubuntu One account and get a"
+                   " personal token for up to 3 machines.")
+
+        widgets: List[Widget] = [
+            Text(header),
+            Text(""),
+            Text("To register an account, visit ubuntu.com/pro on another"
+                 " device"),
```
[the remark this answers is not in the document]
ogayot: Thanks!

subiquity/ui/views/ubuntu_pro.py
ogayot: Ah, good catch! :eye: 

According to the mockups, this one should be uppercase ; but the one in the "How to register" screen title should be lowercase.

I will fix it now and ask Elio to confirm is this is what we want.

ogayot: Screenshot updated

subiquity/ui/views/ubuntu_pro.py
ogayot: Thanks, screenshot updated.

subiquity/ui/views/ubuntu_pro.py
dbungert: Nice!

subiquity/ui/views/ubuntu_pro.py
dbungert: Ah yes, I was looking at the dialog title.
Good idea to ask about the ambiguity.

subiquity/client/controllers/ubuntu_pro.py
```diff
@@ -70,6 +70,7 @@ def check_token(self, token: str,
         async def inner() -> None:
             answer = await self.endpoint.check_token.GET(token)
             if answer.status == TokenStatus.VALID_TOKEN:
+                await self.endpoint.POST(UbuntuProInfo(token=token))
```
mwhudson, line 73: Doesn't this mean that if the user enables pro and then goes back and selects "do this later" they will still have pro enabled? That doesn't seem quite right.

ogayot: No, because it would go through this code again:

```python
    def upgrade_yes_no_done(self, form: UpgradeYesNoForm) -> None:
        """ If skip is selected, move on to the next screen.
        Otherwise, show the form requesting the contract token. """
        if form.skip.value:
            self.controller.done("")
```
which in turn would do:
```python
 await self.endpoint.POST(UbuntuProInfo(token=""))
```

but if this isn't clear, I can try to refactor again.

ogayot:
Added more interaction with the GUI in the run_answers method for Ubuntu Pro.
Supplying an invalid or expired token through answers _will_ block the GUI since it will refuse to move forward with an invalid token.
If we want to cover more scenarios, maybe we should add support for a fallback token in the answers. The fallback token would be used if the original token was considered invalid or expired:

```yaml
ubuntu-pro:
    token: xpired
    fallback-token: C1NWcZTHLteJXGVMM6YhvHDpGrhyy7
```

any other idea?

subiquity/ui/views/ubuntu_pro.py
ogayot: I moved this line down for now. Changing the indent seems challenging without leaning on dirty hacks.

dbungert (approved):

ogayot:
> > Supplying an invalid or expired token through answers _will_ block the GUI since it will refuse to move forward with an invalid token.
> 
> If you have an answers file for the Success path, let's include that, otherwise let's just move on and revisit in later PRs.

Sounds good to me, thanks!

These files go through the scenario where the user skips Ubuntu Pro:

- examples/answers-bond.yaml:  token: ""
- examples/answers-preserve.yaml:  token: ""
- examples/answers-serial.yaml:  token: ""
- examples/answers.yaml:  token: ""

And these files cover the scenario where the user supplies a token and succeeds:

- examples/answers-guided-lvm.yaml:  token: C1NWcZTHLteJXGVMM6YhvHDpGrhyy7
- examples/answers-imsm.yaml:  token: C1NWcZTHLteJXGVMM6YhvHDpGrhyy7
- examples/answers-lvm-dmcrypt.yaml:  token: C1NWcZTHLteJXGVMM6YhvHDpGrhyy7
- examples/answers-lvm.yaml:  token: C1NWcZTHLteJXGVMM6YhvHDpGrhyy7
- examples/answers-raid-lvm.yaml:  token: C1NWcZTHLteJXGVMM6YhvHDpGrhyy7
- examples/answers-raid.yaml:  token: C1NWcZTHLteJXGVMM6YhvHDpGrhyy7
- examples/answers-swap.yaml:  token: C1NWcZTHLteJXGVMM6YhvHDpGrhyy7
- examples/autoinstall.yaml:  token: C1NWcZTHLteJXGVMM6YhvHDpGrhyy7

ogayot:
Rebased on top of main to resolve conflict.

PR #1334, opened 2022-06-23
identity: add leading capital letter to error messages

ogayot:
Marking a draft because something weird happens.

In the LVM form, the error message should be "Passphrases do not match" but instead it is "Passwords do not match". It seems that the validator used is the one from IdentityForm, for some reason.

### DIFF
--- a/subiquity/ui/views/filesystem/lvm.py
+++ b/subiquity/ui/views/filesystem/lvm.py
@@ -83,7 +83,7 @@ class VolGroupForm(CompoundDiskForm):
         self.deleted_vg_names = deleted_vg_names
         super().__init__(model, possible_components, initial)
         connect_signal(self.encrypt.widget, 'change', self._change_encrypt)
-        setup_password_validation(self, _("passphrases"))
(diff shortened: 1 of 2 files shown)


subiquity/ui/views/filesystem/guided.py
```diff
@@ -64,11 +64,11 @@ class LUKSOptionsForm(SubForm):
 
     def validate_password(self):
         if len(self.password.value) < 1:
-            return _("Password must be set")
+            return _("Passphrase must be set")
```
dbungert, line 67: That was mighty confusing in isolation but makes sense in the larger file.
A future refactor to s/password/passphrase would be welcome!

dbungert (approved):

subiquity/ui/views/filesystem/guided.py
ogayot: Yes, the UI got me confused and by fixing it, I made the code confusing :upside_down_face:.
That was not so clever, I'll refactor in a follow on PR.

PR #1335, opened 2022-06-23
tests: s/run_coro/IsolatedAsyncioTestCase/

dbungert:
Tests that use run_coro are at risk of being broken by the introduction
of another test using IsolatedAsyncioTestCase.  Switch over to only use
IsolatedAsyncioTestCase.


dbungert:
(`git diff -b` is moderately easier to read)

dbungert:
Bonus - using `asyncio.get_event_loop` to create an event loop is deprecated as of py3.10, so this removes many instances of that usage.

ogayot (approved):
Thanks! I was always worried of unexpected behavior when using the `unittest.TestCase` + `run_coro` construct.
It looks way cleaner and more robust now :+1:

CarlosNihelton (approved):
I had personal experience with the fraility of TestCase + run_coro :smiling_face_with_tear: 
Thanks for doing this! Great job!

mwhudson (approved):

PR #1336, opened 2022-06-24
Refactor how we handle confirmation fields - use passphrase instead of password where relevant

ogayot:
The identity screen implements a mechanism to make sure that the confirmation password is the same as the original password. This implementation leans on hard-coded field names (i.e., "password" and "password_confirm") which is perfectly understandable since this is what we have in the identity screen.

Nevertheless, in the storage screen, we decided to import this mechanism and use it as-is to compare passphrases. The implementation prevents us from using different field names such as "passphrase" and "passphrase_confirm" ; which makes the code confusing.

To make things more developer-friendly (is that a thing?), I exported the confirmation mechanism from the identity screen to `subiquitycore.ui` and made it a method of `BoundFormField`.

Before:
```python
class FormWithPassphrase(Form):
    # XXX we are dealing with passphrases but setup_password_validation forces us to name fields
    # "password" and "confirm_passwords"
    password: PasswordField("Passphrase:")
    confirm_password: PasswordField("Passphrase confirmation:")


from subiquity.ui.views.identity import setup_password_validation
form = FormWithPassphrase(...)
setup_password_validation(form, desc=_("Passphrases"))
```

After:
```python
class FormWithPassphrase(Form):
    passphrase: PasswordField("Passphrase:")
    confirm_passphrase: PasswordField("Passphrase confirmation:")


form = FormWithPassphrase(...)
form.confirm_passphrase.use_as_confirmation(for_field=form.passphrase, desc=_("Passphrases"))
```

This removes the coupling between identity and storage ; which is a good thing.
It also makes the confirmation mechanism usable without using fields with hard-coded names. Yay! Maybe we might use it for email + confirmation email some day?

Where applicable, I've replaced the name of the values from "password" to "passphrase" but only on the client side ; so it does not break the backward compatibility of the API.

### DIFF
--- a/examples/answers-lvm-dmcrypt.yaml
+++ b/examples/answers-lvm-dmcrypt.yaml
@@ -40,8 +40,8 @@ Filesystem:
           - [disk index 0, part 1]
           - [disk index 0, part 2]
         encrypt: true
-        password: "passw0rd"
-        confirm_password: "passw0rd"
+        passphrase: "passw0rd"
+        confirm_passphrase: "passw0rd"
     - obj: [volgroup name vg-1]
       action: EDIT
       data:
@@ -50,8 +50,8 @@ Filesystem:
       action: EDIT
       data:
         encrypt: true
-        password: "passw0rd"
-        confirm_password: "passw0rd"
+        passphrase: "passw0rd"
+        confirm_passphrase: "passw0rd"
     - obj: [volgroup name vg-1]
       action: CREATE_LV
       data:
--- a/subiquity/client/controllers/filesystem.py
+++ b/subiquity/client/controllers/filesystem.py
@@ -114,8 +114,8 @@ class FilesystemController(SubiquityTuiController, FilesystemManipulator):
                 value['lvm_options'] = {
                     'encrypt': True,
                     'luks_options': {
-                        'password': passphrase,
-                        'confirm_password': passphrase,
+                        'passphrase': passphrase,
+                        'confirm_passphrase': passphrase,
                         }
                     }
             self.ui.body.form.guided_choice.value = value
--- a/subiquity/ui/views/filesystem/guided.py
+++ b/subiquity/ui/views/filesystem/guided.py
@@ -59,15 +59,15 @@ subtitle = _("Configure a guided storage layout, or create a custom one:")
 
 class LUKSOptionsForm(SubForm):
 
-    password = PasswordField(_("Passphrase:"))
-    confirm_password = PasswordField(_("Confirm passphrase:"))
+    passphrase = PasswordField(_("Passphrase:"))
+    confirm_passphrase = PasswordField(_("Confirm passphrase:"))
 
-    def validate_password(self):
-        if len(self.password.value) < 1:
+    def validate_passphrase(self):
+        if len(self.passphrase.value) < 1:
             return _("Passphrase must be set")
 
-    def validate_confirm_password(self):
-        if self.password.value != self.confirm_password.value:
+    def validate_confirm_passphrase(self):
+        if self.passphrase.value != self.confirm_passphrase.value:
             return _("Passphrases do not match")
 
 
@@ -185,7 +185,7 @@ can easily be enlarged with standard LVM command line tools (or on the
 next screen).
 
 You can also choose to encrypt LVM volume group. This will require
-setting a password, that one will need to type on every boot before
+setting a passphrase, that one will need to type on every boot before
 the system boots.
 
 If you do not choose to use LVM, a single partition is created covering the
@@ -252,7 +252,7 @@ class GuidedDiskSelectionView(BaseView):
                 use_lvm=results['guided_choice']['use_lvm'])
             opts = results['guided_choice'].get('lvm_options', {})
             if opts.get('encrypt', False):
-                choice.password = opts['luks_options']['password']
+                choice.password = opts['luks_options']['passphrase']
         self.controller.guided_choice(choice)
 
     def manual(self, sender):
--- a/subiquity/ui/views/filesystem/lvm.py
+++ b/subiquity/ui/views/filesystem/lvm.py
@@ -48,9 +48,6 @@ from subiquity.ui.views.filesystem.compound import (
     get_possible_components,
     MultiDeviceField,
     )
-from subiquity.ui.views.identity import (
-    setup_password_validation,
-    )
 
 log = logging.getLogger('subiquity.ui.views.filesystem.lvm')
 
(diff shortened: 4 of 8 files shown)


CarlosNihelton (approved):
LGTM!

ogayot:
Fixed wrong behavior caused by unused `for_field` variable. The variable was meant to be used of course.
Thanks @dbungert for noticing~

dbungert (approved):

mwhudson (approved):
> It also makes the confirmation mechanism usable without using fields with hard-coded names. Yay! Maybe we might use it for email + confirmation email some day?

https://xkcd.com/970/

> ```python
> form.confirm_passphrase.use_as_confirmation(for_field=form.passphrase, desc=_("Passphrases"))
> ```

It would be nice if we could come up with some more declarative way of doing this but well. It would mean introducing a hook that gets called on a BoundFormField after all the other fields are set up I think? (Could (ab)use the fact that validate gets called just after all the fields get bound but well. Let's not do that!)

ogayot:
> > It also makes the confirmation mechanism usable without using fields with hard-coded names. Yay! Maybe we might use it for email + confirmation email some day?
> 
> https://xkcd.com/970/
> 
:laughing: 

> > ```python
> > form.confirm_passphrase.use_as_confirmation(for_field=form.passphrase, desc=_("Passphrases"))
> > ```
> 
> It would be nice if we could come up with some more declarative way of doing this but well. It would mean introducing a hook that gets called on a BoundFormField after all the other fields are set up I think? (Could (ab)use the fact that validate gets called just after all the fields get bound but well. Let's not do that!)

I'd like to see something like the below

```python
class FormWithPassphrase(Form):
    passphrase: PasswordField("Passphrase:")
    confirm_passphrase: PasswordField("Passphrase confirmation:", confirms="passphrase")
    # where "passphrase" is the name of the other (bound) field.
```
but adding the name of another field in the initializer of `FormField` (or `PasswordField`) feels wrong to me. After all, a `FormField` instance does not have any reference to the parent form. It does not even know its own name.

Adding a `confirm: FormField` (or `confirm: PasswordField`) could make sense but using class attributes makes this challenging ; especially because the fields are replaced by boundfields in the Form's initializer.

any other idea?

mwhudson:
I guess I agree that having it as an argument to the field doesn't really make sense (although you can pass the field itself, you don't have to pass its name?). This reminds me of that other PR you made where I wondered aloud about better ways to declare relationships between fields. Something like:

```
class MyForm(Form):
    choice = ChoiceField(['a', 'b', 'other'])
    other = TextField()
    passphrase: PasswordField("Passphrase:")
    confirm_passphrase: PasswordField("Passphrase confirmation:")

    relationships = [
        MustMatch(passphrase, confirm_passphrase),
        EnabledOnValue(choice, 'other', other),
    ]
```

But I don't know if this would be Too Much Magic. Might be worth a 30 minute play to see how it feels?

ogayot:
> I guess I agree that having it as an argument to the field doesn't really make sense (although you can pass the field itself, you don't have to pass its name?).

TIL that class attributes can refer to other class attributes as long as they are defined before. :+1: 

> This reminds me of that other PR you made where I wondered aloud about better ways to declare relationships between fields. Something like:
> 
> ```
> class MyForm(Form):
>     choice = ChoiceField(['a', 'b', 'other'])
>     other = TextField()
>     passphrase: PasswordField("Passphrase:")
>     confirm_passphrase: PasswordField("Passphrase confirmation:")
> 
>     relationships = [
>         MustMatch(passphrase, confirm_passphrase),
>         EnabledOnValue(choice, 'other', other),
>     ]
> ```
> 
> But I don't know if this would be Too Much Magic. Might be worth a 30 minute play to see how it feels?

yep, I will do some experiments.

ogayot:
> > I guess I agree that having it as an argument to the field doesn't really make sense (although you can pass the field itself, you don't have to pass its name?). 
> > This reminds me of that other PR you made where I wondered aloud about better ways to declare relationships between fields. Something like:
> > ```
> > class MyForm(Form):
> >     choice = ChoiceField(['a', 'b', 'other'])
> >     other = TextField()
> >     passphrase: PasswordField("Passphrase:")
> >     confirm_passphrase: PasswordField("Passphrase confirmation:")
> > 
> >     relationships = [
> >         MustMatch(passphrase, confirm_passphrase),
> >         EnabledOnValue(choice, 'other', other),

> > But I don't know if this would be Too Much Magic. Might be worth a 30 minute play to see how it feels?

Tried this and although this needs more thinking to implement properly ; I think the approach is better than what we have currently. The use of radio buttons to enable/disable subforms make things unnecessarily complex though. I wonder if we should try to address this part first.

Coming back to the main topic, should we merge this PR as is or should we try the following approach first?
```python
class FormWithPassphrase(Form):
    passphrase: PasswordField("Passphrase:")
    confirm_passphrase: PasswordField("Passphrase confirmation:", confirms=passphrase)
```

mwhudson:
> > > I guess I agree that having it as an argument to the field doesn't really make sense (although you can pass the field itself, you don't have to pass its name?).
> > > This reminds me of that other PR you made where I wondered aloud about better ways to declare relationships between fields. Something like:
> > > ```
> > > class MyForm(Form):
> > >     choice = ChoiceField(['a', 'b', 'other'])
> > >     other = TextField()
> > >     passphrase: PasswordField("Passphrase:")
> > >     confirm_passphrase: PasswordField("Passphrase confirmation:")
> > > 
> > >     relationships = [
> > >         MustMatch(passphrase, confirm_passphrase),
> > >         EnabledOnValue(choice, 'other', other),
> > > ```
> 
> > > But I don't know if this would be Too Much Magic. Might be worth a 30 minute play to see how it feels?
> 
> Tried this and although this needs more thinking to implement properly ;

Thanks for trying :)

>  I think the approach is better than what we have currently.

Describing things like this with data rather than code almost always is, in my experience.

> The use of radio buttons to enable/disable subforms make things unnecessarily complex though. I wonder if we should try to address this part first.

Certainly that seems a fairly common pattern.

> Coming back to the main topic, should we merge this PR as is or should we try the following approach first?
> 
> ```python
> class FormWithPassphrase(Form):
>     passphrase: PasswordField("Passphrase:")
>     confirm_passphrase: PasswordField("Passphrase confirmation:", confirms=passphrase)
> ```

I don't know. Maybe that would be worth a try? I think what you have here is OK though.

ogayot:
I noticed this PR addresses a failure in integration tests when we attempt to fetch an SSH key but fail to contact github/launchpad. The SSH client controller does not open an overlay but it attempts to close an overlay on failure. 

Although in this case, the PR "solves" the issue, I'm afraid we're also going to mask other failures that would need to be addressed differently.

mwhudson:
> I noticed this PR addresses a failure in integration tests when we attempt to fetch an SSH key but fail to contact github/launchpad. The SSH client controller does not open an overlay but it attempts to close an overlay in case of failure.

Are you sure this is the correct PR for this comment?

ogayot:
> > I noticed this PR addresses a failure in integration tests when we attempt to fetch an SSH key but fail to contact github/launchpad. The SSH client controller does not open an overlay but it attempts to close an overlay in case of failure.
> 
> Are you sure this is the correct PR for this comment?

Indeed, wrong PR. My bad!
https://github.com/canonical/subiquity/pull/1324 is where I intended to comment.

PR #1337, opened 2022-06-27
sizes: calculations for install_min and resize

dbungert:
Add calculations for a suggested minimum install size that contains
padding for swap, updates, boot.
Add calculations for helping resize, to examine the space available and
make suggestions about what the existing partition might be resized to.
(not used yet)

### DIFF
--- a/subiquity/common/filesystem/sizes.py
+++ b/subiquity/common/filesystem/sizes.py
@@ -13,11 +13,19 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import math
+
 import attr
 
+from curtin import swap
+
 from subiquity.models.filesystem import (
+    align_up,
+    align_down,
     MiB,
+    GiB,
     )
+from subiquity.common.types import GuidedResizeValues
 
 
 BIOS_GRUB_SIZE_BYTES = 1 * MiB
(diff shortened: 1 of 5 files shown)


subiquity/common/filesystem/sizes.py
```diff
@@ -75,3 +82,34 @@ def get_efi_size(available_space):
 def get_bootfs_size(available_space):
     all_factors = (uefi_scale, bootfs_scale, rootfs_scale)
     return scale_partitions(all_factors, available_space)[1]
+
+
+def calculate_guided_resize(part_min: int, part_size: int, install_min: int,
+                            part_align: int = 1 << 20) -> GuidedResizeValues:
```
mwhudson, line 88: Can we spell 1<<20 as MiB and 1<<30 as GiB do you think?

subiquity/common/filesystem/tests/test_sizes.py
```diff
@@ -74,3 +77,70 @@ def test_efi(self):
...
+        actual = calculate_guided_resize(
+                part_min=95 << 30, part_size=100 << 30, install_min=10 << 30)
+        self.assertIsNone(actual)
+
+    def test_even_split(self):
+        # 8 GiB * 1.25 == 10 GiB
+        size = 10 << 30
+        actual = calculate_guided_resize(
+                part_min=8 << 30, part_size=100 << 30, install_min=size)
+        self.assertEqual(size, actual.minimum)
```
mwhudson, line 98: Did you think about writing these as a comparison of GuidedResizeValues instances?

subiquity/common/filesystem/sizes.py
```diff
@@ -75,3 +82,34 @@ def get_efi_size(available_space):
 def get_bootfs_size(available_space):
     all_factors = (uefi_scale, bootfs_scale, rootfs_scale)
     return scale_partitions(all_factors, available_space)[1]
+
+
+def calculate_guided_resize(part_min: int, part_size: int, install_min: int,
```
mwhudson, line 87: How will this API be used when a partition is followed by a gap?

subiquity/common/filesystem/sizes.py
```diff
@@ -75,3 +82,34 @@ def get_efi_size(available_space):
...
+                            part_align: int = 1 << 20) -> GuidedResizeValues:
+    if part_min < 0:
+        return None
+
```
mwhudson, line 91: Do you think you could paste in the description of what is going on here from the spec?

mwhudson (approved):
Hi a few comments but nothing too major.

subiquity/common/filesystem/sizes.py
dbungert: Ah, we should at least use a gap that immediately follows a resized partition.  I think that should be a follow-on PR to extend gaps a bit so that we can easily acquire a gap after a partition, if one exists.  Armed with that it should be easy to allow that to be factored in to guided resize.

subiquity/common/filesystem/sizes.py
dbungert: Done for the code, left tests alone.

subiquity/common/filesystem/tests/test_sizes.py
dbungert: Done

PR #1340, opened 2022-06-30
ubuntu-pro: add subscription info in response to /ubuntu_pro/check_token

ogayot:
The response to `/ubuntu_pro/check_token` now includes information about the subscription: the name of the contract and the "name" of the account.

Instead of returning the list of services as an optional field, we now include the list of services in the subscription object. The
subscription object is itself marked optional. This is a backward incompatible change. CC: @jpnurmi 
A backward compatible implementation is possible but it will lead to multiple optional keys. I think the backward incompatible approach is better here.

```json
  {
    "status": "VALID_TOKEN",
    "services": [
      {
        "name": "esm-infra",
        "description": "UA Infra: Extended Security Maintenance (ESM)",
        "auto_enabled": true
      }
    ]
  }
```
  =>
```json
  {
    "status": "VALID_TOKEN",
    "subscription": {
        "account_name": "user@domain.com",
        "contract_name": "UA Apps - Essential (Virtual)",
        "services": [
          {
            "name": "esm-infra",
            "description": "UA Infra: Extended Security Maintenance (ESM)",
            "auto_enabled": true
          }
        ]
     }
  }
```

If the token is not valid, the subscription object will be null:

```python
  {
    "status": "EXPIRED_TOKEN",
    "subscription": null
  }
```

### DIFF
--- a/examples/uaclient-status-expired.json
+++ b/examples/uaclient-status-expired.json
@@ -1,6 +1,12 @@
 {
     "version": "27.4.2~21.10.1",
     "effective": null,
+    "contract": {
+        "name": "UA Apps - Essential (Virtual)"
+    },
+    "account": {
+        "name": "user@domain.com"
+    },
     "expires": "2010-12-31T00:00:00+00:00",
     "services": [
         {
--- a/examples/uaclient-status-valid.json
+++ b/examples/uaclient-status-valid.json
@@ -2,6 +2,12 @@
     "version": "27.4.2~21.10.1",
     "effective": null,
     "expires": "2035-12-31T00:00:00+00:00",
+    "contract": {
+        "name": "UA Apps - Essential (Virtual)"
+    },
+    "account": {
+        "name": "user@domain.com"
+    },
     "services": [
         {
             "name": "cis",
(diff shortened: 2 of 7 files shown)


jpnurmi:
Thanks for the heads up! The Ubuntu Pro API is still not officially in use in the GUI so you don't need to worry about backward compatibility. :)

ogayot:
> Thanks for the heads up! The Ubuntu Pro API is still not officially in use in the GUI so you don't need to worry about backward compatibility. :)

Thanks! Noted.

dbungert (approved):

examples/uaclient-status-expired.json
```diff
@@ -1,6 +1,12 @@
 {
     "version": "27.4.2~21.10.1",
     "effective": null,
+    "contract": {
+        "name": "UA Apps - Essential (Virtual)"
+    },
+    "account": {
+        "name": "user@domain.com"
+    },
```
mwhudson, line 9: my quick play suggests that ua actually has more fields here. Is it simple for you to just regenerate these files?

mwhudson (approved):
Didn't read the code in depth, general idea seems fine!

examples/uaclient-status-expired.json
ogayot: Yes, we can regenerate the files for sure. ua has way more fields indeed :)
I originally generated this file using the "canonical - staff" contract token and I don't have a clear vision if any of the fields contains sensitive information. There's a bunch of contract / account IDs in there that I did not want to risk disclosing.

mwhudson: > I originally generated this file using the "canonical - staff" contract token and I don't have a clear vision if any of the fields contains sensitive information.

Ah yes that's a fair point, I don't know either!

PR #1341, opened 2022-07-05
filesystem: overhaul v2 guided

dbungert:
* Add support on v2 guided for providing and acting upon different scenarios
* Supports resize, use_gap, and reformat
* Extends boot handler to allow resizing a specific partition
* Existing v2 guided moved to deprecated endpoint


subiquity/common/filesystem/gaps.py
```diff
@@ -230,3 +230,14 @@ def at_offset(device, offset):
             if pg.offset == offset:
                 return pg
     return None
+
+
+def intersects(device, gap):
+    """Find the first gap that is a subset of the supplied gap"""
```
mwhudson, line 236: I have a naming quibble here. intersect to me returns a gap that has overlap with the supplied gap, not one that's contained in it. Maybe "def gap_within" or something?

subiquity/common/filesystem/tests/test_gaps.py
```diff
@@ -82,6 +82,46 @@ def test_two_gaps(self):
         self.assertEqual(g2, gaps.at_offset(d, 60 << 20))
 
 
+class TestInteresects(unittest.TestCase):
```
mwhudson, line 85: I'd point out the typo if I wasn't asking you to change the name of this function anyway.

subiquity/common/filesystem/tests/test_gaps.py
```diff
@@ -82,6 +82,46 @@ def test_two_gaps(self):
...
+        d = make_disk()
+        [gap] = gaps.parts_and_gaps(d)
+        self.assertEqual(gap, gaps.intersects(d, gap))
+
+    def test_front_used(self):
+        m, d = make_model_and_disk(size=100 << 20)
+        m.storage_version = 2
+        [gap] = gaps.parts_and_gaps(d)
+        make_partition(m, d, offset=0, size=20 << 20)
+        self.assertEqual(gaps.Gap(device=d, offset=20 << 20, size=79 << 20),
```
mwhudson, line 96: I wonder if these tests would be clearer if the expected value was the gap returned by parts_and_gaps rather than constructed by hand (not sure tbh).

subiquity/common/types.py
```diff
@@ -333,11 +333,69 @@ class StorageResponseV2:
...
+class GuidedStorageTargetUseGap:
+    disk_id: str
+    gap: Gap
+
+
+@attr.s(auto_attribs=True)
+class GuidedChoiceV2:
+    target: Union[GuidedStorageTargetReformat,
+                  GuidedStorageTargetResize,
+                  GuidedStorageTargetUseGap]
```
mwhudson, line 378: I think I had expected that Union[GuidedStorageTargetReformat, GuidedStorageTargetResize,  GuidedStorageTargetUseGap] would be called GuidedStorageTarget, but I guess it's not really important.

subiquity/common/apidef.py
```diff
@@ -253,12 +255,18 @@ class has_bitlocker:
             def GET() -> List[Disk]: ...
 
         class v2:
+            class deprecated:
```
mwhudson, line 258: I think the fact that we have /storage/guided /storage/v2/guided and /storage/v2/deprecated/guided is getting a little silly now... hopefully we can clean all this up soon.

subiquity/server/controllers/filesystem.py
```diff
@@ -335,11 +386,62 @@ async def v2_reset_POST(self) -> StorageResponseV2:
...
+
+        scenarios = []
+        install_min = self.calculate_suggested_install_min()
+
+        for disk in self.get_guided_disks(with_reformatting=True):
+            reformat = GuidedStorageTargetReformat(disk_id=disk.id)
+            scenarios.append((disk.size, reformat))
+
+        for disk in self.get_guided_disks(with_reformatting=False):
+            if len(disk.partitions()) < 1:
```
mwhudson, line 408: Can you explain why you are doing this with a comment? I think I see but it would be nice to be clear.

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -127,3 +137,154 @@ def test_good_modes(self, mode):
     def test_bad_modes(self, mode):
         with self.assertRaises(ValueError):
             self.fsc.validate_layout_mode(mode)
+
+
+bootloaders = [(bl, ) for bl in list(Bootloader) if bl != Bootloader.NONE]
```
mwhudson, line 142: Eh. Some of these tests should run in the NONE case surely?

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -127,3 +137,154 @@ def test_good_modes(self, mode):
...
+
+
+class TestGuidedV2(IsolatedAsyncioTestCase):
```
mwhudson, line 145: I realize it would be painful to write but shouldn't this test the POST method as well? It feels like we should post all the options back and check that the model things it's configured at least?

subiquity/tests/api/test_api.py
```diff
@@ -247,7 +247,7 @@ async def test_server_flow(self):
             resp = await inst.get('/storage/guided')
             disk_id = resp['disks'][0]['id']
             choice = {"disk_id": disk_id}
-            await inst.post('/storage/v2/guided', choice)
+            await inst.post('/storage/v2/deprecated/guided', choice)
```
mwhudson, line 250: I presume you'll write some api tests for the new API at some point?

mwhudson (commented):
Mostly this looks good. It is a pretty large PR and I won't complain if you split it up, but well I've read it all now. Several comments but nothing major.

Getting close to having this line of work all done I think? That's pretty exciting.

subiquity/common/apidef.py
dbungert: @jpnurmi - is the deprecated v2 guided method still useful?  perhaps after you do the merge we can drop this?

jpnurmi: the deprecated API can be removed right away but a separate PR would be much appreciated to let us update subiquity and redo the guided partitioning UI in separate steps.

subiquity/tests/api/test_api.py
dbungert: Thanks, I meant to do that.
See new commits.
I lean a little bit too much on the API tests so I made a point of doing similarish things in the unit test layer, and I'm happy with the result, but a few at the API layer are a good idea.

subiquity/server/controllers/tests/test_filesystem.py
dbungert: I had a few already, so adding another isn't too painful.
See new commits.

subiquity/server/controllers/filesystem.py
dbungert: Done, see new commits

subiquity/common/types.py
dbungert: Done, see new commits

subiquity/common/filesystem/tests/test_gaps.py
dbungert: I worry a bit about relying on in test the same methods used in the implementation, but it does look nice.
See new commits.

subiquity/common/filesystem/tests/test_gaps.py
dbungert: Done

subiquity/common/filesystem/gaps.py
dbungert: Went with gaps.within, since most things access these methods as gaps.X.
See new commits.

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -139,152 +138,192 @@ def test_bad_modes(self, mode):
             self.fsc.validate_layout_mode(mode)
 
 
-bootloaders = [(bl, ) for bl in list(Bootloader) if bl != Bootloader.NONE]
+bootloaders_and_ptables = [(bl, pt)
+                           for bl in list(Bootloader) if bl != Bootloader.NONE
+                           for pt in ('gpt', 'msdos', 'vtoc')]
```
mwhudson, line 143: Oh yikes, there is no v2 partitioning support for vtoc. That needs fixing!

subiquity/server/controllers/tests/test_filesystem.py
mwhudson: Uh right. Well guided partitioning should certainly work in the bl==None case! I guess can_be_boot_disk should always return True in this case and add_boot_disk should be a noop? (either by checks in those functions or get_boot_part_plan returning some kind of noop plan in the bl==None case?)

subiquity/server/controllers/tests/test_filesystem.py
dbungert: Will investigate in follow-on PR

subiquity/server/controllers/tests/test_filesystem.py
dbungert: Added `NoOpBootPlan`

mwhudson (approved):
Looks good. I think I should make you split things up a bit more next time though!

PR #1342, opened 2022-07-06
ubuntu-advantage: determine on error if token was invalid

ogayot:
In case of error, u-a-c exits with status 1. When the contract token supplied is invalid, it is also considered an error ; which previously made us unable to make the distinction between:
 * an invalid token
 * a more generic error such as network error or service unavailable error
    
Thanks to an update in u-a-c, we can now determine if the token was invalid, by parsing the standard output of the process, even when it exits with status 1.
    
The output is expected to be a JSON object that includes an array of errors where each error has a specific error code.
    
The error code for an invalid token is `attach-invalid-token` ; which we now look for in the output to determine if the contract token was invalid.

### DIFF
--- a/subiquity/server/ubuntu_advantage.py
+++ b/subiquity/server/ubuntu_advantage.py
@@ -17,9 +17,10 @@ helper. """
 
 from abc import ABC, abstractmethod
 from datetime import datetime as dt
+import contextlib
 import json
 import logging
-from subprocess import CalledProcessError, CompletedProcess
+from subprocess import CompletedProcess
 from typing import List, Sequence, Union
 import asyncio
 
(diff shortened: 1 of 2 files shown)


subiquity/server/ubuntu_advantage.py
```diff
@@ -132,20 +133,31 @@ async def query_info(self, token: str) -> dict:
...
+                log.exception("Failed to parse output of command %r", command)
+        elif proc.returncode == 1:
+            try:
+                data = json.loads(proc.stdout)
+            except json.JSONDecodeError:
+                log.exception("Failed to parse output of command %r", command)
+            else:
+                with contextlib.suppress(KeyError):
+                    for error in data["errors"]:
+                        if error["message_code"] == "attach-invalid-token":
```
dbungert, line 156: Is there anything else interesting here?  Would be nice to log.debug() it.

ogayot:
```
2022-07-06 14:40:57,301 DEBUG subiquitycore.utils:101 arun_command ('/usr/bin/ubuntu-advantage', 'status', '--format', 'json', '--simulate-with-token', 'C32454555555555555555555') exited with code 1
2022-07-06 14:40:57,301 DEBUG subiquity.server.ubuntu_advantage:159 error reported by u-a-c: attach-invalid-token: Invalid token. See https://ubuntu.com/advantage
```

dbungert: Thanks!

dbungert (approved):

renanrodrigo:
As long as you are using the `ua status` command, this works.
We are implementing the `simulate` as part of the API transition (changing the command to use `ua api`), and the main change is that there will be a more coherent error code.

PR #1344, opened 2022-07-08
filesystem: fix use of f-string without placeholder

ogayot:
When running `make lint` with flake8 >= 4.0.0 (available in jammy), the following error is reported:

`subiquity/server/controllers/filesystem.py:158:33:F541 f-string is missing placeholders`

Github runs the lint job in a focal container, where only version 3.8.4 of flake8 is installed. F541 is not implemented in 3.8.4 so it does not fail.


dbungert (approved):

mwhudson:
> Github runs the lint job in a focal container

Maybe we should change that?

PR #1345, opened 2022-07-08
workflow: +kinetic for linter

dbungert:
Add a second linter container to also lint against the newer python.


ogayot (approved):
Thanks!

mwhudson:
Ah haha

PR #1346, opened 2022-07-11
guided resize: improve post-resize gap handling

dbungert:
The example machine config here exposed this issue, where due to manual gap
offset calculation we fail to find the gap due to the 1MiB spacer in between
logical partitions.

Skip that calculation and just get the next gap.


subiquity/server/controllers/filesystem.py
```diff
@@ -223,11 +223,13 @@ def guided(self, choice: GuidedChoiceV2):
             new_size = align_up(choice.target.new_size, part_align)
             if new_size > partition.size:
                 raise Exception(f'Aligned requested size {new_size} too large')
-            gap_offset = partition.offset + new_size
             partition.size = new_size
             partition.resize = True
             mode = 'use_gap'
-            target = gaps.at_offset(disk, gap_offset)
+            target = gaps.after(disk, partition.offset)
```
mwhudson, line 229: I think a comment here about why you are doing this would be a good idea.

.... although does this mean we are creating an ESP that is a logical partition? I think that doesn't work with our efi tools, unfortunately.

mwhudson: https://bugs.launchpad.net/ubuntu/+source/efivar/+bug/1796260

mwhudson (commented):
The changes make sense to me but I worry that we are about to bump into an old efivars bug.

subiquity/server/controllers/filesystem.py
dbungert: I want boot.py to figure that out.  By the time we are in `guided()` we are actually changing the partitions in the model.  We should be able to say in boot.py that it isn't appropriate to create EFI partitions in logical and that there is a primary partition available.  So if we would be forced into such things, I want boot plan to have `can_be_boot_device()` return false, then that scenario isn't even offered as one of the guided possibilities.

I think this PR doesn't make this problem worse, and fixing extended and logical is next on my list, so I'd like to handle it with that work.  What do you think?

mwhudson: Yes that makes sense

mwhudson (approved):

PR #1347, opened 2022-07-12
cloud_init_files: render a clean script to /etc/cloud/clean.d

blackboxsw:
To support golden image creation from the live installer,
provide cloud-init with /etc/cloud/clean.d/99-installer script.

This script will only be invoked when sudo cloud-init clean is
run, which is typically only used when trying to create customized
golden images.


subiquity/models/subiquity.py
```diff
@@ -93,6 +93,14 @@ def setup_yaml():
 ff02::2 ip6-allrouters
 """
 
+CLOUDINIT_CLEAN_FILE_TMPL = """\
+#!/bin/sh
+# Remove live-installer config artifacts when running: sudo cloud-init clean
```
dbungert, line 98: Over in `file_util.py` we have a `Autogenerated by Subiquity` blurb that is added to generated files, I think it would be good to include something similar.
https://github.com/canonical/subiquity/blob/main/subiquitycore/file_util.py#L75

subiquity/models/subiquity.py
```diff
@@ -93,6 +93,14 @@ def setup_yaml():
...
+for cfg_file in {cfg_files};
+  rm -f $cfg_file;
```
dbungert, line 100: Are you concerned with the return code of this script?
I ask because if there are multiple files, you will only see the return code of the last `rm`.  Maybe that's OK.

subiquity/models/subiquity.py
```diff
@@ -314,6 +322,20 @@ def _cloud_init_files(self):
                 ('etc/hostname', hostname + "\n", 0o644),
                 ('etc/hosts', HOSTS_CONTENT.format(hostname=hostname), 0o644),
                 ])
+        # Add cloud-init clean hooks to support golden-image creation.
+        cfg_files = ["/" + path for (path, _content, _cmode) in files]
+        network_write_files = self.network.render().get('write_files').values()
```
dbungert, line 327: I think I'd rather run this in reverse, that `network` is responsible for registering that there is a file to remove.  A quick new function on this model to append to a list of files to be cleaned ought to do the trick.

subiquity/models/tests/test_subiquity.py
```diff
@@ -227,3 +234,37 @@ def test_cloud_init_user_list_merge(self, run_cmd):
...
+            crypted_password='sample_pass',
+            hostname='somehost')
+
+        model = self.make_model()
+        model.identity.add_user(main_user)
+        model.userdata = {}
+        expected_files = {
+            'etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg':
+                'network: {config: disabled}\n',
+            'etc/cloud/cloud.cfg.d/99-installer.cfg': re.compile('datasource:\n  None:\n    metadata:\n      instance-id: .*\n    userdata_raw: "#cloud-config\\\\ngrowpart:\\\\n  mode: \\\'off\\\'\\\\npreserve_hostname: true\\\\n\\\\\n'), # noqa
```
dbungert, line 250: Do you mind if we stuff this in a multiline string somewhere?
For the purpose of this test I'm less concerned about the contents of 99-installer.cfg and wouldn't mind looking elsewhere to find that.
Making it an "r" string would probably help the readability as well, if that works.

dbungert (changes requested):
I have a few suggestions and discussion items.

dbungert:
> depends on cloud-init landing https://github.com/canonical/cloud-init/pull/1581
That isn't a hard dependency, correct?  We would just harmlessly write a file that would not be able to do anything until a suitable cloud-init version is present in the installed system, right?

ogayot (changes requested):
Thanks for doing this.

I would strongly recommend that we avoid generating shell directives.
Couldn't we store somewhere, as machine-readable data, the list of files to clean up and read it back from the clean.d script?

I'm picturing something like:

/etc/cloud/clean.d/99-installer
```python
CLOUDINIT_CLEAN_FILE = """\
#!/usr/bin/env python3

import json
import os

with open("/var/lib/subiquity/artifacts.json", encoding="utf-8") as fh:
    list_of_files = json.load(fh)

for file_to_cleanup in list_of_files:
    os.remove(file_to_cleanup)
"""
```

blackboxsw:
> That isn't a hard dependency, correct? We would just harmlessly write a file that would not be able to do anything until a suitable cloud-init version is present in the installed system, right?

Correct, I just didn't want to land this is the review on the cloud-init PR suggested a different directory path for the run-parts support. Once we agree on `/etc/cloud/clean.d` as the supportable directory name, nothing would prevent this PR from being released in subiquity before cloud-init published it's release.

blackboxsw:
> Thanks for doing this.
> 
> I would strongly recommend that we avoid generating shell directives. Couldn't we store somewhere, as machine-readable data, the list of files to clean up and read it back from the clean.d script?
> 
> I'm picturing something like:
> 
> /etc/cloud/clean.d/99-installer
> 
> ```python
> CLOUDINIT_CLEAN_FILE = """\
> #!/usr/bin/env python3
> 
> import json
> import os
> 
> with open("/var/lib/subiquity/artifacts.json", encoding="utf-8") as fh:
>     list_of_files = json.load(fh)
> 
> for file_to_cleanup in list_of_files:
>     os.remove(file_to_cleanup)
> """
> ```

Thanks for the suggestion @ogayot. I like moving to a python script instead of shell. The problem this might present is that we'd have to write out that artifacts.json into /target somewhere...  the only things that are left around by subiquity in the post-reboot system look to be in /var/log/installer and the /etc/(cloud|netplan files|hostname) files.

I suppose we could emit an /etc/cloud/clean.d/installer-artifacts.json, but both subiquity.models.network and subiquity.models.server would have to merge into that installer-artifacts.json.

Since we are rendering the one script that needs to use it, why not not emit those filenames direct into the python script? Maybe there are other uses for installer-artifacts.json in the target system that I'm overlooking though. I've just pushed changes to at least render a python script instead of shell.

subiquity/models/subiquity.py
blackboxsw: Sounds good to me. hopefully I understood  this suggestion properly  network model should announce/register it's emitted config files which can be removed. For lack of a better name I called the method `NetworkModel.rendered_config_paths`

subiquity/models/subiquity.py
blackboxsw: Wasn't concerned with the failure of multiple files on rm -f as the force flag should get us past any files which are absent from previous runs of this script.  I've shifted the implementation to python per part of @ogayot's suggestion.

subiquity/models/subiquity.py
blackboxsw: I teased out a new file_util.generate_timestamped_hearder` that can be used by generate_config_yaml and non_yaml file writes.

subiquity/models/tests/test_subiquity.py
blackboxsw: Done and fixed up unit tests.

subiquity/models/subiquity.py
dbungert: Nice!

subiquity/models/subiquity.py
dbungert: Yes, this works.  If we have other similar cases we can add a `rendered_config_paths` to other models.  Thanks!

blackboxsw:
syntax error in the rendered python script fixing now

blackboxsw:
https://github.com/canonical/cloud-nit/pull/1581 has published to `ppa:cloud-init-dev/daily`

Validated this PR of the installer against ppa:cloud-init-dev/daily cloud-init version 22.2-3338 which supports /etc/cloud/clean.d run-parts.

Upon reboot into the installed system
`sudo cloud-init clean` removes /etc/netplan/00-installer-config.yaml and various /etc/cloud/cloud.cfg.d/* artificacts leaving cloud-init in an enabled state for any viable datasource on next boot.

ogayot:
> Thanks for the suggestion @ogayot. I like moving to a python script instead of shell. The problem this might present is that we'd have to write out that artifacts.json into /target somewhere... the only things that are left around by subiquity in the post-reboot system look to be in /var/log/installer and the /etc/(cloud|netplan files|hostname) files.
> 
> I suppose we could emit an /etc/cloud/clean.d/installer-artifacts.json, but both subiquity.models.network and subiquity.models.server would have to merge into that installer-artifacts.json.

Yes, I see your point, thanks for elaborating.

> 
> Since we are rendering the one script that needs to use it, why not not emit those filenames direct into the python script? Maybe there are other uses for installer-artifacts.json in the target system that I'm overlooking though. I've just pushed changes to at least render a python script instead of shell.

AFAIK, there is no /right/ way to generate bash instructions.
A quote at the wrong place in the template parameters and we end up with invalid syntax ; and sooner or later we end up executing code we don't intended to. There might be no risk in this context but I'd recommend we don't "try" this approach is we have other alternatives.

I think the implementation that you have currently have with [Python list => JSON array => interpreted back as Python] is good enough though. It's probably not as robust as doing `json.parse()` but I fail to see any corner case that would misbehave. I'm fine with this approach. Thanks!

subiquity/models/subiquity.py
```diff
@@ -308,12 +326,25 @@ def _cloud_init_files(self):
             ('etc/cloud/cloud.cfg.d/99-installer.cfg', config, 0o600),
             ('etc/cloud/ds-identify.cfg', 'policy: enabled\n', 0o644),
             ]
+        # Add cloud-init clean hooks to support golden-image creation.
+        cfg_files = ["/" + path for (path, _content, _cmode) in files]
+        cfg_files.extend(self.network.rendered_config_paths())
```
mwhudson, line 331: It's not necessarily something to fix in this branch, but this feels a bit tightly coupled. We shouldn't ideally have to change the code here if another model class emits config files that should be cleaned up. Other code does "for model in models: if hasattr(model, 'make_cloudconfig'): model.make_cloudconfig()" type games but that's also kind of ugly. A functools.singledispatch with implementations for the models might be nicer... not something for Chad to worry about though :-)

subiquitycore/models/network.py
```diff
@@ -523,6 +523,13 @@ def render_config(self):
 
         return config
 
+    def rendered_config_paths(self):
+        """Return a list of file paths rendered by this model."""
+        return [
+            '/' + write_file['path']
+            for write_file in self.render().get('write_files').values()
```
mwhudson, line 530: This also feels a bit repetitive. Maybe we can somehow indicate in the write_files stanza which files should be removed on clean? (Not sure how this would work though)

mwhudson (commented):
I have some comments but not really ones I expect Chad to do anything about...

subiquity/models/subiquity.py
dbungert: @mwhudson I think Chad's role on this PR is more than complete, can we take these items as notes to address in a future PR?

dbungert:
Merged, with above notes from @mwhudson marked in JIRA for resolution later.

PR #1348, opened 2022-07-12
filesystem: drop deprecated v2 guided

dbungert:
The old guided method used in v2 is now wholly replaced by the new,
multiple target type, counterpart.


mwhudson (approved):
:champagne: 

I guess we could kill even more code by switching the non-v2 guided code to a similar API? (And just filter out non-supported guided options client side for now?)

dbungert:
@jpnurmi FYI

PR #1351, opened 2022-07-14
Pass ignore-running: true when refreshing ourself

mwhudson:
Otherwise the refresh awareness blocks refresh when you are connected
via SSH.


mwhudson:
NB: not tested at all

mwhudson:
This passes my tests now!

dbungert (approved):

PR #1352, opened 2022-07-14
server: read config from /run/cloud-init/instance-data-sensitive.json

blackboxsw:
cloud-init emits a static JSON file early in boot the moment it
detects a valid datasource. This will occur in cloud-init's
early init-local or init boot stages. It will be present before
cloud-init status --wait unblocks.

Move away from calling into stages.Init, read_cfg, cloudify as
these entry points may change in the future. The flat
/run/cloud-init/instance-data*.json files are intended to be a stable
"api" for extenal consumers.

This is part of a series of PRs to support cloud-init/subiquity live-installer usability:
 1.  [Add /etc/cloud-init/clean.d/90-installer script to allow cloud-init clean to remove any installer artifacts for golden image creation](https://github.com/canonical/subiquity/pull/1347)
 2. [This PR] subiquity to read `/run/cloud-init/instance-data-sensitive.json` for autoinstall: config directives instead of initializing `Init.read_cfg()`
 3. [Perform schema initial validation on merged userdata_raw config within the installer environment to catch cloud-init-related user misconfiguration before final target system reboot](https://github.com/canonical/subiquity/pull/1359)
 4. TBD ubuntu-desktop-installer to disable cloud-init after installed system reboot via write_files directive
 5. TBD Consume cloud-init status --format (json/yaml) machine readable output to report more succinct status of cloud-init

### DIFF
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import asyncio
+import json
 import logging
 import os
 import sys
@@ -22,7 +23,7 @@ from typing import List, Optional
 
 from aiohttp import web
 
-from cloudinit import safeyaml, stages
+from cloudinit import safeyaml
 from cloudinit.config.cc_set_passwords import rand_user_password
 from cloudinit.distros import ug_util
 
@@ -88,6 +89,8 @@ NOPROBERARG = "NOPROBER"
 iso_autoinstall_path = 'cdrom/autoinstall.yaml'
 root_autoinstall_path = 'autoinstall.yaml'
 cloud_autoinstall_path = 'run/subiquity/cloud.autoinstall.yaml'
+# cloud-init emits instance-data.json every boot when detecting any datasource
+cloud_init_instance_data_path = '/run/cloud-init/instance-data-sensitive.json'
 
 log = logging.getLogger('subiquity.server.server')
 
@@ -529,13 +532,13 @@ class SubiquityServer(Application):
             self.cloud_init_ok = True
         log.debug("waited %ss for cloud-init", time.time() - ci_start)
         if "status: done" in status_txt:
-            log.debug("loading cloud config")
-            init = stages.Init()
-            init.read_cfg()
-            init.fetch(existing="trust")
-            self.cloud = init.cloudify()
-            if 'autoinstall' in self.cloud.cfg:
-                cfg = self.cloud.cfg['autoinstall']
+            log.debug(
+                "loading cloud config from %s", cloud_init_instance_data_path
+            )
+            with open(cloud_init_instance_data_path) as stream:
+                cloud_init_instance_data = json.loads(stream.read())
+            cfg = cloud_init_instance_data["merged_cfg"].get("autoinstall")
+            if cfg:
                 target = self.base_relative(cloud_autoinstall_path)
                 write_file(target, safeyaml.dumps(cfg))
         else:
--- a/subiquity/server/tests/test_server.py
+++ b/subiquity/server/tests/test_server.py
@@ -13,14 +13,19 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import asyncio
+import json
 import os
 import shlex
-from unittest.mock import Mock
+import subprocess
+from unittest import mock
 
 from subiquitycore.utils import run_command
 from subiquitycore.tests import SubiTestCase
+import subiquity.server.server
 from subiquity.server.server import (
     SubiquityServer,
+    cloud_init_instance_data_path,
     cloud_autoinstall_path,
     iso_autoinstall_path,
     root_autoinstall_path,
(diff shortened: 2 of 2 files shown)


dbungert:
Looks good so far

blackboxsw:
Added obligatory unit tests for the async cloudinit_wait method fixed the stream.read() call and ensured full path /run/cloud-init/instance-data-sensitive.json is read.  I had inadvertently removed the leading `/` during unit test development.


Ran full a manual snapcraft build and test install on kvm to confirm the read of `/run/cloud-init/instance-data-sensitive.json`,

subiquity/server/tests/test_server.py
```diff
@@ -96,8 +101,56 @@ def test_bogus_autoinstall_argument(self):
         with self.assertRaises(Exception):
             self.server.select_autoinstall()
 
+    async def test_wait_for_cloudinit_dryrun_sets_cloud_init_ok(self):
+        self.server.controllers = mock.Mock()
+        self.server.controllers.instances = []
+        self.assertIsNone(self.server.cloud_init_ok)
+        self.server.opts.dry_run = True
+        await self.server.wait_for_cloudinit()
```
dbungert, line 109: This really asks the parent system if cloud-init is OK, I suppose we should mock that out.

subiquity/server/server.py
```diff
@@ -529,13 +532,13 @@ async def wait_for_cloudinit(self):
...
-            init = stages.Init()
-            init.read_cfg()
-            init.fetch(existing="trust")
-            self.cloud = init.cloudify()
-            if 'autoinstall' in self.cloud.cfg:
-                cfg = self.cloud.cfg['autoinstall']
+            log.debug(
+                "loading cloud config from %s", cloud_init_instance_data_path
+            )
+            with open(cloud_init_instance_data_path) as stream:
```
dbungert, line 538: In what scenarios can we see FileNotFoundError due to the instance data not being there?

dbungert (commented):
Thanks for doing this, I feel like you've gone above and beyond here with the cloud-init/subiquity integration items.
I have 2 items for you to consider.

subiquity/server/server.py
blackboxsw: `/run/cloud-init/instance-data*json` will only not exist in the following scenarios:

 1. cloud-init is disabled: by `/etc/cloud/cloud-init.disabled`
 2. kernel cmdline param `cloud-init=disabled` exists 
 3. cloud-init being unable to detect a valid datasource on this current system boot so cloud-init's systemd generator `/lib/systemd/system-generators/cloud-init-generator` (which calls ds-identify)
 4. Some broad bug in cloud-init that breaks initial datasource detection (which we'd need to hotfix)

Generally in that disabled case subiquity should have fallen through the other logical path where it [assumes cloud-init is disabled](https://github.com/canonical/subiquity/blob/main/subiquity/server/server.py#L541-L544)  TBH, once we have JSON/YAML output format from `cloud-init status` I'd want to make the status output parsing a bit more strict than it currently is in server.py.

subiquity/server/tests/test_server.py
blackboxsw: oops missed that mock once I did it on the other test. Fixing now

blackboxsw: added an `self.assertEqual(0, wait_for.call_count)` for this test

subiquity/server/server.py
blackboxsw: Also once we add the ability to disable Desktop installer after `installed boot` then we know that the final booted desktop installed image will no longer surface /run/cloud-init/* files as cloud-init will be in disabled state.

dbungert: OK, for the check if cloud-init is disabled or not, from the Subiquity perspective, the planned machine-readable output will help with that.  So for now I'm prepared to assume that's correct.

subiquity/server/server.py
```diff
@@ -529,13 +532,13 @@ async def wait_for_cloudinit(self):
...
+                cloud_init_instance_data = json.loads(stream.read())
+            cfg = cloud_init_instance_data["merged_cfg"].get("autoinstall")
```
dbungert, line 540: In a test in a VM, I found that this was not providing autoinstall data.
I mixed in extra code to show the results of the old method and this, and the old method is finding the autoinstall data.  https://github.com/dbungert/subiquity/commit/49d249af3829dc2b8334e9f75ad1be8152c27e9d

The kvm-test script may help in reproducing this.  Create a quick config file at ~/.kvm-test.yaml
https://gist.github.com/dbungert/ecb1b053f4f0d493d5ad9b324999fc38

Then you can run:
```
./scripts/kvm-test.py --build --install -o --autoinstall
```

subiquity/server/server.py
```diff
@@ -529,13 +532,13 @@ async def wait_for_cloudinit(self):
             self.cloud_init_ok = True
         log.debug("waited %ss for cloud-init", time.time() - ci_start)
         if "status: done" in status_txt:
-            log.debug("loading cloud config")
-            init = stages.Init()
-            init.read_cfg()
-            init.fetch(existing="trust")
-            self.cloud = init.cloudify()
```
dbungert, line 536: Other areas of the code reference self.cloud.  Probably best to leave these first few lines for now until we can revisit that.

dbungert (changes requested):

subiquity/server/server.py
blackboxsw: @dbungert  you're absultely right. I had forgotten merged_cfg was only merged from filesystem /etc/cloud and /etc/cloud/cloud.cfg.d/ and not yet userdata as well :/. This is also something that came up in investigating this with UA-related changes which need to start looking at userdata specifically. I agree we should hold off on migration to /run/cloud-init/instance-data-sensitve.json until we can provide user-data/vendor-data content in the merged_cfg.

subiquity/server/server.py
blackboxsw: @dbungert Thanks for the review. I think I'll close this PR and put it on back burner until cloud-init also publishes full user-data/vendor-data content into instance-data-sensitive.json: merged_cfg key. Then it'll make this changeset worth it.

blackboxsw:
Closing until cloud-init makes user-data visibile  through /run/cloud-init/instance-data-sensitive.json

subiquity/server/server.py
dbungert: Thanks @blackboxsw - I like where this changeset is going!

PR #1356, opened 2022-07-18
filesystem: add orig_config API

dbungert:
Clients may want a view of the unmodified disk configuration.


subiquity/models/filesystem.py
```diff
@@ -1041,6 +1041,12 @@ def reset(self):
         self.grub = None
         self.guided_configuration = None
 
+    def get_orig_model(self):
+        orig_model = FilesystemModel(self.bootloader)
+        orig_model.target = self.target
+        orig_model.load_probe_data(self._probe_data)
+        return orig_model
```
mwhudson, line 1048: There is already self._orig_config (which appears to be a copy of the data that has been passed through storage_config.extract_storage_config). Maybe the uses of that can switch to this API too?

mwhudson (approved):
Seems fine, worried a bit about creeping duplication though

subiquity/models/filesystem.py
dbungert: self._orig_config is in terms of curtin actions, which we are not using on the V2 API.

In this specific case, getting the original partition size, I think that can be worked around by going to orig_config, selecting the object with the matching path, and then getting the size, but this is all a strange request for software that is otherwise expecting a V2-like set of objects, and other similar-sounding queries may be disproportionately harder.

If no client uses the curtin action API I think the existing _orig_config can go away?

I can add a comment talking about this difference.

subiquity/tests/api/test_api.py
```diff
@@ -1101,6 +1101,22 @@ async def test_est_min_size(self):
...
+        cfg = 'examples/win10-along-ubuntu.json'
+        extra = ['--storage-version', '2']
+        async with start_server(cfg, extra_args=extra) as inst:
+            start_resp = await inst.get('/storage/v2')
+            resp = await inst.get('/storage/v2/guided')
+            resize = match(resp['possible'],
+                           _type='GuidedStorageTargetResize')[0]
+            resize['new_size'] = 30 << 30
+            await inst.post('/storage/v2/guided', {'target': resize})
+            orig_config = await inst.get('/storage/v2/orig_config')
```
dbungert, line 1115: @jpnurmi FYI

PR #1357, opened 2022-07-19
Replace setxkbmap with a POST /keyboard response

jpnurmi:
The main reason to move the `setxkbmap` call to Subiquity was to avoid running the command in dry-run mode because the desktop GUI doesn't know if Subiquity is in dry-run or live mode. When Subiquity is running as a daemon, however, it cannot run `setxkbmap` because it requires X11. Sending a response that indicates whether a keyboard layout was applied to the system allows the GUI to determine whether it should update the desktop settings accordingly i.e. avoid running `setxkbmap` in dry-run mode.

Ref: canonical/ubuntu-desktop-installer#832


dbungert (changes requested):
Thanks for the removal of setxkbmap.
As to the other part, I think the problem here is that the return value is a bit unusual.
It isn't exactly an indication that the value has changed, it isn't exactly a this-is-dryrun indicator, it also isn't conclusively an indication that the change has been applied to the system.
I would like to change this, somehow, so that we can solve the problem you're trying to solve but make the return value more clear.

dbungert:
When Subiquity is not in dry run mode, it is really making changes to the host system.
Should we proceed with this plan, a bug in Subiquity could mean changes to the host system if the client were to receive the wrong value.  We had a version of that, earlier, when timezone settings were applied when we were supposed to be in dryrun mode and it caused chaos.

When we are really installing, I think the client should be fully aware of that.

jpnurmi:
I may have to disagree with that. :) In my opinion, it would be better to keep the dry-run logic in Subiquity and not spread it across the stack. The desktop GUI is supposed to be a thin client and as such, should not touch the system in the first place.

jpnurmi:
Something like this gets us pretty close:
```
sudo su
export DBUS_SESSION_BUS_ADDRESS="unix:path=/run/user/1000/bus"
sudo -u ubuntu gsettings set org.gnome.desktop.input-sources sources "[('xkb','us')]"
```
It writes the setting for the ubuntu user but spins off a new dconf-service, and thus, the user session doesn't get notified of the changes.

jpnurmi:
Superseded by #1361

PR #1359, opened 2022-07-19
subiquitymodel: validate merged cloud-config userdata_raw before reboot

blackboxsw:
This will raise noisy SchemaValidationErrors if either deprecated keys
or schema errors are provided in autoinstall.user-data or in the
merged userdata_raw provided to cloud-init in the installed target.

This is part of a series of PRs to support cloud-init/subiquity live-installer usability:
 1.  [Add /etc/cloud-init/clean.d/90-installer script to allow cloud-init clean to remove any installer artifacts for golden image creation](https://github.com/canonical/subiquity/pull/1347)
 2. [subiquity to read `/run/cloud-init/instance-data-sensitive.json` for autoinstall: config directives instead of initializing `Init.read_cfg()`](https://github.com/canonical/subiquity/pull/1352)
 3. [This PR] Perform schema initial validation on merged userdata_raw config within the installer environment to catch cloud-init-related user misconfiguration before final target system reboot
 4. [#1395] ubuntu-desktop-installer to disable cloud-init after installed system reboot via write_files directive
 5. TBD Consume cloud-init status --format (json/yaml) machine readable output to report more succinct status of cloud-init

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -23,6 +23,18 @@ from typing import Set
 import uuid
 import yaml
 
+from cloudinit.config.schema import (
+        SchemaValidationError,
+        get_schema,
+        validate_cloudconfig_schema
+)
+
+try:
+    from cloudinit.config.schema import SchemaProblem
+except ImportError:
+    def SchemaProblem(x, y): return (x, y)  # TODO(drop on cloud-init 22.3 SRU)
+
+
 from curtin.config import merge_config
 
 from subiquitycore.file_util import (
@@ -274,6 +286,46 @@ class SubiquityModel:
     def confirm(self):
         self._confirmation.set()
 
+    def validate_cloudconfig_schema(self, data: dict, data_source: str):
+        """Validate data config adheres to strict cloud-config schema
+
+        Log warnings on any deprecated cloud-config keys used.
+
+        :param data: dict of valid cloud-config
+        :param data_source: str to present in logs/errors describing
+            where this config came from: autoinstall.user-data or system info
+
+        :raise SchemaValidationError: on invalid cloud-config schema
+        """
+        # cloud-init v. 22.3 will allow for log_deprecations=True to avoid
+        # raising errors on deprecated keys.
+        # In the meantime, iterate over schema_deprecations to log warnings.
+        try:
+            validate_cloudconfig_schema(data, schema=get_schema(), strict=True)
+        except SchemaValidationError as e:
+            if hasattr(e, "schema_deprecations"):
+                warnings = []
+                deprecations = getattr(e, "schema_deprecations")
+                if deprecations:
+                    for schema_path, message in deprecations:
+                        warnings.append(message)
+                if warnings:
+                    log.warning(
+                        "The cloud-init configuration for %s contains"
+                        " deprecated values:\n%s", data_source, "\n".join(
+                            warnings
+                        )
+                    )
+            if e.schema_errors:
+                if data_source == "autoinstall.user-data":
+                    errors = [
+                        SchemaProblem(f"{data_source}.{path}", message)
+                        for (path, message) in e.schema_errors
+                    ]
+                else:
+                    errors = e.schema_errors
+                raise SchemaValidationError(schema_errors=errors)
+
     def _cloud_init_config(self):
         config = {
             'growpart': {
@@ -311,6 +363,9 @@ class SubiquityModel:
             config.setdefault("write_files", []).append(
                 CLOUDINIT_DISABLE_AFTER_INSTALL
             )
+        self.validate_cloudconfig_schema(
+            data=config, data_source="system install"
+        )
         return config
 
     async def target_packages(self):
--- a/subiquity/server/controllers/userdata.py
+++ b/subiquity/server/controllers/userdata.py
@@ -13,7 +13,11 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import logging
+
 from subiquity.server.controller import NonInteractiveController
+
+log = logging.getLogger("subiquity.server.controllers.userdata")
 
 
 class UserdataController(NonInteractiveController):
@@ -27,6 +31,10 @@ class UserdataController(NonInteractiveController):
 
     def load_autoinstall_data(self, data):
         self.model.clear()
+        if data:
+            self.app.base_model.validate_cloudconfig_schema(
+                data=data, data_source="autoinstall.user-data",
+            )
         self.model.update(data)
 
     def make_autoinstall(self):
--- a/examples/autoinstall-user-data.yaml
+++ b/examples/autoinstall-user-data.yaml
@@ -18,5 +18,5 @@ source:
 updates: security
 user-data:
   users:
-    - username: ubuntu
-      password: '$6$wdAcoXrU039hKYPd$508Qvbe7ObUnxoj15DRCkzC3qO7edjH0VV7BPNRDYK4QR8ofJaEEF2heacn0QgD.f8pO8SNp83XNdWG6tocBM1'
+    - name: ubuntu
+      passwd: '$6$wdAcoXrU039hKYPd$508Qvbe7ObUnxoj15DRCkzC3qO7edjH0VV7BPNRDYK4QR8ofJaEEF2heacn0QgD.f8pO8SNp83XNdWG6tocBM1'
--- a/subiquity/models/tests/test_subiquity.py
+++ b/subiquity/models/tests/test_subiquity.py
@@ -13,13 +13,18 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-
 import fnmatch
 import json
 import unittest
 from unittest import mock
 import re
 import yaml
+
+from cloudinit.config.schema import SchemaValidationError
+try:
+    from cloudinit.config.schema import SchemaProblem
+except ImportError:
+    def SchemaProblem(x, y): return (x, y)  # TODO(drop on cloud-init 22.3 SRU)
 
 from subiquitycore.pubsub import MessageHub
 
@@ -236,6 +241,17 @@ class TestSubiquityModel(unittest.IsolatedAsyncioTestCase):
             self.assertEqual(len(cloud_init_config['users']), 1)
             self.assertEqual(cloud_init_config['users'][0]['name'], 'user2')
 
+        with self.subTest('Invalid user-data raises error'):
+            model = self.make_model()
+            model.userdata = {'bootcmd': "nope"}
+            with self.assertRaises(SchemaValidationError) as ctx:
+                model._cloud_init_config()
+            expected_error = (
+                "Cloud config schema errors: bootcmd: 'nope' is not of type"
+                " 'array'"
+            )
+            self.assertEqual(expected_error, str(ctx.exception))
+
     @mock.patch('subiquity.models.subiquity.lsb_release')
     @mock.patch('subiquitycore.file_util.datetime.datetime')
     def test_cloud_init_files_emits_datasource_config_and_clean_script(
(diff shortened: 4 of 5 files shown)


subiquity/models/subiquity.py
```diff
@@ -278,6 +280,12 @@ def _cloud_init_config(self):
             if getattr(model, 'make_cloudconfig', None):
                 merge_config(config, model.make_cloudconfig())
         merge_cloud_init_config(config, self.userdata)
+        # Raises cloudinit.config.schema.SchemaValidationError or
+        # SchemaDeprecationError to alert to merged-cloud-config issues
+        # in installer environment instead of after installed target reboot.
+        validate_cloudconfig_schema(
+            config, strict=True, log_details=False, log_deprecations=True
+        )
```
blackboxsw, line 288: One issue to note is that we would be running this validation agains the current cloud-init installed in the ephemeral environment, not the cloud-init version installed in the target environment. I've seen minor revision drift a few times between the ephemeral environment version of cloud-init and the target  environment that is being installed. While it is unlikely that there is enough of a drift to deprecate or drop keys between those minor versions, there could be bug fixes related to schema that could crop up resulting in invalid schema warnings/errors. Is it worth us ensuring that we source and run the config validation within the target environment (maybe via chroot). Or cloud-init to provide a mechanism to ensure this entry point is confined to only schema files within the `/target` path

dbungert: Something that may help the target environment is that updates should be applied before target environment first boot.  No guarantees on that front, though.

It's pretty corner case, but if someone used a original version LTS iso with no point release updates, allowed Subiquity to update, then blocked other updates from running, that could be a pretty large (years) version drift.   But, someone blocking updates probably isn't allowing Subiquity to update either.

I suggest we start without chroot into target for schema validation artifacts and see how it goes.

subiquity/models/subiquity.py
```diff
@@ -278,6 +280,10 @@ def _cloud_init_config(self):
             if getattr(model, 'make_cloudconfig', None):
                 merge_config(config, model.make_cloudconfig())
         merge_cloud_init_config(config, self.userdata)
+        # Raises cloudinit.config.schema.SchemaValidationError or
+        # SchemaDeprecationError to alert to merged-cloud-config issues
```
dbungert, line 284: Would it be more consistent with cloud-init to make SchemaDeprecationError non-fatal, and just log it?

subiquity/models/subiquity.py
```diff
@@ -278,6 +280,10 @@ def _cloud_init_config(self):
...
+        # in installer environment instead of after installed target reboot.
+        validate_cloudconfig_schema(config, strict=True)
```
dbungert, line 286: This spot is good, because it will catch the mix of what Subiquity generates and what the user provides, if anything.

It's a bit late in the install process though - there is a chance that someone has provided
```
autoinstall:
    user-data:
        something: invalid
```
And then then don't hear about it until things are almost done.

What's your thoughts about an earlier check for this, in subiquity/server/controllers/userdata.py ?

dbungert (commented):
Thanks for doing this.
I have a few questions.

subiquity/models/subiquity.py
blackboxsw: Sure we can definitely do that as you prefer, ideally I'd love to see the live installer grow some sort of user-facing error dialog handler that would pop up an error message in the UX to inform the person of an actionable problem before they commit the the full install and reboot, or have to spelunk in /var/log/installer/* to see what a hard the traceback/runtime error says.  (maybe that type of error dialog already exists in the installer and I'm not aware of how to plug into it).


 For what it's worth, one thing we found in cloud-init was that warning logs generally aren't discovered until sometime later on a deployment when someone happens to recognize that the setup of the VM didn't quite work in every way they expected. Long-term, we've decided to make cloud-init noisier earlier to avoid delayed awareness of misconfiguration which could lead to extended security holes in a deployed system.  That said, cloud-init is not erroring hard at the moment on schema validation errors because we want to make our schema strict and ensure all unexpected use-cases are resolved. I think it's worth us leaving subiquity in non-strict, warning mode for the moment until cloud-init also applies strict/hard errors on schema validation. We can revisit these warnings in the live installer at that time and maybe we'll also have a mechanism to surface those warnings nicely via UX that we can tie into at that point.

subiquity/models/subiquity.py
blackboxsw: @dbungert How will a warning log be visible to the user if generated up in subiquity/server/controllers/userdata.py ? I'm reading this as a warning that happens to be emitted only to /var/log/installer/* files and isn't really "seen" in UX. Is that incorrect?

blackboxsw: We can move it earlier in the install process and we probably don't want to specifically render the full details of the schema failure because that then results in leaked sensitive content into /var/log/installler/* which may be visible on the target system to non-root users I think. The changeset I just pushed basically gives us a redacted warning saying: `Invalid cloud-config provided: Please run 'sudo cloud-init schema --system' for details`. This message would be applicable in the installed system after reboot, but not in the installer/ephemeral environment. Maybe we want to try/except SchemaValidationError in ` subiquity/server/controllers/userdata.py` with a concrete warning about the provided invalid `autoinstall.user-data` key path (and redact the offending value) to point folks specifically to the part of the autoinstall that looks like it will be a problem for cloud-init.

blackboxsw: @dbungert here's a diff of what I was thinking. The problem a live-installer environement has, I think, is that /autoinstall.yaml is  renderered with any cloud-init passthrough user-data  in a top-level `user-data` key right?

I'd like to be able to give a constructive warning message for the ephemeral environment additionally saying something like run `sudo cloud-init schema --config-file /autoinstall.yaml` but cloud-init will not validate the top-level `user-data` key as valid cloud-config.

So, we'd need a YAML artifact emitted that is only the content below autoinstall.user-data. Maybe we emit this on schema error we could emit some file `/autoinstall-invalid-userdata.yaml` which the warning message could point to.


If the folloowing invalid autoinstall is provided:
cat > brokenautoinstall.yaml <<EOF
#cloud-config
autoinstall:
   version:1
   user-data:
       chef: 
           install_type: 1
EOF

Then a `sudo cloud-init schema -c /autoinstall-invalid-userdata.yaml` will be able to emit errors like:
```
$ sudo cloud-init schema -c /autoinstall-invalid-userdata.yaml
Error:
Cloud config schema errors: chef.install_type: 1 is not of type 'string', chef.install_type: 1 is not one of ['packages', 'gems', 'omnibus']
```

The above invalid autoinstall.yaml would generate the following WARNING with the diff below:
WARNING: Invalid autoinstall config schema provided for keys:autoinstall.user-data.chef.install_type"

But there is no easy way currently to tease out the autoinstall.user-data for cloud-init schema validation on the commandline so we can't provide helpful breadcrumbs to lead someone to discover specifically what the schema prob is.

diff --git a/subiquity/server/controllers/userdata.py b/subiquity/server/controllers/userdata.py
index 77e743e0..efa76f52 100644
--- a/subiquity/server/controllers/userdata.py
+++ b/subiquity/server/controllers/userdata.py
@@ -13,9 +13,22 @@
 
+import logging
+
+from cloudinit.config.schema import (
+    SchemaValidationError, validate_cloudconfig_schema
+)
+
 from subiquity.server.controller import NonInteractiveController
 
 
+log = logging.getLogger("subiquity.server.controllers.userdata")
+
+INVALID_USERDATA_SCHEMA_TMPL = (
+   "Invalid autoinstall config schema provided for keys: {invalid_keys}"
+)
+
+
 class UserdataController(NonInteractiveController):
 
     model_name = 'userdata'
@@ -26,6 +39,19 @@ class UserdataController(NonInteractiveController):
         }
 
+        # We are processing autoinstall.user-data valid raw cloud-config to
+        # give a sensible warning to the user based on content they provided.
+        try:
+            validate_cloudconfig_schema(data, strict=True)
+        except SchemaValidationError as exc:
+            invalid_keys = [
+                f"autoinstall.user-data.{schema_path}"
+                for (schema_path, _msg) in exc.schema_errors
+            ]
+            message = INVALID_USERDATA_SCHEMA_TMPL.format(
+                invalid_keys=", ".join(invalid_keys)
+            )
+            log.warning(message)
         self.model.clear()
         self.model.update(data)

subiquity/server/controllers/userdata.py
```diff
@@ -27,6 +35,25 @@ class UserdataController(NonInteractiveController):
...
+        # SchemaDeprecationError depending on version of cloud-init to alert to
+        # cloud-config issues in provided autoinstall.user-data.
+        # cloud-init v. 22.3 will allow for log_deprecations=True to avoid
+        # raising errors on deprecated keys.
+        try:
+            validate_cloudconfig_schema(data, strict=True)
+        except SchemaValidationError as e:
+            if hasattr(e, "schema_deprecations"):
+                warnings = []
+                for schema_path, message in getattr(e, "schema_deprecations"):
```
dbungert, line 48: Per above if check I suppose we can simplify a bit
```suggestion
                for schema_path, message in e.schema_deprecations
```
Thoughts?

dbungert: And I think we can do some refactoring of this block to pull this and the similar one in subiquity/models/subiquity.py

dbungert (approved):
A few suggestions, or we can merge now and handle that later with some of the other cleanups already in queue.

dbungert:
Something very strange is happening when I VM test this branch with autoinstall.  Using a fairly boring autoinstall set https://gist.github.com/dbungert/a48ea4d6291daa2e2bc53fbd2056a33b, one that works fine with main branch, on this branch I see curtin throwing schema validation errors from block probing, which makes no sense whatsoever.

@blackboxsw - would you run a VM test from this branch and throw some autoinstall at it?

subiquity/server/controllers/userdata.py
blackboxsw: @dbungert refactored common code to a public method `SubiquityModel.validate_cloudconfig_schema` which the controller can call too.

subiquity/models/subiquity.py
```diff
@@ -23,6 +23,18 @@
...
 import yaml
 
+from cloudinit.config.schema import (
+        SchemaValidationError,
+        get_schema,
+        validate_cloudconfig_schema
+)
+
+try:
+    from cloudinit.config.schema import SchemaProblem
```
blackboxsw, line 33: SchemaProblem wasn't defined in original release on Bionic or Focal. So we'll hit import errors at runtime if subiquity is installing from the original released images. SchemaProblem is just a NamedTuple, and the original schema_errors were just a tuple, so we can create a function that does the same when handling `ImportError`

subiquity/models/subiquity.py
```diff
@@ -263,6 +275,46 @@ def is_postinstall_only(self, model_name):
...
+
+        :raise SchemaValidationError: on invalid cloud-config schema
+        """
+        # cloud-init v. 22.3 will allow for log_deprecations=True to avoid
+        # raising errors on deprecated keys.
+        # In the meantime, iterate over schema_deprecations to log warnings.
+        try:
+            validate_cloudconfig_schema(data, schema=get_schema(), strict=True)
+        except SchemaValidationError as e:
+            if hasattr(e, "schema_deprecations"):
```
blackboxsw, line 295: Original release of bionic/focal or Jammy will not have schema_deprecations attr on SchemaValidationError. so be defensive about presence of that attribute in the even that subquity is trying to install from orig release images.

dbungert (commented):
Thanks Chad, LGTM, I will VM test this and report back.

dbungert (commented):
For unclear reasons, this PR causes block probing to start failing, so I think it should not be merged yet.
I'm not convinced that's actually a bug of this PR, though.
Hold for now until the root cause is understood.

blackboxsw:
We determined that the 'old' non-distro version of jsonschema pip installed in the subiquity snap was incompatible with cloud-init. Dan put up https://github.com/canonical/subiquity/pull/1396 to remove that duplicate version of jsonschema and I validated that this PR works in a locally built snap based on this branch and 1396

dbungert (approved):

PR #1361, opened 2022-07-21
Run setxkbmap as the ubuntu user

jpnurmi:
This is a partial fix to canonical/ubuntu-desktop-installer#832. The last missing piece is to set the `DISPLAY` environment variable but this can be done in the Subiquity startup script in UDI: https://github.com/canonical/ubuntu-desktop-installer/pull/1021.

### DIFF
--- a/subiquity/server/controllers/keyboard.py
+++ b/subiquity/server/controllers/keyboard.py
@@ -17,6 +17,7 @@ import logging
 from typing import Dict, Optional
 import os
 import shutil
+import sys
 
 import attr
 
@@ -204,9 +205,23 @@ class KeyboardController(SubiquityController):
             [resource_path('bin/subiquity-loadkeys')],
             ]
         if shutil.which('setxkbmap'):
-            setxkbmap = ['setxkbmap', '-layout', self.model.setting.layout]
+            command = """\
+import os
+import pwd
+import subprocess
+import sys
+
+os.seteuid(pwd.getpwnam("ubuntu").pw_uid)
+cmd = ["setxkbmap", "-layout", sys.argv[1]]
+if len(sys.argv) > 2:
+    cmd.extend(["-variant", sys.argv[2]])
+subprocess.run(cmd)
+"""
+            setxkbmap = [
+                sys.executable, '-c', command, self.model.setting.layout
+                ]
             if self.model.setting.variant:
-                setxkbmap.extend(['-variant', self.model.setting.variant])
+                setxkbmap.append(self.model.setting.variant)
             cmds.append(setxkbmap)
         if self.opts.dry_run:
             scale = os.environ.get('SUBIQUITY_REPLAY_TIMESCALE', "1")


jpnurmi:
Together with canonical/ubuntu-desktop-installer#1021:
```
2022-07-21 09:33:46,245 DEBUG root:37 start: subiquity/Keyboard/POST: 
2022-07-21 09:33:46,245 DEBUG subiquity.server.controllers.keyboard:236 KeyboardSetting(layout='fi', variant='', toggle=None)
2022-07-21 09:33:46,245 DEBUG subiquitycore.utils:92 arun_command called: ['setupcon', '--save', '--force', '--keyboard-only']
2022-07-21 09:33:46,517 DEBUG subiquitycore.utils:101 arun_command ['setupcon', '--save', '--force', '--keyboard-only'] exited with code 0
2022-07-21 09:33:46,517 DEBUG subiquitycore.utils:92 arun_command called: ['/snap/ubuntu-desktop-installer/x1/bin/subiquity/bin/subiquity-loadkeys']
2022-07-21 09:33:46,520 DEBUG subiquitycore.utils:101 arun_command ['/snap/ubuntu-desktop-installer/x1/bin/subiquity/bin/subiquity-loadkeys'] exited with code 0
2022-07-21 09:33:46,521 DEBUG subiquitycore.utils:92 arun_command called: ['setxkbmap', '-layout', 'fi']
2022-07-21 09:33:46,535 DEBUG subiquitycore.utils:101 arun_command ['setxkbmap', '-layout', 'fi'] exited with code 0
2022-07-21 09:33:46,535 DEBUG subiquity.models.subiquity:202 model keyboard for install stage is configured, to go {'filesystem'}
2022-07-21 09:33:46,536 DEBUG root:37 finish: subiquity/Keyboard/POST: SUCCESS: 200 null
2022-07-21 09:33:46,536 INFO aiohttp.access:233  [21/Jul/2022:09:33:46 +0000] "POST /keyboard HTTP/1.1" 200 189 "-" "Dart/2.17 (dart:io)"
```

subiquity/server/controllers/keyboard.py
```diff
@@ -203,16 +204,26 @@ async def set_keyboard(self):
...
             cmds = [['sleep', str(1/float(scale))]]
         for cmd in cmds:
             await arun_command(cmd)
+        await self.run_setxkbmap()
+
+    async def run_setxkbmap(self):
+        if not self.opts.dry_run and shutil.which('setxkbmap'):
+            try:
+                # seteuid() because Popen(user=) is new in 3.9 (UDI has 3.8)
+                os.seteuid(pwd.getpwnam('ubuntu').pw_uid)
```
ogayot, line 218: `run_setxkbmap` is an asynchronous function so there is a chance that a context switch occurs before the EUID gets reset by the `finally` block. I think this can lead to some hard to reproduce bugs.
Would you mind trying the following construct instead?
```python
setxkbmap = ["runuser", "-u", "ubuntu", "--", "setxkbmap", "-layout", self.model.setting.layout]
```
`runuser` is provided by `util-linux` ; and I believe it should mimic the `user` parameter of `subprocess.Popen` quite well.

ogayot (changes requested):
I'm fine with the idea. I have a suggestion to avoid the risky `os.seteuid` though

subiquity/server/controllers/keyboard.py
jpnurmi: Nice idea. Manually executing `runuser` as root in a clean environment works, but I'm having problems running `runuser` from Subiquity:
```
runuser: cannot open session: Module is unknown
```
Seems to be an error from `pam_open_session()`. :thinking:

jpnurmi: I suppose it's essentially the same problem why `sudo -u ubuntu` didn't work either but I had to set the effective uid.

jpnurmi: pam_open_session(3):
> It should be noted that the effective uid, geteuid. of the application should be of sufficient privilege to perform such tasks as creating or mounting the user's home directory for example.

ogayot: yikes!
I guess this would essentially work but this feels like a stretch ...
```python
command = """\
import subprocess
import os
import pwd
import sys

os.seteuid(pwd.getpwnam("ubuntu").pw_uid)
subprocess.run(("setxkbmap", "-layout", sys.argv[1]))
"""

setxkbmap = ["python3", "-c", command, self.model.setting.layout]
```

jpnurmi: I'll take a stab at upgrading UDI from core20 to core22. Then we'd have Python 3.10 available...

ogayot: Subiquity on server runs on core20. We haven't been doing testing on python 3.10 AFAIK so I would not recommend it.
Are you sure subprocess.Popen(user="ubuntu") would not fail with a similar error to "cannot open session: Module is unknown" or something?

jpnurmi: `subprocess.Popen(user="ubuntu")` calls `setreuid()` under the hood.

jpnurmi: What if we'd make it sync? setxkbmap takes no time to execute.

jpnurmi: This seems to work
```py
    async def run_setxkbmap(self):
        if not self.opts.dry_run and shutil.which('setxkbmap'):
            command = """\
import subprocess
import os
import pwd
import sys

os.seteuid(pwd.getpwnam("ubuntu").pw_uid)
subprocess.run(["setxkbmap", "-layout", sys.argv[1], "-variant", sys.argv[2]])
"""
            variant = self.model.setting.variant
            setxkbmap = ["python3", "-c", command, self.model.setting.layout,
                         variant if not None else ""]
            await arun_command(setxkbmap)
```

jpnurmi: Coming back to this... I just realized the desktop installer has some logic to run a specific Python interpreter a) depending on whether it's snapped and b) to avoid conflicts with the Python interpreter in the Flutter snap. I'm not sure if it matters here, but this code would run a different Python interpreter.

jpnurmi: It looks like `sys.executable` gives the correct Python interpreter. Should we proceed with this?

ogayot: Yes please, I am more comfortable with this approach.

`sys.executable` - or alternatively `os.environ["PYTHON"]` - is the right way to do it :+1: 

Is setxkbmap okay with receiving an empty string as the value for `-variant` or should we pass `-variant` only if the value is provided?

ogayot (approved):
Thanks!
I like this change a lot know. LGTM!

PR #1362, opened 2022-07-21
meta/status: replace POST_WAIT and POST_RUNNING by generic states

ogayot:
With the plan to call curtin multiple times where each invocation has a specific purpose, we want the client to have a simple way to monitor the status of the installation.

Instead of defining new application states for each stage, we will report the more generic `WAITING` / `NEEDS_CONFIRMATION` / `RUNNING` states in a loop until we get to the final stages of the installation.

To start with, we are dropping `POST_WAIT` and `POST_RUNNING` in favor or `WAITING` and `RUNNING`.

Since this impacts both Subiquity, the desktop installer and console-setup. I'm marking the PR a draft until enough testing is done.

### DIFF
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -56,16 +56,27 @@ class ErrorReportRef:
 
 
 class ApplicationState(enum.Enum):
+    """ Represents the state of the application at a given time. """
+
+    # States reported during the initial stages of the installation.
     STARTING_UP = enum.auto()
     CLOUD_INIT_WAIT = enum.auto()
     EARLY_COMMANDS = enum.auto()
-    WAITING = enum.auto()
+
+    # State reported once before starting destructive actions.
     NEEDS_CONFIRMATION = enum.auto()
+
+    # States reported during installation. This sequence should be expected
+    # multiple times until we reach the late stages.
+    WAITING = enum.auto()
     RUNNING = enum.auto()
-    POST_WAIT = enum.auto()
-    POST_RUNNING = enum.auto()
+
+    # States reported while unattended-upgrades is running.
+    # TODO: check if these should be dropped in favor of RUNNING.
     UU_RUNNING = enum.auto()
     UU_CANCELLING = enum.auto()
+
+    # Final state
     DONE = enum.auto()
     ERROR = enum.auto()
     EXITED = enum.auto()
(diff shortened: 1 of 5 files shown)


subiquity/common/types.py
```diff
@@ -56,16 +56,25 @@ class ErrorReportRef:
...
 
 class ApplicationState(enum.Enum):
+    """ Represents the state of the application at a given time. """
+
+    # States reported during the initial stages of the installation
     STARTING_UP = enum.auto()
     CLOUD_INIT_WAIT = enum.auto()
     EARLY_COMMANDS = enum.auto()
+
+    # States reported during installation. This sequence should be expected
```
dbungert, line 66: Well, NEEDS_CONFIRMATION would be once only I think?

dbungert (commented):
Minor note about a comment, otherwise LGTM

subiquity/common/types.py
ogayot: Yep, I agree. I think I got confused while reading the spec. I'll update this comment.

dbungert (commented):
Looks good so far

ogayot:
I performed some testing with KVM and did not notice any regression in Subiquity itself.
Marking the PR ready for review but some testing in system_setup and flutterstaller would be good if @CarlosNihelton and @jpnurmi are around. Thanks!

dbungert (approved):

CarlosNihelton:
> I performed some testing with KVM and did not notice any regression in Subiquity itself. Marking the PR ready for review but some testing in system_setup and flutterstaller would be good if @CarlosNihelton and @jpnurmi are around. Thanks!

I run all tests in the repository with Subiquity checkout at  `ogayot:fde-curtin-application-state` and all tests passed. JP is on vacation these days, but before he left, he issued a PR already preparing the GUI repository for those changes. So I think we are good to go.

ogayot:
> > I performed some testing with KVM and did not notice any regression in Subiquity itself. Marking the PR ready for review but some testing in system_setup and flutterstaller would be good if @CarlosNihelton and @jpnurmi are around. Thanks!
> 
> I ran all tests in the repository with Subiquity checkout at `ogayot:fde-curtin-application-state` and all tests passed. JP is on vacation these days, but before he left, he issued a PR already preparing the GUI repository for those changes. So I think we are good to go.

Thanks!

PR #1363, opened 2022-07-21
guided reformat: check min size

dbungert:


subiquity/server/controllers/filesystem.py
```diff
@@ -408,7 +408,8 @@ async def v2_guided_GET(self) -> GuidedStorageResponseV2:
 
         for disk in self.get_guided_disks(with_reformatting=True):
             reformat = GuidedStorageTargetReformat(disk_id=disk.id)
```
ogayot, line 410: minor: this could be moved in the if block I think.

ogayot (approved):
LGTM

subiquity/server/controllers/filesystem.py
dbungert: Done, thanks.

PR #1364, opened 2022-07-21
examples: make the 'simple' mc disk bigger

dbungert:
Increase size of the disk in the simple machine config to 100 GiB.
This file is used heavily as a basic starting machine config and it
should be large enough for many types of tests.


dbungert:
@jpnurmi FYI

PR #1365, opened 2022-07-26
snapcraft: set contact field

dbungert:


murraybd (approved):
This is what I had in mind, thanks!

PR #1368, opened 2022-07-27
keyboard: lookup layout & variant from language

dbungert:


dbungert:
I also have some follow-on work with a script that can be used to generate a datafile, instead of building all that into the code.

mwhudson:
> I also have some follow-on work with a script that can be used to generate a datafile, instead of building all that into the code.

Could this go into the languagelist file? It would be good to at least find out in some obvious way if our list of translations and this list get out of sync.

mwhudson (approved):
I agree that this should be in a data file rather than code but this is good as far as it goes!

dbungert:
> Could this go into the languagelist file

I lean toward 'no' on that, for the reason that ubuntu-desktop-installer will probably have a different list of languages and I would prefer to not have to add stub languages to subiquity just so we can have keyboards for ubuntu-desktop-installer.

> It would be good to at least find out in some obvious way if our list of translations and this list get out of sync.

One of the unit tests does check the `languagelist` table, so if we add a language to `languagelist` then that unittest will complain.

PR #1370, opened 2022-07-29
keyboard: layout data & generation script

dbungert:


mwhudson (approved):
Hi, I have to say that the approach scripts/keyboard-scraper.sh takes is not something that had occurred to me but I like it. Did you consider running it at snap build time or do you think that's a bit risky?

I realise this is a total tangent / annoying extension of what you have done here, but subiquity/server/controllers/keyboard.py contains in latinizable a translation of another hunk of keyboard-configuration.config. Do you think there is a chance to make keyboard-scraper.sh output data for that too? I'm afraid I have completely forgotten how all of this works so I don't think I have any intuition for the answer to this...

dbungert:
> Hi, I have to say that the approach scripts/keyboard-scraper.sh takes is not something that had occurred to me but I like it. Did you consider running it at snap build time or do you think that's a bit risky?

Thanks!  My original assumption, before I sat down to actually script it, was that it was all a kind of nonsense hack and that I'd get it working just enough to get the suggested values.

It might be reasonably robust?  Maybe with a bit more testing I could convince myself that isn't going to break the build 20% of the time or something.  The tools and scripts it wants to run aren't changing every month.

> latinizable

Looks like this came from the STATE==4?  I suspect we can get that info out.

PR #1371, opened 2022-07-29
filesystem: use uuid4() for id generation

dbungert:


mwhudson:
Uhh. I mean I guess this will work but it will make the configs awfully hard to read. What's the motivation here?

dbungert:
partition IDs look almost like they tell you the partition number, but don't.
If the cure is worse than the disease I can reconsider, and we can just ask people to not display partition IDs or treat them as if the number on the end was anything meaningful.

mwhudson:
hmm yeah we definitely want clients to treat ids as opaque, that's for sure. But also being able to comprehend the generated curtin config is useful as well. Maybe having some random stuff in the middle of the id would make the point? Like part-c6b7443d-3 or something

dbungert:
https://github.com/canonical/subiquity/pull/1384

PR #1372, opened 2022-08-01
install: invoke curtin multiple times with different stages

ogayot:
_This PR is ready for review but I'm marking it a draft so that it does not get accidentally merged before https://code.launchpad.net/~ogayot/curtin/+git/curtin/+merge/427671 lands_

We now invoke curtin install in a step-by-step mode. Before, we would perform a single invocation of curtin install for the following stages:

* early (but no `early_commands` configured)
* partitioning
* extract
* curthooks
* hook
* late (but no `late_commands` configured)

We now run multiple invocations of curtin install, in 5 different steps. A curtin install step means one invocation of `$ curtin install`. In most cases, one step <=> one stage but if we want to, we can run multiple stages in one step (or zero stage like the _initial_ step does).

* _initial_: does not run any stage per se but prepares curtin for the next invocations (this invocation is not strictly necessary and could be merged with the next step but having it separate makes the flow easier to understand, I think)
* _partitioning_: runs the partitioning stage
* _extract_: runs the extract stage
* _curthooks_: runs the curthooks stage
* ~~_hook_: runs the hook stage~~

The early and late stages were dropped since they would act as no-ops.
We also dropped the hook stage since it does not do anything in Ubuntu.

For each step run, the following files will be available in /var/log (and copied on target):

| Desc                                      | Path                                                            |
|-------------------------------------------|-----------------------------------------------------------------|
| Curtin conf file - generated by Subiquity | ` /var/log/installer/curtin-install/subiquity-{step_name}.conf` |
| Curtin log file                           | ` /var/log/installer/curtin-install.log` (aggregate)           |
| Curtin error tarball (if any)             | ` /var/log/installer/curtin-errors.tar`      |

All files (i.e. configuration files, log files, error-files) are bundled in the apport report in case of crash.

### known limitations / areas of improvement:
- for all steps but _initial_, the same config is supplied to curtin (only the path to log files differ). The config is namely the result of a call to `SubiquityModel.render()` - which does not currently know what curtin config fragment goes to what step/stage. Later, it will be good that model declare what stage they would like their curtin config fragment to apply to upon calling `.render()`
- after a step runs successfully, we have the line `curtin: Installation finished.` present in its log file. In practice, the installation is really finished after subiquity runs the last step and I think it can confuse our users ; but curtin is not step agnostic.


subiquity/server/controllers/install.py
```diff
@@ -101,20 +114,37 @@ def tpath(self, *path):
...
 
-    def write_config(self):
+    def write_config(self, config_file: Path, config: Any) -> None:
+        """ Create a YAML file that represents the curtin install configuration
+        specified.  """
+        config_file.parent.mkdir(parents=True, exist_ok=True)
+        generate_config_yaml(str(config_file), config)
+
+    def acquire_generic_config(self, step: CurtinInstallStep,
+                               resume_data_file: Path) -> Dict[Any, Any]:
```
mwhudson, line 124: Pedantry in the extreme but I think all the keys in a curtin config are strings :-)

subiquity/server/controllers/install.py
```diff
@@ -130,12 +160,97 @@ async def configure_apt(self, *, context):
...
+
     @with_context(
         description="installing system", level="INFO", childlevel="DEBUG")
     async def curtin_install(self, *, context, source):
-        await run_curtin_command(self.app, context, 'install', source,
-                                 config=self.write_config(),
-                                 private_mounts=False)
+        # Until https://bugs.launchpad.net/curtin/+bug/1876984 gets
+        # fixed, the only way to get curtin to leave the network
+        # config entirely alone is to omit the 'network' stage.
```
mwhudson, line 196: I don't think this comment really belongs any more? We are moving to a model where we include things in stages rather than exclude them, I guess.

subiquity/server/controllers/install.py
```diff
@@ -130,12 +160,97 @@ async def configure_apt(self, *, context):
...
+
+        logs_dir_prefix = Path(
```
mwhudson, line 198: In general I would support using Path more in the subiquity codebase btw. There are a few things I don't really like about it but it's better than paths as strings...

subiquity/server/controllers/install.py
```diff
@@ -130,12 +160,97 @@ async def configure_apt(self, *, context):
...
+                error_file=logs_dir / "extract-error.tar",
+                acquire_config=self.acquire_generic_config,
+            ), CurtinInstallStep(
+                name="curthooks",
+                stages=["curthooks"],
+                config_file=logs_dir / "subiquity-curthooks.conf",
+                log_file=logs_dir / "curthooks.log",
+                error_file=logs_dir / "curthooks-error.tar",
+                acquire_config=self.acquire_generic_config,
+            ), CurtinInstallStep(
```
mwhudson, line 240: I think the hook stage only really does things for installing centos or other non-Ubuntu distros btw.

subiquity/server/controllers/install.py
```diff
@@ -130,12 +160,97 @@ async def configure_apt(self, *, context):
...
+                self.app.opts.output_base if self.app.opts.dry_run else "/")
+
+        logs_dir = logs_dir_prefix / "var/log/installer/curtin-install"
+        logs_dir.mkdir(parents=True, exist_ok=True)
+
+        # TODO close FD
+        unused, resume_data_path = tempfile.mkstemp()
+        resume_data_file = Path(resume_data_path)
+
+        await self.run_curtin_install_step(CurtinInstallStep(
```
mwhudson, line 208: So if you do what Dan and I are asking for in the curtin MP you won't need this bit?

subiquity/server/curtin.py
```diff
@@ -132,24 +130,48 @@ async def run(self, context):
...
+    }
 
     def make_command(self, command, *args, config=None):
         if command == 'install':
+            # Lookup the log file from the config if specified
+            try:
+                with open(config, mode="r") as fh:
+                    log_file = yaml.safe_load(fh)["install"]["log_file"]
+            except (TypeError, AttributeError):
+                log_file = "/dev/stderr"
```
mwhudson, line 148: Do we expect to hit this case? It feels a bit strange...

mwhudson (commented):
This seems mostly pretty good, I have some small comments.

Have you thought about a way to integrate the way we run "curtin apt-config" separately into this? It's not needed for the first implementation but in general we are going to want to separate the different curtin invocations more and have different configs for each invocation so a little thinking about how that is going to work might be worth it (I have to say, I don't have any immediate ideas on how this would work!)

subiquity/server/controllers/install.py
ogayot: Cool! I did not find too much info about this stage so decided to include it just in case. 
I will drop it if it's not useful for us. Thanks.

subiquity/server/controllers/install.py
ogayot: I agree. I don't love pathlib but I don't hate it either.

The thing that frustrates me the most is how difficult it is to "concatenate" two absolute paths without "losing" the first one . The following is pretty common in Subiquity:
```python
path1 = Path("/target")
path2 = Path("/var/log/installer")

path1 / path2
# Gives PosixPath('/var/log/installer') but we want PosixPath('/target/var/log/installer')
```

. To be fair, `os.path.join` suffers from the same thing but I was hoping `pathlib` would provide a simpler way to deal with that.

subiquity/server/curtin.py
ogayot: Only when writing a custom test-case (or unit test)... 
and it shows that this code is unused because I meant to use "/dev/null" instead of "/dev/stderr"..

subiquity/server/controllers/install.py
ogayot: Right, I'll drop this comment. Thanks.

subiquity/server/controllers/install.py
ogayot: Correct :)

subiquity/server/controllers/install.py
ogayot: I guess, yes.
Practically speaking, with the current implementation, I can already drop this step and make partitioning export the info needed to resume. It just felt nice (at least in my mind) to call this initialization step to make sure everything was OK (i.e., target is empty - that kind of stuff) before proceeding with the destructive operations.

I'll work on the update on curtin side and if it feels right, we can drop the initialization stage.

ogayot:
> This seems mostly pretty good, I have some small comments.
> 
> Have you thought about a way to integrate the way we run "curtin apt-config" separately into this? It's not needed for the first implementation but in general we are going to want to separate the different curtin invocations more and have different configs for each invocation so a little thinking about how that is going to work might be worth it (I have to say, I don't have any immediate ideas on how this would work!)

TBH, I haven't thought about this too much yet. For sure, the `install` function needs to be split as soon as we want the ability to retry and so on.

dbungert (approved):

subiquity/server/controllers/install.py
```diff
@@ -130,12 +162,92 @@ async def configure_apt(self, *, context):
         configurer = await mirror.wait_config()
         return await configurer.configure_for_install(context)
 
+    async def run_curtin_install_step(
```
mwhudson, line 165: I think it would be good to have each curtin step go through the context machinery, possibly by adding a @with_context decorator here?

subiquity/server/controllers/install.py
```diff
@@ -130,12 +162,92 @@ async def configure_apt(self, *, context):
...
-                                 private_mounts=False)
+        logs_dir_prefix = Path(
+                self.app.opts.output_base if self.app.opts.dry_run else "/")
+
+        logs_dir = logs_dir_prefix / "var/log/installer"
+        logs_dir.mkdir(parents=True, exist_ok=True)
+
+        config_dir = logs_dir / "curtin-install"
+
+        install_log_file = logs_dir / "curtin-install.log"
```
mwhudson, line 204: I'm curious why you chose to have all logs go to the same file?

subiquity/server/controllers/install.py
```diff
@@ -72,6 +73,18 @@ def feed(self, line):
...
+@attr.s(auto_attribs=True)
+class CurtinInstallStep:
+    """ Represents the parameters of a single step (i.e., invocation of curtin
+    install). """
+    name: str
+    stages: List[str]
+    config_file: Path
+    log_file: Path
+    error_file: Path
+    acquire_config: Callable[["CurtinInstallStep", Path], Dict[str, Any]]
```
mwhudson, line 85: Having a callable to get the config feels a bit complicated. But I guess that can be made use of or cleaned up as we continue this work.

mwhudson (approved):
This seems OK as a starting point. I guess I would like us to think about how the different models tie into the curtin invocations at some point (please don't consider anything about how this is done as particularly well-thought out!).

I guess we have INSTALL_MODEL_NAMES and POSTINSTALL_MODEL_NAMES currently, perhaps they need to be split into something more fine grained? Or maybe the models themselves should indicate which stage they are for? Or something, I dunno :)

mwhudson (approved):

subiquity/server/controllers/install.py
dbungert: He started with one log per stage, and I asked that we unify them.  I thought it would be easier for administrators to diagnose issues, instead of having to think about it in the 'stages'.

mwhudson: Fair enough :)

subiquity/server/controllers/install.py
ogayot: Good point, added a `with_context` to this function:

```
+    @with_context(
+        description="executing curtin install {step.name} step")
```

ogayot:
> I guess we have INSTALL_MODEL_NAMES and POSTINSTALL_MODEL_NAMES currently, perhaps they need to be split into something more fine grained? Or maybe the models themselves should indicate which stage they are for? Or something, I dunno :)

Yes, I think this is the right way forward

PR #1373, opened 2022-08-02
boot: do not place an ESP on a logical part

dbungert:


subiquity/common/filesystem/tests/test_manipulator.py
```diff
@@ -648,6 +648,35 @@ def test_add_boot_existing_boot_partition(self, bl, version):
             manipulator.add_boot_disk(disk)
         self.assertIsBootDisk(manipulator, disk)
 
+    def test_logical_no_esp_in_gap(self):
+        manipulator = make_manipulator(Bootloader.UEFI, storage_version=2)
+        m = manipulator.model
+        m._probe_data.setdefault('blockdev', {})
+        d1 = make_disk(m, preserve=True, ptable='msdos')
+        make_partition(m, d1, flag='extended', preserve=True)
+        make_partition(m, d1, flag='logical', preserve=True)
```
mwhudson, line 657: This relies rather on the behaviour of make_partition with different flags, which I had either forgotten about or didn't know about. I think the test should probably be explicit about this. (Also I guess make_partition isn't really robust about extended / logical partitions in general, like I think it would quite happily create a primary partition in an extended gap today?)

subiquity/common/filesystem/boot.py
```diff
@@ -221,10 +221,11 @@ def get_add_part_plan(device, *, spec, args, resize_partition=None):
 
     create_part_plan = CreatePartPlan(gap=None, spec=spec, args=args)
 
-    if gaps.largest_gap_size(device) >= size:
-        create_part_plan.gap = gaps.largest_gap(device).split(size)[0]
+    gap = gaps.largest_gap(device, in_extended=False)
```
mwhudson, line 224: I think this function needs a comment referencing that bug that makes logical ESPs fail.

mwhudson (approved):
This looks good, I'd like a test tweak and a comment please.

subiquity/common/filesystem/tests/test_manipulator.py
dbungert: > make_partition would quite happily create a primary partition in an extended gap today

Yes, yes it would

Fixed, and also fixed all cases of `make_partition(flag='extended')` - every scenario now specifies a size.

subiquity/common/filesystem/boot.py
dbungert: Done

PR #1378, opened 2022-08-05
boot: fix too many primaries case

dbungert:
Partial progress on making guided respect the number of primary partitions.

### DIFF
--- a/subiquity/common/filesystem/boot.py
+++ b/subiquity/common/filesystem/boot.py
@@ -246,7 +246,8 @@ def get_add_part_plan(device, *, spec, args, resize_partition=None):
             ])
     else:
         new_primaries = [p for p in partitions
-                         if not p.preserve and p.is_primary]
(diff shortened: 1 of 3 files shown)


mwhudson (approved):
There is more to go here, right? But this looks correct.

(I guess a looming corner case here is that guided_lvm needs to create more partitions than guided_direct?)

dbungert:
> There is more to go here, right? But this looks correct.
> 
> (I guess a looming corner case here is that guided_lvm needs to create more partitions than guided_direct?)

Yes, definitely part of the puzzle.
I have a branch started where I am building the knowledge for v2_guided_GET so that it can calculate the number of primary partitions we will need for a given scenario, so we can not offer the cases that would run out of primaries.  As you note that's a bit harder for the LVM case, as it creates one more partition.  That can be mitigated by possible creation of an extended.

PR #1379, opened 2022-08-05
filesystem: guided creates logicals if in_extended

dbungert:


subiquity/server/controllers/filesystem.py
```diff
@@ -150,27 +150,31 @@ def setup_disk_for_guided(self, target, mode):
...
         if gap is None:
-            return disk, gaps.largest_gap(disk)
+            return gaps.largest_gap(disk)
         else:
             # find what's left of the gap after adding boot
             gap = gaps.within(disk, gap)
             if gap is None:
                 raise Exception('failed to locate gap after adding boot')
-            return disk, gap
+            return gap
```
mwhudson, line 159: I wonder if this method still has the best name given the change in its signature here?

subiquity/server/controllers/filesystem.py
```diff
@@ -150,27 +150,31 @@ def setup_disk_for_guided(self, target, mode):
...
+
+    def guided_partition_creator(self, gap, spec):
+        flag = None
+        if gap.in_extended:
+            flag = 'logical'
+        return self.create_partition(
```
mwhudson, line 165: Aiee duelling abstractions I guess -- but maybe create_partition should handle (or at least check?) this logic?

mwhudson: That would let you delete this method, which has a funny name. This little corner doesn't feel completely thought out?

subiquity/common/filesystem/gaps.py
```diff
@@ -61,6 +61,8 @@ def split(self, size):
                         size=size,
                         in_extended=self.in_extended,
                         usable=self.usable)
+        if self.in_extended:
+            size += self.device.alignment_data().ebr_space
```
mwhudson, line 65: Good catch!

mwhudson (approved):
This is more or less fine, I have some quibbles about method names though :)

subiquity/server/controllers/filesystem.py
dbungert: OK, I will move this to `create_partition()`.  I actually started with that.
The lower-level `add_partition()` is a bit too clever for my liking (rounds up sizes!), so maybe we start to make `create_partition()` be the smart one.

subiquity/server/controllers/filesystem.py
dbungert: I have a refactor branch that I want to do soon (next?).  I have a slightly better name `setup_gap_for_guided`, but really the refactor branch should sort that out.

subiquity/common/filesystem/manipulator.py
```diff
@@ -78,8 +78,11 @@ def delete_filesystem(self, fs):
     delete_format = delete_filesystem
 
     def create_partition(self, device, gap, spec, **kw):
+        flag = kw.pop('flag', None)
+        if gap.in_extended:
+            flag = 'logical'
```
mwhudson, line 83: Should this yell at you if the passed flag is not None or logical?

dbungert: Yes, I think yelling about invalid flags is sensible.
But this could be any valid curtin partition flag I think?

dbungert: #1385

PR #1380, opened 2022-08-08
kbds: add layout info for u-d-i

dbungert:


kbds/keyboard-configuration.yaml
```diff
@@ -1,101 +1,248 @@
 # File generated by ./scripts/keyboard-scraper.sh, please do not edit manually
```
ogayot, line 1: Should we update/drop this line?

ogayot (approved):
LGTM
One minor observation

kbds/keyboard-configuration.yaml
dbungert: Yes, it's slightly incorrect now with
* more input data
* passing through pyyaml for normalization and merging

I will leave for now, as we are considering generating at build time.  It's still mostly correct.

dbungert:
@jpnurmi merged

kbds/keyboard-configuration.yaml
jpnurmi: Before I forget, this is how I constructed the full list of locales from the .arb files at https://github.com/canonical/ubuntu-flutter-plugins/tree/main/packages/ubuntu_localizations/lib/src/l10n

```bash
for arb in *.arb; do
    lang=$(jq -r '.["@@locale"]' $arb | cut -d_ -f1)
    country=$(jq -r '.["countryCode"]' $arb)
    echo "${lang}_${country}.UTF-8"
done
```

PR #1381, opened 2022-08-08
Execute run-parts on /etc/subiquity/postinst.d before executing any other late_commands

ogayot:
This PR extends the early / late / error controllers so that they can declare builtin commands.
Builtin commands are executed before other commands declared in the relevant autoinstall section.

In the late controller, we now declare a builtin command that runs `run-parts /etc/subiquity/postinst.d`. The desktop installer can use this and place executable scripts under `/etc/subiquity/postinst.d` to copy network-manager connections and do other things.

### DIFF
--- a/subiquity/server/controllers/cmdlist.py
+++ b/subiquity/server/controllers/cmdlist.py
@@ -15,7 +15,9 @@
 
 import asyncio
 import os
+from typing import List, Sequence, Union
 
+import attr
 from systemd import journal
 
 from subiquitycore.context import with_context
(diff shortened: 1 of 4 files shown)


subiquity/server/controllers/cmdlist.py
```diff
@@ -25,6 +27,22 @@
 from subiquity.server.controller import NonInteractiveController
 
 
+@attr.s(auto_attribs=True)
+class Command:
+    args: Union[str, Sequence[str]]
+    check: bool
+
+    def desc(self) -> str:
```
dbungert, line 35: I'm fine with this, just curious - was the built-in representation not great?

dbungert (approved):

subiquity/server/controllers/cmdlist.py
mwhudson: I think it's just that this way of formatting is what is associated with the "start run" event. It's visible to the user iirc, so probably should stay friendly (I wonder if we should run it though shlex.quote though?)

mwhudson (approved):
Seems fine -- I have the same question as Dan.

subiquity/server/controllers/cmdlist.py
ogayot: So, the installation screen displays the commands that are run. When commands are specified as an array of arguments e.g., `["ls", "-l", "/"]`, they are transformed into a string to be more human readable.

I did not change the existing behavior here but moved the implementation in the Command object so we don't need that many branches in the the `run` function.

```diff
-            if isinstance(cmd, str):
-                desc = cmd
-            else:
-                desc = ' '.join(cmd)
+            desc = cmd.desc()
```

I will add some docstrings to make this clearer.

ogayot: > (I wonder if we should run it though shlex.quote though?)

sure, let's `shlex.join` this. That will allow users/us to copy paste without losing information.

ogayot:
Added a `try ... except AttributeError` block in the initializer of LateController.
`system_setup` imports this controller as well and does not have a postinst script mechanism.

PR #1382, opened 2022-08-09
filesystem: add wait arg for v2_guided_GET

dbungert:


dbungert:
@jpnurmi

jpnurmi:
What about `GET /storage/v2?wait=true`? `StorageResponseV2` contains disks and partitions with `OsProber` that the GUI uses to display what OSes are installed.

subiquity/common/types.py
```diff
@@ -341,10 +341,14 @@ class StorageResponse:
 
 @attr.s(auto_attribs=True)
 class StorageResponseV2:
-    disks: List[Disk]
-    need_root: bool  # if True, there is not yet a partition mounted at "/"
-    need_boot: bool  # if True, there is not yet a boot partition
-    install_minimum_size: int
+    status: ProbeStatus
+    error_report: Optional[ErrorReportRef] = None
```
ogayot, line 345: `error_report` is defined twice in this class.
It's sad that we can't reliably use mypy with subiquity yet because it would catch this kind of thing.

subiquity/common/types.py
```diff
@@ -341,10 +341,14 @@ class StorageResponse:
...
+    disks: List[Disk] = attr.Factory(list)
+    # if need_root == True, there is not yet a partition mounted at "/"
+    need_root: Optional[bool] = None
+    # if need_boot == True, there is not yet a boot partition
+    need_boot: Optional[bool] = None
+    install_minimum_size: Optional[int] = None
```
ogayot, line 351: I'd usually tend to wrap the fields that depend on `status` in an object itself marked optional, as in:

```python
@attr.s(auto_attribs=True)
class StorageResponseDoneV2:
    disks: List[Disk]
    need_root: bool  # if True, there is not yet a partition mounted at "/"
    need_boot: bool  # if True, there is not yet a boot partition
    install_minimum_size: int

@attr.s(auto_attribs=True)
class StorageResponseV2:
    status: ProbeStatus
    probe_info: Optional[StorageResponseDoneV2]
    error_report: Optional[ErrorReportRef] = None
```
would it make sense here? I know it breaks backward compatibility for the desktop team but I'm sure it's not a problem since they will add the `wait=True` parameter to their existing invocations anyway.

ogayot (changes requested):
Thanks.
There is a double definition of error_report that I would like to get fixed before merging.
Otherwise, LGTM. I added a suggestion but feel free to ignore it if you like be better the way it is now.

subiquity/common/types.py
dbungert: Well that was silly, thanks!

subiquity/common/types.py
dbungert: Yes, adding two elements to several return types feels a bit awkward.
I think I want to reverse that, and keep the main data the the caller is interested in at the top level.
Something like:
```
@attr.s(auto_attribs=True)
class ProbeResult:
    status: ProbeStatus
    error_report: Optional[ErrorReportRef] = None

@attr.s(auto_attribs=True)
class StorageResponseV2:
    result: ProbeResult
    disks: Optional[List[Disk]]
    ...
```
Then `_probe_response` can just return the result object, and the caller can handle the rest (and even partially populate the parent object instead of using defaults).
Will add to my future-PR list.

ogayot (approved):

PR #1383, opened 2022-08-09
guided: singledispatch and other refactors

dbungert:
* simplify `guided()` with singledispatch
* move gaps.within to a gap object method
* add warning about autoinstall having layout and manual config
* expand a side-by-side unit test to also cover gpt

### DIFF
--- a/subiquity/common/filesystem/gaps.py
+++ b/subiquity/common/filesystem/gaps.py
@@ -70,6 +70,17 @@ class Gap:
                        usable=self.usable)
         return (first_gap, rest_gap)
 
+    def within(self):
+        """Find the first gap that is contained wholly inside the supplied
+        gap."""
+        gap_end = self.offset + self.size
+        for pg in parts_and_gaps(self.device):
+            if isinstance(pg, Gap):
+                pg_end = pg.offset + pg.size
+                if pg.offset >= self.offset and pg_end <= gap_end:
+                    return pg
+        return None
+
 
 @functools.singledispatch
 def parts_and_gaps(device):
@@ -254,17 +265,6 @@ def at_offset(device, offset):
     for pg in parts_and_gaps(device):
         if isinstance(pg, Gap):
             if pg.offset == offset:
-                return pg
-    return None
-
-
-def within(device, gap):
-    """Find the first gap that is contained wholly inside the supplied gap."""
-    gap_end = gap.offset + gap.size
-    for pg in parts_and_gaps(device):
-        if isinstance(pg, Gap):
-            pg_end = pg.offset + pg.size
-            if pg.offset >= gap.offset and pg_end <= gap_end:
                 return pg
     return None
 
--- a/subiquity/common/filesystem/gaps.py
+++ b/subiquity/common/filesystem/gaps.py
@@ -71,8 +71,7 @@ class Gap:
         return (first_gap, rest_gap)
 
     def within(self):
-        """Find the first gap that is contained wholly inside the supplied
-        gap."""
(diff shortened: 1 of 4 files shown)


subiquity/common/filesystem/gaps.py
```diff
@@ -70,6 +70,17 @@ def split(self, size):
                        usable=self.usable)
         return (first_gap, rest_gap)
 
+    def within(self):
+        """Find the first gap that is contained wholly inside the supplied
```
mwhudson, line 74: "inside this gap" i guess?

subiquity/server/controllers/filesystem.py
```diff
@@ -205,48 +184,65 @@ def guided_lvm(self, target, mode=None, lvm_options=None):
                 mount="/",
                 ))
 
+    @functools.singledispatchmethod
+    def start_guided(self, target):
```
mwhudson, line 188: I think a docstring and a return type annotation would be good here.

mwhudson: Docstrings on the implementations would be nice too, perhaps.

subiquity/server/controllers/filesystem.py
```diff
@@ -205,48 +184,65 @@ def guided_lvm(self, target, mode=None, lvm_options=None):
...
-            if target is None:
-                pgs = gaps.parts_and_gaps(disk)
-                raise Exception(f'gap not found after resize, pgs={pgs}')
-        else:
-            raise Exception(f'Unknown guided target {choice.target}')
+        gap = self.start_guided(choice.target, disk)
+        if DeviceAction.TOGGLE_BOOT in DeviceAction.supported(disk):
+            self.add_boot_disk(disk)
+        # find what's left of the gap after adding boot
+        gap = gap.within()
```
mwhudson, line 237: Would it make sense to move the responsibility for calling add_boot_disk into start_guided? (maybe a helper to check supported() first I guess...) This thought was triggered by thinking about how this would work for a guided method that would assemble a bunch of disks into a RAID (where you'd want to do something like reformat each disk, call add_boot_disk on each disk, create a partition in the remaining space, create a raid out of those partitions and return the gap for that raid...) but also I think it would let you get rid of the within() games?

mwhudson (approved):
Hi, this is a really nice refactor, thanks. I have a suggestion for a small change but feel free to ignore it!

subiquity/server/controllers/filesystem.py
dbungert: Yes, I'd love to not have `gap.within()`.

I had an earlier branch that let the boot logic return a gap, I didn't like the resulting code though.  I want to jot this down for future re-exploration.

subiquity/server/controllers/filesystem.py

subiquity/common/filesystem/gaps.py
dbungert: Done

subiquity/server/controllers/filesystem.py
mwhudson: I don't think that's required (although it might be nicer), it would just be changing start_guided_reformat to:

```
    @start_guided.register
    def start_guided_reformat(self, target: GuidedStorageTargetReformat,
                              disk: ModelDisk) -> gaps.Gap:
        """Perform the reformat, and return the resulting gap."""
        self.reformat(disk, wipe='superblock-recursive')
        self.add_boot_disk_if_supported(disk)
        return gaps.largest_gap(disk)
```

I think?

mwhudson (approved):
Meant to select approve still despite my continued quibbling

PR #1384, opened 2022-08-09
filesystem: part ids are not part nums

dbungert:
The filesystem object IDs are really only meant to be a semireadable
unique string, and not for display to end users.  Partitions in
particular were bad in this respect, as they could look like the number
on the end was the partition number.

Uglify the object IDs a bit to try to convey the above.


mwhudson (approved):
Thanks for this.

PR #1386, opened 2022-08-12
filesystem: fix wipe on v2 edit_partition

dbungert:


dbungert:
@jpnurmi - meant to address LP: #1983036, as observed with ubuntu-desktop-installer

subiquity/common/filesystem/tests/test_manipulator.py
```diff
@@ -91,6 +92,25 @@ def test_delete_encrypted_vg(self):
...
+        p = make_partition(manipulator.model, d, preserve=True)
+        fs = make_filesystem(manipulator.model, partition=p, fstype='ext2',
+                             preserve=True)
+        manipulator.model._actions.append(fs)
+        manipulator.delete_filesystem(fs)
+        spec = {'wipe': 'random'}
+        with mock.patch.object(p, 'original_fstype') as original_fstype:
+            original_fstype.return_value = 'ext2'
+            fs = manipulator.create_filesystem(p, spec)
+        self.assertFalse(p.preserve)
```
mwhudson, line 109: This is all very confusing (the gyrations you have to perform in the setup phase of this test is a bad smell!) but I don't think p.preserve should be false after the above sequence of operations? If you take an existing partition and reformat it, the _partition_ shouldn't be destroyed and recreated  -- you would expect the uuid to be preserved, for example. The partition should be wiped though, for sure!

subiquity/common/filesystem/manipulator.py
```diff
@@ -48,15 +48,15 @@ def create_filesystem(self, volume, spec):
         if spec.get('fstype') is None:
             # prep partitions are always wiped (and never have a filesystem)
             if getattr(volume, 'flag', None) != 'prep':
-                volume.wipe = None
+                volume.wipe = spec.get('wipe', None)
             fstype = volume.original_fstype()
             if fstype is None:
                 return None
-            preserve = True
+            volume.preserve = preserve = volume.wipe is None
```
mwhudson, line 55: Ah yeah, not sure about this change?

mwhudson (commented):
Not sure about the change to the partition's preserve flag here?

subiquity/common/filesystem/tests/test_manipulator.py
dbungert: Ah yes, the curtin behavior with preserve == True and a wipe value.  I have to admit I found that confusing the first time I came across it as well, I think I just have to accept that this is how curtin works and it's OK.

Fixed!

subiquity/common/filesystem/manipulator.py
dbungert: I believe the new code is correct.  If you still have concerns then we may have a different vision for this function and should discuss..

corradoventu:
Installing with the old installer (ubiquity) the screen 'Edit partition' allows formatting the partition but with subiquity the checkbox 'format' is grayed.
edit partition

subiquity/common/filesystem/manipulator.py
```diff
@@ -48,15 +48,14 @@ def create_filesystem(self, volume, spec):
         if spec.get('fstype') is None:
             # prep partitions are always wiped (and never have a filesystem)
             if getattr(volume, 'flag', None) != 'prep':
```
mwhudson, line 50: I always find this conditional very confusing somehow. It's just "if the partition is not prep" right? ... Actually, why would this function ever be called for a prep partition? Maybe this should just be an assert or something.

subiquity/common/filesystem/manipulator.py
```diff
@@ -48,15 +48,14 @@ def create_filesystem(self, volume, spec):
...
             fstype = volume.original_fstype()
             if fstype is None:
                 return None
-            preserve = True
         else:
             fstype = spec['fstype']
-            volume.wipe = 'superblock'
-            preserve = False
+            volume.wipe = spec.get('wipe', None)
+        preserve = volume.wipe is None
```
mwhudson, line 58: I think we need to talk about this!

mwhudson (commented):
Let's have a proper chat about this tomorrow.

jpnurmi:
> Installing with the old installer (ubiquity) the screen 'Edit partition' allows formatting the partition but with subiquity the checkbox 'format' is grayed.

After these fixes have landed, what will be the rules for when a partition is allowed to be wiped? Are partitions implicitly wiped under certain conditions meaning the wipe-field cannot be unset? Are preserved partitions allowed to be wiped?

corradoventu:
I may install on a partition already used by Ubuntu, I may skip formatting to keep my home or force format to remove all.
also if I have a separate home partition i don't want formatting.

dbungert:
> > Installing with the old installer (ubiquity) the screen 'Edit partition' allows formatting the partition but with subiquity the checkbox 'format' is grayed.
> 
> After these fixes have landed, what will be the rules for when a partition is allowed to be wiped? 

The edit routine is a bit overcomplicated right now.  Here are the rules once this is merged and I'd like to clean this up to make it all quite a bit more obvious
* 3 fields may be modified: format, mount, and now wipe
* previously, it tried to be smart based on the format value and manage wipe
* wipe should be explicitly set when desired, and always set for changing the format
* mount works exactly as before
* format is a bit different (the above need to set wipe when changing format)
* wipe has the same values as documented in [curtin](https://curtin.readthedocs.io/en/latest/topics/storage.html).  The simple case is to use 'superblock' for a basic wipe, or null for no wipe.

> Are partitions implicitly wiped under certain conditions meaning the wipe-field cannot be unset?

That used to be the case, but no longer.  Clients should have full control of this value now.

> Are preserved partitions allowed to be wiped?

Yes!  A preserved partition retains other properties (offset, size, flags).

> I may install on a partition already used by Ubuntu, I may skip formatting to keep my home or force format to remove all.  also if I have a separate home partition i don't want formatting.

I expect with this PR that this is working correctly.  For instance, maybe the user marks a partition to be reformatted, then later unchecks that format checkbox, the end result is that the partition is not formatted (wipe is null).

subiquity/common/filesystem/manipulator.py
dbungert: Leaving alone for this PR

subiquity/common/filesystem/tests/test_manipulator.py
```diff
@@ -91,6 +92,25 @@ def test_delete_encrypted_vg(self):
...
+        # LP: #1983036 - edit a partition to wipe it and mount it, but not
+        # actually change the fs type already there
+        # example: it's ext4, wipe and make it ext4 again
+        manipulator, d = make_manipulator_and_disk()
+        p = make_partition(manipulator.model, d, preserve=True)
+        fs = make_filesystem(manipulator.model, partition=p, fstype='ext2',
+                             preserve=True)
+        manipulator.model._actions.append(fs)
+        manipulator.delete_filesystem(fs)
+        spec = {'wipe': 'random'}
```
mwhudson, line 105: I think to match our discussion, this spec should have 'fstype': 'ext2'?

subiquity/tests/api/test_api.py
```diff
@@ -1230,6 +1230,111 @@ async def test_osprober_knames(self):
             }
             self.assertEqual(expected, nvme_p2['os'])
 
+    @timeout()
+    async def test_edit_should_trigger_wipe_when_requested(self):
+        # LP: #1983036 - a partition wipe was requested but didn't happen
+        # 1) the wipe value was ignored, as the format key was used instead
+        # 2) the format wasn't actually changed, so the partition was
+        #    considered preserve=True
```
mwhudson, line 1238: I think this comment is out of date now?

mwhudson (approved):
So we talked about this and agreed that this change is correct. I have a couple of very small comments about tests.

IIRC the next steps we agreed on to reduce the general level of confusion around all this:

 * Stop passing spec dicts around everywhere, instead define a dataclass / attrs class for each one
 * Change the server TUI to stop using fstype == None to indicate "use original fs" and instead use wipe like the v2 methods do
 * Stop deleting and recreating format actions all the time

mwhudson:
Oh one more comment, if the client sets wipe to false but sets a fstype that does not match the original fstype, that's an error right? Should we try to catch that?

subiquity/tests/api/test_api.py
dbungert: Updated, thanks.

subiquity/common/filesystem/tests/test_manipulator.py
dbungert: Comment matches the code now.  I did mean for it to say 'ext2', as the default is 'ext4' I wanted to use a different value in the test.

dbungert:
> Oh one more comment, if the client sets wipe to false but sets a fstype that does not match the original fstype, that's an error right? Should we try to catch that?

Enforced now.

dbungert:
Hit the re-review button since there is a logic change on the last bit.

subiquity/server/controllers/filesystem.py
```diff
@@ -531,11 +531,15 @@ async def v2_edit_partition_POST(self, data: ModifyPartitionV2) \
                 and data.partition.boot != partition.boot:
             raise ValueError('edit_partition does not support changing boot')
         spec = {'mount': data.partition.mount or partition.mount}
-        if data.partition.format is not None \
-                and data.partition.format != partition.format:
+        if data.partition.format is not None:
+            if data.partition.format != partition._fs.fstype:
```
mwhudson, line 535: This needs to check the original fs type not the current one I think?

dbungert: surely this time

mwhudson (approved):
gosh this took longer than I or (presumably) you expected!

corradoventu:
Current ubuntu-desktop-installer version is 613, in which version will there be this change?

PR #1387, opened 2022-08-12
Updates for Ubuntu-Pro with magic-attach

ogayot:
Apologies for the size of the diff :/

This adds the support for magic-attach / contract-selection in Subiquity.

Unfortunately, testing this PR on a current ubuntu server ISO will fail because we are relying on an unreleased feature of ubuntu-advantage-tools. Also the ubuntu.com/pro/attach website is not yet available for the user to do the contract selection.

The dry-run implementation exists but is limited. 
Upon reaching the Ubuntu Pro screen, the user is presented the user-code `AAAAAA`.
The user-code expires after 10 minutes (affected by `SUBIQUITY_REPLAY_TIMESCALE`).
When the user-code expires, a new contract selection is initiated and the next user-code (`AAAAAB`) will be presented instead,
The user has no way to select a contract in dry-run mode so the user-code will keep expiring until the user inputs a contract token instead.

More integration testing can be done by following the steps below:

* build the development version of ubuntu-advantage-tools
  * checkout https://github.com/canonical/ubuntu-advantage-client.git
  * `sbuild -d kinetic`
* deploy a LXD container (ubuntu 22.04 for instance)
* install the .deb built previously on the container
* apply the following diff to subiquity (after replacing "pleasant-griffon" with the name of the LXD container):
```diff
diff --git a/subiquity/server/controllers/ubuntu_pro.py b/subiquity/server/controllers/ubuntu_pro.py
index 831ae635..5e2c0f39 100644
--- a/subiquity/server/controllers/ubuntu_pro.py
+++ b/subiquity/server/controllers/ubuntu_pro.py
@@ -89,15 +89,11 @@ class UbuntuProController(SubiquityController):
     def __init__(self, app) -> None:
         """ Initializer for server-side Ubuntu Pro controller. """
         strategy: UAInterfaceStrategy
-        if app.opts.dry_run:
-            strategy = MockedUAInterfaceStrategy(scale_factor=app.scale_factor)
-        else:
-            # Make sure we execute `$PYTHON "$SNAP/usr/bin/ubuntu-advantage"`.
-            executable = (
-                os.environ["PYTHON"],
-                os.path.join(os.environ["SNAP"], "usr/bin/ubuntu-advantage"),
-            )
-            strategy = UAClientUAInterfaceStrategy(executable=executable)
+        # Make sure we execute `$PYTHON "$SNAP/usr/bin/ubuntu-advantage"`.
+        executable = (
+            "lxc", "exec", "pleasant-griffon", "--", "/usr/bin/ubuntu-advantage",
+        )
+        strategy = UAClientUAInterfaceStrategy(executable=executable)
         self.ua_interface = UAInterface(strategy)
         self.cs: Optional[ContractSelection] = None
         self.magic_token = Optional[str]
```
* the contracts CLI tool can be used to simulate what the user could do on ubuntu.com/pro/attach

### DIFF
--- a/subiquity/server/contract_selection.py
+++ b/subiquity/server/contract_selection.py
@@ -1,0 +1,123 @@
+# Copyright 2022 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+""" Module that deals with contract token selection. """
+
+import asyncio
+import logging
+from typing import Any, List
+import time
+
+from subiquity.server.ubuntu_advantage import (
+    UAInterface,
+    APIUsageError,
+    )
+
+from subiquitycore.async_helpers import schedule_task
+
+
+log = logging.getLogger("subiquity.server.contract_selection")
+
+
+class UPCSExpiredError(Exception):
+    """ Exception to be raised when a contract selection expired. """
+
+
+class UnknownError(Exception):
+    """ Exception to be raised in case of unexpected error. """
+    def __init__(self, message: str = "", errors: List[Any] = []) -> None:
+        self.errors = errors
+        super().__init__(message)
+
+
+class ContractSelection:
+    """ Represents an already initiated contract selection. """
+    def __init__(
+            self,
+            client: UAInterface,
+            magic_token: str,
+            user_code: str,
+            validity_seconds: int) -> None:
+        """ Initialize the contract selection. """
+        self.client = client
+        self.magic_token = magic_token
+        self.user_code = user_code
+        self.validity_seconds = validity_seconds
+        self.task = asyncio.create_task(self._run_polling())
+
+    @classmethod
+    async def initiate(cls, client: UAInterface) \
+            -> "ContractSelection":
+        """ Initiate a contract selection and return a ContractSelection
+        request object. """
+        answer = await client.strategy.magic_initiate_v1()
+
+        if answer["result"] != "success":
+            raise UnknownError(errors=answer["errors"])
+
+        return cls(
+                client=client,
+                magic_token=answer["data"]["attributes"]["token"],
+                validity_seconds=answer["data"]["attributes"]["expires_in"],
+                user_code=answer["data"]["attributes"]["user_code"])
+
+    async def _run_polling(self) -> str:
+        """ Runs the polling and eventually return a contract token. """
+        # Wait an initial 30 seconds before sending the first request, then
+        # send requests at regular interval every 10 seconds.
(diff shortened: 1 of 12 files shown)


subiquity/server/contract_selection.py
```diff
@@ -0,0 +1,123 @@
...
+
+        return cls(
+                client=client,
+                magic_token=answer["data"]["attributes"]["token"],
+                validity_seconds=answer["data"]["attributes"]["expires_in"],
+                user_code=answer["data"]["attributes"]["user_code"])
+
+    async def _run_polling(self) -> str:
+        """ Runs the polling and eventually return a contract token. """
+        # Wait an initlal 30 seconds before sending the first request, then
```
dbungert, line 77: minor typo

subiquity/ui/views/ubuntu_pro.py
```diff
@@ -290,7 +391,12 @@ def upgrade_yes_no_screen(self) -> Widget:
         +---------------------------------------------------------+
         """
 
-        excerpt = _("Upgrade this machine to Ubuntu Pro or skip this step.")
+        excerpt = _("Upgrade this machine to Ubuntu Pro for security updates"
+                    " on a much wider range of packages, until 2032. Assists"
```
dbungert, line 395: currentyear + 10?

subiquity/ui/views/ubuntu_pro.py
```diff
@@ -598,16 +759,17 @@ def __init__(self, parent: UbuntuProView) -> None:
...
 
         services = [
-            _("Security patching for over 30.000 packages, with a focus on"
-              " High and Critical CVEs (extended from 2.500)"),
-            _("10 years of security Maintenance (extended from 5 years)"),
-            _("Kernel Livepatch service for increased uptime and security"),
-            _("Ubuntu Security Guide for hardening profiles, including CIS"
-              " and DISA-STIG"),
+            _("Security patch coverage for CVSS critical, high and selected"
+              ' medium vulnerabilities in all 23,000 packages in "universe"'
```
dbungert, line 768: Would be nice to pull these numbers out of the translated string, they are likely to change.

subiquity/tests/api/test_api.py
```diff
@@ -1320,3 +1320,30 @@ async def test_create_primaries(self):
             }
             with self.assertRaises(ClientResponseError):
                 await inst.post('/storage/v2/add_partition', data)
+
+
+class TestUbuntuProContractSelection(TestAPI):
+    async def test_upcs_flow(self):
```
dbungert, line 1326: `@timeout` please

subiquity/server/contract_selection.py
ogayot: Nice catch, I couldn't see it ever after you said it :+1:

subiquity/ui/views/ubuntu_pro.py
ogayot: Yep, makes sense. Ideally, those numbers should be dependent on the version of Ubuntu being installed but I don't see a simple way to do that.

subiquity/ui/views/ubuntu_pro.py
ogayot: I feel like that would give a wrong information to the user, wouldn't it?

If somebody attempts an install of Jammy in say 2025, he would get "until 2035". Sadly, this is probably another instance of an information that should be dependent on the release being installed.

dbungert: /usr/share/distro-info/ubuntu.csv from package distro-info-data ?

dbungert: To not complicate this PR further, let's pull the number out of the string and deal with getting better numbers in a future PR.

dbungert (approved):

PR #1390, opened 2022-08-22
[WSL] Enables users to skip installing language packs

CarlosNihelton:
All this chunk of code because of one boolean flag.

The motivation for this work is detailed in https://github.com/canonical/ubuntu-desktop-installer/issues/1073 and https://github.com/ubuntu/WSL/issues/174. In short, 
setting up Ubuntu on WSL 2 with our OOBE may take a while, specially on slower connections or machines. Also, WSL 2 is known by having some internal issue limiting the internet speed when compared to the host system. That may turn the moment of installing language packs into an eternity.

I can foresee opportunity for extending the customisability of our setup experience with other options that won't reflect in the `/etc/wsl.conf` (where the options set in the types `WSLConfBase` and `WSLConfAdvanced` currently are writtent to). So I chose to create a new type and endpoint for this kind of customization. Since the endpoint is specific for `system_setup` I had to change `subiquity/tests/api/test_api.py` to let me start a `system_setup` server, so I could exercise that endpoint.

I tried rendering a CheckBox for that just below the languages listbox, but that hurts usability, because one would need to visit all the elements inside the list box to be able to reach the flag in the bottom of the screen. Placing it above would hurt the existing look. So, a new page is required, even though the page contains only one boolean flag. For now, at least. I also took the freedom to cache the native language string into the app, so other pieces of the UI can later reuse it. That enables referring to the language in the CheckBox caption, for instance (per this recommendation https://github.com/canonical/ubuntu-desktop-installer/issues/1073#issuecomment-1223914664). This is how it looks:

image

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -133,6 +133,7 @@ class SubiquityClient(TuiApplication):
         self.server_updated = None
         self.restarting = False
         self.global_overlays = []
+        self.native_language = ""
 
         try:
             self.our_tty = os.ttyname(0)
--- a/subiquity/client/controllers/welcome.py
+++ b/subiquity/client/controllers/welcome.py
@@ -40,6 +40,10 @@ class WelcomeController(SubiquityTuiController):
             self.done((self.answers['lang'], ""))
 
     def done(self, lang):
+        """ Completes this controller. lang must be a tuple of strings
+        containing the language code and its native representation
+        respectively.
+        """
         (code, native) = lang
         log.debug("WelcomeController.done %s next_screen", code)
         i18n.switch_language(code)
--- a/subiquity/client/controllers/welcome.py
+++ b/subiquity/client/controllers/welcome.py
@@ -37,11 +37,13 @@ class WelcomeController(SubiquityTuiController):
 
     def run_answers(self):
         if 'lang' in self.answers:
-            self.done(self.answers['lang'])
+            self.done((self.answers['lang'], ""))
 
-    def done(self, code):
+    def done(self, lang):
+        (code, native) = lang
         log.debug("WelcomeController.done %s next_screen", code)
         i18n.switch_language(code)
+        self.app.native_language = native
         self.app.next_screen(self.endpoint.POST(code))
 
     def cancel(self, sender=None):
--- a/subiquity/client/controllers/welcome.py
+++ b/subiquity/client/controllers/welcome.py
@@ -44,10 +44,11 @@ class WelcomeController(SubiquityTuiController):
         containing the language code and its native representation
         respectively.
         """
-        (code, native) = lang
+        # ('de_DE.UTF-8', 'Deutsch')
+        (code, display_name) = lang
         log.debug("WelcomeController.done %s next_screen", code)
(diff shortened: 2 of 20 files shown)


subiquity/client/controllers/welcome.py
```diff
@@ -37,11 +37,13 @@ async def make_ui(self):
 
     def run_answers(self):
         if 'lang' in self.answers:
-            self.done(self.answers['lang'])
+            self.done((self.answers['lang'], ""))
 
-    def done(self, code):
+    def done(self, lang):
+        (code, native) = lang
```
dbungert, line 43: I'm not going to remember what this means, would you do a quick comment showing what the values would look like in intended usage?

dbungert (changes requested):

system_setup/client/controllers/wslsetupoptions.py
```diff
@@ -0,0 +1,48 @@
...
+
+log = logging.getLogger('system_setup.client.controllers.wslsetupoptions')
+
+
+class WSLSetupOptionsController(SubiquityTuiController):
+    endpoint_name = 'wslsetupoptions'
+
+    async def make_ui(self):
+        data = await self.endpoint.GET()
+        log.debug("%s", self.app)
```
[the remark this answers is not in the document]
CarlosNihelton: That was a leftover between rebases.

Thanks for spotting that out.

subiquity/ui/views/welcome.py
```diff
@@ -98,9 +98,10 @@ def make_language_choices(self):
             buttons=[back] if back else None,
             excerpt=_("Use UP, DOWN and ENTER keys to select your language."))
 
-    def choose_language(self, sender, code):
+    def choose_language(self, sender, lang):
+        (code, _) = lang
```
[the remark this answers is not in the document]
CarlosNihelton: Gotcha.

subiquity/client/client.py
```diff
@@ -133,6 +133,7 @@ def __init__(self, opts, about_msg=None):
         self.server_updated = None
         self.restarting = False
         self.global_overlays = []
+        self.native_language = ""
```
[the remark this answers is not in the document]
CarlosNihelton: It looks like it will generate a major refactoring, since locale currently manages a string, and that will require adding a type. I'm all for it, but I guess that would better the done in a separate pull request.

subiquity/client/controllers/welcome.py
CarlosNihelton: Done.

dbungert: Kind of the problem here is that 'native' is not at all clear.  I was hoping to list an example something like `('de_DE.UTF-8', 'Deutsch')`, that the code is the locale value that we wish to set and that native refers to the translated display name of the language.

My goal here is that the next reader doesn't have to backtrack this all the way to the languagelist parser to know what the native value looks like.

CarlosNihelton: I see your point. Thanks for clarifying it further.

dbungert (approved):

PR #1392, opened 2022-08-23
integration: speed up check for listening interfaces in system_setup

ogayot:
This speeds up the system_setup test that makes sure only the loopback interface listens to TCP connections.

The test used to take at most `10 * 2 * number_interfaces` seconds. On my machine, it took over one minute.

We now run all `curl` calls in parallel so it takes 10 seconds at most.

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -248,38 +248,10 @@ if [ "${RELEASE%.*}" -ge 20 ]; then
         kill $subiquity_pid || true
         exit 1
     fi
-    loopback_failed=0
-    unallowed_failed=0
-    # Assert that only loopback interface is accepted.
-    interfaces=($(ip --json link show up | jq -r '.[]["ifname"] | select ( . != null )'))
-    for if in ${interfaces[@]}; do
-        for ipv in 4 6; do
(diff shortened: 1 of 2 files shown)


dbungert (approved):
2 minute speedup on `make integration` for me, thank you!

scripts/runtests.sh
```diff
@@ -248,38 +248,10 @@ if [ "${RELEASE%.*}" -ge 20 ]; then
...
-        exit 1
-    fi
-    if [ $unallowed_failed -ne 0 ]; then
-        echo "Only the loopback interface should be allowed."
-        exit 1
-    fi
+
+    scripts/test-system-setup-loopback-only.py --port "$port" --debug
+
+    kill -- "$subiquity_pid" || true
```
dbungert, line 254: oh right, we want to ensure the subiquity process is killed off, even in case of error, so that might need a tweak

CarlosNihelton (approved):
Beautiful!

scripts/runtests.sh
dbungert: I'll handle this in a follow-on PR since @ogayot can't at the moment.

PR #1393, opened 2022-08-24
Makes our gray ligther to increase contrast.

CarlosNihelton:
Per Design's recommendation in:
https://github.com/canonical/ubuntu-desktop-installer/issues/1073#issuecomment-1225569053

`#808080` choosen due its value being part of the xterm256-color color space.


dbungert (approved):
Thanks, that is nicer contrast.
I like that we're using an xterm256 color, I think that should be the future standard (most of the palette colors don't match that yet).

PR #1394, opened 2022-08-24
Defaults help text style to gray

CarlosNihelton:
Per https://github.com/canonical/ubuntu-desktop-installer/issues/1073#issuecomment-1224493840 help texts should be styled in gray color.

With those canges in place one can still apply custom styles. Otherwise the gray is defaulted accross the board.

### DIFF
--- a/subiquitycore/ui/form.py
+++ b/subiquitycore/ui/form.py
@@ -249,7 +249,7 @@ class BoundFormField(object):
             return self._help
         elif self.field.help is not None:
             if isinstance(self.field.help, str):
-                return _(self.field.help)
+                return ('info_minor', _(self.field.help))
             else:
                 return self.field.help
         else:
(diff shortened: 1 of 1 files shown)


subiquitycore/ui/form.py
```diff
@@ -304,7 +304,11 @@ class FormField(abc.ABC):
 
     def __init__(self, caption=None, help=None):
         self.caption = caption
-        self.help = help
+        # Allows styling at instantiation
+        if help is None or isinstance(help, tuple) or help is NO_HELP:
+            self.help = help
+        else:
+            self.help = ('info_minor',  help)
```
dbungert, line 311: My guess is that every time we pass in a string, we want to apply the 'info_minor', and every other case we can just assign to self.help.  What do you think about:

```suggestion
        if isinstance(help, str):
            self.help = ('info_minor',  help)
        else:
            self.help = help
```

CarlosNihelton: Yeah, I don't see any other use case. Let's go with the suggestion!

dbungert (approved):

mwhudson (approved):

PR #1395, opened 2022-08-24
cloudinit: disable cloud-init after final install completes

blackboxsw:
Provide a write_files deferred artifact to disable cloud-init in the installed machine after final reboot completes.

NOTE:
 - This artifact will disable cloud-init after final reboot in both desktop and server images. 
   If we want to ensure cloud-init server images remain active after reboot (because typical server image behavior)
   we can ensure this only is written on desktop installs.
- Further discussion on this is that we want to have server and desktop disable cloud-init post-install.
- public mailinglist discussion around what to do with stable releases like Jammy for server live install behavior to confirm we aren't breaking use-cases with a behavior change


subiquity/models/subiquity.py
```diff
@@ -295,6 +306,10 @@ def _cloud_init_config(self):
             if getattr(model, 'make_cloudconfig', None):
                 merge_config(config, model.make_cloudconfig())
         merge_cloud_init_config(config, self.userdata)
+        if config.get("write_files"):
+            config["write_files"].append(CLOUDINIT_DISABLE_AFTER_INSTALL)
+        else:
+            config["write_files"] = [CLOUDINIT_DISABLE_AFTER_INSTALL]
```
dbungert, line 312:
```suggestion
        config.setdefault("write_files", []).append(CLOUDINIT_DISABLE_AFTER_INSTALL)
```

dbungert (approved):

blackboxsw (commented):
Manual integration test validates behavior:
- valid autoinstall.yaml on kinetic
- not tracebacks during install
- no tracebacks in cloud-init in ephemeral environment or installed system
- cloud-init status --long reports  disabled for next boot


## terminal 1
```

$ mkdir /tmp/autoinstall; cd /tmp/autoinstall

$ cat > /tmp/autoinstall/user-data.VALID <<EOF
#cloud-config
# Valid cloud-config in autoinstall.user-data specifying correct list value for ssh_import_id
autoinstall:
    version: 1
    identity:
        hostname: ai-test
        password: "/2.b4/b.ZRTIZM30JZ4QrOQ2aOXJ8yk96xpcCof0kxKwuX1kqLG/ygbJ1f8wxED22bTL4F46P0"
        username: ubuntu
    user-data:
        ssh_import_id: [chad.smith]
EOF

$ sudo python3 -m http.server 3003
Serving HTTP on 0.0.0.0 port 3003 (http://0.0.0.0:3003/) ...
```
## terminal 2
```
$ snapcraft snap --output subiquity_test.snap
$ isoname=kinetic-live-server-amd64.iso
$ sudo ./scripts/inject-subiquity-snap.sh ${isoname} subiquity_test.snap custom.iso
$ qemu-img create -f raw target.img 10G
```
 ### test 1: test valid_subschema: no jsonschema.validation errors from any curtin/filesystem controllers/components due to cloudinit creating it's own Draft4SchemaValiditor
```
$ cp /tmp/autoinstall/user-data.VALID /tmp/autoinstall/user-data
$ kvm -no-reboot -m 1024     -drive file=target.img,format=raw,cache=none,if=virtio     -cdrom ~/src/subiquity/custom.iso     -kernel /mnt/casper/vmlinuz     -initrd /mnt/casper/initrd     -append 'autoinstall ds=nocloud-net;s=http://_gateway:3003/'

...#  autoinstall completes without error and shuts down the vm
```
### validate both disable behavior and cloud-init clean behavior on installed kinetic img
```
$ kvm -m 1024 -hda target.img -serial stdio


+ sudo cloud-init status --long
status: disabled
time: Thu, 25 Aug 2022 22:50:01 +0000
detail:
DataSourceNone
+ fgrep -ir Traceback /var/log/installer /var/log/cloud-init-output.log /var/log/cloud-init.log
+ true
+ cat /etc/cloud/cloud-init.disabled
Disabled by Ubuntu live installer after first boot.
To re-enable cloud-init on this image run:
  sudo cloud-init clean --machine-id
+ cat /etc/cloud/clean.d/99-installer
#!/usr/bin/env python3
# Remove live-installer config artifacts when running: sudo cloud-init clean
# Autogenerated by Subiquity: 2022-08-25 22:48:46.349524 UTC


import os

for cfg_file in ["/etc/cloud/cloud-init.disabled", "/etc/cloud/cloud.cfg.d/99-installer.cfg", "/etc/cloud/cloud.cfg.d/subiquity-disable-cloudinit-networking.cfg", "/etc/cloud/ds-identify.cfg", "/etc/netplan/00-installer-config.yaml"]:
    try:
        os.remove(cfg_file)
    except FileNotFoundError:
        pass
+ sudo cat /etc/machine-id
dc07594482f94b23b9041820d2cd08c2
+ sudo cloud-init clean --machine-id
+ cat /etc/machine-id
cat: /etc/machine-id: No such file or directory
+ echo SUCCESS: no machine-id
SUCCESS: no machine-id
+ sudo cloud-init status --long
status: not run
detail:
```



```

mwhudson:
Assuming that for some reason you wanted to autoinstall an image where cloud-init _would_ be active after first boot, how would you do that? Provide a runcmd in your user-data that runs "cloud-init clean"?

(This is like some kind of brain teaser, you provide cloud-init data containing autoinstall data that contains cloud-init data to run on first boot to make sure cloud-init still runs on subsequent boots...)

blackboxsw:
> 

YEah that is funky. I'll double check, but this is what I think subquity could do if you grew a config switch to "leave cloud-init enabled"

runcmd runs before `write_files_deferred` so your clean would happen (mid-cloud-init run which would be baad)

Maybe write a executable script which invokes `cloud-init clean` to /var/lib/cloud/scripts/per-once/<some_script>. That runs very late in cloud-init final stage..... I'll double check this and get back to you as I worry that the clean will disrupt some cache handling (or the read/write to /run/cloud-init/status.json that the cloud-init-final stage references when "finishing" it's boot stage)

PR #1397, opened 2022-08-25
filesystem: fix kname for os prober

dbungert:


mwhudson (approved):
argh

mwhudson:
BTW, for this and other reasons, it would be good to have stronger ties between the filesystem model objects and the probe data that generated them (like maybe storage_config could stick the DEVPATH or DEVNAME into the dict somehow? or return a dict mapping DEVNAME to id or something idk)

PR #1399, opened 2022-08-26
runtests: use trap to ensure backgroun proc killed

dbungert:


CarlosNihelton (approved):
If that cannot stop the server, nothing else can.

LGTM 👍

PR #1400, opened 2022-08-29
Sets WslSetupOptions as an optional controller

CarlosNihelton:
We have strong defaults for that controller, so clients should only need to POST if the desired options are not the default ones.

This PR makes sure the controller is configured even if clients won't POST anything. That's useful for the GUI.

### DIFF
--- a/system_setup/server/controllers/wslsetupoptions.py
+++ b/system_setup/server/controllers/wslsetupoptions.py
@@ -46,12 +46,14 @@ class WSLSetupOptionsController(SubiquityController):
 
         # load the config file
         data = default_loader()
+        conf_data = WSLSetupOptions()
 
         if data:
             proc_data = \
                 {key: convert_if_bool(value) for (key, value) in data.items()}
             conf_data = WSLSetupOptions(**proc_data)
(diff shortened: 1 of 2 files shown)


dbungert (changes requested):
I have two alternate proposals:
1) if this controller is not required, then we should not mark it required in the first place, then it won't be mandatory to configure it - you don't have to list a controller in INSTALL_MODEL_NAMES
2) it would be more similar to how other parts of Subiquity work today to mark this controller configured when you're done with the matching screen, which should work well if the screen is always shown but won't be great if the screen is only shown sometimes.

CarlosNihelton:
> I have two alternate proposals:
> 
> 1. if this controller is not required, then we should not mark it required in the first place, then it won't be mandatory to configure it - you don't have to list a controller in INSTALL_MODEL_NAMES
> 2. it would be more similar to how other parts of Subiquity work today to mark this controller configured when you're done with the matching screen, which should work well if the screen is always shown but won't be great if the screen is only shown sometimes.

The alternative No 1 works best for this case. Thanks for the suggestion!

PR #1401, opened 2022-08-29
underscore is not valid character for hostname

motoshkov:
separated HostnameEditor for hostname field

### DIFF
--- a/subiquity/ui/views/identity.py
+++ b/subiquity/ui/views/identity.py
@@ -42,7 +42,7 @@ from subiquity.common.types import IdentityData, UsernameValidation
 log = logging.getLogger("subiquity.ui.views.identity")
 
 HOSTNAME_MAXLEN = 64
-HOSTNAME_REGEX = r'[a-z0-9_][a-z0-9_-]*'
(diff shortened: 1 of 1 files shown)


dbungert:
Hi.  Thanks for the proposal.
I see the argument for adding period to the permitted character lists.

While it's true that I cannot point to a specification that permits underscore, previous experience suggests that some administrators do use them and expect to continue to use them.

This may be a Windows network thing, as I don't believe that I have encountered it otherwise.
See also [this document](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-2000-server/cc959336(v=technet.10)) from Microsoft.

I lean toward leaving underscore as permitted and those not using an underscore will be unaffected.

motoshkov:
I do not agree with you. Installing Linux OS has nothing to do with Windows NetBIOS name resolution as only Windows is using it.
"_" is non-valid character for hostname, especially taking in account that DHCP client sends its hostname to the server and may break DHCP server hostname update for DNS resolution.

dbungert (changes requested):
See previous discussion

PR #1405, opened 2022-09-02
Respects the output base dir on server.spawn() on API tests

CarlosNihelton:
When removing some unused command line options from the function responsible for spawning the system_setup server under API test, I ended up introducing a hidden dependency on `.subiquity` directory, when the function that starts the server already provides a proper temporary directory to break such dependency.

This addresses that bug by respecting the `output_base` directory supplied by the `start_server` function.


dbungert (approved):

dbungert:
Thanks!

PR #1406, opened 2022-09-06
network: use routes instead of gateway directives

dbungert:


mwhudson (approved):
LGTM, especially if you've tested in a VM :)

dbungert:
I have.  Dryrun is not a bad place to test either, but I have also done a VM test.  With correct input it applies quickly and later steps actually use the network.

PR #1411, opened 2022-09-08
keyboard: fix wrong keyboard layout in initramfs

ogayot:
During the installation, in the curthooks stage, curtin generates the initramfs. Unfortunately, the generated initramfs did not honor the keyboard mapping selected by the user during installation. Therefore, when LUKS encryption is used for the root file-system, the prompt for the passphrase ignores the keyboard layout set by the user.

To ensure that the initramfs uses the right keyboard layout, the `/etc/console-setup/cached_*` files must be updated before the initramfs gets created. At the beginning of the curthooks stage, we now call `setupcon --save-only` in the target to refresh those files if needed.

This PR addresses https://bugs.launchpad.net/subiquity/+bug/1894009

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -442,6 +442,12 @@ class SubiquityModel:
                 '001-mount-cdrom': [
                     'mount', '--bind', '/cdrom', '/target/cdrom',
                     ],
+                # The below command must be run after updating
+                # etc/default/keyboard on the target so that the initramfs uses
+                # the keyboard mapping selected by the user.  See LP #1894009
+                '002-setupcon-save-only': [
+                    'curtin', 'in-target', '--', 'setupcon', '--save-only',
+                    ],
                 },
 
             'grub': {


subiquity/models/subiquity.py
```diff
@@ -442,6 +442,12 @@ def render(self):
                 '001-mount-cdrom': [
                     'mount', '--bind', '/cdrom', '/target/cdrom',
                     ],
+                # The below command must be run after updating
+                # etc/default/keyboard on the target so that the initramfs uses
+                # the keyboard mapping selected by the user.  See LP #1894009
+                '002-setupcon-save-only': [
```
dbungert, line 448: I don't love a keyboard thing being here but I lack a better answer.

dbungert (approved):
I'm good with this if it has been VM tested.

subiquity/models/subiquity.py
ogayot: Good point. I move this hunk to the keyboard model.

This still works because we merge duplicate keys in merge_config.

Do you like it better now?

dbungert: Yes, that's nicer, thanks.

dbungert (approved):

mwhudson:
I realize we don't really have the setup for this now, but given we run each curtin 'stage' separately, we don't need to get curtin to run these commands for us, we can just run them ourselves (this is true for the `mount --bind /cdrom /target/cdrom` command too of course)

PR #1413, opened 2022-09-09
snaplist: avoid showing snaplist screen on failure

ogayot:
When creating the UI for snaplist, if the server returns a FAILED status when we request the list of snaps, we want to move on to the next screen.

However, calling `self.next_screen` in the `make_ui` function while also returning an actual view triggers a race condition. Depending on the order of execution of the instructions (which are asynchronous), the snaplist view can end up being shown even though we moved to the next screen. This ends up doing funny things upon clicking done.

Fixed by raising the `Skip` exception after making sure we mark snaplist configured.

### DIFF
--- a/subiquity/client/controllers/snaplist.py
+++ b/subiquity/client/controllers/snaplist.py
@@ -16,6 +16,10 @@
 import logging
 from typing import List
 
+from subiquitycore.tuicontroller import (
+    Skip,
+    )
+
 from subiquity.client.controller import (
     SubiquityTuiController,
     )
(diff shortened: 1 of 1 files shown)


ogayot:
@dbungert by any chance do you still have a reproducible use-case for what lead to the below patch?

commit ee528fe6a739eedc598be2af9f7e7cc8a731acdf
Author: Dan Bungert <daniel.bungert@canonical.com>
Date:   Mon Jan 24 17:03:46 2022 -0700

    controller/snaplist: protect GET against timeout

ogayot:
Thanks @dbungert !
I tested an install without network and it sent successfully. I am confident with this patch.

It should also address https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1983345

PR #1417, opened 2022-09-14
refactor CurtinInstallStep a bit

mwhudson:
Prep work for fde installs being ... different.

Not sure if this is the right final form but it's probably a step in the right direction.

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -20,7 +20,7 @@ from pathlib import Path
 import re
 import shutil
 import tempfile
-from typing import Any, Callable, Dict, List, Protocol
+from typing import Any, Callable, Dict, List
 
 import attr
 import yaml
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -72,12 +72,6 @@ class TracebackExtractor:
             self.traceback.append(line)
 
 
-class InstallStep(Protocol):
-
-    async def run(self, *, context) -> None:
-        pass
-
-
 @attr.s(auto_attribs=True)
 class CurtinInstallStep:
     """Represents the parameters of a single invocation of curtin install."""
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -224,12 +224,12 @@ class InstallController(SubiquityController):
                 acquire_config=acquire_config,
             )
 
-        await make_curtin_step(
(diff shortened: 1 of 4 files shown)


subiquity/server/controllers/install.py
```diff
@@ -210,45 +218,42 @@ async def curtin_install(self, *, context, source):
...
+                source=source,
+                acquire_config=acquire_config,
+            )
+
+        await make_curtin_step(
+            "initial",
+            stages=[],
+            acquire_config=self.acquire_initial_config).run(context=context)
 
         generic_steps = [
```
ogayot, line 239: if we're storing callables here, I think we don't need to treat the "initial" step differently - and we can drop the _generic_ prefix from this variable.

ogayot (approved):
LGTM, just one suggestion to refactor further.

subiquity/server/controllers/install.py
```diff
@@ -73,16 +73,49 @@ def feed(self, line):
             self.traceback.append(line)
 
 
+class InstallStep(Protocol):
```
ogayot, line 76: We're not using this Protocol, are we?

mwhudson: No, I did at one point but no longer.

subiquity/server/controllers/install.py
mwhudson: Well one of the very next things to do is to start passing different configs to different invocations. Possibly a caller here isn't the best way to do it but it's sufficient I guess?

ogayot: Yes, a caller/callable is fine IMO.

What I meant is that we could add have all the steps (including the initial step in the same sequence):

```diff
-         await make_curtin_step(
-           "initial",
-            stages=[],
-            acquire_config=self.acquire_initial_config).run(context=context)
-
-        generic_steps = [
+       steps: List[Callable[Context], None] = [
+              make_curtin_step(name="initial", stages=[], acquire_config=self.acquire_initial_config).run
               make_curtin_step(name="partitioning", stages=["partitioning"], acquire_config=self.acquire_generic_config).run
               make_curtin_step(name="extract", stages=["extract"], acquire_config=self.acquire_generic_config).run
```

mwhudson: oh right, yes that makes sense

ogayot: thanks!

PR #1418, opened 2022-09-14
filesystem: fix crash when editing LVM logical volume

ogayot:
When editing a LVM logical volume (LV), Subiquity would crash with the following error:

```python
filesystem/gaps.py", line 244, in movable_trailing_partitions_and_gap_size
    pgs = parts_and_gaps(partition.device)
AttributeError: 'LVM_LogicalVolume' object has no attribute 'device'
```
When we changed the implementation to use `movable_trailing_partitions_and_gap_size`, we did not pay attention to the LVM use case.

### DIFF
--- a/subiquity/common/filesystem/gaps.py
+++ b/subiquity/common/filesystem/gaps.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import functools
+from typing import Tuple, List
 
 import attr
 
@@ -23,6 +24,7 @@ from subiquity.models.filesystem import (
     align_down,
     Disk,
     LVM_CHUNK_SIZE,
+    LVM_LogicalVolume,
     LVM_VolGroup,
     Partition,
     Raid,
@@ -240,7 +242,22 @@ def largest_gap_size(device, in_extended=None):
     return 0
 
 
+@functools.singledispatch
 def movable_trailing_partitions_and_gap_size(partition):
+    """ For a given partition (or LVM logical volume), return the total,
+    potentially available, free space immediately following the partition.
+    By potentially available, we mean that to claim that much free space, some
+    other partitions might need to be moved.
+    The return value is a tuple that has two values:
+     * the list of partitions that would need to be moved
+     * the total potentially available free space
+    """
+    raise NotImplementedError
+
+
+@movable_trailing_partitions_and_gap_size.register
+def _movable_trailing_partitions_and_gap_size_partition(partition: Partition) \
+        -> Tuple[List[Partition], int]:
     pgs = parts_and_gaps(partition.device)
     part_idx = pgs.index(partition)
     trailing_partitions = []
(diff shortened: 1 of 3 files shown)


subiquity/common/filesystem/gaps.py
```diff
@@ -260,6 +277,19 @@ def movable_trailing_partitions_and_gap_size(partition):
...
+def _movable_trailing_partitions_and_gap_size_lvm(volume: LVM_LogicalVolume) \
+        -> Tuple[List[LVM_LogicalVolume], int]:
+    # In a Volume Group, there is no need to move partitions around, one can
+    # always use the remaining space.
+
+    def is_gap(x):
+        return isinstance(x, Gap)
+
+    gap = next(filter(is_gap, parts_and_gaps(volume.volgroup)), None)
+    return ([], gap.size if gap is not None else 0)
```
mwhudson, line 290: isn't this size calculation the same as "largest_gap_size(volume.volgroup)"?

subiquity/common/filesystem/tests/test_gaps.py
```diff
@@ -548,6 +563,12 @@ def test_trailing_preserved_partition(self):
             ([], 0),
             gaps.movable_trailing_partitions_and_gap_size(p1))
 
+    def test_trailing_lvm_logical_volume_no_gap(self):
```
mwhudson, line 566: A test with a non-zero free space would be good I think.

mwhudson (commented):
Thanks for this, I think the general direction is right. Couple of comments though.

subiquity/common/filesystem/gaps.py
ogayot: Good point. I guess theoretically it should be, yes. I'll update this function.

AFAIK, from LVM standpoint, there should never be multiple gaps in a volume group, so the largest gap (if any if found) should always be the right one.

I am not familiar with our utilization of Gap objects though ; so I'd assume that:
 * Gap objects of size 0 are not a thing. If they are though, using `largest_gap_size` will do a better job.
 * Two successive Gap objects (which could/should be merged into one) are not a thing.

subiquity/common/filesystem/tests/test_gaps.py
ogayot: Agreed!

subiquity/common/filesystem/gaps.py
dbungert: > Gap objects of size 0 are not a thing. 

correct

> Two successive Gap objects (which could/should be merged into one) are not a thing.

99% correct.  We can split a gap moments before handing a gap off for creation of a partition, and in that moment there are 2 successive gaps, but the part_and_gaps will never respond in that way.

subiquity/common/filesystem/tests/test_gaps.py
ogayot: updated!
I wasn't able to mock the alignment values in a way similar to disks/partitions. But I think the test looks OK.

subiquity/common/filesystem/gaps.py
ogayot: Thanks! Updated.

mwhudson (approved):
Thanks for the updates!

PR #1419, opened 2022-09-14
mirror: only start configuring APT once source & mirror are configured

ogayot:
The mirror controller is responsible from creating the APT configuration once its associated model is marked configured. However, the APT configuration is also dependent on the source model being marked configured.

When subiquity is used for the server installer, the client immediately sends a POST /a/meta/client_variant request. This ends up marking the source model as configured (even though I think it should not - and the default value does not make much sense). So we never end up in a scenario where the mirror model is marked configured before the source model.

When Subiquity is used by a different client, such as the ubuntu-desktop-installer, it is possible that we mark the mirror model configured first. This ends up crashing the installer.

Fixed by adding a dependency on the source model so that we do not end up creating the APT configuration too early.

The following procedure can be used to trigger the crash in dry-run mode:
1. Apply the following patch: https://pastebin.canonical.com/p/xNK3ZtBNFc/
2. Run `make dry-run`
3. Send the following request:
```bash
curl --unix-socket .subiquity/socket \
     -X POST \
    'a/meta/mark_configured?endpoint_names=%5B"mirror"%5D'
```

### DIFF
--- a/subiquity/server/controllers/mirror.py
+++ b/subiquity/server/controllers/mirror.py
@@ -84,7 +84,8 @@ class MirrorController(SubiquityController):
         self.app.hub.subscribe(
             (InstallerChannels.CONFIGURED, 'source'), self.on_source)
         self.cc_event = asyncio.Event()
-        self.configured_once = False
+        self.configured_event = asyncio.Event()
+        self.source_configured_event = asyncio.Event()
         self._apt_config_key = None
         self._apply_apt_config_task = SingleInstanceTask(
             self._apply_apt_config)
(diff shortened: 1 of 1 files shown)


mwhudson:
Oh right, I'm pretty sure the server client could/should stop doing that POST to /a/meta/client_variant now?

mwhudson (approved):
There is lots of logic in subiquity like this and I'm still a bit unhappy that it all has to be so procedural. But you've improved this corner of it for sure!

ogayot:
> Oh right, I'm pretty sure the server client could/should stop doing that POST to /a/meta/client_variant now?

Yeah, I think there is some cleanup that we can do. 
The variant is set to "server" by default through:
```python
supported_variants = ["server", "desktop"]

class SubiquityServer(Application):
    def __init__(self, opt, block_log_dir):
        self.set_source_variant(self.supported_variants[0])
```

But the coupling between the source model and the client_variant thingy is what puzzles me most.

PR #1421, opened 2022-09-15
kvm-test: allow to specify a cloud-init file not meant for autoinstall

ogayot:
The `--cloud-config` option now replaces the old `--autoinstall-file` option.
    
The use of `--autoinstall-file` always made kvm-test pass the `autoinstall` argument to the kernel command line. With the `--cloud-config` option, we now determine if the `autoinstall` argument is needed by reading the cloud config and checking if an autoinstall section is present.
    
The two new options `--force-autoinstall` and `--force-no-autoinstall` can be used to override this behaviour.
    
The `--autoinstall` option is also dropped. If we want to use the default, hardcoded cloud-config, one can use `--cloud-config-default`.

Example of an useful snippet that can be used to debug an installation with more ease:

```yaml
#cloud-config
users:
    - default
    - name: root
      ssh_import_id: lp:ogayot
```

### DIFF
--- a/scripts/kvm-test.py
+++ b/scripts/kvm-test.py
@@ -133,8 +133,6 @@ packages from a cache.
 
 See 'cfg' in script for expected layout of iso files,
 which can be managed with ~/.kvm-test.yaml''')
-parser.add_argument('-a', '--autoinstall', default=False,
-                    action='store_true', help='use autoinstall')
 parser.add_argument('-b', '--base', default=False, action='store_true',
                     help='use base iso')
 parser.add_argument('--basesnap', default=None, action='store',
@@ -147,9 +145,6 @@ parser.add_argument('-c', '--channel', action='store',
                     help='build iso with snap from channel')
 parser.add_argument('-d', '--disksize', default='12G', action='store',
                     help='size of disk to create (12G default)')
-parser.add_argument('-f', '--autoinstall-file', action='store',
-                    type=argparse.FileType(),
-                    help='load autoinstall from file')
 parser.add_argument('-i', '--img', action='store', help='use this img')
 parser.add_argument('-n', '--nets', action='store', default=1, type=int,
                     help='''number of network interfaces.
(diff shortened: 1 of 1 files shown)


scripts/kvm-test.py
```diff
@@ -193,6 +190,16 @@ def load_config(self):
                     help='boot test image')
 
 
+cc_group = parser.add_mutually_exclusive_group()
+cc_group.add_argument('-f', '--autoinstall-file', action='store',
```
dbungert, line 194: My hunch is that we should remove this argument entirely.
Most of the time, if sending autoinstall data, it's interesting to have autoinstall on the kernel command line.
If we did a quick yaml load of the cloud-config and checked for top-level key 'autoinstall', we could switch over to only using --cloud-config-file and send the autoinstall kernel command line (or not) based on autoinstall being in that file.

Also, I wanted to remove the `-f` short option, which somehow didn't get discussed during that kvm-test discussion months ago (or at least didn't make my notes).

scripts/kvm-test.py
```diff
@@ -193,6 +190,16 @@ def load_config(self):
...
+                      type=argparse.FileType(),
+                      help='load autoinstall from file and pass "autoinstall"'
+                           ' to the kernel command line')
+cc_group.add_argument('--cloud-config-file', action='store',
```
dbungert, line 198: What do you think about just calling this arg `--cloud-config`?

dbungert (commented):
Thanks - great idea to make the install environment more easily configurable.
I have a few suggestions.

scripts/kvm-test.py
ogayot: Yeah, I don't mind. I tried to mimic `--autoinstall-file` for consistency but if we are getting rid of it anyway ...

ogayot:
@dbungert I reviewed your suggestions, took some time to think about how to make things consistent ; and I came up with a plan.

I suggest we add the following options:

| option | description | details |
|----------|---------------|------------|
| `--cloud-config` | specify a cloud-config file to use (it may contain an `autoinstall` section or not) | renamed from `--cloud-config-file` |
| `--cloud-config-default` | use hardcoded cloud-config template from kvm-test.py | mutually exclusive with `--cloud-config` |
| `--force-autoinstall` / `--force-no-autoinstall` | pass/do not pass `autoinstall` on the kernel command line | by default, `autoinstall` will be passed if the `autoinstall` key is present in the cloud-config - so we should rarely need those options. |

and drop the following ones
*  `--autoinstall-file` / `-f`
* `--autoinstall`

I think it would cover everything that was possible with the old usage ; and the new use-cases.

| new usage | `autoinstall` on kernel CL | equivalent old usage (if any) | 
|------|-----|-----|
| `kvm-test.py --cloud-config-default` | yes (the default yaml has an autoinstall key) | `kvm-test.py --autoinstall` |
| `kvm-test.py --cloud-config ai.yaml` | yes | `kvm-test.py --autoinstall-file ai.yaml` |
| `kvm-test.py --cloud-config no-ai.yaml` | no | |
| `kvm-test.py --cloud-config no-ai.yaml --force-autoinstall` | yes | `kvm-test.py --autoinstall-file no-ai.yaml` |
| `kvm-test.py --cloud-config ai.yaml --force-no-autoinstall` | no | | 

how do you feel about it?

dbungert:
> how do you feel about it?

This is all great, please proceed.

dbungert (approved):

dbungert:
Thanks!

PR #1423, opened 2022-09-16
serialization enhancements in preparation for talking to snapd api

mwhudson:
I want to start using attrs objects when talking to the snapd API. This requires a bit more flexibility in the serialization code.

### DIFF
--- a/subiquity/common/serialize.py
+++ b/subiquity/common/serialize.py
@@ -27,8 +27,9 @@ import attr
 
 class Serializer:
 
-    def __init__(self, *, compact=False):
+    def __init__(self, *, compact=False, ignore_unknown_fields=False):
         self.compact = compact
+        self.ignore_unknown_fields = ignore_unknown_fields
         self.typing_walkers = {
             typing.Union: self._walk_Union,
             list: self._walk_List,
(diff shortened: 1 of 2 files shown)


subiquity/common/serialize.py
```diff
@@ -127,7 +140,8 @@ def _serialize_datetime(self, annotation, value, metadata, path):
     def _serialize_field(self, field, value, path):
         path = f'{path}.{field.name}'
         return {
-            field.name: self.serialize(field.type, value, field.metadata, path)
+            _field_name(field): self.serialize(
```
dbungert, line 143: When should _field_name be used, and when field.name?

dbungert (approved):

subiquity/common/serialize.py
mwhudson: Well oops. You ask this question and instead of simply explaining, I rewrote everything a bit.

Basically you need to use _field_name() when looking at the serialized form, and field.name when looking at the rich objects. About half the usages only matter for the "path" that is inconsistently included in some error messages, so I rewrote everything to make the path tracking and error messages much more consistent. Sorry about that -- I do think the code is better for it though. I added some very simple tests of the error handling.

dbungert (approved):

PR #1426, opened 2022-09-19
Make snaplist more robust to races and fix snap info prefetch

ogayot:
This general idea about this PR is to make the snaplist more robust and fix some of the known issues.

List of changes in no particular order:

1. The list tasks now raises an exception in case of error.
    * To check if the task was successful, once can call `task.done() and not task.exception()`.
    * To check if it failed, one can call `task.done() and task.exception()`.
    * To check if the task was cancelled, one can call `task.cancelled()`
    * To check if the task is still running, one can call `not task.done()`

2. The use of `SingleInstanceTask` was replaced by a standard `asyncio.Task` object. This allows us to control better what happens in case of cancellation.

3. When waiting in the GET handler for `/snaplist`, if the loader is recreated (i.e., if a network event occurs), we catch the error and wait again on the newly created loader.

4. Fix race condition in GET handler for `/snaplist` that could lead to an empty or partially empty list of snaps when a network event would occur. https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1989193
 
5. After fetching the list of snaps, the loader will automatically start pre-fetching the info about the snaps. This addresses https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1989198

6. Added a mechanism to block until the snap list task gets created. Calling `loader.start()` does not _immediately_ create a task that will fetch the list of snaps. Therefore, calling:
```python
loader = SnapdSnapLoader(...)
loader.start()
await loader.get_snap_list_task()
```
raises a `KeyError`.

Here is the safer alternative:
```python
loader = SnapdSnapLoader(...)
loader.start()
await loader.load_list_task_created.wait()  # Will unblock once the task is created
await loader.get_snap_list_task()
```

### DIFF
--- a/subiquity/server/controllers/snaplist.py
+++ b/subiquity/server/controllers/snaplist.py
@@ -66,6 +66,7 @@ class SnapdSnapInfoLoader:
             task = self.tasks[None] = \
                     SingleInstanceTask(self._load_list, propagate_errors=False)
             task.start_sync()
+            await task.wait()
             self.pending_snaps = self.model.get_snap_list()
             log.debug("fetched list of %s snaps", len(self.pending_snaps))
             while self.pending_snaps:
--- a/subiquity/server/controllers/snaplist.py
+++ b/subiquity/server/controllers/snaplist.py
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import asyncio
 import logging
 from typing import List
 
@@ -22,7 +23,6 @@ import requests.exceptions
 
 from subiquitycore.async_helpers import (
     schedule_task,
-    SingleInstanceTask,
     )
 from subiquitycore.context import with_context
 
@@ -59,15 +59,14 @@ class SnapdSnapInfoLoader:
         self.pending_snaps = []
         self.tasks = {}  # {snap:task}
 
+        self.load_list_task_created = asyncio.Event()
+
     def _fetch_list_ended(self) -> bool:
         """ Tells whether the snap list fetch task has ended without being
         cancelled. """
         if None not in self.tasks:
             return False
-        task = self.get_snap_list_task().task
-        if task is None:
-            # Task is not yet started.
-            return False
+        task = self.get_snap_list_task()
         if task.cancelled():
             return False
         return task.done()
@@ -76,13 +75,13 @@ class SnapdSnapInfoLoader:
         """ Tells whether the snap list fetch task has completed. """
         if not self._fetch_list_ended():
             return False
-        return not self.get_snap_list_task().task.exception()
+        return not self.get_snap_list_task().exception()
 
     def fetch_list_failed(self) -> bool:
         """ Tells whether the snap list fetch task has failed. """
         if not self._fetch_list_ended():
             return False
-        return bool(self.get_snap_list_task().task.exception())
+        return bool(self.get_snap_list_task().exception())
 
     def start(self):
         log.debug("loading list of snaps")
@@ -91,10 +90,10 @@ class SnapdSnapInfoLoader:
     async def _start(self):
         with self.context:
             task = self.tasks[None] = \
-                    SingleInstanceTask(self._load_list, propagate_errors=False)
-            task.start_sync()
+                    asyncio.create_task(self._load_list())
+            self.load_list_task_created.set()
             try:
-                await task.wait()
+                await task
             except SnapListFetchError:
                 log.exception("loading list of snaps failed")
                 return
@@ -205,10 +204,16 @@ class SnapListController(SubiquityController):
             return SnapListResponse(status=SnapCheckState.LOADING)
         # Let's wait for the task to be completed.
         # If the loader gets restarted, the cancellation should be absorbed.
-        try:
-            await self.loader.get_snap_list_task().wait()
-        except SnapListFetchError:
-            return SnapListResponse(status=SnapCheckState.FAILED)
+        while True:
+            await self.loader.load_list_task_created.wait()
+            try:
+                await asyncio.shield(self.loader.get_snap_list_task())
+            except SnapListFetchError:
+                return SnapListResponse(status=SnapCheckState.FAILED)
+            except asyncio.CancelledError:
+                log.warn("load list snaps task was cancelled, retrying...")
+            else:
+                break
         return SnapListResponse(
             status=SnapCheckState.DONE,
             snaps=self.model.get_snap_list(),
--- a/subiquity/server/controllers/snaplist.py
+++ b/subiquity/server/controllers/snaplist.py
@@ -42,6 +42,10 @@ from subiquity.server.types import InstallerChannels
 log = logging.getLogger('subiquity.server.controllers.snaplist')
 
 
+class SnapListFetchError(Exception):
+    """ Exception to raise when the list of snaps could not be fetched. """
+
+
 class SnapdSnapInfoLoader:
 
     def __init__(self, model, snapd, store_section, context):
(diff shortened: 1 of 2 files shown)


subiquity/server/controllers/snaplist.py
```diff
@@ -169,15 +199,24 @@ def make_autoinstall(self):
...
+        # If the loader gets restarted, the cancellation should be absorbed.
+        while True:
+            await self.loader.load_list_task_created.wait()
+            try:
+                await self.loader.get_snap_list_task()
+                break
+            except SnapListFetchError:
+                await self.configured()
+                return SnapListResponse(status=SnapCheckState.FAILED)
+            except asyncio.CancelledError:
```
dbungert, line 218: Infinite retries on cancel + cancel being generally hard to get correct make me question this loop.
What's your thoughts on limiting it to some number of retries before giving up?

dbungert (changes requested):
When testing this in a VM, I encountered a crash when trying to unmount /cdrom.  It's not clear to me how that would be related to this  PR, but an equivalent test using `main` didn't have the same issue.

dbungert:
Sample crash - kinetic + this branch, accept install defaults + install microk8s
[subiquity-server-debug.log](https://github.com/canonical/subiquity/files/9601300/subiquity-server-debug.log)

dbungert:
I'm seeing the crash on other branches as well.  I would like to see a clean vmtest for this first still.

ogayot:
> I'm seeing the crash on other branches as well. I would like to see a clean vmtest for this first still.

Thanks for spotting this. The bug seems related to the last PR (#1417) merged (related to refactoring of install steps for FDE). A partial revert of the following patch fixes the issue:

6d724c0a5be0ccbbf9e3b249e907aa2405993c69 (

<details>
<summary>Click me to show the diff</summary>

```diff
From 8755c4186566a7f1bd8ead92b765dea1f4d819b2 Mon Sep 17 00:00:00 2001
From: Olivier Gayot <olivier.gayot@canonical.com>
Date: Tue, 20 Sep 2022 09:31:20 +0200
Subject: [PATCH] Revert "use new flexibility to remove a bit of mystery code
 from SubiquityModel"

This reverts commit 6d724c0a5be0ccbbf9e3b249e907aa2405993c69.
---
 subiquity/models/subiquity.py           | 6 ++++++
 subiquity/server/apt.py                 | 5 -----
 subiquity/server/controllers/install.py | 5 -----
 3 files changed, 6 insertions(+), 10 deletions(-)

diff --git a/subiquity/models/subiquity.py b/subiquity/models/subiquity.py
index 50e64372..0aabeaa3 100644
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -438,6 +438,12 @@ class SubiquityModel:
 
     def render(self):
         config = {
+            'curthooks_commands': {
+                '001-mount-cdrom': [
+                    'mount', '--bind', '/cdrom', '/target/cdrom',
+                    ],
+                },
+
             'grub': {
                 'terminal': 'unmodified',
                 'probe_additional_os': True
diff --git a/subiquity/server/apt.py b/subiquity/server/apt.py
index 2981f114..11cf46a5 100644
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -278,11 +278,6 @@ class AptConfigurer:
 
         await self.cleanup()
 
-    async def setup_target(self):
-        # Call this after the rootfs has been extracted to the real target
-        # system but before any configuration is applied to it.
-        await self.mount('/cdrom', '/target/cdrom', options='bind')
-
 
 class DryRunAptConfigurer(AptConfigurer):
 
diff --git a/subiquity/server/controllers/install.py b/subiquity/server/controllers/install.py
index 6b39b334..e33686d8 100644
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -188,10 +188,6 @@ class InstallController(SubiquityController):
         configurer = await mirror.wait_config()
         return await configurer.configure_for_install(context)
 
-    async def setup_target(self, context):
-        mirror = self.app.controllers.Mirror
-        await mirror.apt_configurer.setup_target()
-
     @with_context(
         description="installing system", level="INFO", childlevel="DEBUG")
     async def curtin_install(self, *, context, source):
@@ -238,7 +234,6 @@ class InstallController(SubiquityController):
                 name="extract", stages=["extract"],
                 acquire_config=self.acquire_generic_config,
             ).run,
-            self.setup_target,
             make_curtin_step(
                 name="curthooks", stages=["curthooks"],
                 acquire_config=self.acquire_generic_config,
-- 
2.34.1

```

</details>

I created a new LP bug to follow up on this one: https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1990239

subiquity/server/controllers/snaplist.py
```diff
@@ -50,22 +54,51 @@ def __init__(self, model, snapd, store_section, context):
...
-        self.snap_list_fetched = False
-        self.failed = False
 
         self.snapd = snapd
         self.pending_snaps = []
         self.tasks = {}  # {snap:task}
 
+        self.load_list_task_created = asyncio.Event()
+
+    def fetch_list_completed(self) -> bool:
```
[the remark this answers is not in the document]
ogayot: Refactored, thanks.
I also removed the `if task is None` block since it was only useful when a `SingleInstanceTask` was used.
`SingleInstanceTask` have their `.task` set to `None` until `.start_sync()` is called.

subiquity/server/controllers/snaplist.py
ogayot: What would you suggest as a number of retries?
IMHO, this is a good scenario where it makes sense to go for infinite retries on cancel ; since:
* the client explicitly provided the `wait=True` ; meaning they expect to wait
* the client provides a loading animation with a mechanism to skip the screen if the user does not want to wait longer
* cancellation on the server side should be the result of:
    * a network event
    * the app stopping - in which case the GET handler will be cancelled anyway

I think the worst case scenario would be a burst of network events that would keep the server busy-waiting. But the outcome would be the same with or without a limit in the GET handler.

dbungert: All right.  The log statement helps - if we see a long scroll of 'load list snap task ...' we will know roughly what happened.

dbungert (approved):

subiquity/server/controllers/snaplist.py
```diff
@@ -169,13 +197,22 @@ def make_autoinstall(self):
...
-        await self.loader.get_snap_list_task().wait()
-        if self.loader.failed or not self.app.base_model.network.has_network:
-            return SnapListResponse(status=SnapCheckState.FAILED)
+        # Let's wait for the task to be completed.
+        # If the loader gets restarted, the cancellation should be absorbed.
+        while True:
+            await self.loader.load_list_task_created.wait()
+            try:
+                await asyncio.shield(self.loader.get_snap_list_task())
+                break
```
mwhudson, line 211: I slightly prefer to express this pattern by having the try: block end with "else: break" but that's very minor.

mwhudson (approved):
Thanks for this. Cancellation with asyncio is such a mess. Maybe we can move to trio one day :-) (would need a new http library first though...)

mwhudson (approved):

subiquity/server/controllers/snaplist.py
ogayot: Oh yeah, that is better!

ogayot:
@mwhudson It looks like the CI failed because of #1423:

```python
SerializationError(..., '2019-01-20T13:28:33.861403Z' is not a <class 'datetime.datetime'>)
```

but I'm confused because the fix-snaplist branch does not have the serialization changes yet ...

mwhudson:
> @mwhudson It looks like the CI failed because of #1423:

argh https://github.com/canonical/subiquity/pull/1432

I guess github tests PRs by merging them into the target branch?

ogayot:
> > @mwhudson It looks like the CI failed because of #1423:
> 
> argh #1432
> 
> I guess github tests PRs by merging them into the target branch?

Oh yeah, that would make sense!

ogayot:
Well, the CI still fails after merging the serialization fix to main so I don't know what github does when testing the branch. Anyway, rebasing on top of main ..

PR #1428, opened 2022-09-20
kernel: fix autoinstall and run file

dbungert:
Marking draft due to ongoing crashes related to /cdrom.


ogayot (approved):
LGTM!

Thanks for getting rid of the `if flavor is None: package = 'generic'` block. It was confusing.

PR #1430, opened 2022-09-21
reserved-usernames: add groups created by udev (src:systemd)

ogayot:
The package udev (part of the systemd source package) creates system groups at installation.

In version 251.4, the following system groups are created:

 * input
 * sgx
 * kvm
 * render

When a user accidentally picks a username that corresponds to one of these udev groups, Subiquity gladly accepts it. However, on first boot, cloud-init fails to create the user because a group of the same name already exists.

Added these 4 groups to reserved-usernames so that they get rejected in the identity screen.

fixes https://bugs.launchpad.net/subiquity/+bug/1987341

### DIFF
--- a/reserved-usernames
+++ b/reserved-usernames
@@ -1,4 +1,4 @@
-# Static users from base-passwd/passwd.master (3.5.11).
+# Static users from base-passwd/passwd.master (3.5.41).
 root
 daemon
 bin
@@ -18,7 +18,7 @@ irc
 gnats
 nobody
 
-# Other static groups from base-passwd/group.master (3.5.11).
+# Other static groups from base-passwd/group.master (3.5.41).
 adm
 tty
 disk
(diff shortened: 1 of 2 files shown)


ogayot:
> This file is overwritten by /usr/lib/user-setup/reserved-usernames from the user-setup package during snap build so this isn't going to work as is. If that file is getting out of date though we should be doing something different (maybe we should just get the file from the user-setup git?).

oh snap! I didn't realize that, thanks.

yeah, I think it would make sense to get it from https://git.launchpad.net/ubuntu/+source/user-setup/tree/reserved-usernames

I would also like to have a CI job that informs us if the copy we have in subiquity.git is outdated - so that we can keep it on sync with upstream.

ogayot:
Updated!
* local copy of reserved-usernamesrefreshed (+ added sgx)
* we now fetch the version of reserved-usernames from https://git.launchpad.net/ubuntu/+source/user-setup when building the snap

mwhudson (approved):
I think I had in mind getting this from user-setup upstream git on salsa but this is probably more coherent, it's certainly better than current main!

dbungert (approved):

PR #1431, opened 2022-09-22
remove special cases around mounting/unmounting /target/cdrom

mwhudson:


mwhudson:
I haven't vmtested this yet and after what happened last time I definitely should!

ogayot:
LGTM, thanks!

ogayot (approved):

dbungert (approved):
I won't block merge for this, but if you have thoughts on how we might automatically test this, given that we've been bit by this before, that would be nice.

On the other hand I want to see some level of VM testing in CI, so maybe that's the answer.

mwhudson:
Yes, I think some kind of automated VM testing for PRs is the right answer here. If you can make that happen, I'll be very happy!

mwhudson:
finally managed to test this in a vm

PR #1433, opened 2022-09-23
filesystem: accept extended partition with size -1 in autoinstall

ogayot:
~Marking a draft until tested in KVM. Using dry-run looked promising so far.~ :heavy_check_mark: 

Our documentation of autoinstall mentions that one can specify the size of a partition as -1 to use the remaining space on the disk.

Because it only makes sense in practice to support this option for the right-most partition of a given disk, specifying size: -1 is only accepted when defining the last partition of the disk.

Having said that, when making an autoinstall configuration that includes an extended partition, the user must define the logical partitions after the extended partition.

For the following configuration:

```
|                                  disk                                |
+--------------+--------------+--------------+-------------------------|
| p1 (primary) | p2 (primary) | p3 (primary) | p4 (extended)           |
|              |              |              | p5 (logic) | p6 (logic) |
```
, the representation in Subiquity will be analogous to:
```
[
  Partition(p1, "primary"),
  Partition(p2, "primary"),
  Partition(p3, "primary"),
  Partition(p4, "extended"),
  Partition(p5, "logical"),
  Partition(p6, "logical"),
]
```

This means that specifying size: -1 on p4 gets rejected because it is not the last partition defined.p6 is the last partition defined.

This is however a valid use-case because p5 and p6 are contained within p4. So p4 should be able to use the remaining space and there is no reason to reject this configuration

Fixed by ignoring logical partitions when checking if a size of -1 is allowed on an extended partition.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -16,6 +16,7 @@
 from abc import ABC, abstractmethod
 import attr
 import collections
+import copy
 import fnmatch
 import itertools
 import logging
@@ -38,6 +39,11 @@ log = logging.getLogger('subiquity.models.filesystem')
 
 MiB = 1024 * 1024
 GiB = 1024 * 1024 * 1024
+
+
+class NotFinalPartitionError(Exception):
+    """ Exception to raise when guessing the size of a partition that is not
+    the last one. """
 
 
 def _set_backlinks(obj):
@@ -170,6 +176,13 @@ def reverse_dependencies(obj):
             yield from v
         elif v is not None:
             yield v
+
+
+def is_logical_partition(obj):
+    try:
+        return obj.is_logical
+    except AttributeError:
+        return False
 
 
 @attr.s(eq=False)
@@ -1180,17 +1193,30 @@ class FilesystemModel(object):
         self._actions = self._actions_from_config(
             ai_config, self._probe_data['blockdev'], is_probe_data=False)
         for p in self._all(type="partition") + self._all(type="lvm_partition"):
+            # NOTE For logical partitions (DOS), the parent is set to the disk,
+            # not the extended partition.
             [parent] = list(dependencies(p))
             if isinstance(p.size, int):
                 if p.size < 0:
-                    if p is not parent.partitions()[-1]:
-                        raise Exception(
+                    if p.flag == "extended":
+                        # For extended partitions, we use this filter to create
+                        # a temporary copy of the disk that excludes all
+                        # logical partitions.
+                        def filter_(x): return not is_logical_partition(x)
+                    else:
+                        def filter_(x): return True
+                    filtered_parent = copy.copy(parent)
+                    filtered_parent._partitions = list(filter(
+                        filter_, parent.partitions()))
+                    if p is not filtered_parent.partitions()[-1]:
+                        raise NotFinalPartitionError(
                             "{} has negative size but is not final partition "
                             "of {}".format(p, parent))
+
                     from subiquity.common.filesystem.gaps import (
                         largest_gap_size,
                         )
-                    p.size = largest_gap_size(parent)
+                    p.size = largest_gap_size(filtered_parent)
             elif isinstance(p.size, str):
                 if p.size.endswith("%"):
                     percentage = int(p.size[:-1])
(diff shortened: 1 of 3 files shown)


ogayot:
* Fixed logical partitions with `size: -1` sometimes expanding to the size the wrong gap (when there is one gap inside the extended partition and another gap outside). Added test to cover the use case.
* Fixed extended partitions with `size: 1` failing to determine the size properly - by ignoring logical partitions when trying to find the largest gap.

ogayot:
Tested with success on KVM with a use-case inspired from 
https://bugs.launchpad.net/subiquity/+bug/1989977

To get the installation to succeed, I had to make the following modifications:
* ran Subiquity with storage version 2 (no support for gaps in extended with storage version 1)
* removed `wipe: superblock` on the extended partition since it discards the extended partition table.
* hardcoded offsets of all partitions. This is far from ideal but is necessary with storage version 2 because of bug https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1991413 

This is the partition table as it looks after a Subiquity install:

```
Disk /dev/vda: 20 GiB, 21474836480 bytes, 41943040 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0xc1a3fe6e

Device     Boot    Start      End  Sectors  Size Id Type
/dev/vda1  *        2048  2099199  2097152    1G ef EFI (FAT-12/16/32)
/dev/vda2        2099200 14682111 12582912    6G 83 Linux
/dev/vda3       14682112 14886911   204800  100M 83 Linux
/dev/vda4       14886912 41943039 27056128 12.9G  5 Extended
/dev/vda5       14888960 16986111  2097152    1G 83 Linux
/dev/vda6       16988160 21182463  4194304    2G 83 Linux
/dev/vda7       21184512 41940991 20756480  9.9G 83 Linux
```
Below is a partition table created manually with fdisk (where fdisk guessed all offsets and the size of vda4 and vda7):

```diff
Disk test.img: 20 GiB, 21474836480 bytes, 41943040 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x092efef4

Device     Boot    Start      End  Sectors  Size Id Type
test.img1           2048  2099199  2097152    1G ef EFI (FAT-12/16/32)
test.img2        2099200 14682111 12582912    6G 83 Linux
test.img3       14682112 14886911   204800  100M 83 Linux
test.img4       14886912 41943039 27056128 12,9G  5 Extended
test.img5       14888960 16986111  2097152    1G 83 Linux
test.img6       16988160 21182463  4194304    2G 83 Linux
test.img7       21184512 41943039 20758528  9,9G 83 Linux
```

The only difference is that the final logical partition is 2048 sectors (1MiB) shorter than the one created by fdisk.
There is probably something we can do to claim those few extra sectors but the benefit would be minimal.

subiquity/models/filesystem.py
```diff
@@ -40,6 +41,11 @@
 GiB = 1024 * 1024 * 1024
 
 
+class NotFinalPartitionError(Exception):
+    """ Exception to raise when guessing the size of a partition that is not
+    the right-most one. """
```
dbungert, line 46:
```suggestion
    the last one. """
```

dbungert (approved):

dbungert:
> The only difference is that the final logical partition is 2048 sectors (1MiB) shorter than the one created by fdisk.

I wonder if some chunk of code is looking at the gpt min_end_offset.  A problem for another day.

ogayot:
> > The only difference is that the final logical partition is 2048 sectors (1MiB) shorter than the one created by fdisk.
> 
> I wonder if some chunk of code is looking at the gpt min_end_offset.

Very much likely ! :)

subiquity/models/filesystem.py
ogayot: Applied, thanks!

PR #1434, opened 2022-09-27
Fix unset SNAP_VERSION in WSL

EduardGomezEscandell:
In WSL, the environment variable SNAP_VERSION is unknown. This causes issue https://github.com/ubuntu/WSL/issues/212 to happen.

This PR fixes this by writing the revision instead of the version.
The revision is known by WSL thanks to [this line](https://github.com/ubuntu/wsl-setup/blob/7e6c9abb46cd7f370fc7d5647804278ebd03c976/wsl-setup#L71)


dbungert:
Maybe some better commit messages in the future, please.

PR #1435, opened 2022-09-28
create and use a more richly typed snapd api

mwhudson:
Some bits of this should probably be split off and maybe a little rebasing should be done but I'm interested in what you think about this.

### DIFF
--- a/subiquity/server/controllers/refresh.py
+++ b/subiquity/server/controllers/refresh.py
@@ -213,9 +213,9 @@ class RefreshController(SubiquityController):
 
     @with_context()
     async def start_update(self, context):
-        change = await self.app.snapd.post(
+        change = (await self.app.snapd.post(
             'v2/snaps/{}'.format(self.snap_name),
-            {'action': 'refresh', 'ignore-running': True})
+            {'action': 'refresh', 'ignore-running': True}))['change']
         context.description = "change id: {}".format(change)
         return change
 
--- a/subiquity/server/controllers/refresh.py
+++ b/subiquity/server/controllers/refresh.py
@@ -27,6 +27,7 @@ from subiquitycore.context import with_context
 
 from subiquity.common.apidef import API
 from subiquity.common.types import (
+    Change,
     RefreshCheckState,
     RefreshStatus,
     )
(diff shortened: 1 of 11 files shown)


dbungert:
I'll want to read in more detail but at first glance seems straightforward.
The type annotation stuff is a bit mysterious.

subiquity/server/controllers/refresh.py
```diff
@@ -213,16 +221,14 @@ async def check_for_update(self, context):
...
-    async def get_progress(self, change):
-        result = await self.app.snapd.get('v2/changes/{}'.format(change))
-        change = result['result']
-        if change['status'] == 'Done':
+        change_id = await self.app.snapdapi.v2.snaps[self.snap_name].POST(
+            SnapActionRequest(action=SnapAction.REFRESH, ignore_running=True))
+        context.description = "change id: {}".format(change_id)
+        return change_id
+
+    async def get_progress(self, change_id):
```
dbungert, line 229: Suggest annotating this one.

subiquity/server/server.py
```diff
@@ -313,9 +314,11 @@ def __init__(self, opts, block_log_dir):
                     "examples", "snaps"),
                 self.scale_factor, opts.output_base)
             self.snapd = AsyncSnapd(connection)
+            self.snapdapi = make_api_client(self.snapd)
         elif os.path.exists(self.snapd_socket_path):
             connection = SnapdConnection(self.root, self.snapd_socket_path)
             self.snapd = AsyncSnapd(connection)
+            self.snapdapi = make_api_client(self.snapd)
         else:
             log.info("no snapd socket found. Snap support is disabled")
```
dbungert, line 323: For the current support this is appropriate but that might need to be upgraded to a more serious error later on.

dbungert (commented):
The changes are pretty straightforward.  Just my previous comment on some of the type assertion changes and a few more comments.

subiquity/server/controllers/refresh.py
mwhudson: done

subiquity/server/server.py
mwhudson: this is here because the wsl system setup code comes through here (we need to fix this somehow or other!)

mwhudson:
> The type annotation stuff is a bit mysterious.

Do you mean the stuff I have now split out into a separate PR? (#1440) Or something else?

dbungert:
> Do you mean the stuff I have now split out into a separate PR?

Yes

dbungert (approved):

mwhudson:
ahh poo this breaks in a vm :( i think the snapd output has changed a bit.

PR #1449, opened 2022-10-11
apt: workaround umount failure

dbungert:
When the apt overlay fails to umount, this directory removal will also fail due to the mount still being present.

LP: #1992531


mwhudson:
Um. OK I guess -- would like to understand why this is happening though. I thought the point of running ubuntu-drivers in a private mount namespace was supposed to avoid this kind of nonsense :(

mwhudson (approved):

dbungert:
> Um. OK I guess -- would like to understand why this is happening though.

Agree.  I filed https://bugs.launchpad.net/bugs/1992533 as a reminder.

PR #1450, opened 2022-10-12
filesystem: ensure bootfs at aligned size

dbungert:
Certain disk sizes could trigger unaligned bootfs size.  This can result in the next partition overlapping into the bootfs partition.

LP: #1992541


ogayot (approved):
Wouldn't it be safer to make `add_partition()` call `real_size = align_down(size)` rather than `real_size = align_up(size)`?
I can picture similar issues occurring if we create partitions ever so slightly bigger than requested.

mwhudson:
Blarg having add_partition round up seems like a trap. The handling of alignments and so on has gotten a lot better over the last little while, we must keep pushing on to make bugs like this impossible!

PR #1453, opened 2022-10-17
drivers: make overlay unmount error not fatal

ogayot:
Sometimes, unmounting the overlay used by ubuntu-drivers fails with EBUSY. Heinrich witnessed it on riscv64 but Dan also reproduced it on amd64. The root cause is not clear but the most obvious reason would be that a process is still running with its CWD set to the overlay. It could be due to:

* ubuntu-drivers spawning a subprocess that had not exited when the main process exited
*  `subprocess` unblocking before calling wait on the child process (seems unlikely)
* the lack of guarantee that all resources are freed released after a process exits?
* an issue with `umount` itself - as pointed out in the manpage:

>       Note that a filesystem cannot be unmounted when it is 'busy' - for example,
>       when there are open files on it, or when some process has its working
>       directory there, or when a swap file on it is in use. The offending process
>       could even be umount itself - it opens libc, and libc in its turn may open
>       for example locale files. A lazy unmount avoids this problem, but it may
>       introduce other issues. See --lazy description below.

There must be something that we can do about it but we need to conduct a root cause investigation. In the meantime, this PR logs and ignores unmount errors when cleaning up the drivers overlay.

Addresses https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1992977

### DIFF
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -18,6 +18,7 @@ import functools
 import logging
 import os
 import shutil
+import subprocess
 import tempfile
 from typing import List, Optional, Union
 
@@ -32,6 +33,10 @@ from subiquitycore.utils import arun_command
 from subiquity.server.curtin import run_curtin_command
 
 log = logging.getLogger('subiquity.server.apt')
+
+
+class OverlayCleanupError(Exception):
+    """ Exception to raise when an overlay could not be cleaned up. """
 
 
 class _MountBase:
@@ -245,7 +250,10 @@ class AptConfigurer:
             # Mountpoint object and (thanks to attr.s) make sure that it
             # compares equal to the one we discarded earlier.
             # But really, there should be better ways to handle this.
-            await self.unmount(Mountpoint(mountpoint=overlay.mountpoint))
+            try:
+                await self.unmount(Mountpoint(mountpoint=overlay.mountpoint))
+            except subprocess.CalledProcessError as exc:
+                raise OverlayCleanupError from exc
 
     async def cleanup(self):
         for m in reversed(self._mounts):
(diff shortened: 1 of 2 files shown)


ogayot:
Tested in a VM after forcing the unmmount call to fail:

```diff
diff --git a/subiquity/server/apt.py b/subiquity/server/apt.py
index c4402cba..094fd392 100644
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -251,7 +251,7 @@ class AptConfigurer:
             # compares equal to the one we discarded earlier.
             # But really, there should be better ways to handle this.
             try:
-                await self.unmount(Mountpoint(mountpoint=overlay.mountpoint))
+                subprocess.check_call(["/bin/false"])
             except subprocess.CalledProcessError as exc:
                 raise OverlayCleanupError from exc
```

The installation succeeded without showing an error report on the client.
The server log recorded the error as intended:

```python
2022-10-17 12:10:10,702 ERROR subiquity.server.controllers.drivers:99 Failed to cleanup overlay. Continuing anyway.
  File "/snap/subiquity/x1/lib/python3.8/site-packages/subiquity/server/apt.py", line 254, in overlay
    subprocess.check_call(["/bin/false"])
  File "/snap/subiquity/x1/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/bin/false']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

  File "/snap/subiquity/x1/lib/python3.8/site-packages/subiquity/server/controllers/drivers.py", line 95, in _list_drivers
    self.drivers = await self.ubuntu_drivers.list_drivers(
  File "/snap/subiquity/x1/usr/lib/python3.8/contextlib.py", line 178, in __aexit__
    await self.gen.__anext__()
  File "/snap/subiquity/x1/lib/python3.8/site-packages/subiquity/server/apt.py", line 256, in overlay
    raise OverlayCleanupError from exc
subiquity.server.apt.OverlayCleanupError
2022-10-17 12:10:10,716 DEBUG subiquity.server.controllers.drivers:100 Available drivers to install: []
```

dbungert (approved):

PR #1455, opened 2022-10-17
Adds language-selector-common to the snap

CarlosNihelton:
Would it be acceptable to add that package to the snap?

The system_setup configure controller requires it. UDI used to provide that, but now we are Subiquity-only-snap.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -117,6 +117,7 @@ parts:
       - libsystemd0
       - iso-codes
       - lsb-release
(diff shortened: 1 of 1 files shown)


CarlosNihelton:
We can close this, since we took another approach.

https://git.launchpad.net/~ubuntu-core-dev/ubuntu-seeds/+git/ubuntu/commit/?id=f7dbd1fd419e9cda92436bb780fd14e1bc0a694a

PR #1456, opened 2022-10-18
storage: make GET /storage return the last successful probe result when requested

ogayot:
In the following patch:

  4c59e1ee never return a PROBING response from guided_POST

we fixed a UI crash occurring when a storage probing operation was ongoing while the user would enter the partitioning screen with "Use an entire disk".

To address the issue, we made `POST /storage/guided` return the last (cached) successful probe result rather than returning a potential PROBING status result.

That being said, when "Custom Storage Layout" is selected, we are using a different HTTP call: GET /storage which can also return a PROBING status result. So our fix was only partial.

This patch adds a "use_cached_result" parameter (which defaults to false) to GET /storage. This makes the call return the last successful probe result.

When the client selects "Custom Storage Layout" we now explicitly request a cached result so that we don't have to deal with a potential PROBING status response.

Ideally, we should either:

 * make the client aware that a PROBING status can be returned and act accordingly.
 * pass the wait=true argument to GET /storage or POST /storage/guided

But doing so has other implications and the kinetic release is imminent.

### DIFF
--- a/subiquity/client/controllers/filesystem.py
+++ b/subiquity/client/controllers/filesystem.py
@@ -236,10 +236,17 @@ class FilesystemController(SubiquityTuiController, FilesystemManipulator):
             raise Exception("could not process action {}".format(action))
 
     async def _guided_choice(self, choice):
+        # FIXME It would seem natural here to pass the wait=true flag to the
+        # below HTTP calls, especially because we wrap the coroutine in
+        # wait_with_progress.
+        # Having said that, making the server return a cached result seems like
+        # the least risky option to address https://launchpad.net/bugs/1993257
+        # before the kinetic release. This is also similar to what we did for
+        # https://launchpad.net/bugs/1962205
         if choice is not None:
             coro = self.endpoint.guided.POST(choice)
         else:
-            coro = self.endpoint.GET()
+            coro = self.endpoint.GET(use_cached_result=True)
         status = await self.app.wait_with_progress(coro)
         self.model = FilesystemModel(status.bootloader)
         self.model.load_server_data(status)
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -251,7 +251,9 @@ class API:
                     -> StorageResponse:
                 pass
 
-        def GET(wait: bool = False) -> StorageResponse: ...
+        def GET(wait: bool = False, use_cached_result: bool = False) \
+            -> StorageResponse: ...
+
         def POST(config: Payload[list]): ...
 
         class reset:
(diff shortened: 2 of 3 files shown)


ogayot:
I applied the following patch to reproduce the issue in dry-run mode. And I am about to do the same test in a VM
```dfff
commit b77a2c0a666e9dfc61cae279113d86a59ef4f880
Author: Olivier Gayot <olivier.gayot@canonical.com>
Date:   Tue Oct 18 15:04:32 2022 +0200

    force bug 1993257 to trigger
    
    Select "Custom Storage Layout" and it should crash.
    
diff --git a/subiquity/client/controllers/filesystem.py b/subiquity/client/controllers/filesystem.py
index 02c120f8..5dddba2f 100644
--- a/subiquity/client/controllers/filesystem.py
+++ b/subiquity/client/controllers/filesystem.py
@@ -236,6 +236,7 @@ class FilesystemController(SubiquityTuiController, FilesystemManipulator):
             raise Exception("could not process action {}".format(action))
 
+        await self.endpoint.probe_again.POST()
         if choice is not None:
             coro = self.endpoint.guided.POST(choice)
         else:
diff --git a/subiquity/common/apidef.py b/subiquity/common/apidef.py
index 85821a3d..03a4daa6 100644
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -253,6 +253,9 @@ class API:
         def GET(wait: bool = False) -> StorageResponse: ...
         def POST(config: Payload[list]): ...
 
+        class probe_again:
+            def POST() -> None: ...
+
         class reset:
             def POST() -> StorageResponse: ...
 
diff --git a/subiquity/server/controllers/filesystem.py b/subiquity/server/controllers/filesystem.py
index 0974d537..63758586 100644
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -305,6 +305,11 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             config, self.model._probe_data['blockdev'], is_probe_data=False)
         await self.configured()
 
+    async def probe_again_POST(self) -> None:
+        log.debug("requesting to start probing again!")
+        self._udev_event()
+
+
     def get_guided_disks(self, check_boot=True, with_reformatting=False):
         disks = []
         for raid in self.model._all(type='raid'):
@@ -543,6 +548,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
 
     @with_context(name='probe_once', description='restricted={restricted}')
     async def _probe_once(self, *, context, restricted):
+        await asyncio.sleep(5)
         if restricted:
             probe_types = {'blockdev'}
             fname = 'probe-data-restricted.json'
```

dbungert (approved):

PR #1460, opened 2022-10-20
autoinstall reference documentation update

ogayot:
This PR updates the autoinstall-reference documentation with the new keys that were added after the 20.04 release.

### DIFF
--- a/documentation/autoinstall-reference.md
+++ b/documentation/autoinstall-reference.md
@@ -473,12 +473,25 @@ The timezone to configure on the system. The special value "geoip" can be used t
 
 **type:** string (enumeration)
 **default:** `security`
+**can be interactive:** no
 
 The type of updates that will be downloaded and installed after the system install.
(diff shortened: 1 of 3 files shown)


documentation/autoinstall-reference.md
```diff
@@ -416,6 +461,58 @@ The installer will update the target with debconf set-selection values. Users wi
...
+
+**type:** string
+
+The flavor of the kernel, e.g., `generic` or `hwe`.
+
+### timezone
+
+**type:** string
+**default:** no timezone
+**can be interactive:** no <!-- TODO Should it be yes? There is no screen in the UI but a HTTP API -->
```
dbungert, line 488: I think "no" is the correct answer here.  No screen (yet).  There are aspirations to add something like that - the notion of having a time setting in the top orange bar has been thrown around.

dbungert (approved):

PR #1462, opened 2022-10-24
Start of UI for TPM-backed FDE

mwhudson:
This is definitely missing a few pieces (it should display an error in some situations vs just crashing) but it's a start. I'd also like to provide sample data for various different varieties of system and some integration (at least...) tests.

The main decision here are that we will change the source catalog to indicate the snapd label for the system being installed. We should probably extend this check if the system being installed defines a kernel and gadget snap -- a system that does not do this probably does not require any extra effort.

### DIFF
--- a/examples/tpm-sources.yaml
+++ b/examples/tpm-sources.yaml
@@ -43,3 +43,14 @@
   type: fsimage
   variant: server
   snapd_system_label: defective
+- description:
+    en: This test source requires TPM backed full disk encryption.
+  id: src-mandatory
+  locale_support: none
+  name:
+    en: TPM encryption required.
+  path: foo.squashfs
+  size: 530485248
+  type: fsimage
+  variant: server
+  snapd_system_label: mandatory
--- a/examples/tpm-sources.yaml
+++ b/examples/tpm-sources.yaml
@@ -1,6 +1,6 @@
 - description:
     en: This test source has encryption support set to "unavailable"
-  id: ubuntu-server-minimal
+  id: src-unavailable
   locale_support: none
   name:
     en: TPM encryption unavailable.
@@ -8,4 +8,38 @@
   size: 530485248
   type: fsimage
   variant: server
-  snapd_system_label: classic
+  snapd_system_label: unavailable
+- description:
+    en: This test source has encryption support set to "prefer-encrypted" (and encryption available)
+  id: src-prefer-encrypted
+  locale_support: none
+  name:
+    en: TPM encryption preferred.
+  path: foo.squashfs
+  size: 530485248
+  type: fsimage
+  variant: server
+  snapd_system_label: prefer-encrypted
+  default: true
+- description:
+    en: This test source has encryption support set to "prefer-unencrypted" (and encryption available)
+  id: src-prefer-unencrypted
+  locale_support: none
+  name:
+    en: TPM encryption not preferred.
+  path: foo.squashfs
+  size: 530485248
+  type: fsimage
+  variant: server
+  snapd_system_label: prefer-unencrypted
+- description:
+    en: This test source has the invalid combination of encryption reqired and unavailable.
+  id: src-defective
+  locale_support: none
+  name:
+    en: TPM encryption defective.
+  path: foo.squashfs
+  size: 530485248
+  type: fsimage
+  variant: server
+  snapd_system_label: defective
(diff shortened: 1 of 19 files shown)


subiquity/client/controllers/filesystem.py
```diff
@@ -85,6 +90,9 @@ async def _wait_for_probing(self):
                       self.ui.body)
 
     def make_guided_ui(self, status):
+        se = self.storage_encryption = status.storage_encryption
+        if se is not None and se.support == StorageEncryptionSupport.DEFECTIVE:
+            1/0  # should show an error page here
```
dbungert, line 95: Adding a note as a reminder to come back to this.  Even throwing an Exception with a message would be a good placeholder.

subiquity/client/controllers/filesystem.py
```diff
@@ -236,6 +244,9 @@ async def _answers_action(self, action):
             raise Exception("could not process action {}".format(action))
 
     async def _guided_choice(self, choice):
+        if self.storage_encryption is not None:
+            self.app.next_screen(self.endpoint.guided.POST(choice))
```
dbungert, line 248: The lines below have a code path for choice==None, I suppose we should plan for that here.

mwhudson: Do you mean we should error in the choice == None case? I'm not sure we'll ever support a completely custom layout for this stuff (maybe??? ) but it's certainly all a good way off.

subiquity/ui/views/filesystem/guided.py
```diff
@@ -144,6 +149,50 @@ def __init__(self, parent):
         self.disk.widget.index = initial
         connect_signal(self.use_lvm.widget, 'change', self._toggle)
         self.lvm_options.enabled = self.use_lvm.value
+        se = parent.storage_encryption
+        if se is not None:
+            self.remove_field('use_lvm')
+            self.remove_field('lvm_options')
+            if se.support == StorageEncryptionSupport.DISABLED:
```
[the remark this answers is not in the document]
mwhudson: Pushed something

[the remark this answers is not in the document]
dbungert: I do like this, it helps to shorten the init routine.  Thanks!

mwhudson:
This has an integration test now so I think it's perhaps ready for review. Maybe there are some more tests we could write?

examples/tpm-sources.yaml
```diff
@@ -1,11 +1,56 @@
...
+  id: src-prefer-unencrypted
+  locale_support: none
+  name:
+    en: TPM encryption not preferred.
+  path: foo.squashfs
+  size: 530485248
+  type: fsimage
+  variant: server
+  snapd_system_label: prefer-unencrypted
+- description:
```
dbungert, line 35: This one crashes with:
```
   File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
     self._context.run(self._callback, *self._args)
   File "/home/dbungert/dev/github/canonical/subiquity/subiquitycore/async_helpers.py", line 26, in _done
     fut.result()
   File "/home/dbungert/dev/github/canonical/subiquity/subiquity/server/controllers/filesystem.py", line 144, in _get_system
     self._system = await self.app.snapdapi.v2.systems[label].GET()
   File "/home/dbungert/dev/github/canonical/subiquity/subiquity/common/api/client.py", line 47, in impl
     async with make_request(
   File "/usr/lib/python3.10/contextlib.py", line 199, in __aenter__
     return await anext(self.gen)
   File "/home/dbungert/dev/github/canonical/subiquity/subiquity/server/snapdapi.py", line 252, in make_request
     content = await async_snapd.get(path[1:], **params)
   File "/home/dbungert/dev/github/canonical/subiquity/subiquitycore/snapd.py", line 179, in get
     response = await run_in_thread(
   File "/home/dbungert/dev/github/canonical/subiquity/subiquitycore/async_helpers.py", line 46, in run_in_thread
     return await loop.run_in_executor(None, func, *args)
   File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
     result = self.fn(*self.args, **self.kwargs)
   File "/home/dbungert/dev/github/canonical/subiquity/subiquitycore/snapd.py", line 169, in get
     raise Exception(
 Exception: Don't know how to fake GET response to ('v2/systems/mandatory', {})
```

dbungert (approved):
Code LGTM.
One of the source options crashes.

examples/tpm-sources.yaml
mwhudson: Added the missing file for this.

dbungert (approved):

PR #1463, opened 2022-10-24
Autoinstall doc update

ogayot:
This PR namely updates the quickstart guides (both the generic one and the one specific for s390x).
The instructions still apply well with the new kinetic release. I did testing with 1024MiB of RAM and it worked surprisingly well but I still decided to bump this value to 2048MiB ; which seems reasonable, The s390x quickstart guide already used 2048MiB of RAM so it is now consistent.

I didn't find a good way to avoid making the documentation dependent on a specific ubuntu release. Therefore, the documentation explicitly mentions 22.10 / kinetic but the instructions work the same on a jammy or focal image (tested).

It also fixes a mistake in the shutdown autoinstall section ; where the default value that I documented was wrong.

I think once I address any potential comments, the document will be ready to be published on discourse / wiki.

### DIFF
--- a/documentation/autoinstall-quickstart-s390x.md
+++ b/documentation/autoinstall-quickstart-s390x.md
@@ -6,15 +6,15 @@ This page is just a slightly adapted page of [the autoinstall quickstart page](a
 
 ## Download an ISO
 
-At the time of writing (just before focal release), the best place to go is here:
-<http://cdimage.ubuntu.com/ubuntu/releases/20.04/release/>
+At the time of writing (just after the kinetic release), the best place to go is here:
+<https://cdimage.ubuntu.com/ubuntu/releases/22.10/release/>
 
-<pre><code>wget http://cdimage.ubuntu.com/ubuntu/releases/20.04/release/ubuntu-20.04-live-server-s390x.iso -P ~/Downloads</code></pre>
+<pre><code>wget https://cdimage.ubuntu.com/ubuntu/releases/22.10/release/ubuntu-22.10-live-server-s390x.iso -P ~/Downloads</code></pre>
 
 ## Mount the ISO
 
 <pre><code>mkdir -p ~/iso
-sudo mount -r ~/Downloads/ubuntu-20.04-live-server-s390x.iso ~/iso</code></pre>
+sudo mount -r ~/Downloads/ubuntu-22.10-live-server-s390x.iso ~/iso</code></pre>
 
 ## Write your autoinstall config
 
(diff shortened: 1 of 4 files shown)


documentation/autoinstall-quickstart-s390x.md
```diff
@@ -6,15 +6,15 @@ This page is just a slightly adapted page of [the autoinstall quickstart page](a
 
 ## Download an ISO
 
-At the time of writing (just before focal release), the best place to go is here:
-<http://cdimage.ubuntu.com/ubuntu/releases/20.04/release/>
+At the time of writing (just after the kinetic release), the best place to go is here:
+<https://cdimage.ubuntu.com/ubuntu/releases/22.10/release/>
```
dbungert, line 10: Certainly good to update this bit.
What do you think about generalizing this mention and changing it to point to https://releases.ubuntu.com/
If we do leave this, there is the mild inconsistency of sometimes pointing to cdimage and sometimes releases, we should pick one (releases would be my preference).

documentation/autoinstall.md
```diff
@@ -2,7 +2,9 @@
 
 ## Introduction
 
-The server installer for 20.04 supports a new mode of operation: automated installation, autoinstallation for short. You might also know this feature as unattended or handsoff or preseeded installation.
+Since version 20.04, the server installer supports the automated installation
```
dbungert, line 5: Is Discourse smart enough to flow these correctly like it was one long line?  I think the answer is no, it would be nice to 80 column these files but I don't think we can without a post-processing step.

documentation/autoinstall-reference.md
```diff
@@ -504,7 +504,7 @@ Supported values are:
 ### shutdown
 
 **type:** string (enumeration)
-**default:** do nothing
```
dbungert, line 507: That reminds me that we actually want a do-nothing power mode (not this PR of course).

documentation/autoinstall-quickstart.md
```diff
@@ -2,19 +2,24 @@
 
 The intent of this page is to provide simple instructions to perform an autoinstall in a VM on your machine.
 
-This page assumes you are on the amd64 architecture. There is a version for [s390x](autoinstall-quickstart-s390x.md) too.
+This page assumes that you are willing to install the latest Ubuntu release
```
dbungert, line 5: See the other comment about long lines.

dbungert (approved):
A few tweaks and we can call this done I think.

documentation/autoinstall-quickstart-s390x.md
ogayot: Thanks!
Sadly, releases.ubuntu.com only offers downloads for amd64 whereas cdimage.ubuntu.com offers downloads for everything **but** amd64...

There is a link on releases.ubuntu.com:

>The cdimage server also hosts releases of other Ubuntu images not found on this server, such as builds for less popular architectures and other non-standard and unsupported images. For Ubuntu Desktop and Server on popular architectures, please see the links above instead.
>
>[Ports, Unsupported, and Experimental Images for Ubuntu](http://cdimage.ubuntu.com/releases/)

documentation/autoinstall-reference.md
ogayot: :+1: that would be good to have!

documentation/autoinstall.md
ogayot: Thanks! Looks like you're right .. Well I'll discard the whitespace changes for now then.

documentation/autoinstall-quickstart-s390x.md
dbungert: oh right

PR #1466, opened 2022-10-27
extract mount handling bits out of AptConfigurer

mwhudson:
I want this for some FDE bits. I gave it a vm test successfully, score one for the unit and integration tests which certainly found plenty of bugs :)

### DIFF
--- a/subiquity/server/mounter.py
+++ b/subiquity/server/mounter.py
@@ -1,0 +1,186 @@
+# Copyright 2022 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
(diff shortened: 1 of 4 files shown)


subiquity/server/mounter.py
```diff
@@ -0,0 +1,186 @@
...
+
+    async def cleanup(self):
+        for m in reversed(self._mounts):
+            await self.unmount(m, remove=False)
+        self.tmpfiles.cleanup()
+
+
+class DryRunMounter(Mounter):
+
+    async def setup_overlay(self, lowers: List[Lower]) -> OverlayMountpoint:
```
ogayot, line 164: We'll probably want to rework this function to make it less coupled with the APT stuff.
Would you add a TODO ?

ogayot (approved):
All this overlay & mount stuff in APT felt out of place. For sure, this is a nice improvement and it makes the code easier to read. Thanks!

PR #1467, opened 2022-10-30
shutdown: add 'wait' mode

dbungert:
Add a 'wait' mode for shutdown for autoinstall.  The idea here is that an install can be automated but it waits at the end before rebooting.

During autoinstall, a new prompt may be found at the end, allowing the user to initiate poweroff or reboot.


dbungert:
While I mostly consider this ready I'm not sure yet what's going on with system setup.

PR #1468, opened 2022-11-01
support setting filesystem label and partition name

mwhudson:


subiquity/models/filesystem.py
```diff
@@ -1554,7 +1556,7 @@ def add_dm_crypt(self, volume, key):
     def remove_dm_crypt(self, dm_crypt):
         self._remove(dm_crypt)
 
-    def add_filesystem(self, volume, fstype, preserve=False):
+    def add_filesystem(self, volume, fstype, preserve=False, label=False):
```
dbungert, line 1559:
```suggestion
    def add_filesystem(self, volume, fstype, preserve=False, label=None):
```

dbungert (changes requested):

subiquity/models/filesystem.py
mwhudson: doh

dbungert (approved):

PR #1469, opened 2022-11-01
notice errors when an async snapd operation fails

mwhudson:


dbungert (approved):
I'm sure this is fine.  What sort of things can trigger this?

mwhudson:
An example of this I have seen is snapd's "finish installation" step failing with a device busy error. I should see if I can commit some appropriate sample data ...

mwhudson:
Oh right, branch ordering things. I'll add some failing data to my tpm-install branch.

PR #1470, opened 2022-11-02
do partitioning based on information from gadget

mwhudson:


subiquity/server/controllers/filesystem.py
```diff
@@ -377,11 +380,59 @@ async def guided_GET(self, wait: bool = False) -> GuidedStorageResponse:
...
+                       structure: snapdapi.VolumeStructure):
+        ptype = structure.type.split(',', 1)[1].upper()
+        if structure.offset is not None:
+            offset = structure.offset
+        else:
+            offset = next_offset
+        if structure.role == snapdapi.Role.SYSTEM_DATA and is_last:
+            size = gaps.largest_gap(disk).size
+        else:
+            size = structure.size
```
dbungert, line 393: Do we have checks at some point that there is enough room?

subiquity/server/controllers/filesystem.py
```diff
@@ -377,11 +380,59 @@ async def guided_GET(self, wait: bool = False) -> GuidedStorageResponse:
...
+        next_offset = offset + size
```
dbungert, line 394: Please tell me this is gpt only.

subiquity/server/controllers/filesystem.py
```diff
@@ -377,11 +380,59 @@ async def guided_GET(self, wait: bool = False) -> GuidedStorageResponse:
             disks=[labels.for_client(d, min_size=min_size) for d in disks],
             storage_encryption=se)
 
+    def _add_structure(self, *, disk: Disk, next_offset: int, is_last: bool,
+                       structure: snapdapi.VolumeStructure):
+        ptype = structure.type.split(',', 1)[1].upper()
```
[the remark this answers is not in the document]
mwhudson: sure

subiquity/server/controllers/filesystem.py
mwhudson: this comment prompted me to refactor how we make the assertions to need to about the system we're installing. and yes, gpt only!

subiquity/server/controllers/filesystem.py
mwhudson: no. i guess we should filter the disks we present to only show ones that the gadget will git on.

dbungert: Yes, that would go a long way.  For now we can assume reformatting the entire disk and not do side-by-side, correct?

subiquity/server/controllers/filesystem.py
dbungert: :tada:

subiquity/server/controllers/filesystem.py
mwhudson: now done

mwhudson: > Yes, that would go a long way. For now we can assume reformatting the entire disk and not do side-by-side, correct?

weellll there is an incoming wibble about recovery partitions but for now lets say yes

dbungert: That concerns me less, that's a predictable size section as opposed to coping with arbitrary user disk layout.

subiquity/server/controllers/filesystem.py
```diff
@@ -371,17 +405,64 @@ async def guided_GET(self, wait: bool = False) -> GuidedStorageResponse:
         se = None
         if self._system is not None:
             se = self._system.storage_encryption
+            for _, offset, size in self._offsets_and_sizes_for_system():
```
dbungert, line 408: I'm nitpicking at this point but we don't need to iterate to assign the min size if we only care about  the last entry.  This structure does give the reader the impression that we need to accumulate all the values.

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -457,3 +458,68 @@ async def test_lvm_20G_bad_offset(self, disk_size):
...
+
+    def test_from_sample_data(self):
+        self.fsc.model = model = make_model(Bootloader.UEFI)
+        disk = make_disk(model)
+        with open('examples/snaps/v2-systems-unavailable.json') as fp:
+            self.fsc._system = snapdapi.snapd_serializer.deserialize(
+                snapdapi.SystemDetails, json.load(fp)['result'])
+        self.fsc.apply_system(disk.id)
+        self.assertEqual(
+            len(self.fsc._system.volumes['pc'].structure) - 1,
```
dbungert, line 522: The -1 is due to mbr and bios boot being split in the json, correct?
A quick comment would be nice.

dbungert (approved):

subiquity/server/controllers/filesystem.py
mwhudson: fair enough.

subiquity/server/controllers/tests/test_filesystem.py
mwhudson: I've tried to make this clearer

subiquity/server/controllers/filesystem.py
dbungert: LGTM

subiquity/server/controllers/tests/test_filesystem.py
dbungert: LGTM

PR #1471, opened 2022-11-03
Ensure the first user is in the users group and not in the lxd group

kenvandine:
Ensure the first user is in the users group and not in the lxd group.  This is necessary for new installs to support unprivileged lxd

### DIFF
--- a/subiquity/models/tests/test_subiquity.py
+++ b/subiquity/models/tests/test_subiquity.py
@@ -56,7 +56,7 @@ uucp:x:10:
 man:x:12:
 sudo:x:27:
 ssh:x:118:
-lxd:x:132:
+users:x:100:
 '''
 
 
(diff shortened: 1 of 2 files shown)


mwhudson:
This isn't going to work as is - the users-and-groups is generated at snap build time from data in other packages. As subiquity becomes the single installer maybe it's time to change that...

kenvandine:
> This isn't going to work as is - the users-and-groups is generated at snap build time from data in other packages. As subiquity becomes the single installer maybe it's time to change that...

@mwhudson I was worried about that, suggestions?

dbungert:
To elaborate on status quo, this file is generated at snap build time from data in the distro package `user-setup`.  Is this a change that should be done in the larger distro?

seb128:
I've uploaded the change to lunar as https://launchpad.net/ubuntu/+source/user-setup/1.90ubuntu3 , how do we get that picked up by subiquity?

ogayot:
> I've uploaded the change to lunar as https://launchpad.net/ubuntu/+source/user-setup/1.90ubuntu3 , how do we get that picked up by subiquity?

It looks like for users-and-groups, we're installing user-setup in the build environment and then query the list of users using `echo "get passwd/user-default-groups" | debconf-communicate user-setup`.

To me that means we're fetching the package from focal / focal-updates since we're based on core20.

We might need to change the build system if we want to pick it up from some other location. Maybe we need to align with what we're doing for reserved-usernames:
https://github.com/canonical/subiquity/blob/main/snapcraft.yaml#L165

PR #1473, opened 2022-11-03
codecs: add ability to install ubuntu-restricted-addons

ogayot:
Subiquity now supports a new endpoint that can be used by the desktop installer to configure whether the `ubuntu-restricted-addons` package should be installed. This package contains third-party codecs commonly used on a desktop install.

```bash
$ curl --unix-socket /run/subiquity/socket http://a/codecs
{"install": false}

$ curl --unix-socket /run/subiquity/socket http://a/codecs -d '{"install": true}'

$ curl --unix-socket /run/subiquity/socket http://a/codecs
{"install": true}
```
I was considering naming this endpoint something else than `codecs`, like `restricted-addons` maybe?

### DIFF
--- a/documentation/autoinstall-reference.md
+++ b/documentation/autoinstall-reference.md
@@ -411,6 +411,23 @@ A list of SSH public keys to install in the initial user's account.
 **type:** boolean
 **default:** `true` if `authorized_keys` is empty, `false` otherwise
 
+<a name="codecs"></a>
+
+### codecs
+
+**type:** mapping, see below
+**default:** see below
+**can be interactive:** no
+
+Configure whether common restricted packages (including codecs) from [multiverse] should be installed.
+
+#### install
+
+**type:** boolean
+**default:** `false`
+
+Whether to install the ubuntu-restricted-addons package.
+
 <a name="drivers"></a>
 
 ### drivers
--- a/autoinstall-schema.json
+++ b/autoinstall-schema.json
@@ -420,6 +420,14 @@
                 "additionalProperties": false
             }
         },
+        "codecs": {
+            "type": "object",
+            "properties": {
+                "install": {
+                    "type": "boolean"
+                }
+            }
+        },
         "drivers": {
             "type": "object",
             "properties": {
--- a/documentation/autoinstall-schema.md
+++ b/documentation/autoinstall-schema.md
@@ -442,6 +442,14 @@ The [JSON schema](https://json-schema.org/) for autoinstall data is as follows:
                 "additionalProperties": false
             }
         },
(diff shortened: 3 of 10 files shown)


documentation/autoinstall-reference.md
```diff
@@ -411,6 +411,23 @@ A list of SSH public keys to install in the initial user's account.
 **type:** boolean
 **default:** `true` if `authorized_keys` is empty, `false` otherwise
 
+<a name="codecs"></a>
```
dbungert, line 414: Would you merge the documentation directory to ubuntu/kinetic?  As we add main branch features, we should not use that as the basis for the documentation on Discourse.

subiquity/server/controllers/install.py
```diff
@@ -339,6 +339,13 @@ async def postinstall(self, *, context):
                 ubuntu_drivers = self.app.controllers.Drivers.ubuntu_drivers
                 await ubuntu_drivers.install_drivers(root_dir=self.tpath(),
                                                      context=child)
+        if self.model.codecs.do_install:
+            # NOTE currently, ubuntu-restricted-addons is an empty package that
```
dbungert, line 343: Does this imply that we need curtin changes?

subiquity/server/controllers/install.py
```diff
@@ -339,6 +339,13 @@ async def postinstall(self, *, context):
...
+            # pulls relevant packages through Recommends:
+            # Ideally, we should make sure to run the APT command with
+            # the --install-recommends option.
+            await self.install_package(context=context,
```
mwhudson, line 347: Instead of doing this here you can implement target_packages on CodecsModel

subiquity/server/controllers/install.py
ogayot: I tried with a kinetic server ISO and apt installed the recommended packages so I think there is no need for an immediate change in curtin. If we were to build an image with `APT::Install-Recommends "false"`, the codecs would not get installed.

subiquity/server/controllers/install.py
ogayot: Oh, right, that would be much cleaner. Thanks

documentation/autoinstall-reference.md
ogayot: Good point. I've just opened https://github.com/canonical/subiquity/pull/1474.

mwhudson:
This seems fine. I guess we should tag @jpnurmi to see the API changes before approving/merging.

dbungert (approved):

jpnurmi:
:+1:

PR #1476, opened 2022-11-07
support installing an unencrypted core boot classic system

mwhudson:
I have used code like this (but not exactly this) to install a core boot classic system in a VM.


subiquity/server/controllers/install.py
```diff
@@ -245,30 +245,49 @@ def make_curtin_step(name, stages, acquire_config, *,
...
+                make_curtin_step(
+                    name="extract", stages=["extract"],
+                    acquire_config=self.acquire_generic_config,
+                    ).run,
+                fs_controller.finish_install,
+                self.setup_target,
+                ])
+        else:
+            steps.extend([
+                make_curtin_step(
```
dbungert, line 273: duplication with if block

dbungert (approved):
This all looks fine but some sort of automated testing would be nice.

subiquity/server/controllers/install.py
mwhudson: ah yes, that will go away in the next branch

mwhudson:
> This all looks fine but some sort of automated testing would be nice.

Added something and fixed the bug this inevitably revealed...

dbungert (approved):

PR #1479, opened 2022-11-09
call into snapd to set up encryption when required

mwhudson:


subiquity/server/controllers/install.py
```diff
@@ -258,7 +259,19 @@ def make_curtin_step(name, stages, acquire_config, *,
             steps.extend([
```
ogayot, line 259: steps.append() ?

ogayot (approved):
The logic seems good.  Just a few minor suggestions.

subiquity/models/filesystem.py
```diff
@@ -958,6 +958,19 @@ def size(self):
         return self.volume.size - LUKS_OVERHEAD
 
 
+@fsobj("device")
+class ArbitraryDevice(_Device):
+    ptable = attr.ib(default=None)
```
dbungert, line 963: not that it's a new problem but we're overdue for a ptable enum

subiquity/models/filesystem.py
```diff
@@ -1444,13 +1457,29 @@ def can_emit(obj):
                 raise Exception("\n".join(msg))
             work = next_work
 
+        if early:
+            r = [act for act in r if act['type'] not in ('format', 'mount')]
+        if late:
+            r = [act for act in r if act['type'] in ('format', 'mount')]
```
dbungert, line 1463: The connection between early & late & mounts is not immediately clear, I suggest a comment.
if early and later are both true, we'll drop the list from r, which I assume is always an error but would go unnoticed.

scripts/validate-yaml.py
```diff
@@ -75,13 +75,17 @@ def final_checks(self):
...
 
 
-config = yaml.safe_load(open(sys.argv[1]))
 
 
 def main():
     storage_checker = StorageChecker()
 
-    for action in config['storage']['config']:
+    actions = []
```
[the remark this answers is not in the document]
mwhudson: well ok. I think Dan is going to make this script obsolete soon so I'll leave it for now :)

subiquity/server/controllers/install.py
mwhudson: fair

subiquity/models/filesystem.py
mwhudson: heh yes. not right now though!

subiquity/models/filesystem.py
mwhudson: I did a bit of refactoring here to make this all a bit clearer (basically _render_actions now takes an enum).

dbungert: Thank you, this leaves less room for surprises for the caller.

dbungert (approved):

PR #1480, opened 2022-11-11
do not switch snap channel from .disk/info if not on ubuntu/stable-XX.YY

mwhudson:
The point of switching snap channel based on .disk/info is so that we can avoid releasing updates to point release media if they won't work there. If the snap is not tracking ubuntu/stable-XX.YY then there is funny business afoot and we are not on point release media, so switching channels as if we are is just unhelpful (for example, if you are working on a project that is going to involve giving someone an image with a subiquity from a non default track).

If the channel to switch to comes from the kernel command line or autoinstall though, it should still be honoured.

### DIFF
--- a/subiquity/server/controllers/tests/test_refresh.py
+++ b/subiquity/server/controllers/tests/test_refresh.py
@@ -1,0 +1,112 @@
+# Copyright 2022 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


subiquity/server/controllers/tests/test_refresh.py
```diff
@@ -0,0 +1,112 @@
...
+            AsyncSnapd(get_fake_connection()))
+        self.rc = RefreshController(app=self.app)
+
+    async def test_configure_snapd_kernel_autoinstall(self):
+        # If a snap channel is supplied via kernel command line or
+        # autoinstall data, switch to it.
+
+        for src in SnapChannelSource.CMDLINE, SnapChannelSource.AUTOINSTALL:
+            with mock.patch.object(self.rc, 'get_refresh_channel') as grc:
+                grc.return_value = ('newchan', SnapChannelSource.CMDLINE)
```
dbungert, line 48: src is unused, intended for use here presumably

dbungert (approved):
LGTM with quick test fix

subiquity/server/controllers/tests/test_refresh.py
mwhudson: oops, thanks

PR #1481, opened 2022-11-11
force storage version to 2 for a core boot classic install

mwhudson:
also a barely-related tweak to more accurately decide which path to take
in installation


mwhudson:
This is based on top of mwhudson/tmp-encrypt so will be easier to review after that is merge.

subiquity/server/controllers/install.py
```diff
@@ -187,10 +189,15 @@ def acquire_initial_config(self,
...
-                                  ) -> Dict[str, Any]:
+    def acquire_filesystem_config(
+            self, step: CurtinPartitioningStep,
+            mode: ActionRenderMode = ActionRenderMode.DEFAULT,
+            ) -> Dict[str, Any]:
         cfg = self.acquire_initial_config(step)
-        cfg.update(self.model.filesystem.render())
+        version = None
+        if mode != ActionRenderMode.DEFAULT:
+            version = 2
```
dbungert, line 199: Isn't this too late?  I worry about earlier parts of the code having made different decisions (the gaps code for example having a completely different code path) than it would have if it was v2.

mwhudson: I think probably the way the code is now it isn't actually a problem but you're definitely right that this is a surprise waiting to happen. I'll try something else :)

(v2 by default everywhere when? :) )

dbungert (approved):

PR #1482, opened 2022-11-11
configure controllers tui does not support when installing desktop

mwhudson:
We could definitely make this more elegant. But what do you think?

### DIFF
--- a/examples/tpm-sources.yaml
+++ b/examples/tpm-sources.yaml
@@ -7,7 +7,7 @@
   path: foo.squashfs
   size: 530485248
   type: fsimage
-  variant: server
+  variant: desktop
   snapd_system_label: unavailable
 - description:
     en: This test source has encryption support set to "prefer-encrypted" (and encryption available)
@@ -18,7 +18,7 @@
   path: foo.squashfs
   size: 530485248
   type: fsimage
-  variant: server
+  variant: desktop
   snapd_system_label: prefer-encrypted
   default: true
 - description:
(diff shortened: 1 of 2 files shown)


subiquity/ui/views/source.py
```diff
@@ -54,6 +55,7 @@ def __init__(self, controller, sources, current_id, search_drivers: bool):
                 ns[source.id] = RadioButtonField(
                     group, source.name, '\n' + source.description)
                 initial[source.id] = source.id == current_id
+                self.sources_by_id[source.id] = source
```
dbungert, line 58: A unit test on population of sources_by_id would be nice to have.

subiquity/ui/views/source.py
```diff
@@ -93,7 +95,8 @@ def done(self, result):
             if k == "search_drivers":
                 continue
             if v:
-                self.controller.done(k, search_drivers=search_drivers)
+                self.controller.done(
```
dbungert, line 98: This done method is a bit unclear but that's not new to this PR.

dbungert (approved):

mwhudson:
I accidentally rewrote this to be more generic, more local and probably even dirtier...

mwhudson:
I changed this again in a way that probably depends on #1483 but I'm thoroughly confused by now.

mwhudson:
OK back to a variant (ha!) that doesn't depend on understanding the difference between the two variants.

dbungert (approved):

PR #1483, opened 2022-11-14
simplify variant handling between client and server

mwhudson:


mwhudson:
Hmm actually maybe the real point here is that app.variant and source.variant are different?

dbungert:
My long-term goal with all of this might be summarized as - if we had a install-sources.yaml where some entries were 'server' and some where 'desktop', would we handle it appropriately?

PR #1484, opened 2022-11-15
remove any support for a desktop installer without a source catalog

mwhudson:
The code in subiquity main could plausibly be offered as an upgrade to users installing focal server and focal server ISOs do not have a source catalog so we need to support that. But we will never support a desktop ISO that does not have a source catalog so we can remove some slightly confusing code that attempted to cover that case.

### DIFF
--- a/subiquity/models/source.py
+++ b/subiquity/models/source.py
@@ -38,35 +38,23 @@ class CatalogEntry:
     snapd_system_label: typing.Optional[str] = None
 
 
-fake_entries = {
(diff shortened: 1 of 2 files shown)


mwhudson:
I'm still a bit confused about the other variant handling stuff, but I'm pretty sure this change is sound and it fixes the visible problem that started me looking at all of this.

dbungert (approved):
Oh thank you, I didn't understand the fake source entries.

PR #1485, opened 2022-11-16
reuse an existing partition if possible in apply_system

mwhudson:
This required a bit of a refactor which lead to rewriting tests as well as adding a new one.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -496,6 +496,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
                 part.partition_name = structure.name
             if structure.filesystem:
                 part.wipe = 'superblock'
+                self.delete_filesystem(part.fs())
                 fs = self.model.add_filesystem(
                     part, structure.filesystem, label=structure.label)
                 if structure.role == snapdapi.Role.SYSTEM_DATA:
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -452,43 +452,60 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         for structure in volume.structure:
             if structure.role == snapdapi.Role.MBR:
                 continue
-            if ',' not in structure.type:
-                continue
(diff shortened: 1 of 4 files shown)


mwhudson:
grumble something i have done has made the integration tests flaky :(

subiquity/server/controllers/filesystem.py
```diff
@@ -452,43 +452,60 @@ def _offsets_and_sizes_for_system(self):
...
-            self._add_structure(
-                disk=disk, offset=offset, size=size, structure=structure,
-                is_last=structure == volume.structure[-1])
+            if (offset, size) in parts_by_offset_size:
+                part = parts_by_offset_size[(offset, size)]
+            else:
+                if structure.role == snapdapi.Role.SYSTEM_DATA and \
+                   structure == volume.structure[-1]:
+                    gap = gaps.largest_gap(disk)
+                    size = gap.size - (offset - gap.offset)
```
dbungert, line 492: This is just trying to use the rest of the disk, right?  Why the decrease in size?  Wouldn't it just be gap.size?

subiquity/server/snapdapi.py
```diff
@@ -142,7 +142,10 @@ class VolumeStructure:
     update: VolumeUpdate = attr.Factory(VolumeUpdate)
 
     def gpt_part_uuid(self):
-        return self.type.split(',', 1)[1].upper()
+        if ',' in self.type:
```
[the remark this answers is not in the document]
mwhudson: I hope not... Maybe I should just add code that rejects a non-gpt schema.

subiquity/server/controllers/filesystem.py
mwhudson: Because the offset in the gadget isn't necessarily immediately after the preceding structure.

dbungert: I think that works then, so long as the preceding structure obeys 1MiB alignment rules.

dbungert (approved):

mwhudson:
This contains another unrelated fix, but with this branch I was able to install an encrypted system! Hurrah!

PR #1486, opened 2022-11-17
workflow: +lunar

dbungert:


ogayot (approved):

mwhudson:
I wonder why this isn't working? Something wrong in cloud-init?

ogayot:
> I wonder why this isn't working? Something wrong in cloud-init?

seems like it. I've opened https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1997559 to track the issue. The regression might end up SRUed to stable releases, which would affect us even more.

ogayot:
This should be fixed once cloud-init 22.4.2-0ubuntu1 makes it out of -proposed:

https://launchpad.net/ubuntu/+source/cloud-init/22.4.2-0ubuntu1

PR #1488, opened 2022-11-21
add a new member of the snapdapi.Role enum

mwhudson:
This kind of change is a bit of a problem with the current approach to talking to snapd I guess -- deserialization will fail if we encounter a new value for one of these fields.


dbungert (approved):
Is anything else in https://github.com/snapcore/snapd/blob/e51daa8fa5ebb90d3336b0863531c169dd5d293d/gadget/gadget.go#L49 interesting?

dbungert (approved):
That does sound safer.

PR #1490, opened 2022-11-21
Fetch SSH keys through the configured proxy

ogayot:
The proxy settings specified by the user were ignored when importing SSH keys from github / launchpad. This PR fixes the issue by explicitly passing the https_proxy environment variable when a proxy is set.

The calls to `ssh-import-id` used to be done on the client side of Subiquity. Our design makes it difficult to know in a given screen what proxy settings are required to access the network.

I decided to move the calls to `ssh-import-id` on the server side and expose an API to do it. Not only does it allow us to pass the relevant proxy variables in the environment, it could also be used to pass locale-related variables (e.g., LC_MESSAGES) which we are not doing as part of this PR.

From the client's standpoint, the behavior should be identical to status quo when no proxy is needed. The screens should look exactly the same.

Also, I noticed that when importing multiple keys from Github / Launchpad, we did not discard empty lines. This resulted in python objects and yaml objects having empty strings. e.g.:

```python
      authorized_keys = [
        'ssh-rsa AAAA[...] user@hostname',
        '',
        'ssh-ed255129 AAAA[...] user@hostname2',
      ]
```
I added some logic to discard these empty lines /  keys.

I haven't yet been able to test in a VM because of an unrelated issue when bind-mounting /var/lib/snapd/seed/systems so I'm leaving the PR as a draft for now. Also, the PR depends on https://github.com/canonical/subiquity/pull/1489 - which we use to display the error message from `ssh-import-id` and `ssh-keygen`. I'll rebase on top of main once this other PR is merged.

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -52,6 +52,7 @@ from subiquity.common.types import (
     SnapSelection,
     SourceSelectionAndSetting,
     SSHData,
+    SSHFetchIdResponse,
     LiveSessionSSHInfo,
     StorageResponse,
     StorageResponseV2,
@@ -75,7 +76,6 @@ class API:
     """The API offered by the subiquity installer process."""
     locale = simple_endpoint(str)
     proxy = simple_endpoint(str)
-    ssh = simple_endpoint(SSHData)
     updates = simple_endpoint(str)
     wslconfbase = simple_endpoint(WSLConfigurationBase)
     wslconfadvanced = simple_endpoint(WSLConfigurationAdvanced)
@@ -374,6 +374,13 @@ class API:
 
         class validate_username:
             def GET(username: str) -> UsernameValidation: ...
+
+    class ssh:
+        def GET() -> SSHData: ...
+        def POST(data: Payload[SSHData]) -> None: ...
+
+        class fetch_id:
+            def GET(user_id: str) -> SSHFetchIdResponse: ...
 
 
 class LinkAction(enum.Enum):
--- a/subiquity/server/controllers/ssh.py
+++ b/subiquity/server/controllers/ssh.py
@@ -31,6 +31,13 @@ from subiquity.server.controller import SubiquityController
 log = logging.getLogger('subiquity.server.controllers.ssh')
 
 
+class SSHFetchError(Exception):
+    def __init__(self, status: SSHFetchIdStatus, reason: str) -> None:
+        self.reason = reason
+        self.status = status
+        super().__init__()
+
+
 class SSHController(SubiquityController):
 
     endpoint = API.ssh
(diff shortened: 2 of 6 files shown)


subiquity/server/controllers/ssh.py
```diff
@@ -59,8 +68,52 @@ async def GET(self) -> SSHData:
...
+            return SSHFetchIdResponse(
+                    tatus=SSHFetchIdStatus.FINGERPRINT_ERROR,
+                    identities=None, error=e.stderr)
+
+        fingerprints: str = cp.stdout.replace(
+                f'#ssh-import-id {user_id}', '').strip()
+
+        zipped = zip(
+                [mat for mat in keys_material.splitlines() if mat],
+                fingerprints.splitlines())
```
mwhudson, line 110: I'm a tiny bit worried that this might end up matching the key with the wrong fingerprint or (probably worse) end up  dropping a key or a fingerprint if the lists end up mismatched. Maybe it's better to run each key through ssh-keygen individually? (maybe not something for this branch I dunno)

mwhudson (commented):
Hi, thanks for changing this, it is of course silly that this was being done client side, I must have missed / skipped that when doing the client server split and never got back to this.

BTW, I've wanted to change this screen forever to have an "import SSH key" button that fetches the key rather than doing it as part of moving to the next screen but never gotten around to it. It doesn't make sense to do that in this branch but maybe now you've got your head into this code it wouldn't be so bad?

subiquity/common/apidef.py
```diff
@@ -375,6 +376,13 @@ def POST(data: Payload[IdentityData]): ...
         class validate_username:
             def GET(username: str) -> UsernameValidation: ...
 
+    class ssh:
```
mwhudson, line 379: oh, you forgot to remove the initial definition of the ssh endpoint i think?

ogayot: oh yeah, good catch!

ogayot:
> BTW, I've wanted to change this screen forever to have an "import SSH key" button that fetches the key rather than doing it as part of moving to the next screen but never gotten around to it. It doesn't make sense to do that in this branch but maybe now you've got your head into this code it wouldn't be so bad?

Right, I think it would make sense too. Submitting data when clicking next prevents the user from adding multiple keys. Also, if the user navigates to the next screen and then presses Back, he would come back to the SSH screen with previously added keys discarded.

subiquity/server/controllers/ssh.py
```diff
@@ -59,8 +68,50 @@ async def GET(self) -> SSHData:
...
+        # ssh-keygen multiple times.
+        fingerprint_command = ('ssh-keygen', '-l', '-f', '-')
+        for key_material in (mat for mat in keys_material.splitlines() if mat):
+            try:
+                cp = await arun_command(fingerprint_command, check=True,
+                                        input=key_material)
+            except subprocess.CalledProcessError as e:
+                log.exception("ssh-import-id failed. stderr: %s", e.stderr)
+                return SSHFetchIdResponse(
+                        tatus=SSHFetchIdStatus.FINGERPRINT_ERROR,
```
mwhudson, line 105: tatus? :)

ogayot: :facepalm: Thanks, I will take this as an opportunity to include unit tests..

mwhudson (approved):
Thanks, this looks great!

ogayot:
Testing in a VM kept failing because of https://bugs.launchpad.net/subiquity/+bug/1997538. So I did testing of the SSH stuff based on ubuntu/kinetic. Works as intended :+1: .

PR #1491, opened 2022-11-22
fixes from testing core boot classic install

mwhudson:
The removal of the align in add_partition is definitely a bit sneaky but I think / hope all alignment issues are handled in other code now. I guess we should check that before landing this. The other changes are sound I think.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1588,12 +1588,12 @@ class FilesystemModel(object):
     def add_partition(self, device, *, size, offset, flag="", wipe=None,
                       grub_device=None, partition_name=None):
         from subiquity.common.filesystem import boot
-        real_size = align_up(size)
-        log.debug("add_partition: rounded size from %s to %s", size, real_size)
+        #real_size = align_up(size)
+        #log.debug("add_partition: rounded size from %s to %s", size, real_size)
         if device._fs is not None:
             raise Exception("%s is already formatted" % (device,))
         p = Partition(
-            m=self, device=device, size=real_size, flag=flag, wipe=wipe,
+            m=self, device=device, size=size, flag=flag, wipe=wipe,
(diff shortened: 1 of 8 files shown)


subiquity/models/filesystem.py
```diff
@@ -1588,12 +1588,12 @@ def _remove(self, obj):
...
         from subiquity.common.filesystem import boot
-        real_size = align_up(size)
-        log.debug("add_partition: rounded size from %s to %s", size, real_size)
+        #real_size = align_up(size)
+        #log.debug("add_partition: rounded size from %s to %s", size, real_size)
         if device._fs is not None:
             raise Exception("%s is already formatted" % (device,))
         p = Partition(
-            m=self, device=device, size=real_size, flag=flag, wipe=wipe,
+            m=self, device=device, size=size, flag=flag, wipe=wipe,
```
dbungert, line 1596: While I'm in agreement with this change, I don't think we should batch it with other things and I want to ensure it's tested very well.  I suggest splitting this PR up a bit, particularly with a PR focused on this change and the fallout.

dbungert (changes requested):

subiquity/models/filesystem.py
mwhudson: oh 100% agreed

mwhudson: When you say that you want to see this be very well tested, do you have specific things in mind? I can stare at all the code paths that end up in add_partition and i'm very very nearly convinced there are no problems lurking (possibly there is an issue around the very fine details of resizing a pre-existing partition with an unaligned size, but that's not really very likely to be hit in real life and not actually to do with the proposed change). But it is hard to be absolutely sure!

mwhudson:
All of the good stuff from this branch is in review in other PRs now.

PR #1492, opened 2022-11-25
test_api: remove execution bit from the test_api.py script

ogayot:
The test_api.py is not meant to be used as en entry point. There is no shebang so executing the script from a sh-compatible shell can be pretty messy.

I accidentally executed the script from bash and it sort of froze all my windows. Turns out it's because it executed `import` (from the imagemagick package) which takes screen captures.


mwhudson (approved):
Hah oops.

PR #1493, opened 2022-11-25
Fix crash when /var/lib/snapd/seed/systems does not exist on source

ogayot:
Subiquity attempts to bind-mount `/var/lib/snapd/seed/systems` from the source to the host. Unfortunately, in standard ubuntu-server installs, the directory does not exist so the install crashes early with a `CalledProcessError` (mount: No such file or directory). 

https://bugs.launchpad.net/subiquity/+bug/1997538

This PR essentially checks if the directory exists before trying to bind-mount it.

Unfortunately, adding this check impacts dry-run executions (and unit tests) because the behavior becomes dependent on the presence of `/var/lib/snapd/seed/systems` on the host (in dry-run mode, the source path is set to `/` so we look for the `systems` directory on the host).

I took the opportunity to introduce a configuration object that controls what Subiquity does in dry-run mode (I have wanted to do this for some time). This object is meant to contain default values for what a typical dry-run execution would do. For now it contains a single variable named `systems_directory_exists` but we can add more going forward. The values in the object can be overridden in unit tests (i.e.: `self.app.dr_cfg.systems_directory_exists = True`) and values can also be loaded from a YAML file passed to the subiquity server process via the `--dry-run-config` argument.

I am sure it feels over-engineered for now with only a single configuration item available. That said, it is the kind of architecture I wanted to have so we can cover many more use-cases in ubuntu-pro, ubuntu-drivers, etc. Here's a list of potential ideas I would like to use this mechanism for:

* drivers:
    * to tell Subiquity the list of third-party drivers it should suggest (or none)
    * to request Subiquity to run ubuntu-drivers on the host
    * to specify an alternative sysfs directory containing predefined modalias files
* ubuntu-pro:
    * to specify the ua-contracts test environment URL 
    * to specify automatic replies for the server
* to specify the Ubuntu release that is returned by `lsb_release` which we can be used to test behavior on LTS vs non LTS releases.
* to specify the current Subiquity version number and the version number that we would offer as a refresh (or not offer if it is lower)

### DIFF
--- a/examples/answers-tpm.yaml
+++ b/examples/answers-tpm.yaml
@@ -1,4 +1,5 @@
 #source-catalog: examples/tpm-sources.yaml
+#dr-config: examples/tpm-dr-config.yaml
 Source:
   source: src-prefer-encrypted
 Serial:
--- a/examples/tpm-dr-config.yaml
+++ b/examples/tpm-dr-config.yaml
@@ -1,0 +1,1 @@
+systems_dir_exists: true
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -159,6 +159,7 @@ for answers in examples/answers*.yaml; do
     if echo $answers|grep -vq system-setup; then
         config=$(sed -n 's/^#machine-config: \(.*\)/\1/p' $answers || true)
         catalog=$(sed -n 's/^#source-catalog: \(.*\)/\1/p' $answers || true)
+        dr_config=$(sed -n 's/^#dr-config: \(.*\)/\1/p' "$answers" || true)
         if [ -z "$config" ]; then
             config=examples/simple.json
         fi
@@ -169,6 +170,9 @@ for answers in examples/answers*.yaml; do
         opts=()
         if [ -n "$serial" ]; then
             opts+=(--serial)
+        fi
+        if [ -n "$dr_config" ]; then
+            opts+=(--dry-run-config "$dr_config")
         fi
         # The --foreground is important to avoid subiquity getting SIGTTOU-ed.
         LANG=C.UTF-8 timeout --foreground 60 \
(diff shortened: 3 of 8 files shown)


mwhudson (approved):
Hi, this looks fine. There is some prior art for this sort of thing in the SUBIQUITY_DEBUG handling, did you find that code? Structured data beats an environment variable in most ways of course, but I wonder if we might want to have a way to specify the value of one of the dry-run config way that is more suited to a single run, like --dry-run-config system_seed_dir_exists=1 vs --dry-run-config-file like you use it here?

ogayot:
> Hi, this looks fine. There is some prior art for this sort of thing in the SUBIQUITY_DEBUG handling, did you find that code? Structured data beats an environment variable in most ways of course, but I wonder if we might want to have a way to specify the value of one of the dry-run config way that is more suited to a single run, like --dry-run-config system_seed_dir_exists=1 vs --dry-run-config-file like you use it here?

Hi, yes I did encounter the SUBIQUITY_DEBUG variable a couple times, namely in the context of ubuntu-drivers. I wasn't aware of the number of uses that we have for this variable though. For some reason, I thought that I couldn't use this mechanism for this fix but I definitely could/should have.

I do think that having the flexibility to specify a variable is important. But I also think it's difficult to come up with something generic enough. For some variables, we would want to convert the value to a boolean (as in your example). For others, it would have to be kept as a string, converted to a number (or even stored as a list of said type?)

We're doing something similar in curtin with `--set key=val` and `--set json:key=['val']` (which supports assigning a boolean, a number,  `null`, a json array or a json object to the variable but it is still limiting with structured data in many scenarios).

There is also the IMO underrated `(< command)` syntax from the shell which allows doing things like: `--dry-run-config <(echo "systems_dir_exists: true")`.

I would like to think about this more and come with the right approach. 

Since we already have multiple crash reports for the bug though, I would like to avoid delaying the merge too much. I'm happy to quickly rewrite the fix using `SUBIQUITY_DEBUG=systems_dir_exists` only and keep the dry-run-config discussion for later. Or do your think the current situation is better @mwhudson ?

mwhudson:
Honestly I don't really mind! I agree this needs fixing and quickly though

ogayot:
Okay, I'll just merge it as is and we can iterate to make it better.

PR #1496, opened 2022-11-29
do not round up partition size in add_partition

mwhudson:
I'm like 99% sure that all call paths to here align the size properly, apart from the one in apply_system, which I want to be able to create improperly sized partitions.

### DIFF
--- a/subiquity/common/filesystem/tests/test_sizes.py
+++ b/subiquity/common/filesystem/tests/test_sizes.py
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import random
 import unittest
 from unittest import mock
 
@@ -27,6 +28,9 @@ from subiquity.common.filesystem.sizes import (
     uefi_scale,
     )
 from subiquity.common.types import GuidedResizeValues
+from subiquity.models.filesystem import (
+    MiB,
+    )
 
 
 class TestPartitionSizeScaling(unittest.TestCase):
@@ -108,6 +112,25 @@ class TestCalculateGuidedResize(unittest.TestCase):
                 install_max=190 << 30,
                 minimum=50 << 30, recommended=200 << 30, maximum=230 << 30)
         self.assertEqual(expected, actual)
+
+    def test_unaligned_size(self):
+
+        def assertAligned(val):
+            if val % MiB != 0:
+                self.fail(
+                    "val of %s not aligned (%s) with delta %s" % (
+                        val, val % MiB, delta))
+
+        for i in range(1000):
+            delta = random.randrange(MiB)
+            actual = calculate_guided_resize(
+                part_min=40 << 30,
+                part_size=(240 << 30) + delta,
+                install_min=10 << 30)
+            assertAligned(actual.install_max)
+            assertAligned(actual.minimum)
+            assertAligned(actual.recommended)
+            assertAligned(actual.maximum)
 
 
 class TestCalculateInstallMin(unittest.TestCase):
--- a/subiquity/common/filesystem/boot.py
+++ b/subiquity/common/filesystem/boot.py
@@ -171,7 +171,7 @@ def get_boot_device_plan_bios(device) -> Optional[MakeBootDevicePlan]:
             return attr_plan
 
     gap = gaps.Gap(device=device,
-                   offset=sizes.BIOS_GRUB_SIZE_BYTES,
+                   offset=device.alignment_data().min_start_offset,
                    size=sizes.BIOS_GRUB_SIZE_BYTES)
     create_part_plan = CreatePartPlan(
         gap=gap,
--- a/subiquity/common/filesystem/sizes.py
+++ b/subiquity/common/filesystem/sizes.py
@@ -101,6 +101,8 @@ def calculate_guided_resize(part_min: int, part_size: int, install_min: int,
     if part_min < 0:
         return None
 
+    part_size = align_up(part_size, part_align)
+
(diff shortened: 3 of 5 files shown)


mwhudson:
This should probably have more tests but unfortunately I'm not sure exactly which.

dbungert:
Subiquity should be usually generating aligned offsets and sizes, and if we find a case that doesn't do that then I would consider that to be a bug.

> apart from the one in apply_system
Interesting.  If we need it that's fine, but I suppose it would be better long term if those values could be made aligned.

My concern does seem to be around existing partitions.  Sometimes there is [weird stuff](https://bugs.launchpad.net/bugs/1929087).   Do we have test cases covering resizing a partition with unaligned start or end?  Or adding a partition after one with the unaligned start/end?

mwhudson:
We could do something like
```diff
diff --git a/subiquity/models/filesystem.py b/subiquity/models/filesystem.py
index cf1bb68c..252003ce 100644
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1586,7 +1586,14 @@ class FilesystemModel(object):
         self._actions.remove(obj)
 
     def add_partition(self, device, *, size, offset, flag="", wipe=None,
-                      grub_device=None, partition_name=None):
+                      grub_device=None, partition_name=None,
+                      check_alignment=True):
+        align = device.alignment_data().part_align
+        if check_alignment:
+            if offset % align != 0 or size % align != 0:
+                raise Exception(
+                    "size %s or offset %s not aligned to %s",
+                    size, offset, align)
         from subiquity.common.filesystem import boot
         if device._fs is not None:
             raise Exception("%s is already formatted" % (device,))
diff --git a/subiquity/server/controllers/filesystem.py b/subiquity/server/controllers/filesystem.py
index 72ab677f..c4697131 100644
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -505,7 +505,8 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
                    structure == self._on_volume.structure[-1]:
                     gap = gaps.largest_gap(disk)
                     size = gap.size - (offset - gap.offset)
-                part = self.model.add_partition(disk, offset=offset, size=size)
+                part = self.model.add_partition(
+                    disk, offset=offset, size=size, check_alignment=False)
 
             type_uuid = structure.gpt_part_type_uuid()
             if type_uuid:
```
and see if the integration tests pass (they do) but honestly I'm not really sure how much that would prove.

> Interesting. If we need it that's fine, but I suppose it would be better long term if those values could be made aligned.

Yes. FWIW the recovery partition we're skating around here was created by the OEM installer, lets make subiquity's version of that behave better wrt alignment :)

> My concern does seem to be around existing partitions. Sometimes there is [weird stuff](https://bugs.launchpad.net/bugs/1929087). Do we have test cases covering resizing a partition with unaligned start or end?

We could / should add some cases to TestCalculateGuidedResize with unaligned partition size I guess, although I convinced myself it always returns an aligned size in practice.

(There are some silly edge cases around things like resizing a partition with an unaligned offset will create slightly less space on the disk than we expect and so we might end up creating a rootfs partition that's up to a megabyte less than the calculated minimum size for the install. It's hard for me to see why we should care about this!)

> Or adding a partition after one with the unaligned start/end?

We do have tests that ensure gaps are created with aligned offset and size even if the pre-existing partitions are out of alignment, and we almost always create partitions into gaps so this should be ok ... Gap.split doesn't explicitly handle an unaligned size, maybe it should?

Also it's not clear to me what we should do if the desktop installer sends an unaligned size.

dbungert (approved):
> so we might end up creating a rootfs partition that's up to a megabyte less than the calculated minimum size for the install

I did consider that scenario during the resize work and I decided that I would let it go because the only 'minimum' sizes we offer are severely padded.

dbungert:
> Gap.split doesn't explicitly handle an unaligned size, maybe it should

I lean toward a 'no' on that with the reasoning that the caller may be working with one of these weird cases and need an unaligned region.  I can probably be convinced otherwise.

subiquity/common/filesystem/tests/test_sizes.py
```diff
@@ -109,6 +113,17 @@ def test_weighted_split(self):
                 minimum=50 << 30, recommended=200 << 30, maximum=230 << 30)
         self.assertEqual(expected, actual)
 
+    def test_unaligned_size(self):
+        for i in range(100):
+            actual = calculate_guided_resize(
+                part_min=40 << 30,
+                part_size=(240 << 30) + random.randrange(MiB),
```
dbungert, line 120: When a test has a random element, I like to see that value be logged.  If it ever fails that value will be critical.
```
diff --git a/subiquity/common/filesystem/tests/test_sizes.py b/subiquity/common/filesystem/tests/test_sizes.py
index b1a7d9b8..2d96bfb4 100644
--- a/subiquity/common/filesystem/tests/test_sizes.py
+++ b/subiquity/common/filesystem/tests/test_sizes.py
@@ -115,11 +115,12 @@ class TestCalculateGuidedResize(unittest.TestCase):
 
     def test_unaligned_size(self):
         for i in range(100):
+            rr = random.randrange(MiB)
             actual = calculate_guided_resize(
                 part_min=40 << 30,
-                part_size=(240 << 30) + random.randrange(MiB),
+                part_size=(240 << 30) + rr,
                 install_min=10 << 30)
-            self.assertEqual(actual.install_max % MiB, 0)
+            self.assertEqual(actual.install_max % MiB, 0, rr)
             self.assertEqual(actual.minimum % MiB, 0)
             self.assertEqual(actual.recommended % MiB, 0)
             self.assertEqual(actual.maximum % MiB, 0)
```

PR #1499, opened 2022-11-29
fix the value passed as on-volumes to snapd

mwhudson:
The data passed to on-volumes needs to indicate, for each "structure" (partition) defined by the gadget the path to the underlying device. The current code attempts to this by tracking the partition for each role but this doesn't work: there maybe be more than one partition with no role. So refactor to have the controller convert the "volume" structure to an "on-volume" structure early and update the device fields after curtin runs.


mwhudson:
I haven't VM tested this yet.

mwhudson:
This has been tested now.

dbungert (approved):

PR #1500, opened 2022-11-29
a few fixes for apply_system

mwhudson:


subiquity/server/controllers/filesystem.py
```diff
@@ -477,7 +477,7 @@ def apply_system(self, disk_id):
                 if (offset, size) in parts_by_offset_size:
                     preserved_parts.add(parts_by_offset_size[(offset, size)])
 
-            for part in disk.partitions():
+            for part in list(disk.partitions()):
```
dbungert, line 478: Would you help me understand this change?  I thought it was a list already.

dbungert (approved):

subiquity/server/controllers/filesystem.py
mwhudson: This is a classic case of mutating a list as you're iterating over it. An alternative fix would be for Disk.partitions() to return a copy of the partition list.

dbungert: Thanks!

dbungert (approved):

PR #1502, opened 2022-11-29
Support for autoinstalling a core boot classic system

mwhudson:
This adds support for a 'hybrid' layout, which can only be used with a core boot classic system (and no other style can be used).

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -72,6 +72,7 @@ from subiquity.common.types import (
     StorageEncryptionSupport,
     StorageResponse,
     StorageResponseV2,
+    StorageSafety,
     )
 from subiquity.models.filesystem import (
     ActionRenderMode,
@@ -155,15 +156,6 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
 
     def load_autoinstall_data(self, data):
         log.debug("load_autoinstall_data %s", data)
-        if data is None:
-            if not self.interactive():
-                data = {
-                    'layout': {
-                        'name': 'lvm',
-                        },
-                    }
-            else:
-                data = {}
         log.debug("self.ai_data = %s", data)
         self.ai_data = data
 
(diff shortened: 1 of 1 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -150,15 +151,6 @@ def is_core_boot_classic(self):
 
     def load_autoinstall_data(self, data):
         log.debug("load_autoinstall_data %s", data)
-        if data is None:
-            if not self.interactive():
```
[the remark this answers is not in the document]
mwhudson: No because apply_autoinstall_config is only called in the non-interactive case.

subiquity/server/controllers/filesystem.py
```diff
@@ -881,6 +922,9 @@ def convert_autoinstall_config(self, context=None):
                             "'layout' and 'config', using 'layout'")
             self.run_autoinstall_guided(self.ai_data['layout'])
         elif 'config' in self.ai_data:
+            if self._system is not None:
```
dbungert, line 925: the 'mode' key is also invalid, a similar check would be nice

subiquity/server/controllers/filesystem.py
```diff
@@ -847,6 +854,40 @@ async def _probe(self, *, context=None):
...
+            if self._system is None:
+                raise Exception(
+                    "can only use name: hybrid when installing core boot "
+                    "classic")
+            encrypted = layout.get('encrypted', None)
+            safety = self._system.storage_encryption.storage_safety
+            support = self._system.storage_encryption.support
+            if encrypted is None:
+                if safety == StorageSafety.ENCRYPTED:
+                    self.use_tpm = True
```
dbungert, line 868: Does anyone do a check for `StorageEncryptionSupport.AVAILABLE` in this case?  If we're demanding it but it's not available then maybe that's a fatal error for autoinstall

subiquity/server/controllers/filesystem.py
mwhudson: ok

subiquity/server/controllers/filesystem.py
mwhudson: This is confusing, but if storage safety is encrypted and encryption is not available then the encryption support ends up as DEFECTIVE which means that _core_boot_classic_error gets set which means that we get an exception raised before we get here. I'll add a comment about this.

dbungert (approved):

PR #1505, opened 2022-12-01
Drivers: make sure we suggest non-GPGPU drivers to the desktop installer

ogayot:
The list of drivers suggested to the user may vary based on whether we are installing a server or desktop image.

In the current implementation of the drivers controller, the value of the source variant (e.g., server or desktop) is read early in the initializer.

This is a problem because it happens before the client gets the opportunity to tell us what we are installing.

As the result, we always suggest drivers for the server variant because it is the default value in Subiquity.

This PR makes sure to query the list of drivers when the source variant is set (or set again).

Launchpad bug: https://bugs.launchpad.net/subiquity/+bug/1998376

### DIFF
--- a/examples/mixed-sources.yaml
+++ b/examples/mixed-sources.yaml
@@ -1,0 +1,23 @@
+- description:
+    en: This version has been customized to have a small runtime footprint in environments
+      where humans are not expected to log in.
+  id: ubuntu-server-minimal
+  locale_support: none
+  name:
+    en: Ubuntu Server (minimized)
+  path: ubuntu-server-minimal.squashfs
+  size: 530485248
+  type: fsimage
+  variant: server
+- default: true
+  description:
+    en: The default install contains a curated set of packages that provide a comfortable
+      experience for operating your desktop.
+  id: ubuntu-desktop
+  locale_support: locale-only
+  name:
+    en: Ubuntu Desktop
+  path: ubuntu-desktop-minimal.ubuntu-desktop.squashfs
+  size: 1066115072
+  type: fsimage-layered
+  variant: desktop
--- a/subiquity/server/controllers/drivers.py
+++ b/subiquity/server/controllers/drivers.py
@@ -52,8 +52,12 @@ class DriversController(SubiquityController):
 
     def __init__(self, app) -> None:
         super().__init__(app)
-        self.ubuntu_drivers: UbuntuDriversInterface = \
-            get_ubuntu_drivers_interface(app)
+        self.ubuntu_drivers: Optional[UbuntuDriversInterface] = None
+
+        self._list_drivers_task: Optional[asyncio.Task] = None
+        self.list_drivers_done_event = asyncio.Event()
+
+        self.drivers: List[str] = []
 
     def make_autoinstall(self):
         return {
@@ -69,7 +73,24 @@ class DriversController(SubiquityController):
         self.app.hub.subscribe(
             InstallerChannels.APT_CONFIGURED,
             self._wait_apt.set)
-        self._drivers_task = asyncio.create_task(self._list_drivers())
+        self.app.hub.subscribe(
+            (InstallerChannels.CONFIGURED, "source"),
+            self.restart_querying_drivers_list)
+
+    def restart_querying_drivers_list(self):
+        """ Start querying the list of available drivers. This method can be
+        invoked multiple times so we need to stop ongoing operations if the
+        variant changes. """
+
+        self.ubuntu_drivers = get_ubuntu_drivers_interface(self.app)
+
+        self.drivers = []
+        self.list_drivers_done_event.clear()
+        if self._list_drivers_task is not None:
+            self._list_drivers_task.cancel()
+
+        log.debug("source variant has been set. Querying list of drivers.")
+        self._list_drivers_task = asyncio.create_task(self._list_drivers())
 
     @with_context()
     async def _list_drivers(self, context):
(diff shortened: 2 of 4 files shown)


subiquity/server/types.py
```diff
@@ -22,3 +22,4 @@ class InstallerChannels(CoreChannels):
     GEOIP = 'geoip'
     CONFIGURED = 'configured'
     APT_CONFIGURED = 'apt-configured'
+    SOURCE_VARIANT_SET = 'client-variant-set'
```
dbungert, line 25: nitpick - one says source, one says client

dbungert (approved):

subiquity/server/types.py
ogayot: nice catch, thanks! I obviously couldn't decide which one to use :D

mwhudson:
So this bumps into something I got confused and upset about last week. Must be time to straighten this out once and for all!

The confusion is between **client variant** and **source variant**. The client variant is TUI or flutter basically and the source variant is server or desktop (and probably one day core) but the current codebase muddles these a lot.

Obviously the default is for the TUI client to install server and the flutter client to install desktop but I've just finished a bunch of work to allow the TUI client to install desktop and I would like to push on to clearing up this muddle properly.  The way the system setup code uses this stuff doesn't help either unfortunately.

In the case of this branch, I think it's reasonably clear that it the source variant that matters. If you're using the TUI to install a desktop, you should get offered the desktop drivers, so ideally you wouldn't need this SOURCE_VARIANT_SET event and you'd just subscribe to (CONFIGURED, 'source'). The wrinkle here is that POSTing to client_variant sets the source variant too -- basically I think we can drop this now, but I guess we can check that with the desktop installer people. I'm pretty sure they added a POST to /source as part of the last subiquity update though.

mwhudson:
Or for a more meta point: why does the server need to know the client variant at all? I don't think it does any more.

ogayot:
> So this bumps into something I got confused and upset about last week. Must be time to straighten this out once and for all!
> 
> The confusion is between **client variant** and **source variant**. The client variant is TUI or flutter basically and the source variant is server or desktop (and probably one day core) but the current codebase muddles these a lot.
> 
> Obviously the default is for the TUI client to install server and the flutter client to install desktop but I've just finished a bunch of work to allow the TUI client to install desktop and I would like to push on to clearing up this muddle properly. The way the system setup code uses this stuff doesn't help either unfortunately.
> 
> In the case of this branch, I think it's reasonably clear that it the source variant that matters. If you're using the TUI to install a desktop, you should get offered the desktop drivers, so ideally you wouldn't need this SOURCE_VARIANT_SET event and you'd just subscribe to (CONFIGURED, 'source'). 

Thanks for clearing things up for me. I really got confused and was not heading the right direction! 
I've done the adjustments needed and it seems to work as intended. I added a source catalog file that has desktop and server variants with tests that check the drivers based on the variant selected.

> The wrinkle here is that POSTing to client_variant sets the source variant too

Does it really? When testing earlier I noticed that app.base_model.source.current.variant did not change after doing a POST query to /meta/client_variant so I was even more confused.

> -- basically I think we can drop this now, but I guess we can check that with the desktop installer people. I'm pretty sure they added a POST to /source as part of the last subiquity update though.

fwiw I confirm that there is a POST to /source in the recent bugs reports of the daily canary :)

dbungert (approved):

examples/mixed-sources.yaml
```diff
@@ -0,0 +1,23 @@
+- description:
```
mwhudson, line 1: Good idea to have a catalog like this. Maybe update the names and descriptions a bit?

subiquity/server/controllers/drivers.py
```diff
@@ -69,7 +73,24 @@ def start(self):
         self.app.hub.subscribe(
             InstallerChannels.APT_CONFIGURED,
             self._wait_apt.set)
```
mwhudson, line 75: Looking at this line... are we sure that APT_CONFIGURED is going to fire only once in the future? It's certainly true today, so maybe this is a worry for another day.

subiquity/server/controllers/drivers.py
```diff
@@ -97,14 +118,13 @@ async def _list_drivers(self, context):
...
+        self.list_drivers_done_event.set()
         log.debug("Available drivers to install: %s", self.drivers)
-        if not self.drivers:
-            await self.configured()
 
     async def GET(self, wait: bool = False) -> DriversResponse:
         local_only = not self.app.base_model.network.has_network
         if wait:
-            await asyncio.shield(self._drivers_task)
+            await self.list_drivers_done_event.wait()
```
mwhudson, line 127: Any change that removes a use of shield is a good change :-)

mwhudson (approved):

mwhudson:
> > The wrinkle here is that POSTing to client_variant sets the source variant too
> 
> Does it really? When testing earlier I noticed that app.base_model.source.current.variant did not change after doing a POST query to /meta/client_variant so I was even more confused.

Ah no it's even more confusing than that. POSTing to client_variant calls app.base_model.set_source_variant, which sets which controllers the install controller will wait for but you're right it does *not* set the value of app.base_model.source.current.variant. It should probably stop (although i have a niggling feeling that this is the behaviour system_setup depends on).

subiquity/server/controllers/drivers.py
ogayot: I think the answer is yes as long as we fire this event after _successfully_ determining which mirror we will use.

examples/mixed-sources.yaml
ogayot: Oops, I forgot to push my change before merging this PR so I've opened https://github.com/canonical/subiquity/pull/1507

jpnurmi:
@ogayot Nice, thanks! We'll update right away: https://github.com/canonical/ubuntu-desktop-installer/pull/1254

PR #1506, opened 2022-12-05
integrity: add md5 check integration

dbungert:
* log the results of the md5 check
* if a failure and a crash occurs, mention that in the crash dialog

### DIFF
--- a/subiquity/server/controllers/integrity.py
+++ b/subiquity/server/controllers/integrity.py
@@ -1,0 +1,83 @@
+# Copyright 2022 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import json
+import logging
+
+from subiquitycore.async_helpers import schedule_task
+from subiquitycore.utils import astart_command
+from subiquity.common.apidef import API
+from subiquity.common.types import CasperMd5Results
+from subiquity.server.controller import SubiquityController
+
+
+log = logging.getLogger('subiquity.server.controllers.integrity')
+
+mock_pass = {'checksum_missmatch': [], 'result': 'pass'}
+mock_skip = {'checksum_missmatch': [], 'result': 'skip'}
+mock_fail = {'checksum_missmatch': ['./casper/initrd'], 'result': 'fail'}
+
+
+class IntegrityController(SubiquityController):
+
+    endpoint = API.integrity
+
+    model_name = 'integrity'
+    result_filepath = '/run/casper-md5check.json'
+
+    @property
+    def result(self):
+        return CasperMd5Results(
+                self.model.md5check_results.get('result', 'unknown'))
+
+    async def GET(self) -> CasperMd5Results:
+        return self.result
+
+    async def wait_casper_md5check(self):
+        if self.app.opts.dry_run:
+            return
+        proc = await astart_command([
+            'journalctl',
+            '--follow',
+            '--output', 'json',
+            '_PID=1',
+            'UNIT=casper-md5check.service',
+        ])
+        while True:
+            jsonbytes = await proc.stdout.readline()
+            data = json.loads(jsonbytes.decode('utf-8'))
+            if data.get('JOB_RESULT') == 'done':
+                break
+        proc.terminate()
+
+    async def get_md5check_results(self):
+        if self.app.opts.dry_run:
+            return mock_fail
+        with open(self.result_filepath) as fp:
+            try:
+                ret = json.load(fp)
+            except json.JSONDecodeError as jde:
+                log.debug(f'error reading casper-md5check results: {jde}')
+                return {}
+            else:
+                log.debug(f'casper-md5check results: {ret}')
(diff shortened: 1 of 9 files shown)


subiquity/server/controllers/integrity.py
```diff
@@ -0,0 +1,84 @@
...
+            '--output', 'json',
+            '_PID=1',
+            'UNIT=casper-md5check.service',
+        ])
+        while True:
+            jsonbytes = await proc.stdout.readline()
+            data = json.loads(jsonbytes.decode('utf-8'))
+            if data.get('JOB_RESULT') == 'done':
+                break
+        proc.terminate()
```
mwhudson, line 63: Other code in subiquity uses subiquity.journald to do this in process. Did you consider that?

subiquity/server/controllers/integrity.py
```diff
@@ -0,0 +1,84 @@
...
+
+    model_name = 'integrity'
+    result_filepath = '/run/casper-md5check.json'
+
+    @property
+    def result(self):
+        return CasperMd5Results(
+                self.model.md5check_results.get('result', 'unknown'))
+
+    async def GET(self) -> CasperMd5Results:
```
mwhudson, line 45: I guess at some point we may want a version of this that blocks until the check is done? But still a good start.

mwhudson (approved):
This looks fine, one comment, one question. I presume this will all work fine as an update to an existing ISO? I think so but I just want to be sure.

subiquity/server/controllers/integrity.py
dbungert: I'll look into it.  Next PR?

dbungert:
> This looks fine, one comment, one question. I presume this will all work fine as an update to an existing ISO? I think so but I just want to be sure.

Yes, I have done testing that way in the form of livefs-edit.  This always fails as the initrd no longer matches.

subiquity/server/controllers/integrity.py
dbungert: An optional block, perhaps.  If a slow verifying system experiences a crash, I wouldn't want to block the dialog waiting on the result.  Or is that desired?

subiquity/server/controllers/integrity.py
mwhudson: sure

subiquity/server/controllers/integrity.py
mwhudson: no i think how you have things currently is right for this use case.

mwhudson:
> Yes, I have done testing that way in the form of livefs-edit. This always fails as the initrd no longer matches

Hm did I not fix livefs-edit to update the checksums...

subiquity/server/controllers/integrity.py
```diff
@@ -0,0 +1,84 @@
...
+        proc.terminate()
+
+    async def get_md5check_results(self):
+        if self.app.opts.dry_run:
+            return mock_fail
+        with open(self.result_filepath) as fp:
+            try:
+                ret = json.load(fp)
+            except json.decoder.JSONDecodeError as jde:
+                log.debug(f'error reading casper-md5check results: {jde}')
```
ogayot, line 72: should this be a log.error or event log.exception?

subiquity/server/controllers/integrity.py
```diff
@@ -0,0 +1,84 @@
...
+                break
+        proc.terminate()
+
+    async def get_md5check_results(self):
+        if self.app.opts.dry_run:
+            return mock_fail
+        with open(self.result_filepath) as fp:
+            try:
+                ret = json.load(fp)
+            except json.decoder.JSONDecodeError as jde:
```
ogayot, line 71: minor: json.JSONDecodeError is the documented exception. I guess it is more future proof ?

subiquity/server/controllers/integrity.py
```diff
@@ -0,0 +1,84 @@
...
+                log.debug('casper-md5check results: '
+                          f'{self.model.md5check_results}')
+                return ret
+
+    async def md5check(self):
+        await self.wait_casper_md5check()
+        self.model.md5check_results = await self.get_md5check_results()
+
+    def start(self):
+        schedule_task(self.md5check())
```
ogayot, line 84: I have yet to see it occur in Subiquity but the asyncio doc says:

> Important
>
> Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:
https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

so I think we should store this in a self._md5check_task and just forget about it

ogayot (changes requested):

ogayot (approved):
just minor things from my previous review (I accidentally clicked "request changes")

subiquity/server/controllers/integrity.py
dbungert: Ah, interesting.  I actually did this in an earlier version of this changeset but removed it since I was never going to await it.  I will store it and think that's good practice for the future.

subiquity/server/controllers/integrity.py
dbungert: I would like to discontinue the practice of having things called "Traceback" in subiquity for non-fatal errors.  Users get confused if there is a crash later - they see a Traceback and assume that this is the cause of their problem.

casper-md5check writes the results file incrementally.  If that hung or exited prematurely it could leave a fragment of json, which would trigger this exception, but wouldn't necessarily be the cause of a possible later subiquity crash.

subiquity/server/controllers/integrity.py
dbungert:
```
>>> json.JSONDecodeError
<class 'json.decoder.JSONDecodeError'>
```
I'm going with json.JSONDecodeError since that's the documented one.

subiquity/server/controllers/integrity.py
dbungert: #1508

PR #1509, opened 2022-12-07
journald: stop passing a loop

dbungert:
Per python 3.10 changelog[1], passing a loop around is no longer considered a good idea.  Remove the loop argument and request the running loop.

[1]: https://docs.python.org/3/whatsnew/3.10.html#removed


mwhudson (approved):
Thanks for this. I'm sure there's still more of it we could do!

I am a bit annoyed that there isn't a asyncio.add_reader() function. But afaict there isn't.

dbungert:
Oh yes, there are still methods passing around loop arguments.  I wanted to focus on the journald corner of this while it was fresh.

PR #1517, opened 2022-12-09
Removes the switch to shell widget

CarlosNihelton:
From the summary controller when installation is in progress. The shell is inoperable on WSL due the fake snap environment.


jibel (approved):
LGTM. Approved.

PR #1519, opened 2022-12-13
refresh: remove the cancel update button since it is broken

ogayot:
Cancelling the installer self refresh does not work. Clicking on the button simply goes back to the screen asking the user if they want to update ; but the refresh operation still continues in the background.

Since we do not have a simple way to implement refresh cancellation at the moment, we are just removing the button.

### DIFF
--- a/subiquity/ui/views/refresh.py
+++ b/subiquity/ui/views/refresh.py
@@ -238,13 +238,11 @@ class RefreshView(BaseView):
         self.lb_tasks = ListBox([])
         self.task_to_bar = {}
 
-        buttons = [
-            other_btn(_("Cancel update"), on_press=self.check_state_available),
-            ]
-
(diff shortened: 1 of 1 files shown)


dbungert (commented):
This is an improvement, but I'd also like a comment explaining why we aren't offering a cancel.  There is still cancel related logic around and it would look weird that we weren't using it.

ogayot:
> This is an improvement, but I'd also like a comment explaining why we aren't offering a cancel.

Sure, I will add a comment.

> There is still cancel related logic around and it would look weird that we weren't using it.

@dbungert Are you referring to the logic allowing the user to cancel the _check for updates_ operation? If there is some logic in place to cancel an actual update operation, I might have overlooked it.

dbungert (approved):

PR #1520, opened 2022-12-14
distutils migration related items

dbungert:
* Start migration away from distutils
* Declare dependency on gettext
* Needed to get the lunar CI running


mwhudson (approved):
I guess we could do the build_i18n stuff in snapcraft.yaml if we need to or something. But seems like an improvement for now.

ogayot (approved):

PR #1521, opened 2022-12-14
snapcraft.yaml: Get all Python dependencies from the archive

mwhudson:
Once we do this, there is no reason to use the 'python' plugin, so switch to the 'nil' plugin with an override-build that calls pip for each of the subiquity, curtin, and probert parts.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -57,103 +57,80 @@ apps:
 
 parts:
   curtin:
+    plugin: nil
+
+    source: https://git.launchpad.net/curtin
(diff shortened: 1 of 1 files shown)


mwhudson:
Frankly this was surprisingly easy once I decided on calling pip from override-build which almost makes me suspicious ;-) Obviously this needs some careful ISO testing and probably shouldn't hit edge just before the xmas break.

dbungert (approved):
All sounds reasonable to me, +1 on not messing with Edge days before an extended break.

PR #1522, opened 2022-12-16
[WSL] WslConfigAdvanced only on reconfiguration variant

CarlosNihelton:
This pull request addresses the feedback we received that most users won't interact with the WSL configuration knobs during setup, but later if their workflows reveal the need for customisation.
With the set of changes herein proposed the advanced configuration screen is no longer shown in the setup variant of the TUI client and the integration tests are updated to reflect that.


didrocks (approved):
looks good, I didn’t spot anything that looks wrong to my eyes and you tested it. Let’s merge.

PR #1523, opened 2023-01-03
kvm-test: add option to add extra disks for install

ogayot:
I tried to reproduce https://bugs.launchpad.net/subiquity/+bug/2000066 in a VM where the original test case involves multiple disks. Sadly, I didn't manage to reproduce the crash but I think having the option in kvm-test to add extra disks would be good for future testing.

The patch introduces a new option `--extra-disk-size` to create additional disks in kvm-test. The option can be specified multiple times and will not affect the first disk (the one boots in presence of the `--boot` option).
```bash
# will create the first disk normally and will add an extra 10GiB disk
kvm-test.py --install --extra-disk-size 10G
```
```bash
# will create the first disk normally and will two extra disks (one of 10 and one of 20 GiB).
kvm-test.py --extra-disk-size 10G --extra-disk-size 20G
```

The size of the first disk (that can be honor the --boot option) can still be configured using the `--disksize` option).

### DIFF
--- a/scripts/kvm-test.py
+++ b/scripts/kvm-test.py
@@ -70,6 +70,7 @@ class Context:
         self.iso = f'/tmp/kvm-test/{self.release}-test.iso'
         self.hostname = f'{self.release}-test'
         self.target = f'/tmp/kvm-test/{self.hostname}.img'
+        self.extra_disk_fmt = f'/tmp/kvm-test/{self.hostname}-' + '{id}.img'
         self.password = salted_crypt('ubuntu')
         self.cloudconfig = f'''\
 #cloud-config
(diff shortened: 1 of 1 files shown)


dbungert (approved):
Excellent.  I was able to do the raid1 isotracker test with this config.

scripts/kvm-test.py
```diff
@@ -198,11 +199,18 @@ def load_config(self):
...
                     help='build iso with snap from channel')
-disk_group = parser.add_mutually_exclusive_group()
-disk_group.add_argument('-d', '--disksize', help='size of disk to create')
-disk_group.add_argument('--no-disk', action='store_true', help='attach no local storage')
-parser.add_argument('--disk-interface', help='type of interface for the disk',
+
+parser.add_argument('--disk-count', type=int, dest='disk_count',
+                    default=1, help='how many local disks should we have')
+parser.add_argument('--one-disk', action='store_const', const=1,
+                    dest='disk_count', help='attach no local storage')
```
Chris-Peterson444, line 206: suggestion: `s/attach no local storage/attach one local storage disk/` ?

Chris-Peterson444 (approved):
> ogayot requested a review from dbungert >2 years ago 

A long time coming! :eyes:

This is great, thanks!

scripts/kvm-test.py
ogayot: Thanks :D

PR #1524, opened 2023-01-03
utils: fix unused stdin parameter in astart_command

ogayot:
Specifying the value of `subprocess.PIPE` as the `stdin` argument of `astart_command` did not have any effect. This happened because the parameter was not forwarded to the subprocess function. The parameter was effectively unused.


mwhudson (approved):
oops

PR #1525, opened 2023-01-04
Add helper to create confirmation dialog and use it for Ubuntu Pro

ogayot:
While working on mirror testing, I realized that I needed  yet another confirmation dialog to display.
I decided to rework the implementation I did for Ubuntu Pro and moved it to subiquitycore to create these confirmation dialogs.

Using the new helpers, we can open a confirmation dialog with the following code and receive the decision from the user:

```python
confirmed = await view.ask_confirmation(title="Exit",
                                        question="Are you sure you want to exit?",
                                        cancel_label="On second thought...",
                                        confirmation_label="Yes, I am sure!")

if confirmed:
    print("The user confirmed his decision to exit")
else:
    print("The user reconsidered his decision")
```

If the confirmation overlay gets closed as the result of the user pressing ESC, the confirmation coroutine returns False.

### DIFF
--- a/subiquitycore/view.py
+++ b/subiquitycore/view.py
@@ -18,6 +18,7 @@
 Contains some default key navigations
 """
 
+import asyncio
 import logging
 
 from urwid import (
@@ -31,6 +32,8 @@ from subiquitycore.ui.container import (
     Pile,
     WidgetWrap,
     )
+
+from subiquitycore.ui.confirmation import ConfirmationOverlay
 from subiquitycore.ui.stretchy import StretchyOverlay
 from subiquitycore.ui.utils import disabled, undisabled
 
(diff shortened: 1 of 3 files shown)


subiquitycore/view.py
```diff
@@ -82,6 +86,42 @@ def show_stretchy_overlay(self, stretchy):
...
         self._w = StretchyOverlay(disabled(self._w), stretchy)
 
+    async def ask_confirmation(self, title: str, question: str,
+                               confirm_label: str, cancel_label: str) -> bool:
+        """ Open a confirmation dialog using a strechy overlay.
+        If the user selects the "yes" button, the function returns True.
+        If the user selects the "no" button or closes the dialog, the function
+        returns False.
+        """
+        confirm_event = asyncio.Event()
```
mwhudson, line 96: would it be slightly less clunky to use a queue for this?

mwhudson (approved):
thanks, seems like a good API to add!

subiquitycore/view.py
ogayot: Thanks for the suggestion. Yes that seems cleaner!

PR #1526, opened 2023-01-06
show the FDE "unavailable-reason" from snapd to the user

mwhudson:


subiquity/ui/views/filesystem/guided.py
```diff
@@ -155,12 +155,12 @@ class TPMChoice:
             enabled=False, default=False,
             help=_("The model being installed prefers but does not require "
                    "TPM backed full-disk encryption and it is not available "
-                   "on this device")),
+                   "on this device (the reason given was \"{reason}\")")),
```
ogayot, line 158: minor: Full stop for consistency with other similar messages?

subiquity/ui/views/filesystem/guided.py
```diff
@@ -155,12 +155,12 @@ class TPMChoice:
...
         StorageSafety.PREFER_UNENCRYPTED: TPMChoice(
             enabled=False, default=False,
             help=_("The model being installed does not prefer TPM backed "
                    "full-disk encryption and it is not available on this "
-                   "device.")),
+                   "device (the reason given was \"{reason}\").")),
```
ogayot, line 163: Minor:
As a non-native speaker, I find it confusing whether the reason given applies to:
* FDE not being available on this device
* The model _not_ preferring TPM backed FDE

but an example would probably make it clear, so it's just a potential obstacle for translators I guess :)

ogayot (approved):

subiquity/ui/views/filesystem/guided.py
mwhudson: The messages are not especially likely to be human friendly (which is why they were not shown originally) but here is the example from the sample data:

image

In this case at least, I think it's reasonably clear that this is a reason for FDE not being available. I can add a note for the translator.

subiquity/ui/views/filesystem/guided.py
mwhudson: :+1:

dbungert (approved):
Those are some ugly messages, but better to show them then to add another reason to send someone to the logs.  Thanks!

PR #1529, opened 2023-01-09
include .pyc files in the snap

mwhudson:


dbungert:
Decided to look at the effect of the snap size on this.  We were running at 25-28M, which dropped to 18M apparently?  Must be the removal of pip packages in favor of archive ones.

This would add back 6M.  

Thoughts?  I guess the advantage to having this would be Subiquity should start up faster?

mwhudson:
I guess the interesting question is whether this makes a measurable difference to the start up speed of subiquity. Not sure how to do that -- I guess we can compare the timestamp of "starting subiquity" message in the journal with some message from the subiquity logs?

PR #1530, opened 2023-01-09
subiquity.network: cloud-init networking when netplan root-readonly

blackboxsw:
subiquity.network: cloud-init networking when netplan root-readonly

When cloudinit.features.NETPLAN_CONFIG_ROOT_READ_ONLY is True,
cloud-init will write /etc/netplan/50-cloud-init.yaml as read-only
root.

This added security allows for subiquity to use cloud-init's
network renderer directly allowing both datasource and network
configuration passed in one place.

Read cloud-init features from
/run/cloud-init/combined-cloud-config.json when present.

Any netplan wifi configuration can be specified in a single
root-read-only network config file
/etc/cloud/cloud.cfg.d/90-installer-network.cfg instead of
having a separate config file for wifi, which could contain
credentials.

This simplifies golden image creation from images installed using
subiquity because image builders will not need to track down and 
purge separate /etc/netplan/00-installer-config.yaml and 
/etc/netplan/subiquity-disable-cloudinit-networking.cfg when preparing
a golden image.

Eventually, netplan config validation and cloudinit will support
separation of sensitive configuration by cloud-init without needing
to pre-categorize sensitive information.

This will allow cloud-init to grow to ability to write separate
world-readable configuration from config which is security sensitive
with no change needed in subiquity.


subiquity/models/network.py
```diff
@@ -15,6 +15,7 @@
 
 import logging
 
+from cloudinit import features
```
mwhudson, line 18: So this is cloudinit from the snap, where we really want the features from the cloudinit on the target system which is a bit harder to access. It also makes things a bit more annoying because now (pedantically speaking) the result of NetworkModel.render now depends on the configuration of the source model (now I'm off wondering about how this would work when we support downloading the install artefacts over the network but well that's definitely one for the future)

mwhudson (changes requested):
just recording this point already made out of band

subiquity/models/network.py
```diff
@@ -13,17 +13,17 @@
...
 
+import json
 import logging
 import subprocess
 
 from subiquitycore.models.network import NetworkModel as CoreNetworkModel
 from subiquitycore.utils import arun_command
 
-log = logging.getLogger('subiquity.models.network')
+log = logging.getLogger("subiquity.models.network")
```
dbungert, line 23: A fair amount of the diff ends up being just this sort of s/'/"/ - for a lot of the Subiquity code base we're standardized on single quote, and having this sort of change in here makes the logic changes harder to see.

Do you mind if we revert the quote changes and just focus on the rest?

subiquity/models/network.py
```diff
@@ -34,53 +34,86 @@ def render_config(self):
         else:
             return super().render_config()
 
+    def _host_combined_cloud_config(self) -> dict:
```
dbungert, line 37: I see what you mean about a utility function.  Yes, we'll want to share this with the server/server.py code.  Maybe lift this to a new file subiquity/cloudinit.py ?  Or I'm happy to extract this for my autoinstall consume-combined-cloud-config branch if we don't do it here.

subiquity/models/tests/test_subiquity.py
```diff
@@ -116,210 +118,250 @@ def assertConfigHasVal(self, config, path, val):
...
         config = model.render()
-        self.assertNotIn('sources', config)
+        self.assertNotIn("sources", config)
 
     def test_mirror(self):
         model = self.make_model()
-        mirror_val = 'http://my-mirror'
-        model.mirror.create_primary_candidate(mirror_val).elect()
+        mirror_val = "http://my-mirror"
+        model.mirror.set_mirror(mirror_val)
```
dbungert, line 215: Is this the part that tripped the tests?  A failed merge maybe?

subiquity/models/network.py
blackboxsw: My mistake running python black formatter on this before committing. Regenerating this PR now without those formatting changes.

subiquity/models/tests/test_subiquity.py
blackboxsw: yes failed merge. I've rebased and have pushed the changeset now

subiquity/models/network.py
blackboxsw: Moved this function to `subiquity.cloudinit` if needed in the future we can define a class for interacting w/ cloud-init on the host

dbungert (approved):

subiquity/models/network.py
```diff
@@ -41,31 +42,56 @@ def render(self):
...
+        use_cloudinit_net = cloud_cfg.get('features', {}).get(
+            'NETPLAN_CONFIG_ROOT_READ_ONLY', False
+        )
+
+        if use_cloudinit_net:
+            r = {
+                'write_files': {
+                    'etc_netplan_installer': {
+                        'path': (
+                            'etc/cloud/cloud.cfg.d' '/90-installer-network.cfg'
```
mwhudson, line 58: Why is there a space on this line?

subiquity/models/network.py
```diff
@@ -41,31 +42,56 @@ def render(self):
...
+                        ),
+                        'content': self.stringify_config(netplan),
```
mwhudson, line 60: Presumably this file should be written as 0o600?

mwhudson (approved):
A couple of nits, nothing major

subiquity/models/network.py
blackboxsw: Good thought and correct. added 'permissions': '0600'. Defeats the purpose if we leave this world-readable.

subiquity/models/network.py
blackboxsw: oops leftover from my accidental black formatting run due to a stray pre-commit hook setup I had. I manually reverted, but didn't squash the string all the way back to one from the multi-line that was setup due to earlier iteration of this change. Dropper/

PR #1531, opened 2023-01-10
storage: fix creation of encrypted VG in manual partitioning

ogayot:
While working on https://bugs.launchpad.net/subiquity/+bug/2000066, I found out that my patch renaming the passphrase fields from "password" to "passphrase" broke the ability to create an encrypted VG in manual partitioning.

It is only in main so no need to cherry-pick the fix anywhere AFAICT.

--

The following patch changed the name of the luks passphrase field from "password" to "passphrase" to make it consistent across the screens:
```
commit d63b44c014ebaa2831f683f7397cdf141e2641b4
Author: Olivier Gayot <olivier.gayot@canonical.com>
Date:   Fri Jun 24 10:56:25 2022 +0200

  storage: use fields named passphrase for passphrases
  
  Because the storage views lean on the implementation of
  setup_password_validation from the identity screen, we were forced to
  use a form with fields named "password" and "password_confirm".
  
  This makes the code confusing because we use the "passphrase"
  terminology in the storage forms.
  
  We now leave up to the caller to specify which fields he wants to be
  part of the validation ; instead of making him provide the full form.
```
However, this change broke the ability to create an encrypted VG in manual partitioning mode.

This happened because the values input in the VG edit form are passed as-is (mostly) to functions that are also used by the server.

On one hand, we have the client which deals with passphrases named "passphrase" and "confirm_passphrase".
On the other hand, we have the server which deals with passphrases named "password" and "confirm_password".

This inconsistency makes it hard to work with shared code.

~To work around the issue, we now rename the passphrase key (passphrase -> password) before passing it to the shared code.~

EDIT: the second patch replaces password by passphrase in the server too ; making the implementation consistent. Only the guided/guided_v2 API still use "password" instead of "passphrase" for compatibility reasons.

### DIFF
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -123,7 +123,7 @@ class FilesystemManipulator:
 
     def create_volgroup(self, spec):
         devices = set()
-        key = spec.get('password')
+        key = spec.get('passphrase')
         for device in spec['devices']:
             self.clear(device)
             if key:
@@ -277,7 +277,7 @@ class FilesystemManipulator:
 
     def volgroup_handler(self, existing, spec):
         if existing is not None:
-            key = spec.get('password')
+            key = spec.get('passphrase')
             for d in existing.devices:
                 if d.type == "dm_crypt":
                     self.model.remove_dm_crypt(d)
--- a/subiquity/common/filesystem/tests/test_manipulator.py
+++ b/subiquity/common/filesystem/tests/test_manipulator.py
@@ -82,7 +82,7 @@ class TestFilesystemManipulator(unittest.TestCase):
     def test_delete_encrypted_vg(self):
         manipulator, disk = make_manipulator_and_disk()
         spec = {
-            'password': 'passw0rd',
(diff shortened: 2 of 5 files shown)


dbungert:
> On the other hand, we have the server which deals with passphrases named "password" and "confirm_password".

What would it mean to fix the server to use "passphrase"?

ogayot:
> > On the other hand, we have the server which deals with passphrases named "password" and "confirm_password".
> 
> What would it mean to fix the server to use "passphrase"?

Apparently not as much as I anticipated. I expected the server and curtin to share the "password" terminology but they use "key" (or "keyfile") already. "password" seems only used in a couple places plus the API.

The POST storage/guided and storage/v2/guided endpoints expect the "password" field so we would have to convert from/to the API unless we decide to also break/add a compat layer to the API.

Definitely, it's not as much work as I expected.

ogayot:
Pushed a second patch which makes the implementation consistent between the client and the server ; only the guided & v2.guided endpoints still use "password" for backward compatibility with other clients.

dbungert (approved):
Thanks!

PR #1532, opened 2023-01-10
storage: fix crash when editing encrypted VG created in guided storage

ogayot:
When editing an encrypted VG that was created in the guided storage screen, the VG information is originating from the server. However, the server does not send the LUKS key over the wire. Instead it sends the path to a keyfile which contains the key. The client may or may not have read access to this keyfile so it does not have a reliable way to determine the key.

This causes problem when editing the VG because the GUI expects to receive a key when encryption is enabled.

If the VG object only contain a keyfile, the passphrase is set to None and this result in the GUI crashing.

This patch fixes the crash by passing an empty passphrase instead of a None value when the VG object only contains a keyfile.

This means the user gets forced to supply a passphrase again when editing an encrypted VG that was created in the guided partition screen.

The crash only occurs if the encrypted VG was created in the guided storage screen. In the manual partitioning scenario, the LVM_VolGroup instance is created from the information submitted in the guided storage form ; rather than obtained from the server through the HTTP API.


mwhudson:
Argh. I wonder what the right ui is for this. I think it's fairly obvious that if you edit the vg and just hit ok, the encryption data should be unchanged. Also if you put a new password in or disable encryption, that should be honoured. But I don't like the idea of passing the passphrase over the API either.

Maybe it would be best to render this password as "<unchanged>" or something initially and switch to the entry field as soon as a change is made?

ogayot:
> [...] if you edit the vg and just hit ok, the encryption data should be unchanged. Also if you put a new password in or disable encryption, that should be honoured. 

+1

> Maybe it would be best to render this password as "" or something initially and switch to the entry field as soon as a change is made?

Right something along those lines seems good. I was contemplating making the passphrase fields disabled and adding a "Clear passphrase" button below the form but I didn't want to throw myself at it without getting some feedback first.

edit-form

dbungert (approved):
Agree on not the best UX, but a crash is worse.  Thanks!

PR #1534, opened 2023-01-12
Fix wslsetupotions being None

CarlosNihelton:
Recent changes in GUI turned possible to have that field as None, leading to a crash.
This avoids the crash while preserving the default behavior of installing the language packs, unless the client sets otherwise.


EduardGomezEscandell (approved):
Looking good 👍

PR #1536, opened 2023-01-18
load grub config from ai_data

ITJamie:
launchpad bug: [1952926](https://bugs.launchpad.net/subiquity/+bug/1952926)

This change should make it so that user supplied grub config in the autoinstall data will be used even when the layout option is supplied.

This would allow users to supply the simpler storage configs (see launchpad bug) and disable the grub reordering (which currently moves network boot devices to the first boot item making the newly installed os not the default boot option)

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -1241,7 +1241,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
                 raise Exception(
                     "must not use config: when installing core boot classic")
             self.model.apply_autoinstall_config(self.ai_data['config'])
-            self.model.swap = self.ai_data.get('swap')
+        self.model.swap = self.ai_data.get('swap')
         self.model.grub = self.ai_data.get('grub')
 
     def start(self):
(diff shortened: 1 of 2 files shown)


dbungert:
Thanks for the contribution.  Looks correct.

I took the liberty of adding some tests - I was going to ask you to look into it, but tests on the filesystem controller can be kind of nasty and I wasn't sure how to explain it, so I wrote it.  Then I realized the same should be done for the swap declaration.  Then I realized I was kind of taking over your PR.  Sorry!

documentation/autoinstall-reference.md would benefit from an update for the above, if you're interested in helping further a text update would be nice, or I can take a look soon.

ITJamie:
Absolutely no problem. I had looked at the tests and instantly realised i was in over my head to figure out initial testing logic for this. Thank you!

m-shibata:
Could you please update me on the status of this Pull Request?

The code changes seem fine, is it just the documentation that needs to be updated? If you are expecting someone to make the documentation changes, should they issue a Pull Request to ITJamie:layout_grub_config?

ITJamie:
I dont think there is any doco to change. There are examples showing its uses already but they didnt work as the config block was indented

m-shibata:
Is it okay if I don't make any changes to the document at the URL below?

https://github.com/canonical/subiquity/blob/95e520ddf8317ac9c9f2193c0b1ef1a7ae60bde1/documentation/autoinstall-reference.md?plain=1#L360-L362

From what I understand, the document explains that even with a layout set, you can write "grub" and "swap" under the config block. With the proposed modification, I believe that "grub" and "swap" will be applied regardless of the presence of a layout.

dbungert (approved):

mwhudson (approved):

PR #1537, opened 2023-01-20
Test the applied apt configuration (mirror testing)

ogayot:
This is the first step to make subiquity more robust to APT errors. The supplied APT configuration (i.e., mirror location, proxy, network settings) is now tested before proceeding with the install.

In the text installer, the output of apt-get update is visible in the mirror screen ; which automatically starts testing the mirror selected. If the test is successful, the user can proceed further. The user is also allowed to proceed with a failed or ongoing mirror test, but they would have to confirm their choice (a modal dialog opens warning them that the install might/may fail based on the status of said test).

In autoinstalls, the test starts automatically as part of the `apply_autoinstall_config`. If the test fails, the mirror is tested again after 10 seconds. If it fails again, we abort the install.

I originally wanted the mirror test to consist of a full `apt-get update` command execution. Sadly, this seems to consume too much data and would take too long on a slow connection. Therefore, we attempt to limit what apt-get update downloads.

Notes about dry-run:
* we can define rules to control if the mirror check should:
     * succeed (the output of APT is faked)
     * fail (the output of APT is faked)
     * randomly succeed or fail (the output of APT is faked)
     * run on the host (this will effectively bypass the apt parameters supplied, except for the proxy)
 * The default rules are as follows:
     * https://archive.ubuntu.com/ubuntu will succeed
     * `https://<country-code>.archive.ubuntu.com/ubuntu` will succeed
     * any URL that ends in /success will succeed
     * any URL that ends in /fail or /failed will fail
     * any URL that ends in /rand or /random will randomly succeed or fail
     * any URL that ends in /host will run on the host (bypassing the URL configured)
     * anything else with run on the host
 * Different rules can be provided by supplying a `--dry-run-config` YAML file

Nice to have improvements:
* the test of the mirror could be started before reaching the mirror screen
* going back to the proxy or network settings and changing settings should interrupt any ongoing test and start a new one with the new settings

Known issues:
* in a real install `apt-get update` complains about the `_apt` user not having read permissions to `<overlay>/usr/lib/apt/lists/partial`. This is only a warning but can be confusing for users. We do change ownership of this directory but the problem is most likely caused by the overlay not being accessible by `_apt`.

This PR is vastly inspired from @mwhudson's POC. Kudos to him :)

NOTE: The PR also contains some refactoring of the mirror model that are not strictly necessary for mirror testing AFAICT. They are more targeted towards easing the implementation of the next step (i.e., automatic failover to a different mirror). If these changes make the PR too difficult to review, please let me know and I will try to separate them from the rest of the implementation.

Thanks!


subiquity/client/controllers/mirror.py
```diff
@@ -27,13 +29,29 @@ class MirrorController(SubiquityTuiController):
 
     async def make_ui(self):
         mirror = await self.endpoint.GET()
-        return MirrorView(self, mirror)
+        has_network = await self.app.client.network.has_network.GET()
+        if has_network:
+            check = await self.endpoint.check_mirror.progress.GET()
+        else:
+            check = None
```
mwhudson, line 36: Feels like the check for network could be done server side inside progress.GET() but OK.

subiquity/common/types.py
```diff
@@ -732,3 +732,16 @@ class CasperMd5Results(enum.Enum):
     FAIL = 'fail'
     PASS = 'pass'
     SKIP = 'skip'
+
+
+class MirrorCheckStatus(enum.Enum):
+    OK = enum.auto()
+    RUNNING = enum.auto()
+    FAILED = enum.auto()
```
mwhudson, line 740: fwiw I think something more explicit than enum.auto() is a good idea here (esp if we ever expect the desktop installer to call these apis!)

subiquity/server/apt.py
```diff
@@ -99,6 +130,67 @@ async def apply_apt_config(self, context):
...
+    async def run_apt_config_check(self, output: io.StringIO) -> None:
+        """ Run apt-get update (with various options limiting the amount of
+        data donwloaded) in the overlay where the apt configuration was
+        previously deployed. The output of apt-get (stdout + stderr) will be
+        written to the output parameter.
+        Raises a AptConfigCheckError exception if the apt-get command exited
+        with non-zero. """
+        assert self.configured_tree is not None
+
+        pfx = pathlib.Path(self.configured_tree.p())
```
mwhudson, line 142: So my spike on this topic (or at least one of them) did this checking in an overlay over the configured_tree because running apt with Acquire::IndexTargets::deb::*::DefaultEnabled=false actually deleted the indices for all the types that were disabled, so that the entire e.g. release Packages files got downloaded again during the install. It would be good to check that this isn't happening in your branch, or to switch to doing the check in an overlay (on an overlay!) again if it is.

mwhudson: Otherwise this method is very nice and easy to read :-)

subiquity/server/controllers/mirror.py
```diff
@@ -156,9 +199,51 @@ async def POST(self, data: str):
...
-        self.model.config['disable_components'] = data
+        self.model.disabled_components = set(data)
+
+    async def check_mirror_start_POST(
+            self, cancel_ongoing: bool = False) -> None:
+        if self.mirror_check is not None and not self.mirror_check.task.done():
+            if cancel_ongoing:
+                await self.check_mirror_abort_POST()
+            else:
+                assert False
```
mwhudson, line 219: er. This probably shouldn't still be here?

mwhudson (commented):
I haven't read every line (and I definitely want Dan to look at this too) but what's here looks pretty good!

subiquity/common/types.py
ogayot: oh yeah, definitely!

subiquity/client/controllers/mirror.py
ogayot: I have the impression that we have more and more screens that behave differently based on the check for network and so far the implementation has been pretty inconsistent:

* GET /drivers returns a "local_only" boolean that tells if we should look online or only in the media
* GET /ubuntu_pro returns a "has_network" boolean
* GET /snaplist returns SnapCheckState.FAILED if no network is available

so I tried to define a consistent way for the client to get that information.

subiquity/server/controllers/mirror.py
ogayot: Right, fixed!

subiquitycore/async_helpers.py
```diff
@@ -40,6 +40,20 @@ def schedule_task(coro, propagate_errors=True):
...
 
 
+# Collection of tasks that we want to fire and forget.
+# Keeping a reference to all background tasks ensures that the tasks don't get
+# garbage collected before they are done.
+# https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
+background_tasks = set()
+
+
+def run_bg_task(coro, *args, **kwargs) -> None:
```
dbungert, line 50: much nicer than a class member that we'll never use, great idea

subiquity/server/apt.py
```diff
@@ -99,6 +130,67 @@ async def apply_apt_config(self, context):
...
+        }
+
+        # Need to ensure the "partial" directory exists.
+        partial_dir = apt_dirs["State::Lists"] / "partial"
+        partial_dir.mkdir(
+                parents=True, exist_ok=True)
+        try:
+            shutil.chown(partial_dir, user="_apt")
+        except (PermissionError, LookupError) as exc:
+            log.warning("could to set owner of file %s: %r", partial_dir, exc)
```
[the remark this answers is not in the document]
ogayot: This call can run in dry-run but it will fail (unless run as root).
.. which is okay since apt will only try to `setuid("_apt")` if it is running as root.

OTOH, when doing a real install:
 * the chown call succeeds (or should succeed)
 * apt still fails to use the sandbox _apt user because the overlay is not accessible. so it warns and runs unsandboxed. I would like to get this fixed.

subiquity/server/apt.py
ogayot: > So my spike on this topic (or at least one of them) did this checking in an overlay over the configured_tree because running apt with Acquire::IndexTargets::deb::*::DefaultEnabled=false actually deleted the indices for all the types that were disabled, so that the entire e.g. release Packages files got downloaded again during the install. It would be good to check that this isn't happening in your branch, or to switch to doing the check in an overlay (on an overlay!) again if it is.

I think I see what you mean. Currently, I'm calling apply_apt_config each time I want to run a mirror check, therefore a fresh configured_tree gets deployed every time. We might see the issue you describe going forward if we decide to avoid creating new overlays all the time.

dbungert (approved):
Nice work!

subiquity/server/apt.py
mwhudson: oh i see, that makes sense too

subiquity/server/controllers/mirror.py
```diff
@@ -88,32 +105,64 @@ def __init__(self, app):
...
+            # If the test fails a second time, consider it fatal.
+            await try_mirror_checking_once()
 
     def on_geoip(self):
         if self.geoip_enabled:
             self.model.set_country(self.app.geoip.countrycode)
         self.cc_event.set()
 
     def on_source(self):
+        # FIXME disabled until we can sort out umount
```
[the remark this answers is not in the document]
mwhudson: no. i slightly suspect kernel bugs actually but hopefully it's not that

mwhudson (approved):

PR #1540, opened 2023-01-24
autoinstall: adjust x-make-view-request behavior

dbungert:
With autoinstall, a controller that was not interactive would decline
outright to provide the response to GET queries, with the notion that the
controller needed to be skipped.  Limit this behavior to only when
`x-make-view-request: yes` is set, which lets the client be able to get
the real response, or the skip info, depending on need.

### DIFF
--- a/subiquity/tests/api/test_api.py
+++ b/subiquity/tests/api/test_api.py
@@ -89,10 +89,28 @@ class Client:
         return await self.request('POST', query, data, **kwargs)
 
     async def request(self, method, query, data=None, **kwargs):
+        """send a GET or POST to the test instance
+        args:
+            method: 'GET' or 'POST'
+            query: endpoint, such as '/locale'
+
+        keyword arguments:
+            data: body of request
+            headers: dict of custom headers to include in request
+            all other keyword arguments are turned into query arguments
+
+            get('/meta/status', cur='WAITING') is equivalent to
+            get('/meta/status?cur="WAITING"')
+
+        returns:
+            python data of response content
+        """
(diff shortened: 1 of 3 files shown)


subiquity/tests/api/test_api.py
```diff
@@ -1561,6 +1584,27 @@ async def test_upcs_flow(self):
...
+            view_request_unspecified = await inst.get('/locale')
+            self.assertEqual('en_US.UTF-8', view_request_unspecified)
+
+            view_request_no = await inst.get(
+                    '/locale', headers={'x-make-view-request': 'no'})
+            self.assertEqual('en_US.UTF-8', view_request_no)
+
+            view_request_yes = await inst.get(
+                    '/locale', headers={'x-make-view-request': 'yes'})
+            self.assertIsNone(view_request_yes)
```
mwhudson, line 1605: do you have any way to test the value of the x-skip header in the response?

mwhudson: if the answer is "no" that's fine, having any tests >> having no tests!

mwhudson (approved):
Thanks for this.

It's almost as if the client-server interaction was designed by a single developer without much review and in a hurry!

subiquity/tests/api/test_api.py
dbungert: I debated doing so.  I'll spend a little time on it and explore.

dbungert: Implemented.

mwhudson (approved):
thanks!

PR #1541, opened 2023-01-26
More fde strings

dbungert:
@mwhudson is a reason available in the DISABLED case?

### DIFF
--- a/subiquity/ui/views/filesystem/guided.py
+++ b/subiquity/ui/views/filesystem/guided.py
@@ -129,14 +129,16 @@ class TPMChoice:
     help: str
 
 
-help_texts = {
-    StorageEncryptionSupport.DISABLED:
+tpm_help_texts = {
+    "DISABLED":
         _("TPM backed full-disk encryption has been disabled."),
-    StorageEncryptionSupport.AVAILABLE:
+    "AVAILABLE_CAN_BE_DESELECTED":
         _("The entire disk will be encrypted and protected by the "
           "TPM. If this option is deselected, the disk will be "
           "unencrypted and without any protection."),
-    StorageEncryptionSupport.UNAVAILABLE:
+    "AVAILABLE_CANNOT_BE_DESELECTED":
+        _("The entire disk will be encrypted and protected by the TPM."),
+    "UNAVAILABLE":
         # for translators: 'reason' is the reason FDE is unavailable.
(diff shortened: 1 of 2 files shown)


mwhudson:
> @mwhudson is a reason available in the DISABLED case?

(some time passes, sorry)

I think not, and it seems from looking at snapd that the DISABLED case is a bit of an odd one that we are unlikely to run into in a "real" environment:

```
	if dangerous && osutil.FileExists(filepath.Join(boot.InitramfsUbuntuSeedDir, ".force-unencrypted")) {
```

Looks like it's basically there to support a hackish unencrypted install of a dangerous model, presumably for development purposes only?

mwhudson:
what is the status of this branch btw?

dbungert:
I think it's ready, albeit not tested terribly well as I'm not all that up to speed on this work.

dbungert:
Mostly I think it was the hope that there was more info for the disabled case, and if there isn't we should move on.

subiquity/ui/views/filesystem/guided.py
```diff
@@ -129,41 +129,43 @@ class TPMChoice:
...
-            help=_("The entire disk will be encrypted and protected by the "
-                   "TPM. If this option is deselected, the disk will be "
-                   "unencrypted and without any protection.")),
+            help=help_texts[StorageEncryptionSupport.AVAILABLE]),
         StorageSafety.PREFER_UNENCRYPTED: TPMChoice(
             enabled=True, default=False,
-            help=_("The model being installed does not prefer but allows TPM "
-                   "backed full-disk encryption")),
-            },
+            help=help_texts[StorageEncryptionSupport.AVAILABLE]),
```
mwhudson, line 161: I'm not sure using the same text in all AVAILABLE cases makes sense, in particular for the ENCRYPTED/AVAILABLE case the text will say "If this option is deselected" but that's not going to be possible. Maybe this is OK, but it feels a bit strange to me.

mwhudson (approved):
This is an improvement but I do have one quibble (which I am happy for you to ignore)

subiquity/ui/views/filesystem/guided.py
dbungert: That sounds like a common case and talking about an impossible deselection would look awkward.  Happily fixed.

PR #1542, opened 2023-01-27
snapcraft: add pkg_resources to the snap

ogayot:
From subiquity, we sometimes execute entry point python scripts such as:

 * `usr/bin/ubuntu-advantage`
 * `usr/bin/ssh-import-id`

Those scripts require pkg_resources on core20 to run properly:

```python
from pkg_resources import load_entry_point
```

On core22, the scripts will be a bit more flexible but for now, we need python3-pkg-resources in the snap.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -109,6 +109,7 @@ parts:
       - python3-jsonschema
       - python3-minimal
       - python3-oauthlib
+      - python3-pkg-resources
       - python3-pyrsistent
       - python3-pyudev
       - python3-requests


dbungert (approved):
I'm sure this is fine, but what sort of problems did you run in to?

ogayot:
> I'm sure this is fine, but what sort of problems did you run in to?

When you get to the Ubuntu Pro screen, subiquity crashes and the logs indicate something like:

```
/snap/.../usr/bin/ua: ImportError: module pkg_resources not found.
```

and about the same thing when importing a SSH key, except that the error is shown in a UI dialog instead of in the logs.

I believe the regression was introduced by ceff8a6095ffb58e67b4181eb0dc4675ed294a0a

mwhudson:
Oops, sorry for missing that!

PR #1544, opened 2023-01-31
Fixes and more tests for mirror testing

ogayot:
I noticed a few things:
* if a mirror check fails and the error gets logged in response to a GET /mirror/check_mirror/progress call, the type of the error is not printed. This makes exceptions that do not have a description appear as blank; which does not make the diagnostic easy. Print the type of exception as well using `%r` instead of `%s`.
* Some types hints are wrong (or not accurate) in the UI / view.
* the `primary` and `disable_components` were not present in the generated autoinstall configuration. This was a regression introduced by [c13edb07f, 8c29f3475]

I also added more tests to cover `MirrorController.run_mirror_testing`. The code only runs in autoinstall and testing in manually in dry-run is clunky.

### DIFF
--- a/subiquity/models/mirror.py
+++ b/subiquity/models/mirror.py
@@ -106,3 +106,6 @@ class MirrorModel(object):
 
     def render(self):
         return {}
+
+    def make_autoinstall(self):
+        return self.get_apt_config()
--- a/subiquity/models/tests/test_mirror.py
+++ b/subiquity/models/tests/test_mirror.py
@@ -99,3 +99,11 @@ class TestMirrorModel(unittest.TestCase):
         actual.sort()
         expected.sort()
         self.assertEqual(expected, actual)
+
+    def test_make_autoinstall(self):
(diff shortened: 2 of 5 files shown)


mwhudson:
This looks fine to me but obviously I didn't spot any of this when reviewing the previous version so that that with what you will!

> Some types hints are wrong (or not accurate) in the UI / view.

Do you have any thoughts about running mypy or some other type checker over the codebase in CI?

ogayot:
> Do you have any thoughts about running mypy or some other type checker over the codebase in CI?

@mwhudson 
I would love to but I think it's unrealistic to expect mypy to be fully happy with our codebase.

There's current a lot of false positives and I can't find a way to disable them selectively.

That said, we could start running it as part of the CI though, as long as we don't block PR's based on its output. I'm just scared we will just ignore its output if there's too much nonsense.

If you know any alternative tool that could to the job (better?), I'm all ears.

dbungert (approved):

PR #1545, opened 2023-01-31
tui: turn off systemd ShowStatus

dbungert:
During TUI launch, turn systemd ShowStatus off. We don't want to do that earlier, if we did then the easier path would be quiet boot or something similar. This is done here so that service completion messages don't overwrite the console.

### DIFF
--- a/bin/subiquity-service
+++ b/bin/subiquity-service
@@ -5,7 +5,16 @@ export PYTHONPATH=$SNAP/lib/python3.8/site-packages
 if [ -n "$1" ]; then
    port=$1
 fi
+
+# Stops some dmesg from overwriting the TUI.
 /bin/dmesg -n 1
+
+# Stop the systemd service completion messages from doing the same.
+# The systemd(1) manpage documents SIGRTMIN+21 as a method to set
+# show_status=0.  Without this, ongoing services such as the casper-md5check
+# can result in service status message being written on top of the TUI.
+/bin/kill "-SIGRTMIN+21" 1
+
 if [ "$port" = "tty1" ]; then
 	$SNAP/bin/subiquity-loadkeys
 	setfont $SNAP/subiquity.psf


dbungert:
Other methods that were considered:
* more dmesg arguments - `--clear`, `--console-off` - these had no useful effect.  The existing `dmesg -n 1` is helpful, without it even more junk can appear over the TUI
* changes to casper-md5check.service LogLevelMax - this worked, but wasn't great because it silenced all md5check logs.  We could tune that further and change loglevels in casper-md5check but the possibility of another service causing the same problem remains.
* setting ShowStatus=no earlier - does work but then that is more-or-less `quiet` boot.

Side effects:
On shutdown the console is more quiet, showing only the "remove media" message.  I like the result.  If we don't like that change, then we can add logic to shutdown to clear this file and reset ShowStatus to enabled.

bin/subiquity-service
```diff
@@ -5,7 +5,18 @@ export PYTHONPATH=$SNAP/lib/python3.8/site-packages
...
+# stops some dmesg from overwriting the TUI
 /bin/dmesg -n 1
+
+# stop the systemd service completion messages from doing the same
+mkdir -p /etc/systemd/system.conf.d/
+cat > /etc/systemd/system.conf.d/subiquity.conf <<EOF
+[Manager]
+ShowStatus=no
+EOF
+systemctl daemon-reload
```
mwhudson, line 18: it appears (from man systemd) you can also do this via sending SIGRTMIN+21 to pid 1, which sounds less disruptive than daemon-reload. Could you give that a try?

mwhudson (changes requested):
Ah of course dmesg -c only affects kernel messages, not systemd. I have a different suggestion I'd like you to try out first though.

bin/subiquity-service
dbungert: Works, implemented.

dbungert: actually, I tested incorrectly.  Will follow-up with a conclusion.

dbungert: OK, per manpage, your solution looks like it should be better, but it does not appear to actually work.  The original solution does.

bin/subiquity-service
```diff
@@ -5,7 +5,13 @@ export PYTHONPATH=$SNAP/lib/python3.8/site-packages
 if [ -n "$1" ]; then
    port=$1
 fi
+
+# stops some dmesg from overwriting the TUI
 /bin/dmesg -n 1
+
+# stop the systemd service completion messages from doing the same
+kill "-SIGRTMIN+21" 1
```
mwhudson, line 13: hnnngh i don't think dash's kill builtin supports this syntax (hilariously, dash's kill builtin wasn't documented at all until a few weeks ago). I'm also not sure this syntax is right for dash even if it was supported. Anyway, try changing it to `/bin/kill -SIGRTMIN+21 1` ?

ogayot: yep.. dash's kill does not support the `kill -SIGINT` syntax. Only `kill -INT` or `kill -s INT`

I tried it on my system and it seems like it can't go beyond RTMIN+15.
```
$ dash -c 'kill -s RTMIN+14 99999'
dash: 1: kill: No such process
$ dash -c 'kill -s RTMIN+15 99999'
dash: 1: kill: No such process
$ dash -c 'kill -s RTMIN+16 99999'
dash: 1: kill: invalid signal number or name: RTMIN+16
```
```
$ dash -c 'kill -l' | grep RTMIN
RTMIN
RTMIN+1
[...]
RTMIN+14
RTMIN+15
```
> Anyway, try changing it to /bin/kill -SIGRTMIN+21 1 ?

+1

bin/subiquity-service
dbungert: Marking resolved as other kill invocations are producing the desired result, and we're deciding on the preferred version of that invocation.

bin/subiquity-service
dbungert: I'm proposing `/bin/kill "-55" 1`, as in dash in this environment, neither `kill -l` nor `/bin/kill -l` are giving me high confidence.  But they both seem to work with the number just fine.

ogayot: fwiw man 7 signal is pretty strict regarding this rule but I don't know if we're really at risk of doing something wrong... so I'd say pick what you think it best.

>  programs should  never refer to real-time signals using hard-coded numbers, but instead should always refer to real-time signals using the notation SIGRTMIN+n

[1] https://www.man7.org/linux/man-pages/man7/signal.7.html#DESCRIPTION § Real-time signals

dbungert: gross.  I'm going to read the source for /bin/kill and do whatever that suggests.

mwhudson: Sorry, why don't you like "/bin/kill -SIGRTMIN+21 1" ? I agree it doesn't appear in the -l output but I don't think it's going to stop working. I guess you could do something like kill $(($(/bin/kill -l SIGRTMIN)+21)...

A NACK on hardcoding 55 here in any case.

mwhudson: eh that doesn't work either. maybe we should just use bash for this helper :-)

dbungert: Yes, because it doesn't match what /bin/kill shows as expected.  If the goal here is to maximize forward compatibility then my hope is that by conforming to what /bin/kill wants, we can do that.

dbungert: https://git.launchpad.net/ubuntu/+source/procps/tree/proc/sig.c#n139 suggests that support for a construction like "-SIGRTMIN+21" is reasonable to assume.

mwhudson (approved):
Assuming this works it gets my vote. Might be worth a comment on why you are using /bin/kill?

dbungert:
This is the final version unless there is additional feedback.

ogayot (approved):
Thanks!

mwhudson:
That was more involved than I expected, glad we got there :-)

PR #1547, opened 2023-02-02
Make sure mirror test output is translated if translation is available

ogayot:
When a mirror test failure occurs, the user is expected to read the apt-get output to determine what the root cause could be. Users who don't use English as their installer's language should get the output of apt-get translated (provided there is a translation available).

I made 3 changes to make it possible:
1. when we execute a command through `arun_command` or another helper, the environment is passed to `subiquitycore.utils._clean_env()` ; which unconditionally hard-coded the value of LC_ALL to C. This behavior is still preserved by default, but the _clean_env function now takes an optional locale boolean, which can be set to False to disable the override of LC_* variables. The helpers have been updated to take an optional `clean_locale` boolean (defaulting to True).
2. in an autoinstall scenario we set the LANG env variable to the user's selected language. For some reason, we don't do that for normal interactive installs. I believe we should be more consistent but I didn't want to proceed with such a change without prior investigation. Therefore, I now explicitly set the LANG variable for the mirror check process (i.e., apt-get).
3. for translations to work, we need to have the relevant locale definition compiled. We now call the `localedef` function in the background when setting the locale. A warning is reported if this operation fails, but it does not prevent the install from pursuing.

### DIFF
--- a/subiquity/models/locale.py
+++ b/subiquity/models/locale.py
@@ -13,9 +13,11 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import locale
 import logging
+import subprocess
 
-from subiquitycore.utils import split_cmd_output
+from subiquitycore.utils import arun_command, split_cmd_output
 
 log = logging.getLogger('subiquity.models.locale')
 
@@ -34,6 +36,23 @@ class LocaleModel:
 
     def switch_language(self, code):
         self.selected_language = code
+
+    async def gen_localedef(self) -> None:
+        language, charmap = locale.normalize(self.selected_language).split(".")
+        cmd = [
+            "localedef",
+            "-f", charmap,
+            "-i", language,
+            "--",
+            f"{language}.{charmap}",
+        ]
+        await arun_command(cmd, check=True)
+
+    async def try_gen_localedef(self) -> None:
+        try:
+            await self.gen_localedef()
+        except subprocess.CalledProcessError as exc:
+            log.warning("Could not generate locale: %r", exc)
 
     def __repr__(self):
         return "<Selected: {}>".format(self.selected_language)
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -173,7 +173,10 @@ class AptConfigurer:
             apt_cmd.append(
                     f"-o{target}::DefaultEnabled=false")
 
-        proc = await astart_command(apt_cmd, stderr=subprocess.STDOUT)
+        env = os.environ.copy()
+        env["LANG"] = self.app.base_model.locale.selected_language
+        proc = await astart_command(apt_cmd, stderr=subprocess.STDOUT,
+                                    clean_locale=False, env=env)
 
         async def _reader():
             while not proc.stdout.at_eof():
--- a/subiquity/server/controllers/locale.py
+++ b/subiquity/server/controllers/locale.py
@@ -17,6 +17,7 @@ import asyncio
 import logging
 import os
 
+from subiquitycore import async_helpers as async_helpers
 from subiquity.common.apidef import API
 from subiquity.server.controller import SubiquityController
 from subiquity.server.types import InstallerChannels
@@ -65,5 +66,7 @@ class LocaleController(SubiquityController):
         return self.model.selected_language
 
     async def POST(self, data: str):
+        log.debug(data)
         self.model.switch_language(data)
+        async_helpers.run_bg_task(self.model.try_gen_localedef())
         await self.configured()
--- a/subiquity/models/tests/test_locale.py
+++ b/subiquity/models/tests/test_locale.py
@@ -1,0 +1,58 @@
+# Copyright 2023 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import subprocess
+import unittest
+from unittest import mock
+
+from subiquity.models.locale import LocaleModel
+
+
+class TestLocaleModel(unittest.IsolatedAsyncioTestCase):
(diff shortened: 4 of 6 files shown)


subiquity/models/locale.py
```diff
@@ -35,6 +36,23 @@ def __init__(self, chroot_prefix):
...
         self.selected_language = code
 
+    async def gen_localedef(self) -> None:
+        language, charmap = self.selected_language.split(".")
+        cmd = [
+            "localedef",
+            "-f", charmap,
+            "-i", language,
+            "--",
+            f"{language}.{charmap}",
```
dbungert, line 46: self.selected_language?

subiquity/models/locale.py
```diff
@@ -35,6 +36,23 @@ def __init__(self, chroot_prefix):
...
+        ]
+        await arun_command(cmd, check=True)
+
+    async def try_gen_localedef(self) -> None:
+        try:
+            await self.gen_localedef()
+        except subprocess.CalledProcessError as exc:
+            log.warning("Could not generate locale: %r", exc)
```
dbungert, line 54: What state does this leave the system?  Sounds like you ran it but it doesn't always work, does it leave the localedef as half-present or something?

subiquity/models/locale.py
ogayot: Good catch, that should world too :/

subiquity/models/locale.py
ogayot: In my tests, I've only seen the call to localedef fail when running subiquity as non-root (e.g., dry-run).
As far as I know, `localedef` does not alter any locale from the archive except the one specified. That said, the man page says:

```
       One of the following exit values can be returned by localedef:
       0      Command completed successfully.
       1      Warnings or errors occurred, output files were written.
       4      Errors encountered, no output created.
```
and here's what the glib says about exit code 1:
```
     * Some warnings: Write the output file and exit with status 1 to 
     indicate there may be problems using the output file e.g. missing 
     data that makes it difficult to use
```
I feel like the worst thing that can happen would be an incomplete translation but I don't have a read test-case for it.

subiquity/server/controllers/locale.py
```diff
@@ -66,4 +67,9 @@ async def GET(self) -> str:
 
     async def POST(self, data: str):
         self.model.switch_language(data)
+        # In an answers based install, the client will POST "" as the
+        # locale/data if there is no locale section. We should probably use
+        # null instead but ...
+        if not self.model.selected_language:
```
mwhudson, line 73: I don't understand this check. But I got up at 04:30 so it might be my issue.

subiquity/server/controllers/locale.py
```diff
@@ -66,4 +67,9 @@ async def GET(self) -> str:
 
     async def POST(self, data: str):
         self.model.switch_language(data)
+        # In an answers based install, the client will POST "" as the
+        # locale/data if there is no locale section. We should probably use
+        # null instead but ...
```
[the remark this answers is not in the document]
ogayot: Right, that would be the ideal scenario I guess

subiquity/server/apt.py
```diff
@@ -173,7 +173,10 @@ async def run_apt_config_check(self, output: io.StringIO) -> None:
             apt_cmd.append(
                     f"-o{target}::DefaultEnabled=false")
 
-        proc = await astart_command(apt_cmd, stderr=subprocess.STDOUT)
+        env = os.environ.copy()
+        env["LANG"] = self.app.base_model.locale.selected_language
```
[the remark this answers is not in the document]
ogayot: > I thought we tried fairly hard to have LC_ALL=C.UTF-8 by default everywhere in Ubuntu

Interesting.  I didn't know about this effort.

When testing in a jammy ISO, Subiquity is started with LANG=C.UTF-8 but LC_ALL unset, so I assumed LANG was the right candidate.
```
$ strings /proc/$(pgrep subiquity)/environ | grep -e '^LANG' -e '^LC_'
LANG=C.UTF-8
```
But I guess setting LC_ALL would work too, so I don't mind.

subiquity/server/controllers/locale.py
ogayot: answers can /currently/ pass an empty string, so the check is here to avoid calling try_gen_localedef when `self.model.selected_language == ""`.

Otherwise, the `lang, charset = self.model.selected_language.split(".")` would raise an exception.

ogayot: Well, it does not make sense. I guess the condition is inverted. I must have tried before refactoring :facepalm:

subiquity/server/controllers/locale.py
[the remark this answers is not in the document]
ogayot: I think my brain was malfunctioning. If no lang is specified in an answers-base run, the form is not submitted. 

I saw:

```python
        if 'lang' in self.answers:
            self.done((self.answers['lang'], ""))
```
and read
```python
        if 'lang' in self.answers:
            self.done(self.answers.get('lang', ""))
```

ogayot:
I reworked the locale generation part a bit because I misinterpreted the existing behavior.

Some languages in the languages list don't have an encoding specified e.g.:
```
ssh:oc:Occitan (aprèp 1500)
ssh:bo_IN:བོད་ཡིག
console:sr_RS:Српски
console:nb:Norsk bokmål
```
and so we need the server to react nicely to those. I've made use of `locale.normalize` which returns something sensible for these languages:

```python
locale.normalize("nb")    => 'nb_NO.ISO8859-1'
locale.normalize("sr_RS") => 'sr_RS.UTF-8'
locale.normalize("bo_IN") => 'bo_IN.UTF-8'
locale.normalize("oc")    => 'oc_FR.ISO8859-1'
```

IMO using `locale.normalize` is a better approach than what we are doing in `make_cloudconfig` (which does not work well for "oc" and "nb"):

```python
        if '.' not in locale and '_' in locale:
            locale += '.UTF-8'
```

so I am wondering if we should update make_cloudconfig ; but it will potentially select a different default encoding than UTF-8.

```python
loc = "fr_FR"
locale.normalize(loc)   => 'fr_FR.ISO8859-1'
if '.' not in loc and '_' in loc:
     loc += '.UTF-8'    => 'fr_FR.UTF-8'
```

subiquity/server/apt.py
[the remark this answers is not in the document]
mwhudson: Ah no you're right we set LANG by default. I find all of this quite confusing I'm afraid :(

subiquity/server/controllers/locale.py
mwhudson: ah ok :)

mwhudson (approved):
This looks OK to me now. Would like Dan to take a look too though.

dbungert (approved):

dbungert:
Rebase to main should fix CI.

ogayot:
Rebased on top of main.

PR #1548, opened 2023-02-02
kinetic merge: tui: turn off systemd ShowStatus

dbungert:
* cherry-picked 8dd1495 from #1545


mwhudson (approved):
This is what's going to be on the 24.04.2 media, correct?

dbungert:
> This is what's going to be on the 24.04.2 media, correct?

Yes, assuming a stable release is done in time.

PR #1549, opened 2023-02-03
apt: restore unmount

dbungert:
This unmount is needed, as otherwise the sources.list only points to the /cdrom during curthooks.


dbungert:
Experiment to consider: change the output of https://github.com/canonical/subiquity/blob/f47874bdc95912d412c5461843f0341fb932b414/subiquity/server/apt.py#L218 and observe the changes to what happens in the curtin install log at the apt-get update step.

dbungert:
Closing, we want something similar eventually though.

PR #1550, opened 2023-02-06
Active Directory API Part 1

CarlosNihelton:
This is a skeleton implementation of the Active Directory endpoint API. There is more to come, but this was growing too buch to bite in one single code review.

The current state just implements the endpoint `/active_directory` backed by an optional controller, since AD is an opt-in feature. The endpoint is connected to the subiquity server to allow growing and testing the API simultaneously, but should not cause any disruption in current workflows.

Don't bother with the almost empty `POST` method by now. The next pull request will be dedicated to it to validate the user input before configuring the controller, following the ideas implemented in Ubiquity.

A third pull request will address the autoinstall schema to allow interactive section to prefil the admin and domain controller names, but hold on the password during setup.

A forth PR will address early pinging the domain controller and a fitth will bring the blocking endpoint the clients should watch for the result of the actual target system joining.

---

Revision notes: added the auto install PR to the plan.

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -407,6 +407,11 @@ class API:
 
     class active_directory:
         def GET() -> Optional[ADConnectionInfo]: ...
+        # POST must validate the payload before configuring the controller,
+        # which may contain several errors as described in [ADValidationResult]
+        # simultaneously - such as invalid chars on the admin name and DC name
+        # starting with a hyphen or a dot. Thus this must returns a List
+        # of errors [ADValidationResult.OK] on success.
         def POST(data: Payload[ADConnectionInfo]) \
             -> List[ADValidationResult]: ...
 
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -25,6 +25,8 @@ from subiquitycore.models.network import (
 
 from subiquity.common.api.defs import api, Payload, simple_endpoint
 from subiquity.common.types import (
+    ADValidationResult,
+    ADConnectionInfo,
     AddPartitionV2,
     AnyStep,
     ApplicationState,
@@ -402,6 +404,11 @@ class API:
 
     class integrity:
         def GET() -> CasperMd5Results: ...
+
+    class active_directory:
+        def GET() -> Optional[ADConnectionInfo]: ...
+        def POST(data: Payload[ADConnectionInfo]) \
+            -> List[ADValidationResult]: ...
 
 
 class LinkAction(enum.Enum):
--- a/subiquity/models/ad.py
+++ b/subiquity/models/ad.py
@@ -30,7 +30,7 @@ class ADModel:
     async def target_packages(self):
         # NOTE Those packages must be present in the target system to allow
         # joining to a domain.
-        if self.do_join is True:
+        if self.do_join:
             return ["adcli", "realmd", "sssd"]
 
         return []
--- a/subiquity/models/ad.py
+++ b/subiquity/models/ad.py
@@ -30,4 +30,7 @@ class ADModel:
     async def target_packages(self):
         # NOTE Those packages must be present in the target system to allow
         # joining to a domain.
-        return ["adcli", "realmd", "sssd"] if self.do_join else []
+        if self.do_join is True:
+            return ["adcli", "realmd", "sssd"]
+
+        return []
--- a/subiquity/models/ad.py
+++ b/subiquity/models/ad.py
@@ -1,0 +1,33 @@
+# Copyright 2023 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+from typing import Optional
+
+from subiquity.common.types import ADConnectionInfo
+
+log = logging.getLogger('subiquity.models.ad')
+
+
+class ADModel:
+    """ Models the Active Directory feature """
+    def __init__(self) -> None:
+        self.do_join = False
+        self.conn_info: Optional[ADConnectionInfo] = None
+
+    async def target_packages(self):
+        # NOTE Those packages must be present in the target system to allow
+        # joining to a domain.
+        return ["adcli", "realmd", "sssd"] if self.do_join else []
--- a/subiquity/server/controllers/ad.py
+++ b/subiquity/server/controllers/ad.py
@@ -1,0 +1,39 @@
+# Copyright 2023 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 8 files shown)


CarlosNihelton:
Running a lunar container locally reproduces the CI failure, but it's out of our scope for now. Something deep inside building cpython.

subiquity/tests/api/test_api.py
```diff
@@ -1632,3 +1632,21 @@ async def test_wslsetupoptions(self):
 
             resp = await inst.get(endpoint)
             self.assertFalse(resp['install_language_support_packages'])
+
+
+class TestActiveDirectory(TestAPI):
+    async def test_ad(self):
```
dbungert, line 1638: I've been inconsistent on enforcing this, but please add the `@timeout()`

subiquity/models/ad.py
```diff
@@ -0,0 +1,33 @@
...
+log = logging.getLogger('subiquity.models.ad')
+
+
+class ADModel:
+    """ Models the Active Directory feature """
+    def __init__(self) -> None:
+        self.do_join = False
+        self.conn_info: Optional[ADConnectionInfo] = None
+
+    async def target_packages(self):
```
dbungert, line 30: What's the plan for this?  Curtin does have some logic here for carrying a list of needed packages and it dynamically installs them if needed.  Would you expect them to be on the system already, for both the minimal and non source cases?

subiquity/tests/api/test_api.py
```diff
@@ -1632,3 +1632,21 @@ async def test_wslsetupoptions(self):
...
+            self.assertIsNone(ad_dict)
+
+            # POST succeeds
+            ad_dict = {
+                'domain_name': 'ubuntu.com',
+                'admin_name': 'u',
+                'password': 'u'
+            }
+            result = await instance.post(endpoint, ad_dict)
+            self.assertIn('OK', result)
```
dbungert, line 1652:
```suggestion
            self.assertEqual(['OK'], result)
```

Though this depends on the answer to the return type above

subiquity/common/apidef.py
```diff
@@ -403,6 +405,11 @@ def GET(user_id: str) -> SSHFetchIdResponse: ...
     class integrity:
         def GET() -> CasperMd5Results: ...
 
+    class active_directory:
+        def GET() -> Optional[ADConnectionInfo]: ...
+        def POST(data: Payload[ADConnectionInfo]) \
+            -> List[ADValidationResult]: ...
```
dbungert, line 411: In what scenario would we see multiple results?

subiquity/server/controllers/ad.py
```diff
@@ -0,0 +1,39 @@
...
+from subiquity.server.controller import SubiquityController
+
+log = logging.getLogger('subiquity.server.controllers.ad')
+
+
+class ADController(SubiquityController):
+    """ Implements the server part of the Active Directory feature. """
+    model_name = "ad"
+    endpoint = API.active_directory
+    # No auto install key and schema for now due password handling uncertainty.
```
dbungert, line 30: This isn't the agreed upon plan.  I can accept 'follow-on PR' as the plan.

dbungert (changes requested):

dbungert:
> Running a lunar container locally reproduces the CI failure,

And it doesn't seem related to this PR.  I can reproduce it by a no-change re-run of CI on the main branch.  https://github.com/dbungert/subiquity/actions/runs/4108533137

dbungert:
https://bugs.launchpad.net/ubuntu/+source/python3.11/+bug/2006404 opened about the above failure.  I'm testing CI changes as well.

dbungert:
A rebase to main should fix CI.

subiquity/models/ad.py
CarlosNihelton: Those packages must be on the live CD pool. Lots of users won't join AD, thus it makes sense for not having them inthe minimal layer. Having them on the pool enables curtin to install them in the target system for those that opt in for AD.

subiquity/common/apidef.py
CarlosNihelton: Multiple errors. Since this is meant to validate the whole structure at once we may have a combination of errors on username and hostname, for instance. I was in doubt about the success case: empty list versus OK value. My take was in favor of explicitness.

subiquity/server/controllers/ad.py
CarlosNihelton: Let's take the follow-on approach then.

subiquity/models/ad.py
dbungert: Having the packages in the ship-live pool is just fine and matches other similar things.  We just might want to move this list to curtin.

subiquity/common/apidef.py
dbungert: Hmm, I don't know how username validation will work if the hostname is wrong, but I'm in favor of returning all the info we have.  I suggest a comment as this sort of multiple errors return is a bit uncommon.

subiquity/models/ad.py
CarlosNihelton: Is there any other endpoint that moved a similar list to curtin we could follow the trend on? I've seen the pattern herein applied in a few places in subiquity's source. It might be worth doing them altoghether if the procedure is straightforward.

subiquity/common/apidef.py
CarlosNihelton: That validation is just matching regexes. I'm placing that validation in the server to avoid code duplication on the clients.

subiquity/models/ad.py
```diff
@@ -0,0 +1,33 @@
...
+        # NOTE Those packages must be present in the target system to allow
+        # joining to a domain.
+        return ["adcli", "realmd", "sssd"] if self.do_join else []
```
mwhudson, line 33: If this bit stays here can I be old fashioned and ask you to write this out more long hand? I find it clearer for this sort of thing.

subiquity/models/ad.py
mwhudson: I think if the logic for joining the domain etc is going into curtin, the list of package required should too, probably.

mwhudson (commented):
Looks good as far as it goes! Thanks.

sergio-costas (approved):
Don't see any obvious error, so LGTM.

subiquity/server/controllers/ad.py
dbungert: I have been strict in recent times about making sure that autoinstall is implemented for all new features.  I hear now that I'm not getting that.
I want a ticket somewhere that I can track to ensure this doesn't get dropped.

subiquity/models/ad.py
dbungert: curtin has a `detect_required_packages` concept, and I look at this as very similar.  But maybe that particular detection is too late.

dbungert (approved):
My feedback items have been addressed, Michael has valid feedback items so please respond to those as well before merge.

subiquity/models/ad.py
CarlosNihelton: My plan A is to follow the example of codecs, ssh and network, where we keep a list of optional packages in the model to get them installed in the target system by the install controller's postinstall method (if necessary):

https://github.com/canonical/subiquity/blob/f77971238f179cbcb0db8b16279157ef6801db28/subiquity/server/controllers/install.py#L382-L394

I don't think that fits in curtin (as Dan suggested) because we consider those packages optional and can only now they are required when the user interacts with the client UI.

CarlosNihelton: I might be wrong about curtin though. My understanding is that `detect_required_packages` output depends on the configuration data, not matching the interactive use case.

subiquity/models/ad.py
CarlosNihelton: 😄 
I prefer the mouthful if-statement over the comprehension too.

subiquity/models/ad.py
CarlosNihelton: Unless we install the packages as part of the same action (command) used to join the domain. Does that sound as awkward to you as to me? I mean, the command would have the responsibility of installing packages and performing the join. Besides SRP, it feels strange to have distinct moments during the installation when we install packages. It seems more natural do have one single moment of `Installing additional packages...`.

I'm not sure. Let me know your thoughts.

mwhudson (approved):
all good from me once you fix the if self.do_join is True thing!

PR #1551, opened 2023-02-07
requirements: remove

dbungert:
Fixes problems building the yarl wheel we aren't using.
https://github.com/canonical/subiquity/pull/1550#issuecomment-1419726594


mwhudson (approved):
Oh yeah, missed that I could remove that!

PR #1552, opened 2023-02-07
Support of multiple candidates mirrors with automatic selection in autoinstall & desktop installs

ogayot:
This PR introduces a mechanism to automatically elect a working mirror out of multiple candidates. This will be used in autoinstalls (implicitly) and in desktop installs (either explicitly - which is recommended - or implicitly).

### Autoinstall format

The format of the autoinstall apt->primary section has been revisited to support multiple candidates. Because it introduces backward-incompatible changes, the `primary` section must be wrapped in a `mirror-selection` section to enable the new format.
```yaml
apt:
   primary:
       # Usual directives that are inherited from curtin
```

```yaml
apt:
  mirror-selection:
    primary:
    - uri: http://candidate-mirror1
    - country-mirror     # <- will be resolved using geoip, probably http://{cc}.archive.ubuntu.com/ubuntu
    - uri: http://candidate-mirror3
      arches: [amd64]
    - uri: http://candidate-mirror4
```

Mirrors are tested in the order defined and the first one to pass the test will be used for the rest of the install.

### Implementation details

in subiquity/models/mirrors, two new types are introduced:

* `LegacyPrimarySection` which is a wrapper around the old curtin primary section directives (there can only be one).
* `PrimaryEntry` which is a wrapper around _one_ candidate in the new format. In the snippet above, there would be 4 `PrimaryEntry` instances  (including one for the country mirror).

### API changes

* GET /mirror now returns:
  * the URL of all the candidates
  * the URL of the candidate staged for testing (if any - otherwise null)
  * the URL of the elected candidate (if any - otherwise null)
```json
{
    "candidates": ["http://fr.archive.ubuntu.com/ubuntu", "http://archive.ubuntu.com/ubuntu"],
    "staged": "http://fr.archive.ubuntu.com/ubuntu",
    "elected": null
}
```
* POST /mirror can now be invoked in different ways:
```yaml
# Make fr.archive.ubuntu.com be the elected mirror ; and mark the mirror model configured.
{"elected": "http://fr.archive.ubuntu.com/ubuntu"}

# Mark us.archive.ubuntu.com as staged for testing.
# This replaces POST /mirror/candidate that was recently introduced
{"staged": "http://us.archive.ubuntu.com/ubuntu"}

# Replace the list of candidates with the new values. Not used on practice, but can be used for testing / debugging.
{"candidates": ["http://us.archive.ubuntu.com/ubuntu"]}

# Explicit (more or less :D) way to trigger the election of a mirror automatically. 
# If a mirror is elected, the model will be marked configured ; otherwise an exception is raised.
# This would be the recommended option for the desktop installer
# NOTE: maybe something like {"auto_elect": true} is better?
null
```
@jpnurmi ^ 
POST /mirror with null in the body (subject to change to something more explicit) is what I envision the desktop installer doing going forward. This would be a "long polling" call that will find a suitable mirror or raise an exception if no mirror can be used. The next step is to implement the forced offline install if this call fails.

### Things to do before marking ready for review

- [x] More VM testing; including autoinstalls in both the legacy and new format
- [x] Produce usable autoinstall-user-data


subiquity/server/controllers/mirror.py
```diff
@@ -184,28 +237,74 @@ async def configured(self):
...
         await asyncio.gather(self.source_configured_event.wait(),
                              self.configured_event.wait())
-        await self.apt_configurer.apply_apt_config(self.context)
+        if self.model.primary_elected is None:
+            # NOTE: In practice, this should only happen if the mirror was
+            # marked configured using a POST to mark_configured ; which is not
+            # recommended. Clients should do a POST request to /mirror with
+            # null as the body instead.
+            await self.find_and_elect_candidate_mirror(self.context)
+        await self.apt_configurer.apply_apt_config(self.context, elected=True)
```
ogayot, line 246: FIXME: This fails when running an automated install without network:

```python
  File "/snap/subiquity/x1/lib/python3.8/site-packages/subiquity/server/controllers/mirror.py", line 246, in _promote_mirror
    await self.apt_configurer.apply_apt_config(self.context, elected=True)
  File "/snap/subiquity/x1/lib/python3.8/site-packages/subiquity/server/apt.py", line 131, in apply_apt_config
    generate_config_yaml(config_location, self.apt_config(elected))
  File "/snap/subiquity/x1/lib/python3.8/site-packages/subiquity/server/apt.py", line 119, in apt_config
    self.app.base_model.mirror.get_apt_config_elected())
  File "/snap/subiquity/x1/lib/python3.8/site-packages/subiquity/models/mirror.py", line 220, in get_apt_config_elected
    assert self.primary_elected is not None
AssertionError
```

dbungert (commented):
I've given this a partial read so far, I want to have a look, but so far so good.  Hopefully that means any feedback is not large.

subiquity/models/mirror.py
```diff
@@ -32,12 +36,111 @@
...
+]
 
 DEFAULT = {
     "preserve_sources_list": False,
 }
 
 
+@attr.s(auto_attribs=True)
+class PrimaryElement(abc.ABC):
+    parent: "MirrorModel" = attr.ib(kw_only=True)
```
mwhudson, line 59: I think this class needs a docstring to explain what an instance represents because after a few seconds of staring I certainly don't know :-) I think you need some kind of glossary for terms like "primary", "element", "legacy", "staged", "elected". It's not that hard for me to guess at what is meant by each word but it is a bit opaque on first reading and I might be getting something wrong.

subiquity/models/mirror.py
```diff
@@ -32,12 +36,111 @@
...
+    @abc.abstractmethod
+    def serialize_for_ai(self) -> Any:
+        """ Serialize the element for autoinstall. """
+
+
+@attr.s(auto_attribs=True)
+class PrimaryEntry(PrimaryElement):
+    # Having uri set to None is only valid for a country mirror.
+    uri: Optional[str] = None
+    arches: Optional[List[str]] = None
```
mwhudson, line 76: I see the unittest for this tests both arches == [] and None. What do they both mean?

subiquity/server/controllers/mirror.py
```diff
@@ -130,8 +141,16 @@ async def try_mirror_checking_once(self) -> None:
...
-    @with_context()
-    async def apply_autoinstall_config(self, context):
+    async def find_and_elect_candidate_mirror(self, context):
+        # This function assumes that the source and proxy models have
+        # already been marked configured. Currently apply_autoinstall_config
+        # coroutines are called sequentially so it should be okay for a fully
+        # non-interactive install. If we were to invoke the coroutines
+        # concurrently, we would need to wait for the source and proxy to
+        # finish first.
+        # TODO check if the same assumption applies in a partially interactive
```
mwhudson, line 151: I don't think it does. I guess you get to confirm this though!

subiquity/server/controllers/mirror.py
```diff
@@ -143,13 +162,39 @@ async def apply_autoinstall_config(self, context):
...
-            await self.try_mirror_checking_once()
-        except AptConfigCheckError:
-            log.debug("Retrying in 10 seconds...")
+        # Try each mirror one after another.
+        compatibles = self.model.compatible_primary_candidates()
+        for idx, candidate in enumerate(compatibles):
+            if idx != 0:
+                # Sleep before testing the next candidate..
+                log.debug("Will check next candiate mirror after 10 seconds.")
+                await asyncio.sleep(10)
```
mwhudson, line 171: Can't this go at the end of the for loop?

mwhudson (commented):
Looks great so far. I think the new model code is a bit opaque at first glance and the reader needs some hints to get started. I haven't read the controller code closely yet.

subiquity/server/controllers/mirror.py
ogayot: I would definitely prefer to but I also want to avoid sleeping after the final test, because it only makes sense to sleep for 10 seconds (which is a pretty bad penalty if there is still something to test after.

I came up with this enumerate + reversed logic construct to do exactly that, but I hate how this looks so any suggestion is appreciated :)

Basically, I want to sleep between iterations, rather than after each iteration. I guess this is somewhat similar to `'_'.join(("a", "b", "c"))` returning `"a_b_c"` and not `"a_b_c_"`.

subiquity/models/mirror.py
ogayot: * `arches = []` means that no arches are supported. (I hope we don't see this in a real world scenario :+1: ).
* `arches = None` means that the current CPU architecture is supported.

I will make sure to add more documentation!

subiquity/common/types.py
```diff
@@ -747,6 +747,20 @@ class MirrorCheckResponse:
...
+    elected: Optional[str] = None
+    candidates: Optional[List[str]] = None
+    staged: Optional[str] = None
+
+
+@attr.s(auto_attribs=True)
+class MirrorGet:
+    elected: Optional[str]
+    candidates: List[str]
+    staged: Optional[str]
```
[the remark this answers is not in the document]
ogayot: I think the most obvious reason is that when doing a GET, having candidates=None is **not** acceptable, so having a single type with `candidates: Optional[List[str]] = None` is .. meh.

ideally IMO, the `MirrorPost` would be represented by a `TypedDict` with all keys optional ; rather than an `attr.s` with default `None` values everywhere ; but we don't support serialization with `TypedDict` classes (yet?).

subiquity/models/mirror.py
```diff
@@ -49,51 +152,88 @@ class MirrorModel(object):
...
+        assert "primary" not in self.config
+
         config = copy.deepcopy(self.config)
-        config["primary"] = copy.deepcopy(self.primary)
         config["disable_components"] = sorted(self.disabled_components)
         return config
 
-    def mirror_is_default(self):
-        return self.get_mirror() == self.default_mirror
+    def get_apt_config_staged(self) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
ogayot: I added it, thanks. 
I took the opportunity to replace "elected" by "final" in apt code since "elected" is more of a mirror thing, that does not make sense in other contexts.

subiquity/models/mirror.py
ogayot: Updated, let me know if this is clearer now :)

I made some adjustments too, hopefully this is more sensisble:
PrimaryElement => BasePrimaryEntry
LegacyPrimarySection => LegacyPrimaryEntry
PrimaryEntry => PrimaryEntry (stays the same)

subiquity/models/mirror.py
ogayot: Updated. Also fixed issue causing the list of arches passed to curtin to be empty. Since we only give curtin a single URI, passing arches=["default"] ensures it uses it and does not revert to http://archive.ubuntu.com

subiquity/server/controllers/mirror.py
ogayot: Just as you thought, we can't make the same assumptions. I tested with "proxy" as an interactive section and the mirror testing certainly started way before I had a chance to make "proxy" configured. I added an `await self.proxy_configured.wait()` statement and it works better now.

dbungert:
```
    version: 2
    primary:
```
I'm having second thoughts about this part.  I'm trying to keep autoinstall at a minimal of boilerplate.   What conflicts does this scheme have with `version: 1`?

subiquity/server/controllers/mirror.py
```diff
@@ -143,13 +163,39 @@ async def apply_autoinstall_config(self, context):
...
+            if candidate.uri is None:
+                log.debug("Skipping unresolved country mirror")
```
dbungert, line 174: if candidate.uri is ever none for any other reason this log statement will be confusing

dbungert (commented):
This all looks fine but I'd like to resolve the autoinstall question first.

ogayot:
> I'm having second thoughts about this part. I'm trying to keep autoinstall at a minimal of boilerplate. What conflicts does this scheme have with `version: 1`?

Thanks for asking. There are a few reasons, but mainly because we handle the arches differently and because v2 drops support t some of the specifiers that curtin supports:

* Different behavior for the same autoinstall section. For instance, this:
```yaml
primary:
  - uri: http://mirror/ubuntu
```
will parse properly in both versions, but the absence of an `arches` specifier will yield to a different result.
With version 1, this section will cause the install to use http://archive.ubuntu.com/ubuntu because that's what curtin's get_mirror() returns when no arches match.
With version 2, the install will use http://mirror/ubuntu "as instructed".

* version 2 has no support for the `arches: [default]` specifiers that curtin internally supports. Already existing autoinstalls often have this specifier so I didn't want to completely lose support for it.

* version 2 has no support for the `search` directive (a sort of replacement for uri) that curtin internally supports. Subiquity did not produce `search` directives in autoinstall but manually crafted configurations can have it.

dbungert:
I wish I thought of this earlier.

> There are a few reasons, but mainly because we handle the arches differently and because v2 drops support for some of the specifiers that curtin supports ...

What if we used a different word than "primary"?  If we use a new term here then people can opt-in to the new behavior, which they would have to do anyway by setting `version: 2`, but it would mean there is no conflict unless someone both has a "primary" section content and the new term and the resulting autoinstall will be shorter.  

"candidates" maybe, and possibly nested under primary or security?  or "detect"?  what are we doing with security?
could be "primary-candidates"  if you prefer to avoid the list versus object weirdness.

```yaml
apt:
    primary:
        # old behavior, entirely unchanged
```
OR
```yaml
apt:
    primary:
        candidates:
            # new stuff
```

subiquity/server/controllers/mirror.py
ogayot: if it's none for any other reason, this means something has gone very wrong :)
If you think it's a problem, I can add a getter for `uri` and make it `assert self.country_mirror` if uri is None.

dbungert: I'm thinking about my recent experience in Curtin where I believe that no swap was being created, and really it was a swap file problem.
Yes, we should have hit a ValueError before getting this far.
I think it's potentially interesting to store the raw input configuration line with the entry, which should be only a small change and means you would be able to log it here.  WDYT?

ogayot:
> what are we doing with security?

I don't have special support for it. So if there is a `security` section, it should still use the curtin format. It will be merged to the apt config before being passed to curtin apt-config. Otherwise, curtin will use its own default value for the security mirror. When we're doing mirror testing, we test connectivity to the primary and the secondary mirror, so having a bad security mirror will break all tests, but this is a limitation I am ready to accept for now, since we don't have a screen to set the security mirror.

> What if we used a different word than "primary"? If we use a new term here then people can opt-in to the new behavior, which they would have to do anyway by setting `version: 2`, but it would mean there is no conflict unless someone both has a "primary" section content and the new term and the resulting autoinstall will be shorter.

I guess that would work. I'd rather avoid nesting though. This will help make the JSON schema more robust. I think there is a way to express  the notion of "zero or one of" ["primary", "primary-candidates"]
Does `apt: primary-candidates:` work is does it sound too political ? :laughing: 

I need to think a bit more about the implications before going forward with this change though.

dbungert:
> Does apt: primary-candidates: work is does it sound too political ? laughing

hehe, yea I'm not completely sold on that term, alternate suggestions definitely welcome! I'm just trying to come up with a new term that is distinguishable from the old setup, unambiguous, and avoiding the version line.

subiquity/server/controllers/mirror.py
ogayot: I'm not completely sure I got what you're saying. 

By raw input configuration, you mean the original yaml autoinstall section?
There is not always an original yaml. For the desktop installer use case, for instance, the two candidates will be the default ones created automatically by `MirrorModel`.

Would you like to see a logline with the autoinstall representation of each candidate we try? That I can do with `candidate.serialize_for_ai()`

dbungert: > By raw input configuration, you mean the original yaml autoinstall section?

Well, one line from it at least, the relevant object from version-2 primary list.

> Would you like to see a logline with the autoinstall representation of each candidate we try? That I can do with candidate.serialize_for_ai()

Works for me, I'm looking to leverage something we have at hand that is just more clear.

subiquity/models/mirror.py
```diff
@@ -12,15 +12,75 @@
...
+   "entries", each denoting a primary candidate.
+
+primary entry
+-------------
+ * represents a fragment of the 'apt->primary' autoinstall section. Each entry
+ can be used as a primary candidate.
+ * in the legacy format, the primary entry corresponds to the whole primary
+ section.
+ * in the new format, multiple primary entries make the primary section.
+"""
```
mwhudson, line 69: thanks for this!!

subiquity/server/controllers/mirror.py
mwhudson: Yes but in this case the loop is usually exited via break?

mwhudson (approved):
I think this looks pretty good now, thanks. Do get Dan's approval before merging please :-)

subiquity/server/controllers/mirror.py
ogayot: If a mirror is usable, yes. Otherwise, no.

dbungert:
Olivier and I chatted about this, we want to finalize on what's happening with the YAML then I think this is good for merge.  We're both going to put together a few suggestions.

ogayot:
Updated by moving the `primary` section from under `apt:` to `apt: mirror-selection:` to enable the new behavior.
The `security` key (which is optional and undocumented in subiquity) stays under the `apt:` section since it only supports the old behavior.

autoinstall-schema.json
```diff
@@ -313,6 +313,39 @@
                 "primary": {
                     "type": "array"
                 },
+                "mirror-selection": {
```
dbungert, line 316: Would you update autoinstall-reference.md please?

dbungert (approved):
Doc update and I think we're good.

dbungert (approved):

PR #1553, opened 2023-02-08
Active Directory API Part 2 - Validations on POST

CarlosNihelton:
Following up on #1550, this brings the current domain controller and admin user name validations that exists on Ubiquity.

This will later be augmented when the endpoint for testing the connection to the domain controller gets implemented.

The API tests demonstrate the kind of errors we are talking about.

As before, this is an optional controller, thus the current setup experience remains unaffected.


subiquity/tests/api/test_api.py
```diff
@@ -1651,3 +1651,59 @@ async def test_ad(self):
             }
             result = await instance.post(endpoint, ad_dict)
             self.assertEqual(['OK'], result)
+
+            ad_dict = {
```
dbungert, line 1655: it would be nice to move some of this testing to unit tests instead of the API ones, but having at least one API test is still nice so the tests here won't actually be slower.  *shrug*  

The validation methods aren't part of the controller, so just calling those methods direct from unit test shouldn't be too bad.  An example of setting up a controller for unit test can be seen at subiquity/server/controllers/tests/test_filesystem.py  if you need it.

subiquity/server/controllers/ad.py
```diff
@@ -34,6 +39,92 @@ async def GET(self) -> Optional[ADConnectionInfo]:
         return self.model.conn_info
 
     async def POST(self, data: ADConnectionInfo) -> List[ADValidationResult]:
+        """ Configures this controller with the supplied info.
+            Returns a list of errors if the info submitted is invalid or
+            [ADValidationResult.OK] on success. """
+        result = check_domain_name(data.domain_name) | \
```
dbungert, line 45: style nitpick:  I might structure it as
```
result = set()
result |= check_a()
result |= check_b()
```
Also what was done in check_domain_name is fine.
That should make for cleaner diffs if this is tweaked in the future, and I feel like validation rules are a high traffic bit of code anywhere they are found.

dbungert (approved):
I appreciate that you're structuring them in small PRs.  Easier for us and hopefully more timely reviews for you.

I have some suggestions that you are free to implement or decline.

subiquity/tests/api/test_api.py
CarlosNihelton: Indeed, we can take advantage of those out-of-class routines to test them exaustively in unit tests and test the wiring-up in the API with just a few cases. I like the idea.

subiquity/server/controllers/ad.py
```diff
@@ -34,6 +39,92 @@ async def GET(self) -> Optional[ADConnectionInfo]:
         return self.model.conn_info
 
     async def POST(self, data: ADConnectionInfo) -> List[ADValidationResult]:
+        """ Configures this controller with the supplied info.
```
mwhudson, line 42: Perhaps I should have made this comment on the spec or on the last PR or something but well.

Doing this validation as part of POST seems odd. The pattern in the TUI (and just general UX principle) is that invalid data is highlighted in the client as soon as it is entered (or at least as soon as the relevant input widget loses focus) and the "submit" button is not available until the data passes validation.

I can understand the desire to do validation in the server so that it doesn't have to be duplicated between client and server and indeed checking that the data submitted in POST is valid, but the API here does not seem sufficient to provide a good experience. Am I missing something?

mwhudson (commented):
The code looks fine but I do have a design question I'd like an answer to.

subiquity/server/controllers/ad.py
CarlosNihelton: I think you raised a good point. I should have considered it in the spec as well. I'll create specific endpoints clients call immediately after the user finishes input a field, allowing better UX. I was only considering doing so for the DC name, because, besides the regex validation, there will be an attempt to PING it at next. But it makes sense to create the endpoints right away.

CarlosNihelton:
I think the code changed significantly since the last review. I'd be nice to have a second round.

subiquity/common/apidef.py
```diff
@@ -412,8 +414,16 @@ def GET() -> Optional[ADConnectionInfo]: ...
...
+        def POST(data: Payload[ADConnectionInfo]) -> None: ...
+
+        class check_domain_name:
+            def GET(domain_name: str) -> List[AdDomainNameValidation]: ...
+
+        class check_admin_name:
+            def GET(admin_name: str) -> List[AdAdminNameValidation]: ...
+
+        class check_password:
+            def GET(password: str) -> AdPasswordValidation: ...
```
ogayot, line 426: AFAIK this prototype means the password is passed in the query string, so might end up in the access logs. You might want to use Payload[str] so it is passed in the body instead.

It is less of a problem for the other calls, but it feels a bit redundant when you read it:

GET /check_admin_name?admin_name=...
GET /check_domain_name?domain_name=...

subiquity/common/types.py
```diff
@@ -754,12 +754,25 @@ class ADConnectionInfo:
     password: str = attr.ib(repr=False, default="")
 
 
-class ADValidationResult(enum.Enum):
+class AdAdminNameValidation(enum.Enum):
```
ogayot, line 757: Usually, we try to give sensible values for the enums fields that are serialized / deserialized and transit over HTTP.
Right now, the different values will be represented as integers, so this isn't too user friendly.

subiquity/server/controllers/ad.py
```diff
@@ -33,7 +43,124 @@ async def GET(self) -> Optional[ADConnectionInfo]:
...
+        result = set()
+
+        if len(name) == 0:
+            log.debug("admin name is empty")
+            return {AdAdminNameValidation.EMPTY}
+
+        # Ubiquity checks the admin name in two steps:
+        # 1. validate the first char against '[a-zA-Z]'
+        # 2. check the entire string against r'^[-a-zA-Z0-9_]+$'
+        if not re.match('[a-zA-Z]', name[0]):
```
ogayot, line 82: `re.match` is strange and acts as if the pattern is prefixed with "^". Overall, I prefer `re.search` and `re.fullmatch` for this reason but in this specific scenario, `re.match` makes sense. The [0] part is not necessary though:

```suggestion
        if not re.match('[a-zA-Z]', name):
```

subiquity/server/controllers/ad.py
```diff
@@ -33,7 +43,124 @@ async def GET(self) -> Optional[ADConnectionInfo]:
...
+        if len(name) == 1:
+            return result
+
+        regex = re.compile(ADMIN_RE)
+        if not regex.match(name[1:]):
+            log.debug('<%s>: domain admin name contains invalid characters',
+                      name)
+            result.add(AdAdminNameValidation.INVALID_CHARS)
+
+        if len(result) > 0:
```
ogayot, line 94: consider using `if result:`

subiquity/server/controllers/ad.py
```diff
@@ -33,7 +43,124 @@ async def GET(self) -> Optional[ADConnectionInfo]:
...
+            return result
+
+        return {AdAdminNameValidation.OK}
+
+    @staticmethod
+    def password(value: str) -> AdPasswordValidation:
+        """ Validates that the password is not empty. """
+
+        if len(value) < 1:
```
ogayot, line 103: consider using `if not value:`

subiquity/server/controllers/ad.py
```diff
@@ -33,7 +43,124 @@ async def GET(self) -> Optional[ADConnectionInfo]:
...
+        # 1. validate the first char against '[a-zA-Z]'
+        # 2. check the entire string against r'^[-a-zA-Z0-9_]+$'
+        if not re.match('[a-zA-Z]', name[0]):
+            result.add(AdAdminNameValidation.INVALID_FIRST_CHAR)
+
+        if len(name) == 1:
+            return result
+
+        regex = re.compile(ADMIN_RE)
+        if not regex.match(name[1:]):
```
ogayot, line 89: consider using `regex.fullmatch` or `regexp.search` here. It won't make a difference because ADMIN_RE has `^...$` but IMO `regexp.match` is tricky.

subiquity/server/controllers/ad.py
```diff
@@ -33,7 +43,124 @@ async def GET(self) -> Optional[ADConnectionInfo]:
...
+
+        if name.endswith('.'):
+            log.debug('<%s>: %s cannot end with dots (.)',
+                      name, FIELD)
+            result.add(AdDomainNameValidation.END_DOT)
+
+        regex = re.compile(DC_RE)
+        if not regex.search(name):
+            result.add(AdDomainNameValidation.INVALID_CHARS)
+            log.debug('<%s>: %s contain invalid characters', name, FIELD)
```
ogayot, line 161: minor: s/contain/contains

subiquity/server/controllers/ad.py
ogayot: I like that we have client-side validation but I think we should still redo the validation on the server-side, especially because we're seemingly trying to avoid directory traversals issues.

dbungert: The web design best practice is that you do client-side validation for a smooth experience, and server-side validation because trusting the client to only send valid data is how you end up with sql injections and other similar nonsense.

We don't have quite the same level of urgency in Subiquity, as the client is trusted, but validating in both places is the safest choice.

subiquity/server/controllers/ad.py
```diff
@@ -14,14 +14,24 @@
...
+    AdAdminNameValidation,
+    AdDomainNameValidation,
+    AdPasswordValidation
+)
 from subiquity.server.controller import SubiquityController
 
 log = logging.getLogger('subiquity.server.controllers.ad')
 
+DC_RE = r'^[a-zA-Z0-9.-]+$'
+ADMIN_RE = r'^[-a-zA-Z0-9_]+$'
```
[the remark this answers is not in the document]
CarlosNihelton: Interesting. When testing that complain I got this screen and now I wonder whether we should block the invalid characters or replace them by underscores ...

image

This is how it looks after accepting the suggested replacements:
image

It looks like I could go even crazier than that:

image

Attempting to sign in with the converted username just works:
image

There are some docs about that: https://learn.microsoft.com/en-us/windows/win32/adschema/a-samaccountname?redirectedfrom=MSDN
The docs don't mention the `@` symbol, even though it's quite understandable that it shouldn't be in the user name.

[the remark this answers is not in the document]
CarlosNihelton: Even spaces are acceptable

image

🙃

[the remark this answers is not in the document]
dbungert: On the SAM-Account-Name doc is:

> Syntax | String(Unicode)

Does that imply that any arbitrary Unicode is acceptable, minus the reserved characters?  Like you could theoretically paste something in from Unicode 15 as a username and be able to log in with that?  I'm starting to think the validator should be just a check that you're not using the reserved characters and let everything else through.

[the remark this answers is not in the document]
CarlosNihelton: I'll test some other cryptic characters, but my plan now is to go for a `deny` regex instead of an `allow` one.

CarlosNihelton:
Since I'm working from a fork, Olivier's suggestion became less straighforward.

Here is the draft of the next PR https://github.com/CarlosNihelton/subiquity/pull/1

subiquity/server/controllers/ad.py
```diff
@@ -14,14 +14,28 @@
...
 from subiquity.server.controller import SubiquityController
 
 log = logging.getLogger('subiquity.server.controllers.ad')
 
+DC_RE = r'^[a-zA-Z0-9.-]+$'
+# Validation has changed since Ubiquity. See the following links on why:
+# https://bugs.launchpad.net/ubuntu-mate/+bug/1985971
+# https://github.com/canonical/subiquity/pull/1553#discussion_r1103063195
+# https://learn.microsoft.com/en-us/windows/win32/adschema/a-samaccountname
+AD_ACCOUNT_FORBIDDEN_CHARS = re.escape(r'@"/\[]:;|=,+*?<>')
```
ogayot, line 36: I'd prefer if you move the call to `re.escape` where you need it, but it is not a blocker.
```python
AD_ACCOUNT_FORBIDDEN_CHARS = r'@"/\[]:;|=,+*?<>'
[...]
regex = re.compile(f"[{re.escape(AD_ACCOUNT_FORBIDDEN_CHARS)}]")
```

ogayot (approved):

subiquity/common/apidef.py
```diff
@@ -412,8 +414,17 @@ def GET() -> Optional[ADConnectionInfo]: ...
         # simultaneously - such as invalid chars on the admin name and DC name
         # starting with a hyphen or a dot. Thus this must returns a List
         # of errors [ADValidationResult.OK] on success.
-        def POST(data: Payload[ADConnectionInfo]) \
-            -> List[ADValidationResult]: ...
+        def POST(data: Payload[ADConnectionInfo]) -> None: ...
```
mwhudson, line 417: I think this comment is a bit out of date now?

mwhudson (approved):

subiquity/common/apidef.py
CarlosNihelton: Indeed :D Thanks for catching that!

jpnurmi:
Hey, Dart HTTP client sets content-length by default to 0 for `GET` requests. While it's possible to manually override the header, would it be an option to use `POST` for validation just to make it cleaner on the client side?

> HttpException: Content size exceeds specified contentLength. 10 bytes written while expected 0. ["password"]

```dart
// GET and HEAD have 'content-length: 0' by default.
if (method == "GET" || method == "HEAD") {
  contentLength = 0;
}
```
https://github.com/dart-lang/sdk/blob/3f288db57eb80e6289f970ff0ff416f5b72142b2/sdk/lib/_http/http_impl.dart#L1433-L1436

PR #1556, opened 2023-02-08
storage/v2: allow unformatted partition creation

dbungert:
Previously unformatted partition creation was blocked, but supplying None does seem to do the right thing, and we want to support creation of unformatted partitions.  Allow it and add a API test to help document it as an intended feature.


mwhudson (approved):
Looks fine, I see Partition.format was already Optional in types.py so nothing to clean up there

dbungert:
@jpnurmi FYI

PR #1559, opened 2023-02-13
Stop calling ssh-import-id as part of the CI - use mock responses instead

ogayot:
We now have 3 different dry-run strategies for ssh key imports:

* success: fake a successful key import
* failure: fake a failed key import
* run-on-host: actually run ssh-import-id

Through a dry-run-config file, we can choose which strategy applies based on the username.

By default, the policy is as follows:
* username 'heracles' uses the fake successful key import
* username 'sisyphus' uses the fake failed key import
* any other username runs ssh-import-id


dbungert (approved):
Nice.  Add a comment then I think this is good for merge.

PR #1560, opened 2023-02-13
Fix unittest using timedatectl failing in chroot

ogayot:
I've upgraded my desktop to lunar and sadly this broke all the async unittests that use parameterized.

Until what needs fixing is fixed, I will run unittests in a chroot. Sadly, there is one test that fails because its behavior is dependent on the presence of /run/systemd/system ; so I'm making sure to does not.


dbungert (approved):
Interesting!  This does pass for me on Lunar, but I think mocking it is prudent.

mwhudson (approved):

PR #1561, opened 2023-02-13
storage: autoinstall_applied long poll endpoint

dbungert:
Ubuntu-desktop-installer would like to know when the storage autoinstall has been processed, so that it can show the preview information about disk changes.

The existing storage wait methods are insufficient, because that answers a different question - has block probing completed?

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -274,6 +274,11 @@ class API:
         class has_bitlocker:
             def GET() -> List[Disk]: ...
 
+        class autoinstall_applied:
+            def GET(wait: bool = False) -> bool:
+                """Poll or wait for storage autoinstall to be applied to the
+                model."""
+
         class v2:
             def GET(wait: bool = False) -> StorageResponseV2: ...
             def POST() -> StorageResponseV2: ...
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -151,6 +151,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         self._system_mounter: Optional[Mounter] = None
         self._role_to_device: Dict[str: _Device] = {}
         self._device_to_structure: Dict[_Device: snapdapi.OnVolume] = {}
+        self.autoinstall_applied: Optional[asyncio.Event] = None
         self.use_tpm: bool = False
 
     def is_core_boot_classic(self):
@@ -254,6 +255,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             raise Exception(
                 "autoinstall config did not create needed bootloader "
                 "partition")
+        self.autoinstall_applied.set()
 
     def update_devices(self, device_map):
         for action in self.model._actions:
@@ -943,6 +945,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
 
     async def _start(self):
         context = pyudev.Context()
+        self.autoinstall_applied = asyncio.Event()
         self._monitor = pyudev.Monitor.from_netlink(context)
         self._monitor.filter_by(subsystem='block')
         self._monitor.enable_receiving()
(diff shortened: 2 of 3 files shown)


subiquity/common/apidef.py
```diff
@@ -274,6 +274,12 @@ def GET() -> bool:
         class has_bitlocker:
             def GET() -> List[Disk]: ...
 
+        class autoinstall_applied:
+            def GET(wait: bool = False) -> None:
+                """Poll or wait for storage autoinstall to be applied to the
+                model."""
+                ...
```
mwhudson, line 281: obscure but you don't need the ... if you have a docstring :-)

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -56,6 +57,9 @@ def setUp(self):
         self.app.prober = mock.Mock()
         self.fsc = FilesystemController(app=self.app)
         self.fsc._configured = True
+        # _start() would do this normally, but the rest of what _start() does
+        # is going to need mocking out.
+        self.fsc.autoinstall_applied = asyncio.Event()
```
mwhudson, line 62: eh, the assignment is done in `__init__`? is this a leftover from a previous version?

subiquity/server/controllers/filesystem.py
```diff
@@ -988,3 +991,8 @@ def make_autoinstall(self):
         if 'swap' in rendered:
             r['swap'] = rendered['swap']
         return r
+
+    async def autoinstall_applied_GET(self, wait: bool = False) -> None:
+        if wait:
+            await self.autoinstall_applied.wait()
+        return self.autoinstall_applied.is_set()
```
mwhudson, line 998: er this returns bool, contrary to the type annotation.

mwhudson (approved):
A few quibbles but mostly short and sweet. I guess wait for @jpnurmi obvs.

subiquity/server/controllers/tests/test_filesystem.py
dbungert: This is a python 3.8 vs 3.10 distinction.  I originally had this initialized in `__init__`, but py3.8 is grumpy about the loop object.  So instead I need this worse version that risks a race.

subiquity/common/apidef.py
dbungert: Done, thanks.

subiquity/server/controllers/filesystem.py
dbungert: Done, thanks.

subiquity/server/controllers/tests/test_filesystem.py
mwhudson: ah.

mwhudson:
ps: please rebase or something, a commit message of "fixup" isn't great :-)

jpnurmi:
I had somehow managed to misread the proposal

> long poll style endpoint for completion of the autoinstall event

as something that wouldn't literally only wait for autoinstall (and be only usable in autoinstall mode). The goal is to present the storage config summary together with a _Start installing_ button regardless of whether we're running in autoinstall mode or not, so I had imagined this new endpoint would return the storage config as soon as it's available, blocking on autoinstall under the hood if appropriate.

Would this be doable? Does SubiquityController know if it's running in autoinstall mode?

subiquity/server/controllers/tests/test_filesystem.py
ogayot: What's the problem exactly in Python 3.8? Is it because in Python < 3.10, asyncio.Event instances must be created after the event loop is running? 

https://github.com/canonical/subiquity/pull/1446#issuecomment-1294792866
https://github.com/canonical/subiquity/pull/1446#issuecomment-1293254597

dbungert: > What's the problem exactly in Python 3.8? Is it because in Python < 3.10, asyncio.Event instances must be created after the event loop is running?

Yes.  See a Focal CI run, where I attempted to create the Event in __init__, for some test failures.
https://github.com/canonical/subiquity/actions/runs/4166633969/jobs/7211213000#step:3:2839

dbungert:
@jpnurmi is this PR obsolete?

jpnurmi:
@dbungert yes, sorry for wasting your time with this request. we found a simple way to get around the issue by waiting for the `NEEDS_CONFIRMATION` state. i hope that's a sensible solution.

PR #1563, opened 2023-02-14
Active Directory API Part 3 - Pinging the Domain Controller

CarlosNihelton:
In tandem with the Ubiquity this implements a "ping" to the domain controller during validation. It allows clients to validate not only if the DC name is "syntatically" correct, but also whehter it's reachable in the network.

The "ping" relies on realmd, which is part of the minimal layer. The package is not staged in Subiquity nor UDI at this point.

Here it is the domain name validation in action (running in a locally built ubuntu desktop installer snap):

image

### DIFF
--- a/subiquity/server/controllers/ad.py
+++ b/subiquity/server/controllers/ad.py
@@ -16,6 +16,7 @@
 import logging
 import re
 from typing import List, Optional, Set
+from subiquitycore.utils import arun_command
 
 from subiquity.common.apidef import API
 from subiquity.common.types import (
@@ -37,11 +38,44 @@ AD_ACCOUNT_FORBIDDEN_CHARS = r'@"/\[]:;|=,+*?<>'
 FIELD = "Domain Controller name"
 
 
+class DcPingStrategy:
+    # Consider staging in the snap for future proof.
+    cmd = "/usr/sbin/realm"
+    arg = "discover"
+
+    async def ping(self, address: str) -> AdDomainNameValidation:
+        cp = await arun_command([self.cmd, self.arg, address], env={})
+        if cp.returncode:
+            return AdDomainNameValidation.REALM_NOT_FOUND
+
+        return AdDomainNameValidation.OK
+
+
+class StubDcPingStrategy(DcPingStrategy):
+    """ For testing purpose. This class doesn't talk to the network.
+        Instead its response follows the following rule:
+
+        - addresses starting with "r" return REALM_NOT_FOUND;
+        - addresses starting with any other letter return OK. """
+    async def ping(self, address: str) -> AdDomainNameValidation:
+        if address[0] == 'r':
+            return AdDomainNameValidation.REALM_NOT_FOUND
+
+        return AdDomainNameValidation.OK
+
+
 class ADController(SubiquityController):
     """ Implements the server part of the Active Directory feature. """
     model_name = "ad"
     endpoint = API.active_directory
     # No auto install key and schema for now due password handling uncertainty.
+
+    def __init__(self, app):
+        super().__init__(app)
+        if self.app.opts.dry_run:
+            self.ping_strgy = StubDcPingStrategy()
+        else:
+            self.ping_strgy = DcPingStrategy()
 
     async def GET(self) -> Optional[ADConnectionInfo]:
         """Returns the currently configured AD settings"""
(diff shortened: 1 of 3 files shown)


subiquity/server/controllers/ad.py
```diff
@@ -37,12 +38,45 @@
...
 
+class DcPingStrategy:
+    # Consider staging in the snap for future proof.
+    cmd = "/usr/sbin/realm"
+    arg = "discover"
+
+    async def ping(self, address: str) -> AdDomainNameValidation:
+        cp = await arun_command([self.cmd, self.arg, address], env={})
+        if cp.returncode:
+            return AdDomainNameValidation.REALM_NOT_FOUND
```
dbungert, line 49: Is the return value granular, meaning there may be other failure types, or is that pretty much the only result?

dbungert (commented):
Looks nice so far.

subiquity/server/controllers/ad.py
CarlosNihelton: A well formed `realm discover` has two possible exit codes: 0 for success and 1 for any kind of failure (from lack of network to DC not found):

https://gitlab.freedesktop.org/realmd/realmd/-/blob/master/tools/realm-discover.c#L154-163

So other failure scenarios would mean a bug in the formation of the command line or an exception raised by `arun_command` if the binary is not found.

Would you say we should consider those in the enum?

dbungert: OK, so if we're only assuming exit codes 0 or 1, this is fine.

> binary is not found

If we plan to sure that realm is present on the system, either by seeded as part of the installer squashes or in the snap, then that binary not found error shouldn't happen and can be an exception.  Same for bug in the formation of the command line.

This all looks good to me.

jibel: If realm or a required binary is not available, then we shouldn't show the option to enable AD in the installer.

dbungert: That could be done, if desired.   https://github.com/canonical/subiquity/blob/main/DESIGN.md?plain=1#L242

CarlosNihelton: That means the client should `GET /active_directory` from this page to decide whether the `Use Active Directory` flag should be enabled or not, isn't it? I was considering implementing an endpoint for that, but raising is even easier, if that's the idiomatic approach, I'm ok with doing so.

image

CarlosNihelton: (As a side note I only see `raise Skip` in the client controllers, but not on the server ones).

dbungert: We did something like this recently, I'll get back to you.

subiquity/server/controllers/ad.py
```diff
@@ -37,12 +38,45 @@
...
+    """ For testing purpose. This class doesn't talk to the network.
+        Instead its response follows the following rule:
+
+        - addresses starting with "r" return REALM_NOT_FOUND;
+        - addresses starting with any other letter return OK. """
+    async def ping(self, address: str) -> AdDomainNameValidation:
+        if address[0] == 'r':
+            return AdDomainNameValidation.REALM_NOT_FOUND
+
+        return AdDomainNameValidation.OK
```
CarlosNihelton, line 64: @jpnurmi this is what you've requested in the first meeting (based on the Ubuntu Pro mocking) isn't it?

jpnurmi: yes, this is perfect. :) thanks!

subiquity/server/controllers/ad.py
dbungert: here's the bit that I was thinking of - commit fe4b6fde6fe6eb29db172c4fe9cb013d55ab5da5 - though that's client initiated so maybe not so similar.

If you want an endpoint the client can call to know if the screen can be shown then I think that's fine.

CarlosNihelton: I don't have strong opinion about this. @jpnurmi what do you think? We are discussing how the client would know ahead of time that the live system has the dependencies to configure AD, so it could disable the `Use Active Directory` check box accordingly. Options are:

1. `GET /active_directory` (before showing the active directory page) raises `Skip` if lacking support;
2. `GET /active_directory/has_support` (one extra endpoint - better naming suggestion is accepted) returns false if lacking support.

CarlosNihelton: I'll keep that in mind for my next PR that will address those small UX enhancements:

1. new endpoint to check ahead of time if AD is supported
2. auto discover the domain in the controller start

dbungert (approved):

PR #1565, opened 2023-02-16
debconf-selections: pass to curtin's apt-config rather than curthooks

ogayot:
We got multiple bug reports in the past stating that the installer is not honoring some part of the apt settings; either supplied by means of autoinstall (e.g., pinning) or via the subiquity UI (e.g., proxy).

What happens under the hood is that curtin overwrites the APT settings as part of the curthooks stage ; effectively discarding earlier settings applied in the apt-config stage.

Curtin does so because we pass debconf_selections directives to curthooks. In the past, curtin used to handle debconf_selections separately but nowadays it considers that they are part of the APT config. As a result, it decides to run apt-config again (but with a close to empty configuration) as part of the curthooks stage.

We now pass debconf_selections as part of the apt-config stage. This should hint curtin not to run apt-config again as part of curthooks.

I think when this is merged, we should revert the following patches:

https://github.com/canonical/curtin/commit/bfbba202e2cc3b02e4e3953081effb6768da41a8
https://github.com/canonical/curtin/commit/b49f5b1505d6597a5c64e3220bb45a291d9a78f7

Requires some testing but I think this should help on the long run. I want to get rid of this issue once and for all.

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -19,7 +19,7 @@ import functools
 import json
 import logging
 import os
-from typing import Set
+from typing import Any, Dict, Set
 import uuid
 import yaml
 
@@ -159,6 +159,10 @@ class DebconfSelectionsModel:
         self.selections = ''
 
     def render(self):
+        return {}
+
+    def get_apt_config(
+            self, final: bool, has_network: bool) -> Dict[str, Any]:
         return {'debconf_selections': {'subiquity': self.selections}}
 
 
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -223,9 +223,9 @@ python3 scripts/check-yaml-fields.py $tmpdir/var/log/installer/subiquity-curtin-
         apt.preferences[0].pin-priority=200 \
         apt.preferences[0].pin='"origin *ubuntu.com*"' \
         apt.preferences[1].package='"python-*"' \
-        apt.preferences[1].pin-priority=-1
+        apt.preferences[1].pin-priority=-1 \
+        apt.debconf_selections.subiquity='"eek"'
 python3 scripts/check-yaml-fields.py "$tmpdir"/var/log/installer/curtin-install/subiquity-curthooks.conf \
-        debconf_selections.subiquity='"eek"' \
         storage.config[-1].options='"errors=remount-ro"'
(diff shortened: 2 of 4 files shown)


ogayot:
This will not work without first fixing a regression introduced in 2016 by canonical/curtin@1257a38f572771e76ee66b8e7484071c9c77d37f

The regression prevents the following code from ever running the `else` branch, because `translate_old_apt_features` ensures `cfg["apt"]` always has a value.

```python
def do_apt_config(cfg, target):
    cfg = apt_config.translate_old_apt_features(cfg)
    apt_cfg = cfg.get("apt")
    if apt_cfg is not None:
        LOG.info("curthooks handling apt to target %s with config %s",
                 target, apt_cfg)
        apt_config.handle_apt(apt_cfg, target)
    else:
        LOG.info("No apt config provided, skipping")
```
So even without debconf_selections, curtin still decides to run apt_config again.

(back then `cfg["apt"]` was named `cfg["apt_source"]`)

ogayot:
Tested with this patch + the curtin apt and that works :+1: . Curtin stops overwriting apt configuration.

dbungert (commented):
The plan sounds reasonable but I want to digest the curtin side more.

subiquity/models/subiquity.py
```diff
@@ -159,6 +159,10 @@ def __init__(self):
         self.selections = ''
 
     def render(self):
+        return {}
```
mwhudson, line 162: So for a while we've been moving away from the binary install/postinstall model distinction and this just moves it further. No action required for this PR -- but I guess this method name is now too generic and we should at some point change it, or change it to something that takes an argument or something.

mwhudson (approved):
this looks fine, we should keep pushing on removing the install/postinstall binary as opportunities come up!

subiquity/models/subiquity.py
ogayot: Agreed, render() and serialize() are the two serialization functions that I always mix up because of how vague their name are.

Currently, we have:
* `model.make_cloudconfig`
* `model.render` (targets curtin install)
* `model.get_apt_config` (targets curtin apt-config)

* `controller.make_autoinstall`
* `controller.serialize` (for the refresh / restart)

IMO the ones that start with `make_` have better names.

mwhudson: Yeah serialize is far too generic a name for the specific and obscure thing those methods do!

dbungert (approved):

PR #1566, opened 2023-02-17
tests: patch parameterized to handle async

dbungert:
parameterized async tests run into cpython bug gh-101486

We use parameterized.expand, which works by creating new functions and inserting those into the list of tests.  It's a wonder this worked at all before for the async tests.

Update the function generation to create a coroutine function, if appropriate.

LP: #2007554


ogayot (approved):
Great work, thanks!

mwhudson (approved):

PR #1567, opened 2023-02-17
source: honor source if hinted by autoinstall but interactive

ogayot:
If a specific source ID is set via autoinstall but the source controller is marked interactive, the UI would select the default source (i.e., the first one on the list) instead of the one matching the ID specified.

This happens because the matching between the ID and the sources loaded was done in apply_autoinstall_config ; which is only called in non-interactive scenarios.

By moving this matching in the start coroutine instead, the right source is also selected when the source controller is interactive.

Test-case:
```bash
$ cat autoinstall-user-data-test 
interactive-sections: ["source"]
source:
  id: ubuntu-server-minimal
  search_drivers: true
version: 1
```
```diff
diff --git a/Makefile b/Makefile
index 209d6ecce..ce0a7e852 100644
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ PROBERTDIR=./probert
 PROBERT_REPO=https://github.com/canonical/probert
 DRYRUN?=--dry-run --bootloader uefi --machine-config examples/simple.json \
        --source-catalog examples/install-sources.yaml \
-       --postinst-hooks-dir examples/postinst.d/
+       --postinst-hooks-dir examples/postinst.d/ --autoinstall autoinstall-user-data-test
 SYSTEM_SETUP_DRYRUN?=--dry-run
 export PYTHONPATH
 CWD := $(shell pwd)
```

### DIFF
--- a/subiquity/server/controllers/source.py
+++ b/subiquity/server/controllers/source.py
@@ -93,13 +93,9 @@ class SourceController(SubiquityController):
         self.model.search_drivers = data.get("search_drivers", True)
 
         # At this point, the model has not yet loaded the sources from the
-        # catalog. So we store the data and lean on apply_autoinstall_config.
+        # catalog. So we store the ID and lean on self.start to select the
+        # current source accordingly.
         self.ai_source_id = data.get("id")
-
-    async def apply_autoinstall_config(self) -> None:
-        if self.ai_source_id is None:
-            return
-        self.model.current = self.model.get_matching_source(self.ai_source_id)
 
     def start(self):
         path = '/cdrom/casper/install-sources.yaml'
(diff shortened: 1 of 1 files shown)


dbungert (approved):
This is definitely an improvement, but is this an every controller problem for interactive-sections?

ogayot:
> This is definitely an improvement, but is this an every controller problem for interactive-sections?

No no, it seems to be specific to the source controller. Or at least most controllers behave as expected.
I've done a bunch of autoinstalls recently with all sections as interactive and the autoinstall directives were pretty consistently honored.
To some extent, the experience in the SSH screen also fails to reflect the autoinstall directives fully (the imported keys are not shown), but it's a bit harder to adjust, and probably less of a problem.

mwhudson (approved):
Thanks. Annoying that this stuff is so fiddly, makes me feel we haven't quite got the design of this stuff right. Please land this though :-)

PR #1568, opened 2023-02-17
Active Directory API UX Enhancements

CarlosNihelton:
As discussed in #1563 two small enhancements are presented:

1. Automatic domain discovery through the network (powered by realmd) and
2. An endpoint to allow clients to know whether AD is supported in the live system.

The current plans require the existance of the realmd package to discover the domain, thus checking whether the required binary is accessible is enough to represent the AD support.

It is expected that clients should GET this endpoint prior to showing the AD configuration page.

### DIFF
--- a/subiquity/server/controllers/ad.py
+++ b/subiquity/server/controllers/ad.py
@@ -13,12 +13,12 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import asyncio
 import logging
 import os
 import re
 from typing import List, Optional, Set
 from subiquitycore.utils import arun_command
+from subiquitycore.async_helpers import run_bg_task
 
 from subiquity.common.apidef import API
 from subiquity.common.types import (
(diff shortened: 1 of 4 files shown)


subiquity/server/controllers/ad.py
```diff
@@ -77,6 +100,15 @@ def __init__(self, app):
         else:
             self.ping_strgy = DcPingStrategy()
 
+    def start(self):
+        if self.ping_strgy.has_support():
+            asyncio.create_task(self._try_discover_domain())
```
dbungert, line 105: Bare create_task should no longer be done, see the note marked "important" at https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task 

`run_bg_task` is available to mitigate this, otherwise keep a reference here

dbungert (changes requested):

subiquity/server/controllers/ad.py
CarlosNihelton: Thanks for pointing that out!

dbungert (approved):

PR #1569, opened 2023-02-18
network: use pyroute2 to manage default routes

dbungert:
The existing event based method of watching for has_network has a flaw.

The incoming route_change events from probert do not distinguish routes on the same interface but a different metric, so if 2 routes on one interface appear, we only get one event.  Then if one of those routes is removed, we will inappropriately remove this route from the default_routes list.

Aside from the code watching the event stream, the set of default routes is an elaborate boolean value.

Simplify the code by passing around a boolean, and when we get a route_change event, use that to go looking again at the list of default routes.

LP: #2004659


subiquitycore/controllers/network.py
```diff
@@ -87,13 +87,31 @@ def route_change(self, action, data):
...
-        elif action == "DEL" and ifindex in self.default_routes:
-            self.default_routes.remove(ifindex)
-        log.debug('default routes %s', self.default_routes)
-        self.controller.update_default_routes(self.default_routes)
+        self.probe_default_routes()
+        self.controller.update_default_routes(self.has_default_routes)
+
+    def _analyze_routes(self, routes):
+        return any(route['table'] == 254 and not route['dst']
+                   for route in routes)
```
mwhudson, line 95: I think we should consider rejecting routes with metrics > 20000 here, to filter out connections that have not passed NM's connectivity check (unless there is a expectation that such routes might be created by some other mechanism? It seems very unlikely in the context of subiquity though?)

mwhudson (approved):
One more thing i think we should consider.

subiquitycore/controllers/network.py
dbungert: I think we should, yes.  These high-metric routes can stick around.  I think it sensible to filter them.  Implemented.

apt-deps.txt
```diff
@@ -35,6 +35,7 @@ python3-pyflakes
 python3-pytest
 python3-pytest-xdist
 python3-pyudev
+python3-pyroute2
```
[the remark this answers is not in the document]
dbungert: oops

dbungert:
> Can we reword the update_default_routes function without breaking the world? It feels weird to me that it takes a boolean value.

Sure, we are already changing the type so I think changing the name is appropriate.  Implemented.

mwhudson (approved):
This is pedantry to a possibly unhelpful level but this is a bit inconsistent about "has_default_route" vs "has_default_routes". Other than that it looks great!

subiquity/server/controllers/network.py
```diff
@@ -246,8 +246,7 @@ async def apply_autoinstall_config(self, context):
             self.apply_config(context)
         with context.child("wait_for_apply"):
             await self.apply_config_task.wait()
-        self.model.has_network = bool(
-            self.network_event_receiver.default_routes)
+        self.model.has_network = self.network_event_receiver.has_default_route
```
mwhudson, line 249: 100% nothing to do with this branch but I think this line is pointless given that configured does the same and is called immediately after apply_autoinstall_config?

PR #1571, opened 2023-02-20
Turns AD validation methods into POST

CarlosNihelton:
Per the request in https://github.com/canonical/subiquity/pull/1553#issuecomment-1435514498, to make it easy on the GUI side, we need to turn the validation methods that were GET into POST.


mwhudson (approved):
Seems a bit gross but sure

PR #1572, opened 2023-02-22
Active Directory - Implement the actual join

CarlosNihelton:
This intends to use curtin to run in-target commands to perform the real joining and to enable `pam_mkhomedir`.

Changing the live session host name is required because `adcli`, the tool invoked by `realm` under the hood, doesn't understand the chroot target. That's the same recipe done in Ubiquity.

To make it easier for client testing, a stub joining strategy is implemented, returning success or error depending upon the domain name initial character.

This also brings a blocking endpoint to get the result of joining. Once called during any moment during the server lifetime, if the action of joining a domain is not yet completed, it will block until a result is available. The actual join action only starts in the `InstallController.post_install` method, since it has to run in the target.


subiquity/common/types.py
```diff
@@ -801,3 +801,11 @@ class AdDomainNameValidation(enum.Enum):
...
     OK = 'OK'
     EMPTY = 'Empty'
+
+
+class AdJoinResult(enum.Enum):
+    OK = 'OK'
+    JOIN_ERROR = 'Failed to join'
+    EMPTY_HOSTNAME = 'Target hostname cannot be empty'
+    PAM_ERROR = 'Failed to update pam-auth'
+    UNKNOWN = "Didn't attempt to join yet"
```
jibel, line 811: Are these strings displayed to the user? If so, they should be translatable.

subiquity/server/ad_joiner.py
```diff
@@ -0,0 +1,117 @@
...
+    ADConnectionInfo,
+    AdJoinResult,
+)
+
+log = logging.getLogger('subiquity.server.ad_joiner')
+
+
+class AdJoinStrategy():
+    realm = "/usr/sbin/realm"
+    pam = "/usr/sbin/path-auth-update"
```
jibel, line 31: typo: pam-auth-update

CarlosNihelton: Thanks for catching that one!

subiquity/common/types.py
CarlosNihelton: Based on the majority of the types definition in the same file, it doesn't seem they are displayed to the user.

dbungert: We are trending toward descriptive strings rather than what enum.auto() is generating, at minimum if you end up debugging at the API layer it's much nicer to look at the values rather than trying to puzzle out what integer maps to the return value.

subiquity/server/ad_joiner.py
```diff
@@ -0,0 +1,117 @@
...
+                                          "--enable", "mkhomedir",
+                                          private_mounts=True)
+
+            if cp.returncode:
+                result = AdJoinResult.PAM_ERROR
+            else:
+                result = AdJoinResult.OK
+
+        # Restoring the live session hostname.
+        cp = await arun_command(['hostname', hostname_current])
```
dbungert, line 71: It's a bit unusual to set a hostname temporarily, but I'll take it for the moment as a requirement.

It would be nice to structure the set hostname / restore hostname in something that is more certain to get restored.  For instance a context manager could be created which sets the hostname, you run your in-targets, then it is responsible for restoring the hostname.

The subiquity codebase has several context managers, journald_subscriptions is a short example that I think is pretty clear.

subiquity/tests/api/test_api.py
```diff
@@ -1696,3 +1696,20 @@ async def test_ad(self):
...
+            result = await instance.post(endpoint, ad_dict)
+            self.assertIsNone(result)
+            join_result_ep = endpoint + '/join_result'
+            # Without wait this shouldn't block but the result is unknown until
+            # the install controller runs.
+            join_result = await instance.get(join_result_ep, wait=False)
+            self.assertEqual('UNKNOWN', join_result)
+            # And without the installer controller running, a blocking call
+            # should timeout since joining never happens.
+            with self.assertRaises(asyncio.exceptions.CancelledError):
```
dbungert, line 1714: Would this be triggered by the test-level timeout?  If more API calls are added after this I think things are going to get weird.  maybe `asyncio.wait_for` instead, to indicate that this specific call is expected to timeout?

mwhudson (commented):
I agree with Dan :-)

CarlosNihelton:
And finally, after 44d29c5, I got a run of the Ubuntu Desktop Installer with successful AD setup (custom snap build on the live session 😉 and with a little help of `curl` because I didn't implement the GUI part).

image

subiquity/server/controllers/ad.py
```diff
@@ -95,12 +97,19 @@ class ADController(SubiquityController):
...
             self.ping_strgy = StubDcPingStrategy()
         else:
             self.ping_strgy = DcPingStrategy()
 
     def start(self):
+        # AD is by nature an optional feature.
+        # Yet, we need to make its model part of the POSTINSTALL_MODEL_NAMES
+        # set, which would turn this into a required controller.
+        # So, let's start this as configured by default.
+        run_bg_task(self.configured())
```
dbungert, line 112: I think you don't want this, and the client should POST to mark it configured instead.
The defect would be, the user intends to use AD, they do most of the install but let the installer screen sit on the AD page.  The `postinstall` method in the `install` controller runs, no AD join is attempted, and then the user inputs join information that won't be processed.
Instead, the client should POST empty connection info if AD is undesired, and that means that before that POST is done, the `postinstall` method will be blocked from running.

dbungert (changes requested):
Overall this is looking fine, I spotted 2 problems though.

subiquity/server/ad_joiner.py
```diff
@@ -0,0 +1,148 @@
...
+
+@contextmanager
+def hostname_context(hostname: str):
+    """ Temporarily adjusts the host name to [hostname] and restores it
+        back in the end of the caller scope. """
+    hostname_current = gethostname()
+    hostname_process = run_command(['hostname', hostname])
+    yield hostname_process
+    # Restoring the live session hostname.
+    hostname_process = run_command(['hostname', hostname_current])
```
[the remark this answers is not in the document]
dbungert: I wasn't sure either, I had to write the above to clarify :)

dbungert (approved):

PR #1573, opened 2023-02-22
Revert "Compare the snaps versions to avoid suggesting a downgrade"

dbungert:
This reverts commit 2f9c7399ca01980443384ecff5d686e8d8c4bcfa.

The original work was implementing a fix for LP: #1936190, but there are legitimate situations where we really do want to refresh to a version number that appears older at a glance.


mwhudson (approved):
Well as you know I've never really liked this so +1 from me :-)

ogayot (approved):

PR #1575, opened 2023-02-24
when installing packages, try the download operation up-to 3 times

ogayot:
When installing a package, instead of using a single call to `curtin system-install [package]`, we now call:
* `curtin system-install --download-only [package]` (up to three times with a delay between iterations)
* `curtin system-install --assume-downloaded [package]`

curtin's merge request: https://code.launchpad.net/~ogayot/curtin/+git/curtin/+merge/437897

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -61,7 +61,7 @@ parts:
 
     source: https://git.launchpad.net/curtin
     source-type: git
-    source-commit: cc23249707a2773ecb1f10b17de35fff4620de9a
+    source-commit: b1f4da3bec92356e8ef389c1c581cfdcd1b36c42
 
     override-pull: |
       snapcraftctl pull
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -426,10 +426,18 @@ class InstallController(SubiquityController):
         name="install_{package}",
         description="installing {package}")
     async def install_package(self, *, context, package):
-        await run_curtin_command(
-            self.app, context, 'system-install', '-t', self.tpath(),
-            '--', package,
-            private_mounts=False)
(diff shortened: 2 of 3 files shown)


ogayot:
Leaving as a draft since it requires an update on the curtin revision.
Also, probably needs a FFE to file.

snapcraft.yaml
```diff
@@ -59,9 +59,9 @@ parts:
   curtin:
     plugin: nil
 
-    source: https://git.launchpad.net/curtin
+    source: https://git.launchpad.net/~ogayot/curtin
```
ogayot, line 62: leaving a note here so that we use the right remove and revision before merging.

subiquity/server/controllers/install.py
```diff
@@ -426,10 +427,32 @@ async def get_target_packages(self, context):
         name="install_{package}",
         description="installing {package}")
     async def install_package(self, *, context, package):
-        await run_curtin_command(
-            self.app, context, 'system-install', '-t', self.tpath(),
-            '--', package,
-            private_mounts=False)
+        """ Attempt to download the package up-to three times, then install it.
+        """
+        for attempt, i in enumerate(reversed(range(3))):
```
dbungert, line 432: this line bugs me, it doesn't scan.  I think it's the mix of using `i` alongside the other.  I'd prefer a clearer name on `i` - maybe `attempts_remaining` or something, or to reduce this to a single variable.

dbungert (approved):
One tweak then let's merge.

subiquity/server/controllers/install.py
ogayot: yep, this is definitely hard on the eyes.. I really want to come up with a retry mechanism we can use at various places, but haven't come up with the right design for it yet.

dbungert (approved):

PR #1577, opened 2023-02-28
filesystem: don't offer tiny partitions / disks as potential PVs

ogayot:
When creating a virtual group, all disks and partitions that could be used as a raid device were also offered as a potential LVM physical volume.

However, tiny disks and partitions (i.e., < 1 MiB) are too small to store the LVM header. They make the get_lvm_size() function potentially return a negative value; which in turn causes math errors.

We now make sure to only offer disks/partitions that are big enough to store the LVM header.

Technically speaking, a 2MiB disk is still useless as a physical volume since it is not big enough to store any PE (which are 4MiB by default). We can come up with better rules later but this is good enough to prevent subiquity from crashing.

Steps to reproduce the crash:
1. `make dryrun`
2. navigate to the filesystem screen
3. select "Custom storage layout" and next
4. select the free space -> add a GPT partition
5. make the size `0G` and format: `Leave unformatted'
6. redo 4. and 5. to create a second partition
7. click on `Create volume group (LVM)`
8. try to add one of the two partition

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -688,7 +688,9 @@ class Disk(_Device):
             return False
         return True
 
-    ok_for_lvm_vg = ok_for_raid
+    @property
+    def ok_for_lvm_vg(self):
+        return self.ok_for_raid and self.size > LVM_OVERHEAD
 
     @property
     def model(self):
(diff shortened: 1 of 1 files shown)


dbungert (approved):

dbungert:
You said this had to do with using a seed ISO, right? I think this is the second bug that has to do with trying to treat a seed ISO as a real disk.  I think we also want to adjust probert to not show a seed ISO as a disk we could actually do disk operations on - not required for this PR.

ogayot:
> You said this had to do with using a seed ISO, right? I think this is the second bug that has to do with trying to treat a seed ISO as a real disk. I think we also want to adjust probert to not show a seed ISO as a disk we could actually do disk operations on - not required for this PR.

Correct. I originally witnessed the issue with a seed ISO.
The disk would show up as "already formatted as iso9660, not mounted", but it can be formatted, used as a LVM PV or as a raid.

PR #1578, opened 2023-02-28
keyboard: validate layout and variant

dbungert:
In LP: #2008271, an invalid but reasonable-looking layout was supplied in autoinstall, which makes it all the way to curthooks before failing ckbcomp in a way that makes it look like files are missing.

Validate ahead of time these values.

### DIFF
--- a/subiquity/models/keyboard.py
+++ b/subiquity/models/keyboard.py
@@ -20,7 +20,11 @@ import os
 import yaml
 
 from subiquity.common.resources import resource_path
-from subiquity.common.types import KeyboardSetting
+from subiquity.common.serialize import Serializer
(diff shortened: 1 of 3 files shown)


ogayot (approved):
LGTM, thanks!

Suggestion for future work: if you need to move code around (especially across files) and modify the code at the same time, check if you can make separate commits.

It feels easier to review what changed if we can exclude the "move" commits.

dbungert:
> if you need to move code around (especially across files) and modify the code at the same time, check if you can make separate commits

Great idea, will do next time.  It's half this PR :)

dbungert:
@jpnurmi FYI

PR #1581, opened 2023-03-01
Active Directory - Partial autoinstall support

CarlosNihelton:
This aims to allow Subiquity to accept the domain and admin names, but not the password, because that is a very sensitive information and require a careful treatment. That implies in the AD controller always being interactive.
A integration testing case is added to verify that the packages expected by the AD Model are installed in the target system if the AD config is present in the autoinstall file.


subiquity/server/controllers/ad.py
```diff
@@ -91,9 +91,49 @@ def has_support(self) -> bool:
...
+        info = self.model.conn_info
+        if info is None:
+            return None
+
+        return {'admin_name': info.admin_name, 'domain_name': info.domain_name}
+
+    def deserialize(self, state):
+        if state is None:
+            return
+        if 'admin_name' in state and 'domain_name' in state:
```
dbungert, line 121: serialize/deserialize is part of how Subiquity self-update works, or any other service restart.   Having a partial serialize is a bit odd.  While we have generally added serialize/deserialize to new controllers it isn't actually required, and other common ones like identity and filesystem that handle sensitive information don't provide it.

Probably better to just not have serialize/deserialize methods, and if a restart happens subiquity will just show the screen again.

subiquity/server/controllers/ad.py
```diff
@@ -91,9 +91,49 @@ def has_support(self) -> bool:
 
 class ADController(SubiquityController):
     """ Implements the server part of the Active Directory feature. """
-    model_name = "ad"
     endpoint = API.active_directory
     # No auto install key and schema for now due password handling uncertainty.
+    autoinstall_key = model_name = "ad"
+    autoinstall_schema = {
```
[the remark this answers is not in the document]
CarlosNihelton: Good point!

autoinstall-schema.json
```diff
@@ -461,6 +461,18 @@
                 "additionalProperties": false
             }
         },
+        "ad": {
```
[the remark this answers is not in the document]
CarlosNihelton: I did some other renamings for consistency. Classes now always prefixed by 'Ad' (no uppercase D) and fields renamed to the longer form.

dbungert:
I want to create the feature freeze branch.  It will be OK to merge the AD work over to it as we wrap up the details.

CarlosNihelton:
> I want to create the feature freeze branch. It will be OK to merge the AD work over to it as we wrap up the details.

Unless you spot any issue in the code, I'm done for now.

autoinstall-schema.json
```diff
@@ -461,6 +461,18 @@
                 "additionalProperties": false
             }
         },
+        "active_directory": {
```
dbungert, line 464: Most autoinstall things are using dash instead of underscore.  `search_drivers` and `preserve_sources_list` are  exceptions but I want to fix them now.  In the interest on standardizing autoinstall, I'm going to ask you to use a dash.

dbungert (commented):
Last minute autoinstall ask then let's merge.

autoinstall-schema.json
dbungert: Thanks - but examples/autoinstall-ad.yaml needs the tweak, and tests passed?  Are the packages being installed unconditionally?

CarlosNihelton: Well observed. There is a logic error in the test. I'm finishing running the integration tests locally to confirm the fix.

CarlosNihelton: Commit 121cf3a18ac21503e9aa5d12e54deb364773852f reveals the logic error and testing against it would crash. The following aeac073dca0d1998ec1b394fc17a0b3a5e8ef8ea fixes the autoinstall file and should not crash the integration test.

CarlosNihelton: Should we consider exposing the `variant` for the `autoinstall`?

dbungert: I'm not sure what problem exposing the variant would solve.  If this is required in autoinstall then we have made a critical error.

If someone wants to configure active-directory, then they should do so.  If not, the omit the section.

CarlosNihelton: Currently "active_directory" is only available for the desktop variant (because only for that variant we add the model as part of the POST_INSTALL_NAMES set, what causes the extra packages to be installed in the target).

But you're right. That condition is an implementation detail that will be sorted out when the server variant also becomes interested in joining an Active Directory domain.

scripts/runtests.sh
```diff
@@ -254,6 +254,28 @@ python3 scripts/check-yaml-fields.py "$tmpdir"/var/log/installer/autoinstall-use
...
+    --autoinstall examples/autoinstall-ad.yaml \
+    --kernel-cmdline autoinstall \
+    --source-catalog examples/mixed-sources.yaml &
+sub_pid=$!
+until [[ -S "$tmpdir/socket" || $(ps -p "$sub_pid" > /dev/null ) == 1 ]]
+do
+    sleep 1
+done
+curl --unix-socket "$tmpdir/socket" -X POST 'http://localhost/meta/client_variant?variant=%22desktop%22'
+ad_info=$(curl --unix-socket "$tmpdir/socket" -X GET 'http://localhost/active_directory' | jq '.password="PassWord"')
```
dbungert, line 272: passw0rd is our standard canary value for the value we don't want to find in a file

scripts/runtests.sh
```diff
@@ -254,6 +254,28 @@ python3 scripts/check-yaml-fields.py "$tmpdir"/var/log/installer/autoinstall-use
...
+# User and domain name (but not passwd) should come from the autoinstall file.
+curl --unix-socket "$tmpdir/socket" -X POST -d "${ad_info}" 'http://localhost/active_directory'
```
dbungert, line 274: This is what our API tests looked like at first.  I suspect driving this test from there is going to be simpler.  I want this PR done but I want to see you follow up with this moved to an API test.

scripts/runtests.sh
```diff
@@ -254,6 +254,28 @@ python3 scripts/check-yaml-fields.py "$tmpdir"/var/log/installer/autoinstall-use
...
+wait $sub_pid
```
dbungert, line 275: If the test fails to complete this is going to hang CI

scripts/runtests.sh
```diff
@@ -254,6 +254,28 @@ python3 scripts/check-yaml-fields.py "$tmpdir"/var/log/installer/autoinstall-use
...
+LANG=C.UTF-8 timeout --foreground 60 \
+    python3 -m subiquity.cmd.tui \
+    --dry-run \
+    --output-base "$tmpdir" \
+    --machine-config examples/simple.json \
+    --autoinstall examples/autoinstall-ad.yaml \
+    --kernel-cmdline autoinstall \
+    --source-catalog examples/mixed-sources.yaml &
+sub_pid=$!
+until [[ -S "$tmpdir/socket" || $(ps -p "$sub_pid" > /dev/null ) == 1 ]]
```
dbungert, line 267: This construction with ps does not correctly screen for the pid disappearing.  you're always comparing an empty string to 1.

dbungert (changes requested):

scripts/runtests.sh
dbungert: or move it there now if that's simpler

CarlosNihelton: Simpler to move it now, indeed. That results in code much better to reason about. The only gotcha is that I needed to expose the `output_base` dir in the testing Server class, so I could inspect the logs as part of the test case.

subiquity/tests/api/test_api.py
```diff
@@ -1714,3 +1720,73 @@ async def test_ad(self):
...
+                                             wait=True)
+                self.assertEqual('OK', join_result)
+                packages = await self.packages_lookup(logdir)
+                for k, v in packages.items():
+                    print(f"Checking package {k}")
+                    self.assertTrue(v, f"package {k} not found in the target")
+
+        # By the time we reach here the server already exited and the context
+        # manager will fail to POST /shutdown.
+        except aiohttp.client_exceptions.ClientOSError:
```
dbungert, line 1791: Hah, that sounds like something we fix in the API tests in general, but this is entirely fine to work around here.

dbungert (approved):

scripts/runtests.sh
dbungert: Thank you.  And no problem on exposing the output_base value, that's a reasonable change for the test you had in mind.

PR #1584, opened 2023-03-03
apt: force downloads to be unsandboxed when running mirror testing

ogayot:
When apt-get is run as root, it tries to drop privileges by setuid-ing to the APT::Sandbox::User user (i.e., _apt by default). This does not work for us when doing mirror testing because the default sandbox user does not have access to the overlay. Therefore, APT produces a warning and reverts to unsandboxed downloads.

Let's force apt to use unsandboxed downloads by setting root as the sandbox user. This has the same result but avoids showing a warning in the APT output during mirror testing.


mwhudson (approved):
This just removes some scary warnings yes? Or does the update fail without this?

dbungert (approved):

PR #1585, opened 2023-03-03
Active Directory - Split the domain validation endpoint in two

CarlosNihelton:
Ref: https://github.com/canonical/ubuntu-desktop-installer/pull/1522#discussion_r1124397775

GUI could face performance issues in calling the validation on each char the user type.

Having two separate endpoints prevent such issue to ever happen, because we provide more granular control. Regex validation goes at each char typed and pinging the domain controller can happen when whe user leaves the text field or hits the "Test Connection" buttont.

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -426,8 +426,15 @@ class API:
             def GET() -> bool: ...
 
         class check_domain_name:
+            """ Applies basic validation to the candidate domain name,
+                without calling the domain network. """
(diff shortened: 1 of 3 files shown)


dbungert (approved):
I'm fine with this but I wonder about the slight inconsistency in the return type.

CarlosNihelton:
> I'm fine with this but I wonder about the slight inconsistency in the return type.

We could return a list for the sake of enhanced consistency, but the list will always be of one single element. I think we are fine as is.

PR #1579, opened 2023-03-01
Support LUKS with 'lvm' autoinstall storage layout

rmounce:
To support the same behaviour that was available with d-i/preseed.

PR #1587, opened 2023-03-03
runtests: test for autoinstall w/ lvm password

dbungert:
In #1579 we enabled autoinstall to supply a LVM passphrase.  Add a test case looking for that passphrase.


mwhudson (approved):
Why did I never review this? It looks fine (although it conflicts now, unsurprisingly)

PR #1588, opened 2023-03-05
Add 'lvm_use_all_space' autoinstall storage layout

rmounce:
Follow-up to https://github.com/canonical/subiquity/pull/1579

The guided LVM layout assumes that by using it rather than the 'direct' layout, the user will want some free space in the VG to create their own LVs. Often, LVM is only used as it's the supported option for LUKS.

Add a new 'lvm_use_all_space' layout with similar behaviour to the old d-i/preseed 'crypto' method.

Related to: https://bugs.launchpad.net/subiquity/+bug/1913986

### DIFF
--- a/documentation/autoinstall-reference.md
+++ b/documentation/autoinstall-reference.md
@@ -278,11 +278,14 @@ Storage configuration is a complex topic and the description of the desired conf
 
 #### Supported layouts
 
-The two supported layouts at the time of writing are "lvm" and "direct".
+The supported layouts at the time of writing are "lvm", "lvm_use_all_space" and "direct".
 
     storage:
       layout:
         name: lvm
+    storage:
+      layout:
+        name: lvm_use_all_space
     storage:
       layout:
         name: direct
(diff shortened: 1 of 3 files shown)


dbungert:
Thanks for your work on this.

In the end I wanted a different syntax on the autoinstall portion, due to the belief that we will add other volume managers and have the same problem.  So I did https://github.com/canonical/subiquity/pull/1630 which generalizes the solution a bit and handles some other things I needed for ubuntu-desktop-installer.

I have listed your as a co-auther on that work.

I plan to finalize this work this week for inclusion in a Subiquity 23.04.1 release.

xnox:
it has been an explicit design decision to not allow full use of VG in subiquity.

separately, many ubuntu deployments today do LUKS+ext4 without lvm in the mix.

If the end goal is to just have LUKS+ext4 we should just add that as a partitioning option, without forcing LVM onto people, which has been harming us. For example - FDE encrypted Desktop, FDE encrypted Ubuntu Core, Azure CVM all are just LUKS+ext4 deployments, which we should support in subiquity as well.

If LUKS+ext4 exists in subiquity, would you still need this option to use LVM VG in full?

xnox (changes requested):
Needs information w.r.t. prior design choices & future roadmap

xnox:
Also, whilst https://github.com/canonical/subiquity/pull/1579 is basically bringing parity with UI and old installers, but still wondering if LUKS+ext4 would have been a better choice.

dbungert:
@xnox and I had a chat about this, I have landed the work as seen in PR #1630/#1633 but I want to pursue his suggestion in the future.

Closing this PR, as the work has been done with a slightly modified implementation.

PR #1589, opened 2023-03-07
storage/v2: endpoint for boot-plausible disks

dbungert:
Provide a list of disk IDs that are plausible boot targets.
This could also be full disk objects if that makes things simpler.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -433,7 +433,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             config, self.model._probe_data['blockdev'], is_probe_data=False)
         await self.configured()
 
-    def get_guided_disks(self, check_boot=True, with_reformatting=False):
+    def potential_boot_disks(self, check_boot=True, with_reformatting=False):
         disks = []
         for raid in self.model._all(type='raid'):
             if check_boot and not boot.can_be_boot_device(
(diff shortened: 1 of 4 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -731,6 +731,11 @@ async def v2_reformat_disk_POST(self, data: ReformatDisk) \
         self.reformat(self.model._one(id=data.disk_id), data.ptable)
         return await self.v2_GET()
 
+    async def v2_potential_boot_disks_GET(self) -> List[str]:
+        disks = self.get_guided_disks(check_boot=True,
+                                      with_reformatting=False)
+        return [disk.id for disk in disks]
```
mwhudson, line 737: I wonder if get_guided_disks is the right name for the method called here, I thought it might do a size check. Maybe it should be potential_boot_disks!

mwhudson (approved):
LGTM, one bikesheddy name suggestion

subiquity/server/controllers/filesystem.py
dbungert: Right, in practice the size checks are not done by get_guided_disks, but other guided infrastructure.  I like the suggestion to rename it.

jpnurmi:
Thanks, LGTM. A list of IDs is sufficient and well-suitable :+1:

PR #1590, opened 2023-03-08
storage: lengthen and log probing time

dbungert:
A reasonably common type of bug report is hitting the block probing timeout.  That timeout is at 15 seconds today.  Extend that to 90 seconds, and log the time it takes.

As a performance exercise, it would be good to know where that time is being spent and why it takes 10x longer or more for some people.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -20,8 +20,8 @@ import json
 import logging
 import os
 import pathlib
-import platform
 import select
+import time
 from typing import Dict, List, Optional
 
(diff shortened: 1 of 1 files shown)


mwhudson (approved):
I mean, OK. Are we sure that probes that aren't complete after 15 seconds are ever going to complete?

The probert logs should give some indication where the time is going, I hope?

dbungert:
> I mean, OK. Are we sure that probes that aren't complete after 15 seconds are ever going to complete?

I created an ugly partition layout on real hardware with multiple filesystems and various degrees of stuff in them, and was able to see probing take 6.3 seconds.  I think doubling that number is realistic.

PR #1591, opened 2023-03-10
allow autoinstalls to use the disk the installer booted from

mwhudson:
Currently subiquity ignores any device that has a mounted partition when enumerating block devices. This is done to exclude the install media, which is something we are going to need to be more nuanced about.

This fairly simpleminded change does not exclude these devices when scanning the probe data but instead filters them out before presenting them to the UI. This should allow an autoinstall file to install to a gap on the install media, at least. It's a bit hard to know that it never shows the install media in any corner of the UI though.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1124,13 +1124,14 @@ class FilesystemModel(object):
         if self._probe_data is not None:
             self._orig_config = storage_config.extract_storage_config(
                 self._probe_data)["storage"]["config"]
-            self._actions = self._actions_from_config(
+            self._actions, self._exclusions = self._actions_from_config(
                 self._orig_config,
                 self._probe_data['blockdev'],
                 is_probe_data=True)
         else:
             self._orig_config = []
             self._actions = []
+            self._exclusions = set()
         self.swap = None
         self.grub = None
         self.guided_configuration = None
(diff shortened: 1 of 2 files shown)


mwhudson:
@dbungert what do you think? too simple? too hard to be sure is correct?

dbungert:
> @dbungert what do you think? too simple? too hard to be sure is correct?

I think the change is correct, but as written I'm worried we are going to forget about the nuance in a few months.

I think the callers that are interested in this info want more explicit information.  Some cases want the excluded devices, some do not, and in the future we would want to show this to the UI.  That to me sounds like a new keyword on the matcher so that we can more visibly opt-in the cases that want the devices that are filtered today, where the default of the matcher retains current behavior.

The `all()` matcher is sounding a bit unfortunately named.

mwhudson:
Yeah I think you're right. What is the path forward to being able to make this sort of change and being confident that we won't break this? One think that might be easier is to attach information to objects rather than carrying these "excluded" sets around.

It's only tangentially related but I think moving the FilesystemModel methods/functionallity that is only used by the tui client into some subclass that is in subiquity.client would be an improvement.

mwhudson:
> It's only tangentially related but I think moving the FilesystemModel methods/functionallity that is only used by the tui client into some subclass that is in subiquity.client would be an improvement.

Turns out this isn't very interesting -- it's more or less only load_server_data and all_devices that are client only. Hmm.

dbungert:
I think a comment blob to get this started is fine, then we can land this and make tweaks.

dbungert (approved):
Good to merge, let's document though.

mwhudson:
Yes we just need to be very careful reviewing any code that gets in the neighbourhood of iterating or searching through "all devices" until we make this more elegant.

PR #1593, opened 2023-03-11
Single instance task wait

dbungert:
LP: #[2009797](https://bugs.launchpad.net/subiquity/+bug/2009797) has the following backtrace:

```
2023-03-09 06:54:49,973 DEBUG subiquity.server.server:456 request to /storage/v2?wait=true crashed
  File "/snap/ubuntu-desktop-installer/820/bin/subiquity/subiquity/common/api/server.py", line 134, in handler
  File "/snap/ubuntu-desktop-installer/820/bin/subiquity/subiquity/server/controllers/filesystem.py", line 656, in v2_GET
    return await self.get_v2_storage_response(self.model, wait)
  File "/snap/ubuntu-desktop-installer/820/bin/subiquity/subiquity/server/controllers/filesystem.py", line 642, in get_v2_storage_response
    probe_resp = await self._probe_response(wait, StorageResponseV2)
  File "/snap/ubuntu-desktop-installer/820/bin/subiquity/subiquity/server/controllers/filesystem.py", line 389, in _probe_response
    await self._start_task
```

What's happening here is that we lost the race and the installer client asked for filesystem information before the controller itself ran the `start()` method.

While working on that, I realized the behavior around `SingleInstanceTask.wait()` was more complicated than I liked for the callers, and improving that will help the fix for the above.

SingleInstanceTask has distinct steps for creation of the object, and starting the task.  If a different coroutine is waiting on the SingleInstanceTask, it isn't safe to directly call SingleInstanceTask.wait() as the task may or may not have been created yet.
    
Existing code usage of SingleInstanceTask is in 4 categories, with regards to SingleInstanceTask.wait():
1) using SingleInstanceTask without using SingleInstanceTask.wait(). This is unchanged.
2) using SingleInstanceTask.wait without a check on task is not None. This may be safe now, but is fragile in the face of innocent-looking refactors around the SingleInstanceTask.
3) using SingleInstanceTask.wait after confirming that the task is not None.  This is fine but a leaky abstraction.
4) directly waiting on the SingleInstanceTask.task.  Another leaky abstraction, but it's solving a cancellation problem.  Leaving this alone.
    
By enhancing SingleInstanceTask.wait(), cases 2 and 3 are improved.  The code not checking the task today is made safer, and the code checking the task today can be simplified.

Marking some tests async is about giving the FilesystemController an active event loop.

### DIFF
--- a/subiquitycore/tests/test_async_helpers.py
+++ b/subiquitycore/tests/test_async_helpers.py
@@ -56,10 +56,13 @@ class TestSITWait(unittest.IsolatedAsyncioTestCase):
         sit = SingleInstanceTask(fn)
         await sit.start()
         await asyncio.wait_for(sit.wait(), timeout=1.0)
+        self.assertTrue(sit.done())
 
     async def test_wait_not_started(self):
         async def fn():
             self.fail('not supposed to be called')
         sit = SingleInstanceTask(fn)
+        self.assertFalse(sit.done())
         with self.assertRaises(asyncio.TimeoutError):
             await asyncio.wait_for(sit.wait(), timeout=0.1)
+        self.assertFalse(sit.done())
--- a/subiquitycore/tests/test_async_helpers.py
+++ b/subiquitycore/tests/test_async_helpers.py
@@ -45,3 +45,21 @@ class TestSingleInstanceTask(unittest.IsolatedAsyncioTestCase):
         sit.task.cancel()
         self.assertEqual(expected_call_count, mock_fn.call_count)
         self.assertEqual(cancel_restart, restarted)
+
+
+# previously, wait() may or may not have been safe to call, depending
+# on if the task had actually been created yet.
+class TestSITWait(unittest.IsolatedAsyncioTestCase):
+    async def test_wait_started(self):
+        async def fn():
+            pass
+        sit = SingleInstanceTask(fn)
+        await sit.start()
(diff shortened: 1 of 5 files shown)


mwhudson:
This feels like an improvement. It's probably super unrealistic but it almost feels like we should attempt some kind of formal correctness proof for SingleInstanceTask...

ogayot (approved):
LGTM. This was one of the reason why I've avoided relying on SingleInstanceTask so far.

Another reason is the potential busy waiting in `SingleInstanceTask.wait()` if the task gets cancelled without restart. But this is for another time.

subiquitycore/tests/test_async_helpers.py
```diff
@@ -45,3 +45,24 @@ async def fn():
...
+        await sit.start()
+        await asyncio.wait_for(sit.wait(), timeout=1.0)
+        self.assertTrue(sit.done())
+
+    async def test_wait_not_started(self):
+        async def fn():
+            await asyncio.sleep(5)
+        sit = SingleInstanceTask(fn)
+        self.assertFalse(sit.done())
+        with self.assertRaises(TimeoutError):
```
ogayot, line 66: Until python 3.11, `asyncio.TimeoutError` and `TimeoutError` were two distinct types.

Using `asyncio.TimeoutError` should be backward compatible I think.

dbungert: Thank you!

PR #1594, opened 2023-03-13
integration: add background runner script

dbungert:
To eventually run these in parallel, we want to not have a single test instance hogging the console.  Add a background-runner script which uses a detached screen session that also preserves the exit code of the program we're testing.


mwhudson (approved):
seems to work and seems like progress!

PR #1595, opened 2023-03-14
filesystem: ensure probe wait handle start timing

dbungert:
In LP: 2009797, an exception of this form happens:

The installer client, u-d-i, is asking for storage information ASAP after the socket starts listening, and in this case that happened before all controllers were started.  The sync primitive the probe is waiting on wasn't created yet.

Add an event earlier in sequence to ensure that the probe has something to wait on, even before the controller is officially `start()`ed.


mwhudson:
Wait we set up the API server before all calls to the controller's start() methods (which are synchronous)? Is that deliberate or a mistake I wonder...

mwhudson:
oh right, we need to serve the /meta/status API early. I wonder if there's any generic way we can cause all api calls other than that one to block until we are passed more of the initialization

dbungert:
#1596 done instead

PR #1596, opened 2023-03-15
api: delay most endpoints until controllers start

dbungert:
In LP: 2009797, an exception of this form happens:

The installer client, u-d-i, is asking for storage information ASAP after the socket starts listening, and in this case that happened before all controllers were started.  The sync primitive the probe is waiting on wasn't created yet.

With one known exception, /meta/status, we really shouldn't be responding to random API calls, and the startup sequence of the controllers should be relatively quick (sub 1 second to be sure). Just delay them, except for the special one.

A more aggressive version of https://github.com/canonical/subiquity/pull/1595

### DIFF
--- a/subiquity/common/api/server.py
+++ b/subiquity/common/api/server.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import inspect
+import logging
 import os
 import traceback
 
@@ -22,6 +23,8 @@ from aiohttp import web
 from subiquity.common.serialize import Serializer
 
 from .defs import Payload
+
+log = logging.getLogger('subiquity.common.api.server')
 
 
 class BindError(Exception):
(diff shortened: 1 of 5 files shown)


subiquity/common/api/server.py
```diff
@@ -131,6 +134,18 @@ async def handler(request):
...
+                    # Most endpoints should not be responding to requests
+                    # before the controllers have started, that's just bound to
+                    # be a big pool of timing bugs that we want nothing to do
+                    # with.  A few chosen methods should be safe, so allow
+                    # those to opt-in.  Everybody else blocks until the
+                    # controllers complete their start().
+                    if not controller.app.controllers_have_started.is_set():
+                        log.debug(f'{request.path} waiting on start')
+                        await controller.app.controllers_have_started.wait()
+                        log.debug(f'resume {request.path}')
```
mwhudson, line 148: I guess the logging is nice but "if not event.is_set(): await event.wait()" is kind of a strange pattern

mwhudson (approved):
Thanks, always nice to kill off a whole bucket of potential bugs in one go

subiquity/common/api/server.py
ogayot, line 148: cosmetic - maybe `f'{request.path} resuming'` to mirror the `f'{request.path} waiting on start'` message ?

ogayot (approved):
Thanks. I really like this change.

Going forward, I'd like to run the `controller.start` methods (and a few others) concurrently for all the controllers. 
I believe we'd want a controller to respond to HTTP requests as soon as their own start method has run (as opposed to when all controllers have been started). But that's fine for now.

ogayot:
> > Going forward, I'd like to run the controller.start methods (and a few others) concurrently for all the controllers.
> 
> The start methods are synchronous currently, do you think that should change?

I think so, yes, even though none of the `start` method is particularly time consuming currently.

subiquity/common/api/server.py
dbungert: It's the sort of thing that I hope is vary rare to see in the logs, but if it does show up then it may be a clue as to other problems.

dbungert: Done

PR #1597, opened 2023-03-16
snapcraft: upgrade to core22

ogayot:
Bump the base snap to core22 so we only have a single core on live-server isos.

It took me a while to come up with a working hack for override-build. The problem is that in debian and ubuntu, we patched python3.10's sysconfig so that running `sysconfig.get_preferred_scheme('prefix')` returns `posix_local`. Unfortunately, this also affects `pip install` even when we specify a prefix.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: subiquity
-base: core20
+base: core22
 version: git
 summary: Ubuntu installer
 description: The Ubuntu server installer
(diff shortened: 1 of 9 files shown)


ogayot:
@dbungert sadly, I'm not seeing better results regarding the absence of the journal output.

dbungert (approved):

snapcraft.yaml
```diff
@@ -64,13 +64,19 @@ parts:
...
       # We install without dependencies because all dependencies come from
       # archive packages.
-      pip3 install --prefix=$SNAPCRAFT_PART_INSTALL --no-dependencies .
+      # XXX: On core22, running `pip3 install --prefix xxx` does not do the
+      # right thing. The package ends up installed in xxx/local and the modules
+      # get installed to dist-packages instead of site-packages.
+      # See https://discuss.python.org/t/18240
+      # As a workaround, we use a fake user install to get the package
+      # installed in the expected place.
+      PYTHONUSERBASE="$CRAFT_PART_INSTALL" pip3 install --user --no-dependencies .
```
mwhudson, line 79: hah well done

mwhudson (approved):

PR #1604, opened 2023-03-21
use capabilities on the v1 api

mwhudson:
This actually makes the v1 and v2 guided apis a bit closer, which is nice. This area of the code is not very well tested though!

Needs a rebase but going to stick it up for review / comment anyway.

### DIFF
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -26,8 +26,6 @@ import attr
 
 from subiquitycore.models.network import NetDevInfo
 
-from subiquity.common.serialize import named_field
-
 
 class ErrorReportState(enum.Enum):
     INCOMPLETE = enum.auto()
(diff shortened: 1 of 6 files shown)


subiquity/common/types.py
```diff
@@ -315,42 +313,28 @@ class Disk:
...
-    PREFER_ENCRYPTED = 'prefer-encrypted'
-    PREFER_UNENCRYPTED = 'prefer-unencrypted'
-
+class GuidedCapability(enum.Enum):
+    DIRECT = enum.auto()
+    LVM = enum.auto()
+    LVM_LUKS = enum.auto()
+    CORE_BOOT_ENCRYPTED = enum.auto()
+    CORE_BOOT_UNENCRYPTED = enum.auto()
+    # These two are not valid as GuidedChoiceV2.capability:
```
dbungert, line 322: Are these `prefer` capabilities staying or going, eventually?

subiquity/ui/views/filesystem/guided.py
```diff
@@ -327,17 +312,27 @@ def local_help(self):
     def done(self, sender):
         results = sender.as_data()
         choice = None
-        if self.storage_encryption is not None:
+        if self.controller.core_boot_capability is not None:
+            if results.get('use_tpm', sender.tpm_choice.default):
+                capability = GuidedCapability.CORE_BOOT_ENCRYPTED
+            else:
+                capability = GuidedCapability.CORE_BOOT_ENCRYPTED
```
dbungert, line 319: Double check this one

subiquity/server/controllers/filesystem.py
```diff
@@ -367,6 +370,14 @@ def guided(self, choice: GuidedChoiceV2):
         self.model.guided_configuration = choice
 
         disk = self.model._one(id=choice.target.disk_id)
+
+        if choice.capability.is_core_boot():
+            assert isinstance(choice.target, GuidedStorageTargetReformat)
+            self.use_tpm = (
+                choice.capability == GuidedCapability.CORE_BOOT_ENCRYPTED)
```
dbungert, line 377: would we also want use_tpm == True for PREFER_ENCRYPTED?

subiquity/ui/views/filesystem/guided.py
mwhudson: thanks :)

subiquity/common/types.py
mwhudson: We talked about this I think? They will probably go eventually and the list of capabilities end up being replaced by something richer that indicates availability and default.

dbungert (approved):

dbungert:
Friendly reminder to rebase before merge button :)

PR #1605, opened 2023-03-21
do not let size: largest in storage autoinstall config match the inst…

mwhudson:
…all media

cc @chihchun and others

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1185,6 +1185,12 @@ class FilesystemModel(object):
             is_ssd = disk.info_for_display()['rotational'] == 'false'
             return is_ssd == match['ssd']
 
+        def match_install_media(disk):
+            return disk in self._exclusions
+
+        if match.get('install-media', False):
+            matchers.append(match_install_media)
+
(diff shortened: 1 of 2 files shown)


subiquity/models/filesystem.py
```diff
@@ -1185,6 +1185,12 @@ def match_ssd(disk):
             is_ssd = disk.info_for_display()['rotational'] == 'false'
             return is_ssd == match['ssd']
 
+        def match_install_media(disk):
+            return disk in self._exclusions
+
+        if match.get('install-media', False):
+            matchers.append(match_install_media)
```
dbungert, line 1192: Can _exclusions ever have more than one item?

dbungert (approved):

subiquity/models/filesystem.py
mwhudson: Depends how far you want to stretch reality -- in theory if the install media is on RAID or LVM then all the underlying disks that are part of the VG or array will be in _exclusions. But besides this being real stretch, I don't think casper would find the live media in that situation.

PR #1606, opened 2023-03-21
ad: add hack to fix non-autoinstall cases

dbungert:
Due to how top-level interactivity is determined:
https://github.com/canonical/subiquity/blob/903299fb84bf95904bba36f9f7c615b2efc9aebb/subiquity/server/server.py#L655

Simply marking this controller as interactive is not sufficient to address the autoinstall flow.
I am proposing this change now to fix non-active directory autoinstalls being hung waiting on the ad controller.


dbungert:
Lint passes locally, this is the flake8/pycodestyle thing that needs to migrate

mwhudson (approved):
Better than before.

PR #1611, opened 2023-03-23
fix "make lint" when run with the new pycodestyle/flake8

mwhudson:
pyflakes is broken in lunar release currently and if you get python3-flake8 from proposed, we get some new issues.

### DIFF
--- a/subiquity/ui/views/refresh.py
+++ b/subiquity/ui/views/refresh.py
@@ -103,7 +103,7 @@ def exc_message(exc):
         message = result.get("result", {}).get("message")
     if message is not None:
         return message
-    return"Unknown error: {}".format(exc)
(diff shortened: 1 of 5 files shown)


subiquity/ui/views/refresh.py
```diff
@@ -103,7 +103,7 @@ def exc_message(exc):
         message = result.get("result", {}).get("message")
     if message is not None:
         return message
-    return"Unknown error: {}".format(exc)
```
dbungert, line 106: wait that used to work?

dbungert (approved):

subiquity/ui/views/refresh.py
mwhudson: Apparently!

PR #1616, opened 2023-03-24
filesystem: fix path property from Raid - override attribute

ogayot:
The commmit ce146ab28 introduced a (read only) path property for Raid objects, returning an imaginary value to make for_client work with Raid objects.
Later on, the commit 8e658998e introduced a (read/write) path attribute to all filesystem objects. For Raid however, the property still takes precedence over the new attribute of the same name, so doing `raid.path = x` is invalid (no setter) and results in the following exception:
```python
AttributeError: can't set attribute 'path'
```

Fixed by using a @property + @setter attribute for path. The getter returns the actual path if it exists, otherwise returns the same imaginary value as before.


ogayot:
Tested in a VM with the patch and the install was successful. Marking ready for review.

dbungert (approved):

mwhudson (approved):
Argh. How could we have found this in advance? I guess we should try to make the installs in the integration tests call into update_devices somehow.

ogayot:
> Argh. How could we have found this in advance? I guess we should try to make the installs in the integration tests call into update_devices somehow.

Yes, that could be a solution. Units tests too maybe?
Or mypy:
```
class C:
    path: str

    @property
    def path(self):
        return "oops"
```
```
mypy file.py 
file.py:4: error: Name "path" already defined on line 2  [no-redef]
Found 1 error in 1 file (checked 1 source file)
```

PR #1618, opened 2023-03-28
(fix) Join target instead of live session

CarlosNihelton:
LP: #2013079
Found during beta testing that the live session is joining, instead of the target.


dbungert (approved):
LGTM, and per discussion we were going to focus on testing this with a live test.    This change can merge when ready.

PR #1619, opened 2023-03-28
cloud-init: preserve ephemeral cloud-init logs in target to aid in debug

blackboxsw:
If cloud-init were active in the ephmeral install stage due to available user-data,
we want to preserve those logs in /var/log/installer on the target system across first reboot
to aid in triage and debugging what happened in ephemeral stage.


subiquity/server/controllers/shutdown.py
```diff
@@ -102,6 +102,10 @@ async def copy_logs_to_target(self, context):
         else:
             await arun_command(
                 ['cp', '-aT', '/var/log/installer', target_logs])
+            if os.path.exists("/var/log/cloud-init.log"):
+                # Preserve ephemeral boot cloud-init logs if applicable
+                await arun_command(
+                    ['cp', '-aT', '/var/log/cloud-init*.log', target_logs])
```
dbungert, line 108: The above copy is a directory, yours are files.  I think you want this, to avoid the error `cp: extra operand`:
```suggestion
                    ['cp', '-a', '/var/log/cloud-init*.log', target_logs])
```

blackboxsw: Thanks copy-pasta issue there. will test and confirm behavior on this PR .

dbungert: Note that this is a funny one to test, because the above copy is after the logs are copied to target, so the debug logs cut off abruptly before this `arun_command()`

system_setup/ubuntu-wsl-setup
dbungert, line 1: There's a bug in the subiquity snapcraft right now where this file is removed.  We'll need to restore it.

subiquity/server/controllers/shutdown.py
```diff
@@ -100,8 +100,19 @@ async def copy_logs_to_target(self, context):
...
+            cloudint_logs = [
+               cloudinit_log
+               for cloudinit_log in (
+                    "/var/log/cloud-init.log",
+                    "/var/log/cloud-init-output.log"
+               )
+               if os.path.exists(cloudinit_log)
+            ]
+            await arun_command(
+                ['cp', '-a', " ".join(cloudinit_logs), '/var/log/installer'])
```
dbungert, line 112: I'm not sure how it would happen, but this form is invalid if neither log file exists I think.

dbungert (changes requested):

subiquity/server/controllers/shutdown.py
dbungert: The default value of check is False, so I guess just try to copy them over and if they don't exist, no big deal.

blackboxsw:
Having finally linted and tested code, we get success:
2023-03-29 01:09:39,860 DEBUG subiquitycore.utils:103 arun_command ['cp', '-a', '/var/log/cloud-init.log', '/var/log/cloud-init-output.log', '/var/log/installer'] exited with code 0

system_setup/ubuntu-wsl-setup
dbungert, line 1: restore this file please and we can merge

dbungert (approved):

PR #1624, opened 2023-03-29
system-setup: fix disappearing ubuntu-wsl-setup

dbungert:
Make a copy of this file to INSTALL, as the organize step renames it, causing it to disappear from the source directory.


CarlosNihelton (approved):
LGTM 👍🏽 Thanks!

There is a small overhead in having this as its own part, since it checks out the source tree once more, but ithe benefit worths it!

PR #1625, opened 2023-03-30
errorreport: all recent syslog

dbungert:
Some outputs only appear in the journal.


ogayot:
I've opened a PR on apport: https://github.com/canonical/apport/pull/150

The proposal is to introduce:
```python
recent_syslog(re.compile("."), journald_system_only=False)
```

PR #1627, opened 2023-03-30
[Active Directory] Bind mounts in the joining context manager

CarlosNihelton:
This concludes LP: 2013079.

`realm --install <target>` requires system bind-mounts to work properly.

### DIFF
--- a/subiquity/server/ad_joiner.py
+++ b/subiquity/server/ad_joiner.py
@@ -16,6 +16,7 @@
 import asyncio
 from contextlib import contextmanager
 import logging
+import os
 from socket import gethostname
 from subprocess import CalledProcessError
 from subiquitycore.utils import arun_command, run_command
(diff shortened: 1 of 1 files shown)


dbungert:
What problem were you encountering when you were running this through curtin?  If humanely possible I don't want new mount code.  Can you tell me more about this `pam-auth-update` thing?  I'm worried about a more fundamental problem and want to understand that better.

CarlosNihelton:
> What problem were you encountering when you were running this through curtin? If humanely possible I don't want new mount code. Can you tell me more about this `pam-auth-update` thing? I'm worried about a more fundamental problem and want to understand that better.

Sure. This is more related to `realm`, though. When running it with the `--install` option, it temporarily `chroot`s into the target system. I guess I didn't notice that before because I was `chroot`ing (manually, with bind mounts) very often to monitor the target system while post-installation was progressing, like running apt commands to check what packages has been installed. 

I couldn't reproduce the `pam-auth-update` issue I've seen before, so I took a break from it and focused on this one.

Without the bind-mounts, `realm join --install` should fail with a very unhelpful error message.

subiquity/server/ad_joiner.py
```diff
@@ -29,18 +30,28 @@
...
+            bound_dir = os.path.join(root_dir, bind[1:])
+            if bound_dir != bind:
+                run_command(["mount", "--bind", bind, bound_dir])
         yield hostname_process
     finally:
         # Restoring the live session hostname.
         hostname_process = run_command(['hostname', hostname_current])
         if hostname_process.returncode:
             log.info("Failed to restore live session hostname")
+        for bind in binds:
```
dbungert, line 51:
```suggestion
        for bind in reversed(binds):
```
I think for the mounts your doing the order won't matter, but it does sometimes.  Let's reverse the order in case we add more to the list later.

dbungert (approved):

PR #1628, opened 2023-03-30
Ld lib path vs cmds

dbungert:
LD_LIBRARY_PATH is populated with snap things, that we don't want for the commands that run orig_environ.  Saving LD_LIBRARY_PATH to an ORIG variable like I did https://github.com/canonical/ubuntu-desktop-installer/pull/1747 was not going to work, as that particular variable already has the snap things set to it, but does work as intended for PATH and PYTHONPATH.


ogayot (approved):

mwhudson:
> LD_LIBRARY_PATH is populated with snap things,

Huh what is setting that I wonder? It doesn't sound helpful.

PR #1630, opened 2023-04-02
 Sizing policy

dbungert:
Add an API that ubuntu-desktop-installer can use and an autoinstall key to control the size allocation for how much of a VG is allocated to the root LV.

Thanks to Ryan Mounce as well for the related PR #1588

### DIFF
--- a/documentation/autoinstall-reference.md
+++ b/documentation/autoinstall-reference.md
@@ -311,6 +311,24 @@ When using the "lvm" layout, LUKS encryption can be enabled by supplying a passw
 
 The default is to use the lvm layout.
 
+#### sizing-policy
+
+The lvm layout will, by default, attempt to leave room for snapshots and further expansion.  A sizing-policy key may be supplied to control this behavior.
+
+**type:** string (enumeration)
+**default:** scaled
+
+Supported values are:
+
+ * `scaled` -> adjust space allocated to the root LV based on space available to the VG
+ * `all` -> allocate all remaining VG space to the root LV
+
+The scaling system is currently as follows:
+ * Less than 10 GiB: use all remaining space for root filesystem
+ * Between 10-20 GiB: 10 GiB root filesystem
+ * Between 20-200 GiB: use half of remaining space for root filesystem
+ * Greater than 200 GiB: 100 GiB root filesystem
+
 #### action-based config
 
 For full flexibility, the installer allows storage configuration to be done using a syntax which is a superset of that supported by curtin, described at https://curtin.readthedocs.io/en/latest/topics/storage.html.
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -323,6 +323,10 @@ class GuidedCapability(enum.Enum):
     CORE_BOOT_PREFER_ENCRYPTED = enum.auto()
     CORE_BOOT_PREFER_UNENCRYPTED = enum.auto()
 
+    def is_lvm(self) -> bool:
+        return self in [GuidedCapability.LVM,
+                        GuidedCapability.LVM_LUKS]
+
     def is_core_boot(self) -> bool:
         return self in [GuidedCapability.CORE_BOOT_ENCRYPTED,
                         GuidedCapability.CORE_BOOT_UNENCRYPTED,
@@ -369,6 +373,11 @@ class StorageResponseV2:
     # if need_boot == True, there is not yet a boot partition
     need_boot: Optional[bool] = None
     install_minimum_size: Optional[int] = None
+
+
+class SizingPolicy(enum.Enum):
+    SCALED = enum.auto()
+    ALL = enum.auto()
 
 
 @attr.s(auto_attribs=True)
@@ -425,6 +434,8 @@ class GuidedChoiceV2:
     target: GuidedStorageTarget
     capability: GuidedCapability
     password: Optional[str] = attr.ib(default=None, repr=False)
+    sizing_policy: Optional[SizingPolicy] = \
+        attr.ib(default=SizingPolicy.SCALED)
 
     @staticmethod
     def from_guided_choice(choice: GuidedChoice):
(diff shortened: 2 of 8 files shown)


subiquitycore/utils.py
```diff
@@ -176,3 +176,10 @@ def disable_subiquity():
                  "snap.subiquity.subiquity-service.service",
                  "serial-subiquity@*.service"])
     return
+
+
+def child_matches(items, **kw):
```
ogayot, line 181: Is the term "child" really relevant here? If I understand correctly, this function returns a subset of the elements from items, filtering out the ones that don't match the criteria specified.

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -497,6 +500,45 @@ async def test_lvm_20G_bad_offset(self, disk_size):
...
+        size = dehumanize_size(lvm_partition['size'])
+        return size, part_size
+
+    @parameterized.expand(bootloaders_and_ptables)
+    async def test_scaled_disk(self, bootloader, ptable):
+        size, part_size = await self._sizing_setup(
+                bootloader, ptable, 50 << 30, SizingPolicy.SCALED_FREE_SPACE)
+        # expected to be about half, differing by boot and ptable types
+        self.assertLess(20 << 30, size)
+        self.assertLess(size, 30 << 30)
```
ogayot, line 529: I don't know if my suggestion is better, so I'll leave that up to you:

```python
self.assertTrue((20 << 30) < size < (30 << 30))
```

documentation/autoinstall-reference.md
```diff
@@ -311,6 +311,24 @@ When using the "lvm" layout, LUKS encryption can be enabled by supplying a passw
...
+**type:** string (enumeration)
+**default:** scaled
+
+Supported values are:
+
+ * `scaled` -> adjust space allocated to the root LV based on space available to the VG
+ * `all` -> allocate all remaining VG space to the root LV
+
+The scaling system is currently as follows:
+ * Less than 10 GB: use all remaining space for root filesystem
```
ogayot, line 327: If applicable, I would prefer to use the `GiB` notation ; although I understand some people are uncomfortable with it. Being the reference document, I think it should reflect the actual implementation.

subiquity/common/types.py
```diff
@@ -371,6 +375,21 @@ class StorageResponseV2:
...
+    SCALED_FREE_SPACE = enum.auto()
+    ALL = enum.auto()
+
+    @classmethod
+    def from_string(cls, value):
+        if value is None:
+            return cls.SCALED_FREE_SPACE
+        if value == 'all':
+            return cls.ALL
+        if value in ('scaled-free-space', 'scaled'):
```
ogayot, line 388: any specific reasons for having two values here?

ogayot (commented):
Thanks!

subiquity/server/controllers/tests/test_filesystem.py
dbungert: I started with that form but changed to the above, as the result is rather unhelpful if it assertTrue form fails:
```
subiquity/server/controllers/tests/test_filesystem.py:529: in test_scaled_disk
    self.assertTrue((20 << 30) < size < (30 << 30))
E   AssertionError: False is not true
```
In this form we get a size number
```
subiquity/server/controllers/tests/test_filesystem.py:530: in test_scaled_disk
    self.assertLess(20 << 30, size)
E   AssertionError: 21474836480 not less than -1
```

subiquity/common/types.py
dbungert: no, I need to pick.  'scaled' it is.

documentation/autoinstall-reference.md
dbungert: This makes sense.  Subiquity needs to be specific in actual sizing.

subiquitycore/utils.py
dbungert: renamed, thanks!

subiquitycore/utils.py
```diff
@@ -176,3 +176,10 @@ def disable_subiquity():
                  "snap.subiquity.subiquity-service.service",
                  "serial-subiquity@*.service"])
     return
+
+
+def matching_dicts(items, **kw):
+    """Given an input sequence of dictionaries, return a list of dicts where
+    the supplied keyword arguments all match those items."""
+    return [item for item in items
+            if all(item.get(k) == v for k, v in kw.items())]
```
ogayot, line 185: Note that because `.get()` returning `None` when key is not found, any filter that expects a `None` value will essentially be ignored:

```python
print(matching_dicts([{"foo", "bar"}], baz=None))
[{'foo': 'bar'}]
```
I would tend to think something like the below is better, but this means we change the existing implementation of `match`, so I don't know if it's a good thing. Maybe a TODO is better until we figure out?

```python
return [item for item in items
        if all(k in item and item[k] == v for k, v in kw.items())]

File: subiquitycore/utils.py [hunk 6]
```diff
@@ -176,3 +176,10 @@ def disable_subiquity():
                  "snap.subiquity.subiquity-service.service",
                  "serial-subiquity@*.service"])
     return
+
+
+def matching_dicts(items, **kw):
```
ogayot, line 181: suggestions:
* `**criteria` instead of `**kw`
* `items: Sequence[Dict[Any, Any]]` so it's more obvious for the caller what to pass

ogayot (commented):
I'd like your opinion on the use of `.get()` but otherwise looks good to me.

subiquitycore/utils.py
dbungert: I'd rather implement this now, I see it as a strictly better version.
If this had regressed an API test (doesn't seem to) I would have considered that a bug in the respective API test.

subiquitycore/utils.py
dbungert: Done!

dbungert: Done!

subiquitycore/utils.py
ogayot: Perfect, thanks!

ogayot (approved):
LGTM, thanks!

PR #1634, opened 2023-04-03
Bootable disks

dbungert:
Mark the disks as can_be_boot_disk, instead of a separate API call.

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -306,16 +306,10 @@ class API:
                 def POST(data: Payload[ReformatDisk]) \
                     -> StorageResponseV2: ...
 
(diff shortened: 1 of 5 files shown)


subiquity/common/apidef.py
```diff
@@ -306,16 +306,10 @@ class reformat_disk:
                 def POST(data: Payload[ReformatDisk]) \
                     -> StorageResponseV2: ...
 
-            class potential_boot_disks:
```
ogayot, line 309: I assume the desktop installer is the only consumer we know of for this endpoint. When we backport this fix, should we still keep the existing endpoint? This feels like a breaking change.

ogayot (approved):

PR #1639, opened 2023-04-06
network: don't run netplan apply if no autoinstall and nm is enabled

ogayot:
When doing a netboot autoinstall on the desktop installer, NetworkManager (through cloud-init) and subiquity compete to configure the network. When an autoinstall file does not include a network section, make sure subiquity does not try to configure the network if NetworkManager is installed and enabled.

### DIFF
--- a/subiquity/models/network.py
+++ b/subiquity/models/network.py
@@ -14,8 +14,10 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import logging
+import subprocess
 
 from subiquitycore.models.network import NetworkModel as CoreNetworkModel
+from subiquitycore.utils import arun_command
 
 log = logging.getLogger('subiquity.models.network')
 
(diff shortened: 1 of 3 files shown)


subiquity/models/network.py
```diff
@@ -72,3 +74,18 @@ async def target_packages(self):
             return ['wpasupplicant']
         else:
             return []
+
+    async def is_nm_enabled(self):
+        try:
+            cp = await arun_command(("nmcli", "networking"), check=True)
+        except subprocess.CalledProcessError as exc:
+            log.warning("failed to run nmcli networking. Will consider"
```
ogayot, line 82: log.warn

subiquity/models/network.py
```diff
@@ -72,3 +74,16 @@ async def target_packages(self):
...
+            cp = await arun_command(("nmcli", "networking"), check=True)
+        except subprocess.CalledProcessError as exc:
+            log.warn("failed to run nmcli networking, considering"
+                     " NetworkManager disabled.")
+            log.debug("stderr: %s", exc.stderr)
+            return False
+        except FileNotFoundError:
+            return False
+
+        return cp.stdout == "enabled\n":
```
dbungert, line 89:
```suggestion
        return cp.stdout == "enabled\n"
```

ogayot: oops, sorry about that!

dbungert:
I tested this today and it LGTM.  I believe it is harmless on server and doing what is expected on desktop.

dbungert (approved):

PR #1640, opened 2023-04-06
locale: switch to localectl set-locale

dbungert:
Move away from localedef to fully set the locale.  The set-locale operation is also generating the locale.

### DIFF
--- a/subiquity/models/locale.py
+++ b/subiquity/models/locale.py
@@ -37,22 +37,19 @@ class LocaleModel:
     def switch_language(self, code):
         self.selected_language = code
 
-    async def gen_localedef(self) -> None:
-        language, charmap = locale.normalize(self.selected_language).split(".")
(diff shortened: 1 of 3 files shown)


ogayot (approved):
LGTM
Feels like a low risk change.
The only thing I'm worried about is how long `localectl set-locale` will take when many locales are enabled in /etc/locale.gen.
On server, we start with only C.UTF-8 so it should not take too long.

ogayot:
I played around with `localectl` after enabling many locales in /etc/locale.gen to see how long it would take to run. I managed to break `systemd-localed` and submitted a fix upstream: https://github.com/systemd/systemd/pull/27179

It should not happen in practice as long as we're not manually editing /etc/locale.gen though. But just in case we need it, the patch applies cleanly https://launchpadlibrarian.net/660199033/systemd_252.5-2ubuntu3_252.5-2ubuntu4.diff.gz

PR #1641, opened 2023-04-07
gaps: fix gap size in extended

dbungert:
Gap sizing in extended had a flaw when there was a partition after it.  We need the ebr space in between.

Emerged in the private bug LP: #2015521, where a user was given a guided resize option based on a logical partition that was not physically last.  The result was that the end position of the newly created partition was the same spot as the offset of the next partition.

### DIFF
--- a/subiquity/common/filesystem/gaps.py
+++ b/subiquity/common/filesystem/gaps.py
@@ -150,7 +150,10 @@ def find_disk_gaps_v2(device, info=None):
         if part is None:
             gap_end = ad(device.size - info.min_end_offset)
         else:
-            gap_end = ad(part.offset)
+            if part.is_logical:
+                gap_end = ad(part.offset - info.ebr_space)
+            else:
+                gap_end = ad(part.offset)
 
         gap_start = au(prev_end)
 
--- a/subiquity/tests/api/test_api.py
+++ b/subiquity/tests/api/test_api.py
@@ -1469,6 +1469,38 @@ class TestRegression(TestAPI):
             v1resp = await inst.get('/storage')
             self.assertEqual([], match(v1resp['config'], type='format'))
 
+    @timeout()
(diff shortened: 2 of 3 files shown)


subiquity/common/filesystem/gaps.py
```diff
@@ -150,7 +150,7 @@ def maybe_add_gap(start, end, in_extended):
         if part is None:
             gap_end = ad(device.size - info.min_end_offset)
         else:
-            gap_end = ad(part.offset)
+            gap_end = ad(part.offset - info.ebr_space)
```
ogayot, line 153: It does not feel right to me to subtract `info.ebr_space` if we're not dealing with a logical partition.
Shouldn't it be:
```python
if part.is_logical():
    gap_end = ad(part.offset - info.ebr_space)
else:
    gap_end = ad(part.offset)
```

subiquity/tests/api/test_api.py
```diff
@@ -1469,6 +1469,38 @@ async def test_can_create_unformatted_partition(self):
             v1resp = await inst.get('/storage')
             self.assertEqual([], match(v1resp['config'], type='format'))
 
+    @timeout()
+    async def test_guided_v2_resize_logical_middle_partition(self):
+        '''LP: #2015521 - a logical partition that wasn't the phyiscally last
```
ogayot, line 1474: minor: physically

subiquity/common/filesystem/gaps.py
dbungert: Excellent spot.  This would have short-changed primary partitions.  Fixed, and a new unit test.

ogayot (approved):
LGTM, thanks!

subiquity/tests/api/test_api.py
dbungert: I need to just leave the spellchecker on

PR #1642, opened 2023-04-10
scripts: update slimy to handle u-d-i

dbungert:
If the target snap is ubuntu-desktop-installer, put subiquity changes in the appropriate location.

### DIFF
--- a/scripts/slimy-update-snap.sh
+++ b/scripts/slimy-update-snap.sh
@@ -35,7 +35,9 @@ add_overlay() {
         local upper="$(mktemp -dp "${tmpdir}")"
     fi
     chmod go+rx "${work}" "${upper}"
-    do_mount -t overlay overlay -o lowerdir="${lower}",upperdir="${upper}",workdir="${work}" "${mountpoint}"
+    do_mount -t overlay overlay \
+        -o lowerdir="${lower}",upperdir="${upper}",workdir="${work}" \
+        "${mountpoint}"
 }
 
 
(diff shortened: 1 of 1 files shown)


scripts/slimy-update-snap.sh
```diff
@@ -53,11 +55,22 @@ do_mount $old old
...
+    subiquity_dest=new/lib/python3.10/site-packages
+elif [ -d new/bin/subiquity/subiquity ] ; then
+    subiquity_dest=new/bin/subiquity
+else
+    echo what snap is this
+    exit 1
+fi
+
+rm -rf $subiquity_dest/subiquity
+rm -rf $subiquity_dest/subiquitycore
```
ogayot, line 69: Let's be extra careful with this kind of command. A space at the wrong place and we can end up with a wiped system.

```bash
rm -rf -- "${subiquity_dest}"/subiquity
rm -rf -- "${subiquity_dest}"/subiquitycore
```

(you can leave aside the `--` if you want)

scripts/slimy-update-snap.sh
```diff
@@ -53,11 +55,22 @@ do_mount $old old
...
 rm -rf new/lib/python3.10/site-packages/curtin
-rm -rf new/lib/python3.10/site-packages/subiquity
-rm -rf new/lib/python3.10/site-packages/subiquitycore
+
+if [ -d new/lib/python3.10/site-packages/subiquity ] ; then
+    subiquity_dest=new/lib/python3.10/site-packages
+elif [ -d new/bin/subiquity/subiquity ] ; then
+    subiquity_dest=new/bin/subiquity
+else
+    echo what snap is this
```
ogayot, line 64: suggestion:
```bash
echo 'what snap are we building?' >&2
```

ogayot (approved):

PR #1646, opened 2023-04-12
errorreport: get udi log

dbungert:


dbungert:
https://git.launchpad.net/ubuntu/+source/apport/tree/data/package-hooks/ubuntu-desktop-installer.py?h=applied/ubuntu/devel#n9

ogayot (approved):

seb128:
> Snap Revision is included today

ah, right, thanks

> That we'll have to add - a tag of literally "ubuntu-desktop-installer" ?

yes, the same was added in https://code.launchpad.net/~ubuntu-core-dev/ubuntu/+source/apport/+git/apport/+merge/440270 , the idea is to be able to filter our the desktop reports easily / subscribe to the tag

PR #1648, opened 2023-04-12
Fix assertion in autoinstall API test

ogayot:
The API tests unconditionally did a POST request to /source.

This is fine in interactive scenarios, but for autoinstalls, it results in the source controller being "configured" twice:
 * once after the apply_autoinstall_config coroutine finishes executing
 * once after the POST request is handled

By the time the POST request is handled, other models might have started (or even finished) applying their own autoinstall configuration. If that happens, changing the source model will likely invalidate their configuration, resulting in various bugs. Subiquity does not currently support re-running the apply_autoinstall_config coroutines, so it seems fair to say that doing a POST to source in a fully automated install is a pretty bad idea, unless it's done veryyyyy early.

A recent [change](https://github.com/canonical/subiquity/pull/1639) slightly affected the timings in CI, and coupled with the above considerations, lead to a race condition to show up regularly:

* the apply_autoinstall_config coroutine from the mirror model now runs sooner in CI
* the POST request to /source gets handled during mirror testing, or more precisely between the moment we apply the apt configuration in an overlay, and the moment we run the apt-get update command

**Summary of changes**

* The race condition is something that could happen in an interactive install, and therefore the first patch in the series addresses it.
* In autoinstall scenarios though, changing the source during mirror testing is a recipe to disaster, so the second patch makes sure subiquity raises an exception if that happens.
* Doing a POST to /source in API tests if fine for interactive installs, but is wrong for autoinstalls. Added a parameter to `start_server` so that individual tests can disable the POST to /source.

### DIFF
--- a/subiquity/server/controllers/mirror.py
+++ b/subiquity/server/controllers/mirror.py
@@ -156,6 +156,7 @@ class MirrorController(SubiquityController):
         self.test_apt_configurer: Optional[AptConfigurer] = None
         self.final_apt_configurer: Optional[AptConfigurer] = None
         self.mirror_check: Optional[MirrorCheck] = None
+        self.autoinstall_apply_started = False
 
     def load_autoinstall_data(self, data):
         if data is None:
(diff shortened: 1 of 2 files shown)


dbungert (approved):
Thanks!

PR #1650, opened 2023-04-12
storage: LVM sizing must be LVM_CHUNK_SIZE aligned

dbungert:


subiquity/server/controllers/filesystem.py
```diff
@@ -299,11 +299,11 @@ def guided_lvm(self, gap, choice: GuidedChoiceV2):
         vg = self.create_volgroup(spec)
         if choice.sizing_policy == SizingPolicy.SCALED:
             lv_size = sizes.scaled_rootfs_size(vg.size)
-            lv_size = align_down(lv_size, LVM_CHUNK_SIZE)
         elif choice.sizing_policy == SizingPolicy.ALL:
             lv_size = vg.size
         else:
             raise Exception(f'Unhandled size policy {choice.sizing_policy}')
+        lv_size = align_down(lv_size, LVM_CHUNK_SIZE)
```
ogayot, line 306: That seems sound but `vg.size` is a property and returns a size that should already be a multiple of LVM_CHUNK_SIZE if I read the code correctly. So I doubt it will have any effect.

dbungert: OK, I still like this from a code-structure perspective but think it isn't critical.

PR #1651, opened 2023-04-12
errorreport: tag with snap name

dbungert:


subiquity/common/errorreport.py
```diff
@@ -140,6 +141,9 @@ def _bg_add_info():
                 self.pr.add_hooks_info(None)
                 apport.hookutils.attach_hardware(self.pr)
             self.pr['Syslog'] = apport.hookutils.recent_syslog(re.compile('.'))
+            snap_name = os.environ.get('SNAP_NAME', None)
+            if bool(snap_name):
```
ogayot, line 145: did you mean?
```python
if snap_name is not None:
```

`if bool(snap_name)` is equivalent to `if snap_name`.

subiquity/common/errorreport.py
```diff
@@ -140,6 +141,9 @@ def _bg_add_info():
...
+                self.add_tags([snap_name])
```
ogayot, line 146: suggestion:
```python
with contextlib.suppress(KeyError):
    self.add_tags([os.environ["SNAP_NAME"])
```

subiquity/common/errorreport.py
```diff
@@ -329,6 +333,20 @@ def ref(self):
...
+    # with core24 these tag methods can be dropped for equivalent methods
+    # that will be on the report object
+    def get_tags(self) -> Set[str]:
+        """Return the set of tags."""
+        if "Tags" not in self.pr:
+            return set()
+        return set(self.pr["Tags"].split(" "))
+
+    def add_tags(self, tags: Iterable[str]) -> None:
+        """Add tags to the report. Duplicates are dropped."""
```
ogayot, line 345: Maybe a note that tags should not include spaces. That's not going to work very well.

subiquity/common/errorreport.py
ogayot: this would also suppress KeyError if self.add_tags raises one, so in hindsight, it may not be a great idea.

subiquity/common/errorreport.py
```diff
@@ -140,6 +141,9 @@ def _bg_add_info():
                 self.pr.add_hooks_info(None)
                 apport.hookutils.attach_hardware(self.pr)
             self.pr['Syslog'] = apport.hookutils.recent_syslog(re.compile('.'))
+            snap_name = os.environ.get('SNAP_NAME', None)
```
ogayot, line 144: None is the default value for .get() but no problem making it more explicit.

subiquity/common/errorreport.py
dbungert: I was intentionally excluding the empty string as well.

subiquity/common/errorreport.py
dbungert: I'm inclined to leave it alone to stay as close as possible to python3-problem-report - thoughts?

ogayot (approved):

PR #1655, opened 2023-04-13

kubiko:

### DIFF
--- a/bin/console-conf-wrapper
+++ b/bin/console-conf-wrapper
@@ -22,7 +22,14 @@ if [ -e /run/snapd-recovery-chooser-triggered ]; then
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


dbungert:
> `core/focal`

What is the goal with this branch?  It's quite out of date and I can't vouch for the contents.

kubiko:
Unless I am mistaken, this branch is used for Ubuntu Core 20 base snap (core20)
@valentindavid can you please confirm here?

valentindavid:
@kubiko Looks good! And it is `core/focal`, correct.

We will it need it applied on newer branches at some point. I am planning on enabling preseeding on core22+. I think it is a safe change to apply in advance.

dbungert (approved):

PR #1656, opened 2023-04-13

kubiko:

### DIFF
--- a/bin/console-conf-wrapper
+++ b/bin/console-conf-wrapper
@@ -22,7 +22,14 @@ if [ -e /run/snapd-recovery-chooser-triggered ]; then
     fi
 fi
 
-if grep -q 'snapd_recovery_mode=install' /proc/cmdline ; then
+# always prefer to use modeevn if it is available
+if [ -e "/var/lib/snapd/modeenv" ]; then
+    mode="$(sed -n 's/mode=\([^[:space:]]*\)/\1/p' /var/lib/snapd/modeenv)"
+else
+    mode="$(sed 's/.*snapd_recovery_mode=\([^[:space:]]*\)[[:space:]].*/\1/' /proc/cmdline))"
+fi
+
+if [ "${mode}" = "install" ]; then
     echo "Installing the system, please wait for reboot"
     # XXX: replace with something more user friendly after the beta?
     journalctl -u snapd.service -f


bin/console-conf-wrapper
```diff
@@ -22,7 +22,14 @@ if [ -e /run/snapd-recovery-chooser-triggered ]; then
     fi
 fi
 
-if grep -q 'snapd_recovery_mode=install' /proc/cmdline ; then
+# always prefer to use modeevn if it is available
+if [ -e "/var/lib/snapd/modeenv" ]; then
+    mode="$(sed -n 's/mode=\([^[:space:]]*\)/\1/p' /var/lib/snapd/modeenv)"
+else
+    mode="$(sed 's/.*snapd_recovery_mode=\([^[:space:]]*\)[[:space:]].*/\1/' /proc/cmdline))"
```
dbungert, line 29:
```suggestion
    mode="$(sed 's/.*snapd_recovery_mode=\([^[:space:]]*\)[[:space:]].*/\1/' /proc/cmdline)"
```

bin/console-conf-wrapper
```diff
@@ -22,7 +22,14 @@ if [ -e /run/snapd-recovery-chooser-triggered ]; then
     fi
 fi
 
-if grep -q 'snapd_recovery_mode=install' /proc/cmdline ; then
+# always prefer to use modeevn if it is available
+if [ -e "/var/lib/snapd/modeenv" ]; then
```
dbungert, line 26: I don't know much about snapd recovery modes or this modeenv file, can you point me to something relevant?

My big question is on the notion that the kernel command line would not take precedence over the on-disk value, as the /proc/cmdline is presumed to be easier to change on the fly.  What's going on there?

bin/console-conf-wrapper
kubiko: ups, fixed
thank you @dbungert!

kubiko:

bin/console-conf-wrapper
kubiko: @valentindavid can you please comment about use of the kernel cmd line vs modeenv file?

kubiko:
@valentindavid let's have a conversation about the change itself here.
Can you please comment here to address @dbungert's comment?

bin/console-conf-wrapper
valentindavid: `snapd_recovery_mode=install` is to instruct `snap-booststrap` what mode to run. `/var/lib/snapd/modeenv` is the mode for snapd to run, which is set during install by `snap-boostrap` and maintained later by `snapd`. And this file contains the effective mode. So it is more correct than the kernel command line.

Unfortunately there is not `snap` command to ask the current mode.

dbungert (approved):

PR #1658, opened 2023-04-13
Fixed paths for lsc and locales commands

dbungert:
Subiquity snap doesn't stage language-selector-common nor locales packages.
Attempting to run those commands require respecting the env previously prepared by the caller, as well as fixing the commands paths (or changing the PATH env var in the env preparation step). Since the apt commands were already ran with a fixed path, I'm applying the same principle in here.
Yet, for now, Jammy doesn't ship the l-s-c package seeded, so a condition was preserved to avoid future crashes.

(cherry picked from commit 2e67005403be469b9e36b34168fa638dc68f667b)


CarlosNihelton (approved):
LGTM 👍🏽

PR #1660, opened 2023-04-14
storage: don't let probing discard storage config in the middle of partitioning

ogayot:
At least in the desktop installer with storage v2, things go bad if a probert storage run finishes during partitioning. Unless storage is already configured, we automatically discard any change done so far upon receiving a response from a probert storage run. This leads to various problems.

This PR aims at making sure the storage configuration is not reset by a probert run while we are in the middle of partitioning.

Essentially, any request that modifies the partitioning configuration will now ensure that a transaction is started. If we receive the result of a probert run in the middle of a transaction, the new data is not applied. Instead it is stored and will only be used if the user asks for a reset/refresh.

It also introduces a new endpoint storage/v2/ensure_transaction that a client can use before doing any modification to the partitioning configuration. This ensures that the data returned by a subsequent GET /storage/v2/ call will not be invalidated by a probert run.

TODO:

- [x] unit tests
- [x] API test(s)
- [ ] VM testing
- [ ] testing with desktop installer

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -22,7 +22,7 @@ import os
 import pathlib
 import select
 import time
-from typing import Dict, List, Optional
+from typing import Any, Dict, List, Optional
 
 from curtin.commands.extract import AbstractSourceHandler
 from curtin.storage_config import ptable_uuid_to_flag_entry
@@ -160,6 +160,10 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         self._role_to_device: Dict[str: _Device] = {}
         self._device_to_structure: Dict[_Device: snapdapi.OnVolume] = {}
         self.use_tpm: bool = False
+        self.locked_probe_data = False
+        # If probe data come in while we are doing partitioning, store it in
+        # this variable. It will be picked up on next reset will pick it up.
+        self.queued_probe_data: Optional[Dict[str, Any]] = None
 
     def is_core_boot_classic(self):
         return self._system is not None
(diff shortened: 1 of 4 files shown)


dbungert:
This sounds correct but, as you mention, would bear significant testing.

subiquity/server/controllers/filesystem.py
```diff
@@ -680,9 +684,18 @@ async def v2_orig_config_GET(self) -> StorageResponseV2:
 
     async def v2_reset_POST(self) -> StorageResponseV2:
         log.info("Resetting Filesystem model")
-        self.model.reset()
+        self.locked_probe_data = False
```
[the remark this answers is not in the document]
ogayot: Dropped this instruction. Once we have started partitioning, the only way to pickup changes is to call v2_reset_POST.

ogayot:
I think at this point it's ready to review, but I want to run manual tests.

subiquity/server/controllers/filesystem.py
```diff
@@ -160,6 +160,10 @@ def __init__(self, app):
         self._role_to_device: Dict[str: _Device] = {}
         self._device_to_structure: Dict[_Device: snapdapi.OnVolume] = {}
         self.use_tpm: bool = False
+        self.locked_probe_data = False
```
sil2100, line 163: Since I can only review the Python and only a bit of the logic, a small thing: per the style here, shouldn't this be typed:
`self.locked_probe_data: bool = False` ?

sil2100 (commented):
Just a small completely irrelevant inline comment, but still browsing through the rest of the logic.

sil2100 (approved):
A very uh, non-expert approve here. But the logic feels fine as well.

subiquity/server/controllers/filesystem.py
ogayot: Good catch, thanks! I included this change in a follow-on PR: https://github.com/canonical/subiquity/pull/1666

PR #1667, opened 2023-04-20
Merge discourse docs 2023 april

dbungert:


documentation/Makefile
```diff
@@ -4,6 +4,12 @@ DISCOURSE_MD = $(addprefix discourse/, $(SRC))
 HTML_MD = $(addprefix html/, $(SRC))
 HTML = $(HTML_MD:.md=.html)
 
+.PHONY: new
+new: clean all
+
+.PHONY: all
+all: discourse html
+
```
ogayot, line 12: Not sure it's strictly needed, but it sounds ok to me to make `all` the new default recipe.
OTOH, I think using `new` as the default is counter-productive. We shouldn't call the clean recipe on every invocation of make IMO.

Would you mind moving these two recipes down (next to the html and discourse)?

documentation/autoinstall-reference.md
```diff
@@ -1,5 +1,3 @@
-# Automated Server Installs Config File Reference
```
ogayot, line 1: Is this an intended change?

documentation/autoinstall-reference.md
```diff
@@ -157,24 +157,28 @@ Identifier of the source to install (e.g., `"ubuntu-server-minimized"`).
...
+```yaml
+network:
+  version: 2
+  ethernets:
+    enp0s31f6:
+      dhcp6: yes
+```
 
-Note that thanks to a bug, the version of subiquity released with 20.04 GA forces you to write this with an extra "network:" key like so:
+Note that thanks to a bug, the version of Subiquity released with 20.04 GA forces you to write this with an extra `network:` key like so:
```
ogayot, line 172: Unrelated to the change so please feel free to ignore.
Doesn't "thanks to a bug" sound odd? How about "because of a bug"?

documentation/autoinstall-reference.md
```diff
@@ -309,7 +322,7 @@ When using the "lvm" layout, LUKS encryption can be enabled by supplying a passw
         name: lvm
         password: LUKS_PASSPHRASE
 
-The default is to use the lvm layout.
+The default is to use the `lvm` layout.
 
 #### sizing-policy
```
ogayot, line 327: I guess we want to capitalize "Sizing" as well.

ogayot (commented):
Thanks, some minor comments.

Overall, this will make the doc easier to review going forward so it's all beneficial.

documentation/autoinstall-reference.md
dbungert: Yes.  This change was done on Discourse to drop a redundant title being shown on the page, so it should be replicated here.

documentation/Makefile
dbungert: I think I have done as you asked, please confirm.

documentation/autoinstall-reference.md
dbungert: Fixed, thanks.

documentation/Makefile
ogayot: I was thinking of moving the `all` recipe too but at least we don't clean by default. That will do it. Thanks

ogayot (approved):

documentation/Makefile
dbungert: Oh, you were suggesting the `discourse` to be the default.  I get it now, will fix.

PR #1668, opened 2023-04-21
Run probert probes in parallel - resulting in faster block probing

ogayot:


dbungert (commented):
So I'm worried about multiple tools trying to gather information from the same block device at the same time.  What sort of testing have you subjected this to?  I'm picturing running many copies of probert in parallel in a loop on a disk with ext4 and ntfs partitions.

dbungert (approved):

PR #1669, opened 2023-04-24
filesystem: fix logical+swap or logical+boot partition not considered logical

ogayot:
Some people who installed a previous version of Ubuntu on a DOS partition table ended up with a logical, ESP partition (which is marked bootable).

This is an unsupported use-case. An ESP partition should always be a primary partition. Ideally, an ESP partition should be on a GPT.
    
 However, curtin's storage_config makes it clear that expecting the flag of a logical partition to be set to 'logical' is a bad idea.
 
```
      # if the boot flag is set, use this as the flag, logical
      # flag is not required as we can determine logical via
      # partition number
```    
Other people want to create a swap partition in an extended partition. This is a perfectly valid use-case and we set flag='swap' for those partitions.

Therefore, in subiquity code, we cannot just check if a partition has a flag set to 'logical'. We need to look at the partition number too.

Going forward, it feels better to make partitions support multiple flags.

Depends on https://code.launchpad.net/~ogayot/curtin/+git/curtin/+merge/441769

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -797,7 +797,18 @@ class Partition(_Formattable):
 
     @property
     def is_logical(self):
-        return self.flag == 'logical'
+        if self.flag == "logical":
+            return True
+
+        if self.number is None:
+            # Should only be possible during initialization.
+            return False
+
+        # There is not guarantee that a logical partition will have its flag
+        # set to 'logical'. For a swap partition, for instance, the partition's
+        # flag will be set to 'swap'.  For MSDOS partitions tables, we need to
+        # check the partition number.
+        return self.device.ptable == "msdos" and self.number > 4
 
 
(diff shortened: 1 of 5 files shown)


subiquity/models/filesystem.py
```diff
@@ -797,7 +797,18 @@ def os(self):
...
+
+        if self.number is None:
+            # Should only be possible during initialization.
+            return False
+
+        # There is not guarantee that a logical partition will have its flag
+        # set to 'logical'. For a swap partition, for instance, the partition's
+        # flag will be set to 'swap'.  For MSDOS partitions tables, we need to
+        # check the partition number.
+        return self.device.ptable == "msdos" and self.number > 4
```
dbungert, line 811: A sensible compromise.  Nice work.

dbungert (approved):

PR #1672, opened 2023-05-11
canary: fixes for v2/systems call

dbungert:
* Fix label / variation name specification so that we select the correct system.
* Mount the snaps dir, the v2/systems call expects it.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -686,7 +686,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
 
     @with_context(description="configuring TPM-backed full disk encryption")
     async def setup_encryption(self, context):
-        label = self.app.base_model.source.current.snapd_system_label
+        label = self._info.label
         result = await snapdapi.post_and_wait(
             self.app.snapdapi,
             self.app.snapdapi.v2.systems[label].POST,
@@ -705,7 +705,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
 
     @with_context(description="making system bootable")
     async def finish_install(self, context):
-        label = self.app.base_model.source.current.snapd_system_label
+        label = self._info.label
         await snapdapi.post_and_wait(
             self.app.snapdapi,
             self.app.snapdapi.v2.systems[label].POST,
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -208,8 +208,9 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         await super().configured()
         self.stop_listening_udev()
 
-    async def _mount_systems_dir(self):
-        self._source_handler = self.app.controllers.Source.get_handler()
+    async def _mount_systems_dir(self, variation_name):
+        self._source_handler = \
+                self.app.controllers.Source.get_handler(variation_name)
         source_path = self._source_handler.setup()
         cur_systems_dir = '/var/lib/snapd/seed/systems'
         source_systems_dir = os.path.join(source_path, cur_systems_dir[1:])
@@ -231,9 +232,9 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             self._source_handler.cleanup()
             self._source_handler = None
 
-    async def _get_system(self, label):
+    async def _get_system(self, variation_name, label):
         try:
-            await self._mount_systems_dir()
+            await self._mount_systems_dir(variation_name)
         except NoSnapdSystemsOnSource:
             return None
         try:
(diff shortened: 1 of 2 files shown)


subiquity/tests/api/test_api.py
```diff
@@ -590,7 +590,11 @@ async def test_guided_v2_resize_logical(self):
 class TestCore(TestAPI):
     @timeout()
     async def test_basic_core_boot(self):
-        cfg = 'examples/simple.json'
+        cfg = self.machineConfig('examples/simple.json')
+        with cfg.edit() as data:
+            attrs = data['storage']['blockdev']['/dev/sda']['attrs']
+            # The partition layout was bugged for disk sizes < 28GiB
+            attrs['size'] = str(25 << 30)
```
mwhudson, line 597: I don't really understand this comment.

subiquity/tests/api/test_api.py
```diff
@@ -609,14 +613,15 @@ async def test_basic_core_boot(self):
             await inst.post('/storage/v2/guided', data)
             v2resp = await inst.get('/storage/v2')
             [d] = v2resp['disks']
-            [p1, p2, p3, p4, p5] = d['partitions']
-            e1 = dict(offset=1 << 20, size=1 << 20, mount=None)
+            pgs = d['partitions']
+            [p4] = match(pgs, number=4)
+            # FIXME that gap probably isn't actually supposed to be there
+            [p1, g1, p2, p3, p4] = d['partitions']
```
mwhudson, line 619: Or this one. Is this a comment about the gadget in the sample data or ...?

subiquity/server/mounter.py
```diff
@@ -66,6 +68,7 @@ def write(self, path, content):
 @attr.s(auto_attribs=True, kw_only=True)
 class Mountpoint(_MountBase):
     mountpoint: str
+    create: bool
```
mwhudson, line 71: I read "create" as an instruction here, but really it's a record of whether the mountpoint was created by the mount code, right? so "created" would at a minimum be a better name. Also it should probably default to False?

subiquity/server/mounter.py
```diff
@@ -121,9 +124,17 @@ async def mount(self, device, mountpoint, options=None, type=None):
             opts.extend(['-o', options])
         if type is not None:
             opts.extend(['-t', type])
+        if os.path.exists(mountpoint):
+            create = False
+        else:
+            if os.path.isdir(device):
```
mwhudson, line 130: Wait a minute. If device is a block device this isn't going to do the right thing at all?

subiquity/server/controllers/filesystem.py
```diff
@@ -638,8 +646,9 @@ async def guided_core_boot(self, disk: Disk):
             else:
                 if structure.role == snapdapi.Role.SYSTEM_DATA and \
                    structure == self._on_volume.structure[-1]:
-                    gap = gaps.largest_gap(disk)
+                    gap = gaps.at_offset(disk, offset)
                     size = gap.size - (offset - gap.offset)
+                    assert size > 0
```
mwhudson, line 651: These changes don't seem quite the same thing as the others in this branch.

mwhudson (commented):
The changes around passing the variation name around make sense to me. The mounter changes I have some questions about, although I can well believe that some changes are needed in this area...

subiquity/server/mounter.py
dbungert: I do believe this part is correct, though was inconsistent between here and in the removal.  Please see https://github.com/canonical/subiquity/pull/1673

In local tests, bind mounting a device over a file does seem to DTRT.  My test was to go to a tempdir, touch a file, bind mount a drive to that file, then run `sfdisk -J` against the bind mount source and destination and get reasonable results.

subiquity/tests/api/test_api.py
dbungert: Hopefully this commit message is clearer: https://github.com/canonical/subiquity/pull/1674/commits/19ef465d48dcaafcd385103885bfe9497c4dba1e

subiquity/tests/api/test_api.py
dbungert: Updated comment in https://github.com/canonical/subiquity/pull/1674

subiquity/server/controllers/filesystem.py
dbungert: Fair.  Moved to https://github.com/canonical/subiquity/pull/1674

mwhudson (approved):
This looks OK now, thanks :-) Hard to be sure it's complete but it's certainly better.

FilesystemController really is an astonishing bag of stuff. Would be nice to split it up somehow (no idea how right now though).

mwhudson (approved):
This looks OK now, thanks :-) Hard to be sure it's complete but it's certainly better.

FilesystemController really is an astonishing bag of stuff. Would be nice to split it up somehow (no idea how right now though).

PR #1676, opened 2023-05-12
workflow: +mantic

dbungert:


ogayot (approved):
LGTM 

Can we already update the linter too?
```diff
  lint:
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
         image:
           - ubuntu-daily:jammy    # match the core snap we're running against
-          - ubuntu-daily:lunar    # latest
+          - ubuntu-daily:mantic    # latest

PR #1677, opened 2023-05-16
workflow: produce a diff of the mypy errors before/after a PR

ogayot:
A new job now runs on PRs. It will automatically run mypy on the target branch and on the source branch. It will generate a diff of the errors, showing the new ones and showing the ones that have been fixed.

It will also show a summary with the number of errors before/after the PR.

Because we have so many false positive, it makes no sense to mark this job red. So we always make it green (unless mypy can't run).

As for now, it's up to developers to go check if any new error is introduced.

If a line that used to produce a mypy error gets moved, it will be reported as:

 * a fixed error
 * a newly introduced error

This is suboptimal and ideally we should have a way to detect moves.

_NOTE: I don't think this is ideal but I'm sure we can iterate on it to make it better_

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -30,3 +30,23 @@ jobs:
     - uses: actions/checkout@v2
     - name: lint
       run: sudo ./scripts/test-in-lxd.sh ${{ matrix.image }} "make lint"
+
+  static-typing:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Install mypy and typeshed
+        run: sudo apt-get install -y python3-mypy python3-typeshed
+      - name: Checkout the source branch
+        uses: actions/checkout@v3
+      - name: Run mypy on source branch
+        run: python3 -m mypy --ignore-missing-imports --check-untyped-defs subiquity subiquitycore system_setup console_conf scripts/replay-curtin-log.py | tee /tmp/mypy-source.out
+      - name: Checkout the target/base branch
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+      - name: Run mypy on target/base branch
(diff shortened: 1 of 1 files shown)


mwhudson (approved):
I think this is a good first step in a great direction to travel. Fine to land as is IMO but would it be reasonably simple to post a comment to the PR with the before/after number of errors?

ogayot:
> I think this is a good first step in a great direction to travel. Fine to land as is IMO but would it be reasonably simple to post a comment to the PR with the before/after number of errors?

I don't have experience with github bots but I assume it would be relatively easy for someone who has done it already :)

.github/workflows/build.yaml
```diff
@@ -30,3 +30,23 @@ jobs:
...
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Install mypy and typeshed
+        run: sudo apt-get install -y python3-mypy python3-typeshed
+      - name: Checkout the source branch
+        uses: actions/checkout@v3
+      - name: Run mypy on source branch
+        run: python3 -m mypy --ignore-missing-imports --check-untyped-defs subiquity subiquitycore system_setup console_conf scripts/replay-curtin-log.py | tee /tmp/mypy-source.out
+      - name: Checkout the target/base branch
```
ogayot, line 44: I think it makes more sense to checkout the common ancestor of {source, target} rather than target itself.

dbungert:
So I think this is fine, albeit the workflow needs a tweak for the run failure.

Something I'd like is the ability to run this query locally.  That could be a follow-on PR.  Maybe a script that takes an optional branch argument but defaults to a branch of 'main' to compare against?

ogayot:
> Something I'd like is the ability to run this query locally. That could be a follow-on PR. Maybe a script that takes an optional branch argument but defaults to a branch of 'main' to compare against?

Agreed, I'll open a follow on PR. Thanks!

PR #1679, opened 2023-05-16
snapcraft: use latest curtin and probert

ogayot:
This will pull the fixes for the utf-8 partition names.


mwhudson (approved):
I haven't checked the commit hashes but this sounds like a good idea :-)

PR #1680, opened 2023-05-22
simplify InstallController.curtin_install a bit

mwhudson:
curtin_install before this assembled a list of steps and then executed them one by one. As far as I can tell/remember (I was at least partly responsible for this) there is no reason to do it this way: it feels more Pythonic to just execute the steps!

This required removing some indirection in how the config for each step was constructed, which I think also makes things a bit clearer.

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -14,7 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import asyncio
-import functools
+import copy
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


mwhudson:
I haven't tested this in a VM yet.

mwhudson:
Oops I broke CI. Can I ask you to let me know what you think of the general approach before I fix that?

ogayot (approved):
I can't remember why I ended up introducing this indirection. :man_shrugging: 

... possibly because of my obsession for using data structures for everything :)

dbungert (approved):

mwhudson:
I finally remembered how to test this in a VM and it worked fine

PR #1682, opened 2023-05-22
meta: add interactive_sections endpoint

dbungert:
A feature request for Desktop was to add the ability to retrieve, all at once, the raw interactive-sections value from autoinstall.

### DIFF
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -181,7 +181,16 @@ class MetaController:
     async def interactive_sections_GET(self) -> Optional[List[str]]:
         if self.app.autoinstall_config is None:
             return None
-        return self.app.autoinstall_config.get('interactive-sections', None)
+
+        i_sections = self.app.autoinstall_config.get(
+                'interactive-sections', None)
+        if i_sections == ['*']:
+            # expand the asterisk to the actual controller key names
+            return [controller.autoinstall_key
+                    for controller in self.app.controllers.instances
+                    if controller.interactive()]
+
+        return i_sections
 
 
 def get_installer_password_from_cloudinit_log():
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -178,6 +178,11 @@ class MetaController:
         # enabling free only mode means disabling components
         self.app.base_model.mirror.disable_components(to_disable, enable)
 
+    async def interactive_sections_GET(self) -> Optional[List[str]]:
+        if self.app.autoinstall_config is None:
+            return None
+        return self.app.autoinstall_config.get('interactive-sections', None)
+
 
 def get_installer_password_from_cloudinit_log():
     try:
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -188,7 +188,8 @@ class MetaController:
             # expand the asterisk to the actual controller key names
             return [controller.autoinstall_key
                     for controller in self.app.controllers.instances
-                    if controller.interactive()]
+                    if controller.interactive()
+                    if controller.autoinstall_key is not None]
 
         return i_sections
 
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -135,6 +135,9 @@ class API:
                 the list of components.  free-only choice must be made prior to
                 confirmation of filesystem changes"""
 
+        class interactive_sections:
+            def GET() -> Optional[List[str]]: ...
+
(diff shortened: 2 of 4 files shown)


subiquity/server/server.py
```diff
@@ -178,6 +178,11 @@ async def free_only_POST(self, enable: bool) -> None:
         # enabling free only mode means disabling components
         self.app.base_model.mirror.disable_components(to_disable, enable)
 
+    async def interactive_sections_GET(self) -> Optional[List[str]]:
+        if self.app.autoinstall_config is None:
+            return None
+        return self.app.autoinstall_config.get('interactive-sections', None)
```
ogayot, line 184: I strongly feel like:
```yaml
interactive-sections: ["*"]
```
should be deprecated in favor of
```yaml
interactive-sections: "*"
```
I would like to support the latter going forward while keeping subiquity backward compatible (with a warning emitted if the former construct is used maybe?).

Would it be a good time to initiate this change. - so we don't break the GET /interactive-sections API later on? WDYT

dbungert: I'm fine with that except I want to have the support for both `"*"` and `["*"]` long term, I feel like carrying this syntax for compat reasons is a small burdeon.  Which do you think we should document as the intended syntax?

mwhudson (changes requested):
So I have some comments here.

The first is that if we go with an interface like this, the server side should expand "*" to a full list of sections. Having the client interpret that seems like the wrong thing to do. (At some point I'd like to extend autoinstall to support something like:

```
interactive-sections:
 - *
 - "-storage"
```

to have all screens _apart_ from storage be interactive).

The other is that we have some screens that decide whether they are interactive or not based on things like whether the network is up (ssh, snaplist) although the current dynamism is to disable those screens when installing a desktop so its not an actual issue today. Conceptually it's a bit broken though.

subiquity/server/server.py
dbungert: Will explore in future PR

subiquity/server/server.py
```diff
@@ -178,6 +178,20 @@ async def free_only_POST(self, enable: bool) -> None:
...
+            return None
+
+        i_sections = self.app.autoinstall_config.get(
+                'interactive-sections', None)
+        if i_sections == ['*']:
+            # expand the asterisk to the actual controller key names
+            return [controller.autoinstall_key
+                    for controller in self.app.controllers.instances
+                    if controller.interactive()]
+
```
mwhudson, line 192: Argl. Is the autoinstall key really what the desktop team want here? Usually controller_name and autoinstall_key and model_name and the name of the endpoint (and ...) are all the same but they are not always. For example for apt configuration the autoinstall key is "apt" but the endpoint in the api is "/mirror". Which does the client actually want?

Also controller.autoinstall_key will be None for some things (they should be filtered out in this context I think). I think an api test is probably a good idea?

cc @jpnurmi

jpnurmi: According to https://ubuntu.com/server/docs/install/autoinstall-reference the sections that can be interactive are:

- locale
- refresh-installer
- keyboard
- source
- network
- proxy
- apt
- storage
- identity
- active-directory
- ubuntu-pro
- ssh
- drivers
- snaps

This list looks ok to me. It's pretty close and easily mappable to the page route names we have today.

Some of these sections don't exist in the desktop GUI at all, though, and source & drivers are on the same page, for example. Given that there's no 1:1 mapping anyway, it'd be fine to use either autoinstall keys or model names whichever makes more sense.

dbungert: API test added.  We should expand with autoinstall_key, since that's the correct value for specifying the desired sections.

subiquity/server/server.py
```diff
@@ -178,6 +178,21 @@ async def free_only_POST(self, enable: bool) -> None:
         # enabling free only mode means disabling components
         self.app.base_model.mirror.disable_components(to_disable, enable)
 
+    async def interactive_sections_GET(self) -> Optional[List[str]]:
+        if self.app.autoinstall_config is None:
```
mwhudson, line 182: One more comment (sorry), which of returning None here vs all sections do you think is better?

dbungert: I like `None` better, actually.  It more closely reflects the way the autoinstall file was written - that no interactive-sections was supplied.

mwhudson (approved):

PR #1687, opened 2023-06-06
server: fix when cloud-config has no default user

dbungert:
If a cloud-config is sent that adjusts the users but creates no default user, Subiquity crashes with a exception while mishandling the None.

cloudinit.distros.ug_util.extract_default is documented as returning None in the event that the default user cannot be found.

Example bad from the Subiquity point of view but otherwise valid cloud-config:
```yaml
users:
- name: ubuntu-server
```

### DIFF
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -629,7 +629,10 @@ class SubiquityServer(Application):
 
         self.installer_user_name = username
 
-        if self._user_has_password(username):
+        if username is None:
+            # extract_default can return None, if there is no default user
+            self.installer_user_passwd_kind = PasswordKind.NONE
+        elif self._user_has_password(username):
             # Was the password set to a random password by a version of
             # cloud-init that records the username in the log?  (This is the
             # case we hit on upgrading the subiquity snap)
--- a/subiquity/server/tests/test_server.py
+++ b/subiquity/server/tests/test_server.py
@@ -15,7 +15,7 @@
 
 import os
 import shlex
-from unittest.mock import Mock
+from unittest.mock import Mock, patch
 
 from subiquitycore.utils import run_command
 from subiquitycore.tests import SubiTestCase
@@ -27,6 +27,7 @@ from subiquity.server.server import (
     iso_autoinstall_path,
     root_autoinstall_path,
 )
+from subiquity.common.types import PasswordKind
 
 
 class TestAutoinstallLoad(SubiTestCase):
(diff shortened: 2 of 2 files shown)


mwhudson (approved):

subiquity/server/tests/test_server.py
```diff
@@ -150,3 +151,26 @@ async def test_interactive_sections_one(self):
...
+
+
+class TestDefaultUser(SubiTestCase):
+    @patch('subiquity.server.server.ug_util.normalize_users_groups',
+           return_value=(None, None))
+    @patch('subiquity.server.server.ug_util.extract_default',
+           return_value=(None, None))
+    @patch('subiquity.server.server.user_key_fingerprints',
+           side_effect=Exception('should not be called'))
+    async def test_no_default_user(self, *patched_fns):
```
ogayot, line 163: Alternatively, you can specify the second argument of patch (a.k.a. `new`) if you don't want `patched_fns` - although it's a bit obscure.

```diff
    @patch('subiquity.server.server.ug_util.normalize_users_groups',
-           return_value=(None, None))
+           Mock(return_value=(None, None)))
    @patch('subiquity.server.server.ug_util.extract_default',
-           return_value=(None, None))
+           Mock(return_value=(None, None)))
    @patch('subiquity.server.server.user_key_fingerprints',
-           side_effect=Exception('should not be called'))
+           Mock(side_effect=Exception('should not be called')))
-   async def test_no_default_user(self, *patched_fns):
+   async def test_no_default_user(self):
```

> If [patch()](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) is used as a decorator and new is omitted, the created mock is passed in as an extra argument to the decorated function.

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

ogayot (approved):

subiquity/server/tests/test_server.py
dbungert: Nice!  Implemented.

subiquity/server/server.py
```diff
@@ -629,7 +629,10 @@ def use_passwd(passwd):
 
         self.installer_user_name = username
 
-        if self._user_has_password(username):
+        if username is None:
+            # extract_default can return None, if there is no default user
+            self.installer_user_passwd_kind = PasswordKind.NONE
+        elif self._user_has_password(username):
```
blackboxsw, line 635: Makes sense, so the error we are avoiding here is `_user_has_password` fails with something like:
```
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

blackboxsw (approved):

subiquity/server/server.py
dbungert: Yes, that's exactly what happens

PR #1689, opened 2023-06-07
zfs autoinstall

dbungert:
Add basic support necessary for an autoinstall-driven ZFS root to succeed.

The early/late/swap handling needs to be migrated elsewhere, in later PRs or other components.


subiquity/models/filesystem.py
```diff
@@ -1044,6 +1044,28 @@ def can_delete(self):
         return True
 
 
+@fsobj("zpool")
+class ZPool:
+    vdevs = attributes.reflist(backlink="_constructed_device")
+    pool = attr.ib()
```
mwhudson, line 1050: It would be nice if we could add type annotations to the fs model objects, I had to go to the docs to find out that "pool" here is a name.

examples/ai-zfs-efi.yaml
```diff
@@ -0,0 +1,83 @@
...
+      - {type: zfs, id: d1_rpool_var_games, pool: d1_rpool, volume: /var/games}
+      - {type: zfs, id: d1_rpool_var_lib,   pool: d1_rpool, volume: /var/lib}
+      - {type: zfs, id: d1_rpool_var_log,   pool: d1_rpool, volume: /var/log}
+      - {type: zfs, id: d1_rpool_var_mail,  pool: d1_rpool, volume: /var/mail}
+      - {type: zfs, id: d1_rpool_var_snap,  pool: d1_rpool, volume: /var/snap}
+      - {type: zfs, id: d1_rpool_var_spool, pool: d1_rpool, volume: /var/spool}
+      - {type: zfs, id: d1_rpool_var_www,   pool: d1_rpool, volume: /var/www}
+
+      - {type: zfs, id: d1_rpool_USERDATA,  pool: d1_rpool, volume: /USERDATA,
+         properties: {canmount: off, mountpoint: /}}
```
mwhudson, line 78: I know next to nothing about zfs in detail, how do the mountpoint here and the mountpoint on the zpool interact?

mwhudson (approved):
This looks fine as far as it goes. A couple of questions but they're not blocking.

subiquity/models/filesystem.py
dbungert: Done

subiquity/models/filesystem.py
```diff
@@ -1763,7 +1788,13 @@ def needs_bootloader_partition(self):
...
 
     def _mount_for_path(self, path):
-        return self._one(type='mount', path=path)
+        mount = self._one(type='mount', path=path)
+        if mount is not None:
+            return mount
+        zpool = self._one(type='zpool', mountpoint=path)
+        if zpool is not None:
+            return zpool
+        return None
```
ogayot, line 1797: minor: all these lines can be rewritten as:

```python
return self._one(type='zpool', mountpoint=path)
```

but I guess it can be intentional?

subiquity/server/controllers/shutdown.py
```diff
@@ -127,7 +128,10 @@ async def copy_logs_to_target(self, context):
             log.exception("saving journal failed")
 
     @with_context(description='mode={self.mode.name}')
-    def shutdown(self, context):
+    async def shutdown(self, context):
+        await self.app.hub.abroadcast(InstallerChannels.PRE_SHUTDOWN)
```
ogayot, line 132: [how do we guarantee / is there a guarantee] that FilesystemController._pre_shutdown can run before we exit the application?

mwhudson: Create an ephemeral systemd unit that depends on shutdown.target or whatever it is?

PR #1690, opened 2023-06-08
Add controller + model listing OEM meta-packages

ogayot:
This is one step towards what we need for OEM. Natural next steps are:

* Making sure that the drivers screen does not list OEM meta-packages (metapackages are not drivers and should be installed unconditionally).
  * the `--no-oem` option could come to the rescue but currently we use the ubuntu-drivers command from the target rootfs. Old ISOs will not have it and will fail on installer refresh.
* Installing the OEM meta-packages. This is tricky because:
  * running `ubuntu-drivers install oem-somerville-tentacool-meta` or alike does not work - the package does not get installed
  * running `ubuntu-drivers install` will install third-party drivers even if the user decided not to do so.

### DIFF
--- a/subiquity/server/controllers/oem.py
+++ b/subiquity/server/controllers/oem.py
@@ -1,0 +1,76 @@
+# Copyright 2023 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import asyncio
+import logging
+from typing import Optional
+
+from subiquitycore.context import with_context
+
+from subiquity.server.apt import OverlayCleanupError
+from subiquity.server.controller import SubiquityController
+from subiquity.server.types import InstallerChannels
+from subiquity.server.ubuntu_drivers import (
+    CommandNotFoundError,
+    get_ubuntu_drivers_interface,
+    )
+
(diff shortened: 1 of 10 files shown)


subiquity/server/controllers/oem.py
```diff
@@ -0,0 +1,84 @@
...
+
+    model_name = "oem"
+
+    def __init__(self, app) -> None:
+        super().__init__(app)
+        # At this point, the source variant has not been selected but it only
+        # has an impact if we're listing drivers, not OEM metapackages.
+        self.ubuntu_drivers = get_ubuntu_drivers_interface(self.app)
+
+        self.task: Optional[asyncio.Task] = None
```
dbungert, line 47: nitpick: a more verbose name on self.task might be nice, where we have one task we tend to accumulate more.

dbungert (approved):

subiquity/server/controllers/oem.py
ogayot: Updated, thanks!

dbungert (approved):

mwhudson (commented):
Hi I realize I am late coming to this and sorry about that, but what is the API actually for here? Running list-oem when apt is configured --> great, telling the client about it --> why? Didn't I browbeat you into agreeing that we just install the oem packages without any client involvement?

ogayot:
> Hi I realize I am late coming to this and sorry about that, but what is the API actually for here? Running list-oem when apt is configured --> great, telling the client about it --> why? Didn't I browbeat you into agreeing that we just install the oem packages without any client involvement?

Yes, we agreed on that part. The API endpoint is probably not useful for the client and I did not implement it at first. It felt like a natural way to check what's going on (mostly for us) but I don't mind reverting the last patch (which adds the GET endpoint) if needed. What do you think?

mwhudson:
> > Hi I realize I am late coming to this and sorry about that, but what is the API actually for here? Running list-oem when apt is configured --> great, telling the client about it --> why? Didn't I browbeat you into agreeing that we just install the oem packages without any client involvement?
> 
> Yes, we agreed on that part. The API endpoint is probably not useful for the client and I did not implement it at first. It felt like a natural way to check what's going on (mostly for us) but I don't mind reverting the last patch (which adds the GET endpoint) if needed. What do you think?

Eh it's implemented already and it lets you write tests it may as well stay. Just wanted to check that's all :)

PR #1693, opened 2023-06-11
rename GuidedStorageResponseV2.possible to targets

mwhudson:
Fairly soon v2/guided will return information about guided targets that are not, in fact, possible.


mwhudson:
cc @jpnurmi @dbungert

PR #1694, opened 2023-06-12
return some information about disallowed guided options

mwhudson:
This depends on a couple of branches I still have in review.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -134,6 +134,21 @@ class CapabilityInfo:
     allowed: List[GuidedCapability] = attr.Factory(list)
     disallowed: List[GuidedDisallowedCapability] = attr.Factory(list)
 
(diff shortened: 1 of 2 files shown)


subiquity/common/types.py
```diff
@@ -406,7 +418,8 @@ class GuidedResizeValues:
 @attr.s(auto_attribs=True)
 class GuidedStorageTargetReformat:
     disk_id: str
-    capabilities: List[GuidedCapability]
+    allowed: List[GuidedCapability]
```
dbungert, line 421: There can be scenarios where a given Target type will not ever be allowed.  For example, imagine a disk that is too small, nothing ever will change to make that into an allowed capability.  Suggest giving it an default empty list value.

dbungert (approved):

subiquity/common/types.py
mwhudson: OK

PR #1697, opened 2023-06-15
support for creating only a reset partition

mwhudson:
This is not very pretty but it works, and perhaps isn't toooooo bad?

### DIFF
--- a/scripts/validate-yaml.py
+++ b/scripts/validate-yaml.py
@@ -78,10 +78,10 @@ class StorageChecker:
 def main():
     storage_checker = StorageChecker()
 
-    final_checks = True
+    root_mount = True
 
     if sys.argv[1:2] == ['--no-root-mount']:
-        final_checks = False
+        root_mount = False
         sys.argv.pop(1)
 
     actions = []
@@ -96,7 +96,7 @@ def main():
             print('checking {} failed'.format(action))
             raise
 
-    if final_checks:
+    if root_mount:
         storage_checker.final_checks()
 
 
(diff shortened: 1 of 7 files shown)


dbungert (commented):
I think this is fine but I'd like some sort of test automation on this.  Perhaps like what's going on in subiquity/server/controllers/tests/test_filesystem.py TestGuidedV2?  We also have some autoinstall tests going on in the API tests, either of those could probably do the trick.

mwhudson:
Bah the new test is failing. It works on my machine!

scripts/validate-yaml.py
```diff
@@ -75,11 +75,15 @@ def final_checks(self):
...
 
-
-
 def main():
     storage_checker = StorageChecker()
 
+    final_checks = True
+
+    if sys.argv[1:2] == ['--no-root-mount']:
+        final_checks = False
```
dbungert, line 84: nitpick: the connection to the generic `final_checks` and this arg name feels non-obvious.  Suggest renaming to `root_mount`.

subiquity/server/controllers/install.py
```diff
@@ -259,7 +259,14 @@ async def run_curtin_step(name, stages, step_config, source=None):
 
         await run_curtin_step(name="initial", stages=[], step_config={})
 
-        if fs_controller.is_core_boot_classic():
+        if fs_controller.reset_partition_only:
```
dbungert, line 262: This method is getting a bit long.  I suggest we start splitting this method up, maybe by creating an install type and doing single dispatch stuff.  Doesn't have to be this PR.

dbungert (approved):

subiquity/server/controllers/install.py
mwhudson: yeah, it's a long series of involved function calls that are very similar but not quite the same

PR #1699, opened 2023-06-15
Zfs guided

dbungert:
Due to lack of action ordering, BIOS boot should be used.

### DIFF
--- a/subiquity/common/filesystem/labels.py
+++ b/subiquity/common/filesystem/labels.py
@@ -23,6 +23,8 @@ from subiquity.models.filesystem import (
     LVM_VolGroup,
     Partition,
     Raid,
+    ZFS,
+    ZPool,
     )
 
 
@@ -152,6 +154,11 @@ def _desc_lv(lv):
 def _desc_gap(gap):
     # This is only used in text "cannot add partition {desc}"... bit hackish.
     return _("to gap")
+
+
+@desc.register(ZPool)
+def _desc_zpool(zpool):
+    return _("zpool")
 
 
 @functools.singledispatch
@@ -336,3 +343,19 @@ def _for_client_partition(partition, *, min_size=0):
 @for_client.register(gaps.Gap)
 def _for_client_gap(gap, *, min_size=0):
     return types.Gap(offset=gap.offset, size=gap.size, usable=gap.usable)
+
+
+@for_client.register(ZPool)
+def _for_client_zpool(zpool, *, min_size=0):
+    return types.ZPool(
+        pool=zpool.pool,
+        mountpoint=zpool.mountpoint,
+        zfses=[for_client(zfs) for zfs in zpool.zfses],
+        pool_properties=zpool.pool_properties,
+        fs_properties=zpool.fs_properties,
+        )
+
+
+@for_client.register(ZFS)
+def _for_client_zfs(zfs, *, min_size=0):
+    return types.ZFS(volume=zfs.volume, properties=zfs.properties)
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -325,6 +325,8 @@ class GuidedCapability(enum.Enum):
     DIRECT = enum.auto()
     LVM = enum.auto()
     LVM_LUKS = enum.auto()
+    ZFS = enum.auto()
+
     CORE_BOOT_ENCRYPTED = enum.auto()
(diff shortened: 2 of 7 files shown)


subiquity/common/filesystem/labels.py
```diff
@@ -336,3 +342,14 @@ def _for_client_partition(partition, *, min_size=0):
...
 def _for_client_gap(gap, *, min_size=0):
     return types.Gap(offset=gap.offset, size=gap.size, usable=gap.usable)
+
+
+@for_client.register(ZPool)
+def _for_client_zpool(zpool, *, min_size=0):
+    return types.ZPool(
+        pool=zpool.pool,
+        mountpoint=zpool.mountpoint,
+        zfses=[for_client(zfs) for zfs in zpool.zfses],
```
mwhudson, line 352: Er. there's no for_client implementation for Zfs objects yet?

subiquity/common/types.py
```diff
@@ -290,6 +290,15 @@ class Partition:
     path: Optional[str] = None
 
 
+@attr.s(auto_attribs=True)
+class ZPool:
+    pool: str
+    mountpoint: str
+    zfses: Optional[str] = None  # wrong but I'm not doing the ZFS objects yet
```
mwhudson, line 297: Er well ok. Optional[str] definitely seems like the wrong annotation though? It's a list of ... well... somethings?

mwhudson (commented):
Mostly fine. The edges around the bits you haven't done yet are a bit rough though!

subiquity/common/filesystem/labels.py
dbungert: Done

subiquity/common/types.py
dbungert: Done

subiquity/common/types.py
```diff
@@ -325,6 +340,8 @@ class GuidedCapability(enum.Enum):
     DIRECT = enum.auto()
     LVM = enum.auto()
     LVM_LUKS = enum.auto()
+    ZFS = enum.auto()
```
mwhudson, line 343: I presume it's the right call that ZFS.supports_manual_customization() is False? It makes sense but I just wanted to check given that that method appeared after you started working on this branch.

subiquity/common/types.py
```diff
@@ -349,6 +366,9 @@ def supports_manual_customization(self) -> bool:
                         GuidedCapability.LVM,
                         GuidedCapability.LVM_LUKS]
 
+    def is_zfs(self) -> bool:
```
mwhudson, line 369: I guess this method is worth including because there will be GuidedCapability.ZFS_ENCRYPTED at some point?

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -618,11 +654,8 @@ async def test_small_blank_disk(self, bootloader, ptable):
                 disabled_cap.capability
                 for disabled_cap in resp.targets[0].disallowed
             },
-            {
-                GuidedCapability.DIRECT,
-                GuidedCapability.LVM,
-                GuidedCapability.LVM_LUKS,
-            })
+            set(default_capabilities)
+            )
```
mwhudson, line 658: I've refactored this test to be better in one of my branches fwiw.

mwhudson (approved):
Looks good to me. I guess it shouldn't be too hard to have a swing at supporting this in the server UI now?

mwhudson:
> Looks good to me. I guess it shouldn't be too hard to have a swing at supporting this in the server UI now?

In a follow on branch, to be clear.

subiquity/common/types.py
dbungert: We could enable it but it's sort of a weird semi-supported state.  I think it would make more sense when we have zfs manual partitioning support.  For now all you'd be able to do would be to configure other partitions on secondary disks.

I lean toward not including ZFS in supports_manual_customization until we have manual zfs config support.

subiquity/common/types.py
dbungert: Yes, I thought it best to add this method now, as I do want to handle the ENCRYPTED case this cycle.

subiquity/common/types.py
mwhudson: yes that's fair. it's not like the desktop client cares yet :-)

PR #1700, opened 2023-06-21
Install (and optionally upgrade) available OEM meta-packages

ogayot:
* Available OEM meta-packages are now installed unconditionally.
* If online, the meta-packages will also get upgraded to their version from the OEM archive - essentially pulling a kernel with it.
* We don't call `ubuntu-drivers install --no-oem` because it would also install third-party drivers on systems affected by https://bugs.launchpad.net/ubuntu/+source/ubuntu-drivers-common/+bug/1966413
* We don't call `ubuntu-drivers --no-oem install` either because it would crash on affected systems
* Instead we implement more or less the same logic as in ubuntu-drivers-common - using subsequent apt commands

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -306,6 +306,36 @@ class InstallController(SubiquityController):
                 source=source,
                 )
             await self.setup_target(context=context)
+
+            # For OEM, we basically mimic what ubuntu-drivers does:
+            # 1. Install each package with apt-get install
+            # 2. For each package, run apt-get update using only the source
+            # installed by said package.
+            # 3. Run apt-get install again for each package. This will upgrade
+            # them to the version found in the OEM archive.
+
+            # NOTE In ubuntu-drivers, this is done in a single call to apt-get
+            # install.
+            for pkg in self.model.oem.metapkgs:
+                await self.install_package(package=pkg)
+
+            if self.model.network.has_network:
+                for pkg in self.model.oem.metapkgs:
+                    source_list = f"/etc/apt/sources.list.d/{pkg}.list"
+                    await run_curtin_command(
+                        self.app, context,
+                        "in-target", "-t", self.tpath(), "--",
+                        "apt-get", "update",
+                        "-o", f"Dir::Etc::SourceList={source_list}",
+                        "-o", "Dir::Etc::SourceParts=/dev/null",
+                        "--no-list-cleanup",
+                        private_mounts=False)
+
+                # NOTE In ubuntu-drivers, this is done in a single call to
+                # apt-get install.
+                for pkg in self.model.oem.metapkgs:
+                    await self.install_package(package=pkg)
+
             await run_curtin_step(
                 name="curthooks", stages=["curthooks"],
                 step_config=self.generic_config(),
--- a/subiquity/server/tests/test_ubuntu_drivers.py
+++ b/subiquity/server/tests/test_ubuntu_drivers.py
@@ -42,7 +42,7 @@ class TestUbuntuDriversInterface(unittest.IsolatedAsyncioTestCase):
             "--recommended",
             ])
         self.assertEqual(ubuntu_drivers.install_drivers_cmd, [
-            "ubuntu-drivers", "install"])
+            "ubuntu-drivers", "install", "--no-oem"])
 
         ubuntu_drivers = UbuntuDriversInterface(self.app, gpgpu=True)
         self.assertEqual(ubuntu_drivers.list_drivers_cmd, [
@@ -50,7 +50,7 @@ class TestUbuntuDriversInterface(unittest.IsolatedAsyncioTestCase):
             "--recommended", "--gpgpu",
             ])
         self.assertEqual(ubuntu_drivers.install_drivers_cmd, [
-            "ubuntu-drivers", "install", "--gpgpu"])
+            "ubuntu-drivers", "install", "--no-oem", "--gpgpu"])
 
     @patch.multiple(UbuntuDriversInterface, __abstractmethods__=set())
     @patch("subiquity.server.ubuntu_drivers.run_curtin_command")
@@ -63,7 +63,7 @@ class TestUbuntuDriversInterface(unittest.IsolatedAsyncioTestCase):
                 self.app, "installing third-party drivers",
                 "in-target", "-t", "/target",
                 "--",
-                "ubuntu-drivers", "install",
+                "ubuntu-drivers", "install", "--no-oem",
                 private_mounts=True,
                 )
 
--- a/subiquity/server/ubuntu_drivers.py
+++ b/subiquity/server/ubuntu_drivers.py
@@ -44,8 +44,13 @@ class UbuntuDriversInterface(ABC):
             "ubuntu-drivers", "list",
             "--recommended",
         ]
+        # Because of LP #1966413, the following command will also install
+        # relevant OEM meta-packages on affected ubuntu-drivers-common
+        # versions (unless --gpgpu is also passed).
+        # This is not ideal but should be acceptable because we want OEM
+        # meta-packages installed unconditionally (except in autoinstall).
         self.install_drivers_cmd = [
-            "ubuntu-drivers", "install",
+            "ubuntu-drivers", "install", "--no-oem",
         ]
         if gpgpu:
             self.list_drivers_cmd.append("--gpgpu")
--- a/subiquity/server/ubuntu_drivers.py
+++ b/subiquity/server/ubuntu_drivers.py
@@ -73,7 +73,7 @@ class UbuntuDriversInterface(ABC):
         """ Parse the output of ubuntu-drivers list --recommended and return a
         list of drivers. """
         drivers: List[str] = []
-        # Drivers are listed one per line, but each driver is followed by a
+        # Drivers are listed one per line, but some drivers are followed by a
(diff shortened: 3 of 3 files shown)


subiquity/server/apt.py
```diff
@@ -99,8 +99,10 @@ class AptConfigurer:
     #
     # 1. Removing /cdrom from the installed system.
     #
-    # 2. Copying /etc/apt from the 'configured' overlay to the installed
-    #    system.
+    # 2. Copying a set of diverted /etc/apt files from the 'configured' overlay
+    #    to the installed system. Some packages such as OEM meta-packages and
+    #    keyrings will install files in /etc/apt so we cannot simply discard
+    #    everything.
```
mwhudson, line 105: Whelp -- bet this was fun to diagnose!

subiquity/server/controllers/install.py
```diff
@@ -306,6 +306,36 @@ async def run_curtin_step(name, stages, step_config, source=None):
...
+            if self.model.network.has_network:
+                for pkg in self.model.oem.metapkgs:
+                    source_list = f"/etc/apt/sources.list.d/{pkg}.list"
+                    await run_curtin_command(
+                        self.app, context,
+                        "in-target", "-t", self.tpath(), "--",
+                        "apt-get", "update",
+                        "-o", f"Dir::Etc::SourceList={source_list}",
+                        "-o", "Dir::Etc::SourceParts=/dev/null",
+                        "--no-list-cleanup",
```
mwhudson, line 331: TIL --no-list-cleanup, that's interesting!

subiquity/server/controllers/install.py
```diff
@@ -306,6 +306,36 @@ async def run_curtin_step(name, stages, step_config, source=None):
...
+            # them to the version found in the OEM archive.
+
+            # NOTE In ubuntu-drivers, this is done in a single call to apt-get
+            # install.
+            for pkg in self.model.oem.metapkgs:
+                await self.install_package(package=pkg)
+
+            if self.model.network.has_network:
+                for pkg in self.model.oem.metapkgs:
+                    source_list = f"/etc/apt/sources.list.d/{pkg}.list"
```
mwhudson, line 324: I was going to ask if we can be sure that this is what the sources list entry is called, but I see this code is from ubuntu-drivers so presumably yes!

mwhudson (commented):
This is nice, and actually a lot less painful than I thought it would be. Having to copy/paste bits of ubuntu-drivers-common is unfortunate but I guess it's not actually that much code in the end.

I do think you should fix the "might install two kernels" thing before merging though.

mwhudson:
Oh I forgot to say -- might be worth splitting the server/apt.py changes out to a separate PR.

subiquity/server/controllers/install.py
```diff
@@ -306,6 +306,36 @@ async def run_curtin_step(name, stages, step_config, source=None):
...
+                        self.app, context,
+                        "in-target", "-t", self.tpath(), "--",
+                        "apt-get", "update",
+                        "-o", f"Dir::Etc::SourceList={source_list}",
+                        "-o", "Dir::Etc::SourceParts=/dev/null",
+                        "--no-list-cleanup",
+                        private_mounts=False)
+
+                # NOTE In ubuntu-drivers, this is done in a single call to
+                # apt-get install.
```
mwhudson, line 335: I think (but am not sure) that in any realistic scenario there is only one metapackage for any system. Do you know differently?

subiquity/server/controllers/install.py
ogayot: Yes, I'm not sure either. I originally wanted to parse the output of `dpkg -L` or similar but I ended up taking the ubuntu-drivers approach instead.

ogayot:
Thanks for the review!

> This is nice, and actually a lot less painful than I thought it would be. Having to copy/paste bits of ubuntu-drivers-common is unfortunate but I guess it's not actually that much code in the end.

Agreed, I'm a bit worried that implementations might divert in the future so I hope we stay on sync with ubuntu-drivers. It would be good to add a `install-oem` command of some sort to `ubuntu-drivers` but realistically we wouldn't be able to use it on old ISOs and old ubuntu verisons, so I don't know if we should try.

> 
> I do think you should fix the "might install two kernels" thing before merging though.

Agreed. I've made progress on that part yesterday and I believe I have something that could work on ubuntu-desktop but does not on ubuntu-server. More on that later :)

> h I forgot to say -- might be worth splitting the server/apt.py changes out to a separate PR.

Good point, I'll go ahead and do that.

subiquity/server/controllers/install.py
ogayot: I'm not sure. I haven't seen the scenario where we would have more than one in practice but I only played around with a single certified laptop. I guess I should add a note somewhere that only one (or zero) is expected - but I don't think we should bail if ubuntu-drivers gives us two.

mwhudson:
> Agreed. I've made progress on that part yesterday and I believe I have something that could work on ubuntu-desktop but does not on ubuntu-server. More on that later :)

OK, that sounds a bit confusing but I'll wait for you to enlighten me!

> Agreed, I'm a bit worried that implementations might divert in the future so I hope we stay on sync with ubuntu-drivers. It would be good to add a `install-oem` command of some sort to `ubuntu-drivers` but realistically we wouldn't be able to use it on old ISOs and old ubuntu verisons, so I don't know if we should try.

I think we probably should. We aren't going to be making Jammy ISOs for *that* much longer and we do have a mechanism for holding updates back from older releases.

ogayot:
> > Agreed, I'm a bit worried that implementations might divert in the future so I hope we stay on sync with ubuntu-drivers. It would be good to add a `install-oem` command of some sort to `ubuntu-drivers` but realistically we wouldn't be able to use it on old ISOs and old ubuntu verisons, so I don't know if we should try.
> 
> I think we probably should. We aren't going to be making Jammy ISOs for _that_ much longer and we do have a mechanism for holding updates back from older releases.

OK, I filed a feature request against ubuntu-drivers: 

https://bugs.launchpad.net/ubuntu/+source/ubuntu-drivers-common/+bug/2024883

ogayot:
Rebased on top of main to drop from the diff the APT changes that were merged as part of #1701

ogayot:
We seem to be missing a call to `apt-get update` prior to running `ubuntu-drivers list-oem`. Some of the OEM meta-packages (e.g., oem-sutton-balin-meta) are not available in the package index when the installer boots.

mwhudson:
> We seem to be missing a call to `apt-get update` prior to running `ubuntu-drivers list-oem`. Some of the OEM meta-packages (e.g., oem-sutton-balin-meta) are not available in the package index when the installer boots.

Hm I think this might be more of a race condition, perhaps. 

We run apt-get update before the install (InstallController.install -> InstallController.configure_apt -> AptConfigurer.configure_for_install) and then send the APT_CONFIGURED signal, which triggers the OEM controller to start looking for metapackages, but nothing waits on the task that starts that I can see? It seems a bit unlikely that ubuntu-drivers list-oem would take longer than block-meta and extract but could that explain what you are seeing?

ogayot:
> > We seem to be missing a call to `apt-get update` prior to running `ubuntu-drivers list-oem`. Some of the OEM meta-packages (e.g., oem-sutton-balin-meta) are not available in the package index when the installer boots.
> 
> Hm I think this might be more of a race condition, perhaps.
> 
> We run apt-get update before the install (InstallController.install -> InstallController.configure_apt -> AptConfigurer.configure_for_install) and then send the APT_CONFIGURED signal, which triggers the OEM controller to start looking for metapackages, but nothing waits on the task that starts that I can see? It seems a bit unlikely that ubuntu-drivers list-oem would take longer than block-meta and extract but could that explain what you are seeing?

Yes, I think that's possible. It might be about the gymnastics I'm doing to test OEM on non-certified hardware - I'm faking the output of some commands so maybe it does have an impact.

subiquity/server/controllers/install.py
```diff
@@ -306,6 +306,36 @@ async def run_curtin_step(name, stages, step_config, source=None):
...
+            # installed by said package.
+            # 3. Run apt-get install again for each package. This will upgrade
+            # them to the version found in the OEM archive.
+
+            # NOTE In ubuntu-drivers, this is done in a single call to apt-get
+            # install.
+            for pkg in self.model.oem.metapkgs:
+                await self.install_package(package=pkg)
+
+            if self.model.network.has_network:
```
dbungert, line 322: If someone is offline (as in an unfortunately timed network drop), either temporarily or (from the point of view of the install) permanently, what is the desired behavior?

mwhudson: For the general ISO, I doubt we'll include metapackages in the pool, so they won't get suggested. For an OEM image, all the relevant metapackages and their dependencies will be in the pool, so installing them will install all the relevant drivers. It will also install the sources.list.d entry, so we shouldn't run apt-get update as it will fail (hmm... I wonder if that will be a problem the way things are today?)

This is all a bit more complicated than I thought when starting this reply so maybe it should go in a comment somewhere :-)

ogayot: I'll add a comment regarding the pool.

If we're temporarily offline, then there will be retries when trying to download the required packages - and then we will bail. However, I haven't implemented a retry for the apt-get call - and there isn't any in the `ubuntu-drivers install` command either.

dbungert (approved):

PR #1701, opened 2023-06-22
apt: reset only diverted files on deconfigure

ogayot:
Before starting the actual installation, we make changes to etc/apt directory in an overlay so that the packages are pulled from the media instead of the network.

At the end of installation, we need to revert these changes. We used to replace the etc/apt directory on the target by the etc/apt directory from the configured overlay.

This is a simple way to revert the temporary changes. Unfortunately, it also removes configuration files that extra packages may have installed in etc/apt. For instance, when installing an OEM meta-package, a etc/apt/sources.list.d/<oem-meta-package>.list gets installed.

If we reset etc/apt to its content from the configured overlay, we lose this file.

Fixed by selectively discarding changes to the files that subiquity modified.

### DIFF
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -99,8 +99,10 @@ class AptConfigurer:
     #
     # 1. Removing /cdrom from the installed system.
     #
-    # 2. Copying /etc/apt from the 'configured' overlay to the installed
-    #    system.
+    # 2. Copying a set of diverted /etc/apt files from the 'configured' overlay
+    #    to the installed system. Some packages such as OEM meta-packages and
+    #    keyrings will install files in /etc/apt so we cannot simply discard
+    #    everything.
     #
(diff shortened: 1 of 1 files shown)


mwhudson (approved):
Thanks for splitting this out. I'm tempted to suggest (and knowing what you're like you probably thought of this too) the introduction of some kind of stateful object to track the customizations we make to the apt dir and that knows how to undo them but I think we should resist that until / unless this all gets a bit more involved!

ogayot:
> I'm tempted to suggest (and knowing what you're like you probably thought of this too) the introduction of some kind of stateful object to track the customizations we make to the apt dir and that knows how to undo them but I think we should resist that until / unless this all gets a bit more involved!

I resisted the urge to go for this approach :) But yes, just like we do for mounts, I wanted to build a stack of `AptDivert` objects that get undone in reverse order.

PR #1702, opened 2023-06-23
Override requested kernel flavor on certified hardware

ogayot:
On certified hardware, we now honor the Ubuntu-Oem-Kernel-Flavour attribute. If the value is "oem", then we install the OEM kernel flavor. If the value if "default", then we install the kernel flavor that was originally requested.

This was a problem on ubuntu-server because the OEM meta-packages that have `Ubuntu-Oem-Kernel-Flavor: default` have a Depends on `linux-hwe-generic` whereas in ubuntu server, the default is to use the `generic` flavor (not hwe). We decided to disable listing and installation of OEM meta-packages in the server installer (and also if we're running on core boot classic).

Depends: #1700

### DIFF
--- a/subiquity/server/controllers/kernel.py
+++ b/subiquity/server/controllers/kernel.py
@@ -80,6 +80,7 @@ class KernelController(NonInteractiveController):
             package = flavor_to_pkgname(flavor, dry_run=dry_run)
         log.debug(f'Using kernel {package} due to autoinstall')
         self.model.metapkg_name = package
+        self.model.explicitly_requested = True
 
     def make_autoinstall(self):
         return {'package': self.model.metapkg_name}
--- a/subiquity/server/controllers/kernel.py
+++ b/subiquity/server/controllers/kernel.py
@@ -16,9 +16,8 @@
 import logging
 import os
 
-from subiquitycore.lsb_release import lsb_release
-
 from subiquity.server.controller import NonInteractiveController
+from subiquity.server.kernel import flavor_to_pkgname
 
 log = logging.getLogger("subiquity.server.controllers.kernel")
 
@@ -75,18 +74,10 @@ class KernelController(NonInteractiveController):
         package = data.get('package')
         flavor = data.get('flavor')
         if package is None:
-            if flavor is None or flavor == 'generic':
-                package = 'linux-generic'
-            else:
-                if flavor == 'hwe':
-                    flavor = 'generic-hwe'
-                # Should check this package exists really but
-                # that's a bit tricky until we get cleverer about
-                # the apt config in general.
-                dry_run: bool = self.app.opts.dry_run
-                package = 'linux-{flavor}-{release}'.format(
-                    flavor=flavor,
-                    release=lsb_release(dry_run=dry_run)['release'])
+            dry_run: bool = self.app.opts.dry_run
+            if flavor is None:
+                flavor = 'generic'
+            package = flavor_to_pkgname(flavor, dry_run=dry_run)
         log.debug(f'Using kernel {package} due to autoinstall')
         self.model.metapkg_name = package
 
--- a/subiquity/server/controllers/oem.py
+++ b/subiquity/server/controllers/oem.py
@@ -94,6 +94,16 @@ class OEMController(SubiquityController):
     async def load_metapackages_list(self, context) -> None:
         with context.child("wait_apt"):
             await self._wait_apt.wait()
+
+        # Skip looking for OEM meta-packages if we are running ubuntu-server.
+        # OEM meta-packages expect the default kernel flavor to be HWE (which
+        # is only true for ubuntu-desktop).
+        if self.app.base_model.source.current.variant == "server":
+            log.debug("not listing OEM meta-packages since we are installing"
+                      " ubuntu-server")
+            self.model.metapkgs = []
+            return
+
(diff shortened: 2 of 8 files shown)


subiquity/server/kernel.py
```diff
@@ -0,0 +1,29 @@
...
+def flavor_to_pkgname(flavor: str, *, dry_run: bool) -> str:
+    if flavor == 'generic':
+        return 'linux-generic'
+    if flavor == 'hwe':
+        flavor = 'generic-hwe'
+
+    release = lsb_release(dry_run=dry_run)['release']
+    # Should check this package exists really but
+    # that's a bit tricky until we get cleverer about
+    # the apt config in general.
```
mwhudson, line 28: Hm I guess we can probably implement this now :-)

mwhudson:
> This is a problem on ubuntu-server because the OEM meta-packages that have `Ubuntu-Oem-Kernel-Flavor: default` have a Depends on `linux-hwe-generic` whereas in ubuntu server, the default is to use the `generic` flavor (not hwe).
> 
> This means ubuntu-drivers will install the hwe flavor and subiquity will install the generic kernel, resulting in two kernels installed.

I guess the only short term solution here is to not probe for an oem metapackage when installing a source with .variant == 'server'?

mwhudson (commented):
Mostly looks fine!

subiquity/server/controllers/oem.py
```diff
@@ -59,6 +62,30 @@ async def list_and_mark_configured() -> None:
...
+        for line in result.stdout.decode("utf-8").splitlines():
+            if not line.startswith("Ubuntu-Oem-Kernel-Flavor:"):
+                continue
+
+            flavor = line.split("=", maxsplit=1)[1].strip()
+            if flavor == "oem":
+                return True
+            elif flavor == "default":
+                return False
+            else:
```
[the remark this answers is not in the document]
ogayot: Good point. Might be worth using the same logic. I'll go ahead and do that.

mwhudson (approved):
Hi, this looks great. The fact that it is reasonably simple to do this now makes all the ground work seem worth it!!

subiquity/server/controllers/oem.py
```diff
@@ -72,14 +113,34 @@ async def load_metapackages_list(self, context) -> None:
...
         except OverlayCleanupError:
             log.exception("Failed to cleanup overlay. Continuing anyway.")
+
+        for pkg in self.model.metapkgs:
+            if pkg.wants_oem_kernel:
+                kernel_model = self.app.base_model.kernel
+                kernel_model.metapkg_name_override = flavor_to_pkgname(
+                        pkg.name, dry_run=self.app.opts.dry_run)
+
+                log.debug("overriding kernel flavor because of OEM")
```
dbungert, line 135: So the theory seems to be that if we're on OEM hardware, that we use the OEM kernel no matter what?  Even if configured otherwise by autoinstall or another method?  That doesn't sound correct.

It seems that the kernel detection logic should be overriding `linux-generic`, not any value of metapkg.

ogayot: > So the theory seems to be that if we're on OEM hardware, that we use the OEM kernel no matter what? 

If we're on OEM hardware, we will use the OEM flavor or the HWE flavor (depending on the OEM meta package). There's no way around it with the current design I'm afraid. 

> Even if configured otherwise by autoinstall or another method? That doesn't sound correct.

Correct, unless we disable that behavior in autoinstall with `oem: install: false`

dbungert: Then I think we need to raise a runtime error in that case.  The only way to satisfy the request to install a specific kernel is to disable the OEM functionality.  We can detect that scenario I think - if kernel supplied by autoinstall and we have an oem kernel metapackage name, raise and point out the fix (remove the kernel autoinstall directive or use `oem: install: false`).

ogayot: Ack that sounds fine to me.

ogayot: Implemented. That said, the error message still allows the install to continue because:
* oem controller cannot be part of the `install` set. It needs the apt overlay to operate properly
* oem controller is not a `postinstall` set. Doing so would not prevent the `install` from running.

We'd need an intermediate set that is required after the apt overlays are ready.

If that's okay I'll implement this in a subsequent PR

subiquity/server/controllers/kernel.py
```diff
@@ -62,6 +62,7 @@ def start(self):
                 with open(mp_file) as fp:
                     kernel_package = fp.read().strip()
                 self.model.metapkg_name = kernel_package
+                self.model.explicitly_requested = True
```
dbungert, line 65: A fun scenario here is that if we trigger this one, it's (usually) an image construction mismatch, not (usually) autoinstall.  It could be autoinstall with early-commands. 

I think the right answer is that we don't want to set `explicitly_requested = True` on this code path.

Note that https://git.launchpad.net/casper/tree/scripts/casper-bottom/61desktop_canary_tweaks#n36 selects the HWE kernel by way of the /run/kernel-meta-package today, and presumably /etc/subiquity/kernel-meta-package on the long term.

ogayot: Good point, thanks!

dbungert (approved):
Thanks!

PR #1703, opened 2023-06-29
OEM: ensure only one kernel gets installed

ogayot:
Some more testing is needed but it looks mostly ok.

Depends: #1702

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -110,6 +110,7 @@ parts:
       # This list includes the dependencies for curtin and probert as well,
       # there doesn't seem to be any real benefit to listing them separately.
       - cloud-init
+      - dctrl-tools
       - iso-codes
       - libpython3-stdlib
       - libpython3.10-minimal
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -34,7 +34,7 @@ from subiquitycore.async_helpers import (
     )
 from subiquitycore.context import with_context
 from subiquitycore.file_util import write_file, generate_config_yaml
-from subiquitycore.utils import log_process_streams
+from subiquitycore.utils import arun_command, log_process_streams
 
 from subiquity.common.errorreport import ErrorReportKind
 from subiquity.common.types import (
@@ -50,6 +50,9 @@ from subiquity.server.controller import (
 from subiquity.server.curtin import (
     run_curtin_command,
     start_curtin_command,
+    )
+from subiquity.server.kernel import (
+    list_installed_kernels,
     )
 from subiquity.server.mounter import (
     Mounter,
@@ -312,6 +315,17 @@ class InstallController(SubiquityController):
                 step_config=self.generic_config(),
                 source=source,
                 )
+            if self.app.opts.dry_run:
+                # In dry-run, extract does not do anything. Let's create what's
+                # needed manually. Ideally, we would not hardcode
+                # var/lib/dpkg/status because it is an implementation detail.
+                status = "var/lib/dpkg/status"
+                (root / status).parent.mkdir(parents=True, exist_ok=True)
+                await arun_command([
+                    "cp", "-aT", "--",
+                    str(Path("/") / status),
+                    str(root / status),
+                ])
             await self.setup_target(context=context)
 
             # For OEM, we basically mimic what ubuntu-drivers does:
@@ -342,6 +356,12 @@ class InstallController(SubiquityController):
                 # apt-get install.
                 for pkg in self.model.oem.metapkgs:
                     await self.install_package(package=pkg.name)
+
+            # If we already have a kernel installed, don't bother requesting
+            # curthooks to install it again or we might end up with two
+            # kernels.
+            if await list_installed_kernels(Path(self.tpath())):
+                self.model.kernel.curthooks_no_install = True
 
             await run_curtin_step(
                 name="curthooks", stages=["curthooks"],
--- a/subiquity/server/kernel.py
+++ b/subiquity/server/kernel.py
@@ -13,7 +13,12 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import pathlib
+import subprocess
+from typing import List
+
 from subiquitycore.lsb_release import lsb_release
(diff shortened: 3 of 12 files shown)


dbungert (approved):
LGTM, minus the bit already raised on the other PR.

subiquity/server/controllers/install.py
```diff
@@ -336,6 +350,14 @@ async def run_curtin_step(name, stages, step_config, source=None):
                 for pkg in self.model.oem.metapkgs:
                     await self.install_package(package=pkg.name)
 
+            # If we already have the needed kernel installed, don't bother
+            # requesting curthooks to install it or we might end up with two
+            # kernels.
+            for pkg in await list_installed_kernels(Path(self.tpath())):
+                if pkg == self.model.kernel.needed_kernel:
+                    self.model.kernel.curthooks_no_install
```
mwhudson, line 358: Um I think this line is incomplete :-)

subiquity/server/kernel.py
```diff
@@ -27,3 +32,30 @@ def flavor_to_pkgname(flavor: str, *, dry_run: bool) -> str:
     # that's a bit tricky until we get cleverer about
     # the apt config in general.
     return f'linux-{flavor}-{release}'
+
+
+async def list_installed_kernels(
+        rootfs: pathlib.Path,
+        *, grep_status: str = "grep-status") -> List[str]:
```
mwhudson, line 39: This argument is unused.

snapcraft.yaml
```diff
@@ -110,6 +110,7 @@ parts:
       # This list includes the dependencies for curtin and probert as well,
       # there doesn't seem to be any real benefit to listing them separately.
       - cloud-init
+      - dctrl-tools
```
mwhudson, line 113: I guess we could just stage /usr/bin/grep-dctrl but the whole package is fairly tiny.

subiquity/server/controllers/install.py
```diff
@@ -336,6 +350,14 @@ async def run_curtin_step(name, stages, step_config, source=None):
                 for pkg in self.model.oem.metapkgs:
                     await self.install_package(package=pkg.name)
 
+            # If we already have the needed kernel installed, don't bother
+            # requesting curthooks to install it or we might end up with two
+            # kernels.
+            for pkg in await list_installed_kernels(Path(self.tpath())):
+                if pkg == self.model.kernel.needed_kernel:
```
mwhudson, line 357: This is all a bit confusing but I don't think we should check this, for two reasons. One is that pkg here will be something like "linux-image-6.2.0-23-generic" or "linux-image-6.1.0-1012-oem" and I think needed_kernel is going to be a metapackage name like "linux-oem-22.04" or "linux" and the other is that even if we think we want "linux-oem-22.04" installing the metapackage might actually install "linux-oem-22.04d" or something. 

Basically IMO we should instruct curtin to not install a kernel iff there is any kernel installed already. Luckily this makes the code simpler :-)

mwhudson: I'm not sure how things should interact if the autoinstall explicitly requests a different flavour though.

mwhudson (changes requested):

subiquity/server/controllers/install.py
```diff
@@ -336,6 +350,14 @@ async def run_curtin_step(name, stages, step_config, source=None):
                 for pkg in self.model.oem.metapkgs:
                     await self.install_package(package=pkg.name)
 
+            # If we already have the needed kernel installed, don't bother
+            # requesting curthooks to install it or we might end up with two
+            # kernels.
+            for pkg in await list_installed_kernels(Path(self.tpath())):
```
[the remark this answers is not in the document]
ogayot: I think that would only happen on broken or half broken scenarios, like:
* two suggested OEM meta-packages pulling two different kernel flavors (for now there should be only one meta-package per hardware model - in theory)
* one kernel in the squashfs (don't think we'll ever do that) and the OEM metapkg pulling another flavor

subiquity/server/controllers/install.py
ogayot: Ouch! Thanks for noticing!

snapcraft.yaml
ogayot: Yes - and no Depends so I guess I'll keep the whole package.

subiquity/server/kernel.py
ogayot: Oops!

subiquity/server/controllers/install.py
ogayot: Yeah, I think you're right. Dan was suggesting a runtime error if there's an explicit kernel flavor + an OEM meta-package. I think it is a sensible approach.

subiquity/server/controllers/install.py
```diff
@@ -312,6 +315,17 @@ async def run_curtin_step(name, stages, step_config, source=None):
                 step_config=self.generic_config(),
                 source=source,
                 )
+            if self.app.opts.dry_run:
+                # In dry-run, extract does not do anything. Let's create what's
+                # needed manually. Ideally, we would not hardcode
+                # var/lib/dpkg/status because it is an implementation detail.
+                status = Path("var/lib/dpkg/status")
+                (Path(root) / status).parent.mkdir(parents=True, exist_ok=True)
```
mwhudson, line 323: This is nitpicky to the max but a) root is already a Path here b) Path / str works fine? so you can leave status as a string.

subiquity/server/kernel.py
mwhudson: I admit I expected you to fix this by removing the argument :-) What is the use case for passing a different value here?

mwhudson (approved):
Thanks, this is good stuff. I still have a couple of comments but they are very much at the nitpicking level now.

subiquity/server/controllers/install.py
ogayot: Good point, updated :)

ogayot:
> I admit I expected you to fix this by removing the argument :-) What is the use case for passing a different value here?

It felt like it could be a way to substitute a command by another when doing dry-runs - as in
```python
list_installed_kernels(..., grep_status='scripts/replay-grep-status')
```
or something.

But it's not really usable as is - so I guess it should go (and is now gone)

PR #1705, opened 2023-07-04
move server client (mostly) to v2 guided API

mwhudson:
Preserving the v1 guided API is a bit tedious now, so get rid of it.


mwhudson:
cc @jpnurmi -- this branch changes the API but not the bits the desktop client uses I expect / hope

dbungert (approved):
Nice!

PR #1706, opened 2023-07-06
add a target/capability for manual partitioning

mwhudson:
Currently the client special cases the core boot capabilities and does not offer manual partitioning for them. This change makes things a bit more regular, and also prepares the ground for other install methods that cannot be customized.

Next up: a major rewrite of subiquity/ui/views/filesystem/guided.py!

### DIFF
--- a/subiquity/client/controllers/filesystem.py
+++ b/subiquity/client/controllers/filesystem.py
@@ -29,6 +29,7 @@ from subiquity.common.types import (
     GuidedCapability,
     GuidedChoiceV2,
     GuidedStorageResponseV2,
+    GuidedStorageTargetManual,
     GuidedStorageTargetReformat,
     StorageResponseV2,
     )
@@ -177,7 +178,10 @@ class FilesystemController(SubiquityTuiController, FilesystemManipulator):
                 await asyncio.sleep(0.1)
             self.finish()
         elif self.answers['manual']:
-            await self._guided_choice(None)
+            await self._guided_choice(
+                GuidedChoiceV2(
+                    target=GuidedStorageTargetManual(),
+                    capability=GuidedCapability.MANUAL))
             await self._run_actions(self.answers['manual'])
             self.answers['manual'] = []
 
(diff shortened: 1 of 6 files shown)


subiquity/client/controllers/filesystem.py
```diff
@@ -287,21 +291,18 @@ async def _answers_action(self, action):
...
-            self.app.next_screen(self.endpoint.guided.POST(choice))
-            return
+    async def _guided_choice(self, choice: GuidedChoiceV2):
         # FIXME It would seem natural here to pass the wait=true flag to the
         # below HTTP calls, especially because we wrap the coroutine in
         # wait_with_progress.
         # Having said that, making the server return a cached result seems like
         # the least risky option to address https://launchpad.net/bugs/1993257
         # before the kinetic release. This is also similar to what we did for
         # https://launchpad.net/bugs/1962205
```
mwhudson, line 301: @ogayot @dbungert i have a feeling I've removed the code this comment was about. Do we need to add a use_cached_result parameter to guided.POST?

mwhudson:
cc @jpnurmi for api changes -- you might need to ignore a new target type for now

jpnurmi:
nice! :+1:

subiquity/client/controllers/filesystem.py
ogayot: since https://github.com/canonical/subiquity/pull/1234 (addressing 1962205), guided.POST (the v1) always returns a cached response. I think the comment was specifically about the GET, for which we don't always have a cached response.

subiquity/common/types.py
```diff
@@ -340,6 +341,14 @@ def is_core_boot(self) -> bool:
                         GuidedCapability.CORE_BOOT_PREFER_ENCRYPTED,
                         GuidedCapability.CORE_BOOT_PREFER_UNENCRYPTED]
 
+    def supports_manual_customization(self) -> bool:
+        # After posting this capability to guided_POST, is it possible
+        # for the user to customize the layout further?
```
dbungert, line 346: I believe so but I bet there are strange corner case bugs

dbungert (approved):

subiquity/client/controllers/filesystem.py
mwhudson: Thanks for checking for me. I'll delete the comment

subiquity/common/types.py
mwhudson: Um this comment is an explanation, not a question :-)

PR #1707, opened 2023-07-06
do not do any mirror configuration when installing core

mwhudson:
The commit history is a bit messy but well. Maybe worth a look?


mwhudson:
Moved most of the changes that were initially in here to other branches, this little bit might be reviewable now.

dbungert (approved):

PR #1708, opened 2023-07-07
docs: autoinstall tutorial

dbungert:
I think much of this content should be moved into other sections, but as a start I'm bringing in the existing autoinstall doc into rst format.  I couldn't help with some editorial improvements along the way, if that becomes a problem I'll resplit.

Deviations from existing autoinstall doc:
* quick start page not listed yet, we don't have that in rst yet
* reference to autoinstall-generator dropped - that project needs some help
* Interactions between Autoinstall and Cloud-init section rewritten
* possible future direction section dropped, we should consider a proper roadmap but it surely doesn't need to be in a tutorial

Note that the docs are written assuming that they are housing all ubuntu install documentation, but is only starting with autoinstall.   I don't know if that will be awkward when we cut over.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -22,7 +22,7 @@ import sys
 # -- Project information -----------------------------------------------------
 
 project = 'Ubuntu Install Guide'
-copyright = f'Canonical Group Ltd, {datetime.date.today().year}'
+copyright = f'Canonical Group Ltd, 2020 - {datetime.date.today().year}'
 
 # -- General configuration ---------------------------------------------------
 
@@ -47,8 +47,8 @@ templates_path = ['_templates']
 # The suffix of source filenames.
 source_suffix = '.rst'
 
-# The master toctree document.
-master_doc = 'index'
+# The root toctree document.
+root_doc = 'index'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
@@ -62,7 +62,8 @@ master_doc = 'index'
 # This pattern also affects html_static_path and html_extra_path.
 
 exclude_patterns = [
-    ".sphinx/venv/*",
+    '.sphinx/venv/*',
+    'README.md',
 ]
 
 # Sphinx-copybutton config options:
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -34,7 +34,6 @@ needs_sphinx = '5.1.1'
 # ones.
 
 extensions = [
-    'm2r2',
     'sphinx.ext.intersphinx',
     'sphinx_copybutton',
     'sphinx_design',
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -22,7 +22,7 @@ import sys
 # -- Project information -----------------------------------------------------
 
 project = 'Ubuntu Install Guide'
-copyright = f'Canonical Group Ltd, 2020 - {datetime.date.today().year}'
+copyright = 'Canonical Group Ltd'
 
 # -- General configuration ---------------------------------------------------
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -35,10 +35,17 @@ needs_sphinx = '5.1.1'
 
 extensions = [
     'm2r2',
+    'sphinx.ext.intersphinx',
     'sphinx_copybutton',
     'sphinx_design',
 ]
 
+intersphinx_mapping = {
+    'cloud-init': (
+        'https://canonical-cloud-init.readthedocs-hosted.com/en/latest',
+         None
+    )
+}
 
 # Add any paths that contain templates here, relative to this directory.
 
--- a/doc/tutorial/intro-to-autoinstall.rst
+++ b/doc/tutorial/intro-to-autoinstall.rst
@@ -37,7 +37,7 @@ Providing the autoinstall configuration
 =======================================
 
 There are 2 ways to provide the autoinstall configuration:
- * Carried as part of cloud-config
+ * Provide :external+cloud-init:ref:`#cloud-config user-data<user_data_formats-cloud_config>` containing `autoinstall:` configuration directives to cloud-init at boot time
  * Directly on the install media
 
 Autoinstall by way of cloud-config
@@ -54,7 +54,7 @@ user data via the :external+cloud-init:ref:`datasource_nocloud` data source.
 When providing autoinstall via cloud-init, the autoinstall config is provided
 as :external+cloud-init:ref:`user_data_formats-cloud_config`. This
 means we need a :code:`#cloud-config` header. The autoinstall directives are
-placed under a top level :code:`autoinstall` key, like so:
+placed under a top level :code:`autoinstall:` key, like so:
 
 .. code-block:: yaml
 
@@ -98,13 +98,18 @@ level.
 Cloud-init and autoinstall interaction
 ======================================
 
-While cloud-init may provide the autoinstall configuration to the
-Ubuntu installer, it does not process the autoinstall itself.
+Cloud-init runs in both the ephemeral system, during install, and in the target
+system during first boot. Cloud-init then becomes inert for every subsequent
+reboot.
 
-If cloud-init directives are intended to modify the ephemeral system, they
-must appear at the top level of the cloud-config. If instead
-cloud-init directives are intended to modify the system being installed, they
-must appear under a :code:`user-data` section in :code:`autoinstall`.
+While cloud-init may provide the autoinstall configuration to the Ubuntu
+installer, it does not process the autoinstall directives itself.
+
+To modify the ephemeral system with cloud-init, any
+:external+cloud-init:ref:`#cloud-config module schema keys<modules>` can
+be provided. If instead cloud-init directives are intended to modify the system
+being installed, they must appear under a :ref:`ai-user-data` section under
+``autoinstall:``.
 
 .. code-block:: yaml
 
--- a/doc/tutorial/intro-to-autoinstall.rst
+++ b/doc/tutorial/intro-to-autoinstall.rst
@@ -1,10 +1,10 @@
 .. _tutorial_intro-to-autoinstall:
 
-Introduction to Autoinstall
+Introduction to autoinstall
 ***************************
 
-Ubuntu installation automation is performed with the autoinstall format.
-You might also know this feature as unattended or handsoff or preseeded
+Automatic Ubuntu installation is performed with the autoinstall format.
+You might also know this feature as "unattended", "hands-off" or "preseeded"
 installation.
 
 This format is supported in the following installers:
@@ -23,35 +23,36 @@ Differences from debian-installer preseeding
 (also known as d-i).
 
 Autoinstalls differ from preseeds in the following ways:
- * the format is different (yaml vs debconf-set-selections)
- * when the answer to a question is not present in a preseed, d-i stops and
-   asks the user for input. Autoinstalls are not like this:  by default, if
-   there is any autoinstall configuration at all, the installer takes the
-   default for any unanswered question (and fails if there is no default).
- * You can designate particular sections in the configuration as "interactive",
-   which means the installer will still stop and ask about those.
+ * The format is different: autoinstalls use YAML instead of the preseed
+   debconf-set-selections
+ * When the answer to a question is not present in a preseed, d-i stops and
+   asks the user for input. By comparison, if there is any autoinstall
+   configuration at all, the autoinstall takes the default for any
+   unanswered question (and fails if there is no default).
+ * You can designate particular sections in the autoinstall configuration as
+   "interactive", which means the installer will still stop and ask about those.
 
 
 Providing the autoinstall configuration
 =======================================
 
-There are 2 methods of providing the autoinstall configuration:
+There are 2 ways to provide the autoinstall configuration:
  * Carried as part of cloud-config
  * Directly on the install media
 
 Autoinstall by way of cloud-config
 ----------------------------------
 
-The suggested method of providing autoinstall to the Ubuntu installer is by way
-of cloud-init.  This allows the configuration to be applied to the installer
+The suggested way of providing autoinstall config to the Ubuntu installer is
+via cloud-init. This allows the configuration to be applied to the installer
 without having to modify the install media.
 
 The autoinstall config is provided via cloud-init configuration, which is
 almost endlessly flexible. In most scenarios the easiest way will be to provide
-user-data via the :external+cloud-init:ref:`datasource_nocloud` data source.
+user data via the :external+cloud-init:ref:`datasource_nocloud` data source.
 
 When providing autoinstall via cloud-init, the autoinstall config is provided
-as :external+cloud-init:ref:`user_data_formats-cloud_config`.  This
+as :external+cloud-init:ref:`user_data_formats-cloud_config`. This
 means we need a :code:`#cloud-config` header. The autoinstall directives are
 placed under a top level :code:`autoinstall` key, like so:
 
@@ -62,6 +63,12 @@ placed under a top level :code:`autoinstall` key, like so:
         version: 1
         ....
 
+.. note::
+
+   :external+cloud-init:ref:`user_data_formats-cloud_config` files must contain
+   the ``#cloud-config`` header to be recognized as a valid cloud config data
+   file.
+
 Autoinstall on the install media
 --------------------------------
 
@@ -69,17 +76,17 @@ Another option for supplying autoinstall to the Ubuntu installer is to place a
 file named :code:`autoinstall.yaml` on the install media itself.
 
 There are two potential locations for the :code:`autoinstall.yaml` file:
- * At the root of the "cdrom".  When writing the installation ISO to a USB
+ * At the root of the "cdrom". When you write the installation ISO to a USB
    Flash Drive, this can be done by copying the :code:`autoinstall.yaml` to the
-   partition containing the contents of the ISO - i.e., in the same directory
-   containing the :code:`casper` directory.
+   partition containing the contents of the ISO - i.e.,
+   in the directory containing the ``casper`` sub-directory.
  * On the rootfs of the installation system - this option will typically
    require modifying the installation ISO and is not suggested, but is
    supported.
 
 Directly specifying autoinstall as a :code:`autoinstall.yaml` file does not
 require a :code:`#cloud-config` header, and does not use a top level
-:code:`autoinstall` key.  The autoinstall directives are placed at the top
+:code:`autoinstall` key. The autoinstall directives are placed at the top
 level.
 
 .. code-block:: yaml
@@ -88,14 +95,14 @@ level.
     ....
 
 
-Cloud-init and Autoinstall interaction
+Cloud-init and autoinstall interaction
 ======================================
 
-While cloud-init may assist in providing the autoinstall configuration to the
-Ubuntu installer, cloud-init itself is not processing the autoinstall.
+While cloud-init may provide the autoinstall configuration to the
+Ubuntu installer, it does not process the autoinstall itself.
 
 If cloud-init directives are intended to modify the ephemeral system, they
-must appear at the top level of the cloud-config.  If instead
+must appear at the top level of the cloud-config. If instead
 cloud-init directives are intended to modify the system being installed, they
 must appear under a :code:`user-data` section in :code:`autoinstall`.
 
--- a/doc/tutorial/intro-to-autoinstall.rst
+++ b/doc/tutorial/intro-to-autoinstall.rst
@@ -138,6 +138,91 @@ for repeating the install is created at
 :code:`/var/log/installer/autoinstall-user-data`.
 
 
+The structure of an autoinstall configuration
+=============================================
+
+See the :ref:`ai` for full details on the supported autoinstall directives.
+
+A minimal autoinstall configuration is:
+
+.. code-block:: yaml
+
+    version: 1
+    identity:
+        hostname: hostname
+        username: username
+        password: $crypted_pass
+
+Here is an example file that shows off most features:
+
+.. parsed-literal::
+
+    :ref:`ai-version`: 1
+    :ref:`ai-reporting`:
+        hook:
+            type: webhook
+            endpoint: http\://example.com/endpoint/path
+    :ref:`ai-early-commands`:
+        - ping -c1 198.162.1.1
+    :ref:`ai-locale`: en_US
+    :ref:`ai-keyboard`:
+        layout: gb
+        variant: dvorak
+    :ref:`ai-network`:
+        network:
+            version: 2
+            ethernets:
+                enp0s25:
+                   dhcp4: yes
+                enp3s0: {}
+                enp4s0: {}
+            bonds:
+                bond0:
+                    dhcp4: yes
+                    interfaces:
+                        - enp3s0
+                        - enp4s0
+                    parameters:
+                        mode: active-backup
+                        primary: enp3s0
+    :ref:`ai-proxy`: http\://squid.internal:3128/
+    :ref:`ai-apt`:
+        primary:
+            - arches: [default]
+              uri: http\://repo.internal/
+        sources:
+            my-ppa.list:
+                source: "deb http\://ppa.launchpad.net/curtin-dev/test-archive/ubuntu $RELEASE main"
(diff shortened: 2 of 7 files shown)


doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,239 @@
...
+    identity:
+        hostname: hostname
+        username: username
+        password: $crypted_pass
+
+Here is an example file that shows off most features:
+
+.. parsed-literal::
+
+    :ref:`ai-version`: 1
```
mwhudson, line 167: I'm curious how this renders, is there some easy way for me to see this locally?

mwhudson (approved):
Looks like a good start

mwhudson (approved):

doc/tutorial/intro-to-autoinstall.rst
dbungert: `make -C doc install run`

install is setting up the venv.

I'm undecided on this part.  We could just have the standard be that in the sample autoinstalls, we always link to the reference for the top level keys.  If we do blocks as code-formatted yaml, it'll look differently than this part that allows the references.

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,239 @@
...
+
+The structure of an autoinstall configuration
+=============================================
+
+See the :ref:`ai` for full details on the supported autoinstall directives.
+
+A minimal autoinstall configuration is:
+
+.. code-block:: yaml
+
```
ogayot, line 156: I'm a bit torn on providing examples without the `#cloud-config` + autoinstall key when we explicitly suggest to use the cloud-init approach. I know it's boilerplate so we probably don't want to repeat the same thing over and over in the examples.

Maybe a little reminder before or after the examples for people tempted to copy-paste? Something like:

> /!\ Remember to add the `#cloud-config` header and place the configuration within an `autoinstall` section if you are using the cloud-init approach.

```yaml
#cloud-config
autoinstall:
  < place the autoinstall configuration here >
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,239 @@
...
+During autoinstall, a prompt will be shown to confirm that the install really
+should proceed and start making modifications to the target system. ::
+
+    start: subiquity/Meta/status_GET
+    Confirmation is required to continue.
+    Add 'autoinstall' to your kernel command line to avoid this
+
+
+    Continue with autoinstall? (yes|no)
```
ogayot, line 134: Nothing to do here but I wish we'd stop the install if the user types "no". Right now, it keeps asking the same question again until you type "yes" (at least in dry-run)

ogayot (approved):
nice work!

doc/tutorial/intro-to-autoinstall.rst
s-makin: The easiest way to see the preview is with the "build PRs" functionality of readthedocs, but this will need someone with admin permissions on the GitHub repo to set up a webhook to readthedocs. Once this has been done, any PR changes will trigger a preview rebuild which you can access via the "checks" section at the bottom of this page.

doc/tutorial/intro-to-autoinstall.rst
dbungert: A goal I have for the doc overhaul is to reduce people sending autoinstall as cloud-config, so I think something like this would help reduce it.  I'm not sure I want the "Remember to add" part all the time, but to just have 99% of the autoinstall snippets (pretty much all of them except the case showing the non-cloud-init path in this document) just always include
```
#cloud-config
autoinstall:
```
might well help reduce that confusion.

dbungert: So yes, I like the suggestion very much, but think I want to skip the reminder part and just include the above 2 lines in yaml samples.

ogayot: Thanks!

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+
+There are 2 methods of providing the autoinstall configuration:
+ * Carried as part of cloud-config
+ * Directly on the install media
+
+Autoinstall by way of cloud-config
+----------------------------------
+
+The suggested method of providing autoinstall to the Ubuntu installer is by way
+of cloud-init.  This allows the configuration to be applied to the installer
```
s-makin, line 46: You've switched from talking about cloud-config to cloud-init in a way that suggests the two are interchangeable (this might be confusing)

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+=======================================
+
+There are 2 methods of providing the autoinstall configuration:
+ * Carried as part of cloud-config
+ * Directly on the install media
+
+Autoinstall by way of cloud-config
+----------------------------------
+
+The suggested method of providing autoinstall to the Ubuntu installer is by way
```
s-makin, line 45:
```suggestion
The suggested way of providing autoinstall config to the Ubuntu installer is via
```

doc/tutorial/intro-to-autoinstall.rst
s-makin, line 46:
```suggestion
cloud-init. This allows the configuration to be applied to the installer
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+of cloud-init.  This allows the configuration to be applied to the installer
+without having to modify the install media.
+
+The autoinstall config is provided via cloud-init configuration, which is
+almost endlessly flexible. In most scenarios the easiest way will be to provide
+user-data via the :external+cloud-init:ref:`datasource_nocloud` data source.
```
s-makin, line 51:
```suggestion
user data via the :external+cloud-init:ref:`datasource_nocloud` data source.
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+
+When providing autoinstall via cloud-init, the autoinstall config is provided
+as :external+cloud-init:ref:`user_data_formats-cloud_config`.  This
```
s-makin, line 54:
```suggestion
as :external+cloud-init:ref:`user_data_formats-cloud_config`. This
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+means we need a :code:`#cloud-config` header. The autoinstall directives are
+placed under a top level :code:`autoinstall` key, like so:
+
+.. code-block:: yaml
+
+    #cloud-config
+    autoinstall:
+        version: 1
+        ....
+
```
s-makin, line 64: It may be worth doing a note here to point out that the file must start with the #cloud-config header line or it won't work (I realise you mention it two lines before, but it's an important detail that I think should be called out explicitly).

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+Autoinstall on the install media
+--------------------------------
+
+Another option for supplying autoinstall to the Ubuntu installer is to place a
+file named :code:`autoinstall.yaml` on the install media itself.
+
+There are two potential locations for the :code:`autoinstall.yaml` file:
+ * At the root of the "cdrom".  When writing the installation ISO to a USB
```
s-makin, line 72:
```suggestion
 * At the root of the "cdrom". When you write the installation ISO to a USB
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+   Flash Drive, this can be done by copying the :code:`autoinstall.yaml` to the
+   partition containing the contents of the ISO - i.e., in the same directory
+   containing the :code:`casper` directory.
```
s-makin, line 75: "In the same directory containing the :code:`casper` directory" was a bit confusing to me. I'd suggest something like "In the directory containing the :code:`casper` sub-directory".

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+ * On the rootfs of the installation system - this option will typically
+   require modifying the installation ISO and is not suggested, but is
+   supported.
+
+Directly specifying autoinstall as a :code:`autoinstall.yaml` file does not
+require a :code:`#cloud-config` header, and does not use a top level
+:code:`autoinstall` key.  The autoinstall directives are placed at the top
```
s-makin, line 82:
```suggestion
:code:`autoinstall` key. The autoinstall directives are placed at the top
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+level.
+
+.. code-block:: yaml
+
+    version: 1
+    ....
+
+
+Cloud-init and Autoinstall interaction
```
s-makin, line 91:
```suggestion
Cloud-init and autoinstall interaction
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+======================================
+
+While cloud-init may assist in providing the autoinstall configuration to the
+Ubuntu installer, cloud-init itself is not processing the autoinstall.
```
s-makin, line 95:
```suggestion
While cloud-init may provide the autoinstall configuration to the
Ubuntu installer, it does not process the autoinstall itself.
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+
+If cloud-init directives are intended to modify the ephemeral system, they
+must appear at the top level of the cloud-config.  If instead
```
s-makin, line 98:
```suggestion
must appear at the top level of the cloud-config. If instead
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+
+
+Zero-touch deployment with autoinstall
+======================================
+
+The Ubuntu Installer contains a safeguard, intended to prevent USB Flash Drives
+with an :code:`autoinstall.yaml` file from wiping out the wrong system.
+
+During autoinstall, a prompt will be shown to confirm that the install really
+should proceed and start making modifications to the target system. ::
```
s-makin, line 126:
```suggestion
should proceed before making modifications to the target system. ::
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+
+    start: subiquity/Meta/status_GET
+    Confirmation is required to continue.
+    Add 'autoinstall' to your kernel command line to avoid this
+
+
+    Continue with autoinstall? (yes|no)
+
+To bypass this prompt, arrange for the argument :code:`autoinstall` to be
```
s-makin, line 135: Would the typical user know how to do this? If not, may need to include a short how-to guide

doc/tutorial/intro-to-autoinstall.rst
s-makin: I think having a warning at the top where you introduce the #cloud-config and autoinstall lines would achieve the same thing (see my comment around line 64) - any user who wants to use that method will have their attention drawn to the fact that these lines are necessary - but having the lines also included in the examples is a great idea in case users try to copy/paste directly. A tutorial should be foolproof, so if we give them examples to try they shouldn't be able to fail with them if they follow the instructions.

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,243 @@
...
+
+    #cloud-config
+    autoinstall:
+        version: 1
+        identity:
+            hostname: hostname
+            username: username
+            password: $crypted_pass
+
+Here is an example file that shows off most features:
```
s-makin, line 165: I'm not clear on what "features" is referring to here

doc/tutorial/intro-to-autoinstall.rst
dbungert: Thanks so much Sally, the build PRs functionality is now fully connected.  A preview of the docs can be seen with the RTD build check.

s-makin (commented):
Overall it's a really good basis. I would definitely recommend having a separate tutorial for each of the suggested methods; using this page as the introduction, you could then divert users early on towards the tutorial they need in order to avoid confusion that might arise from the two being presented together.

doc/tutorial/intro-to-autoinstall.rst
dbungert: I'm not sure "before" fits there.

What's going on is that Subiquity is just about ready to start modifying disks, and we need to confirm that this is OK.

What about this phrasing?

"Before the Ubuntu Installer actually makes changes to the target system, a prompt is shown."

doc/conf.py
```diff
@@ -35,10 +35,17 @@
 
 extensions = [
     'm2r2',
+    'sphinx.ext.intersphinx',
```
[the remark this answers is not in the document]
dbungert: Great, yes I intend to use rST until we see a need otherwise.  Done.

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,239 @@
...
+   default for any unanswered question (and fails if there is no default).
+ * You can designate particular sections in the configuration as "interactive",
+   which means the installer will still stop and ask about those.
+
+
+Providing the autoinstall configuration
+=======================================
+
+There are 2 methods of providing the autoinstall configuration:
+ * Carried as part of cloud-config
```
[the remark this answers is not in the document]
dbungert: maybe not.

what about:
"* As part of a cloud-init datasource"

doc/tutorial/intro-to-autoinstall.rst
dbungert: Both the Subiquity and Cloud-init teams are interested in reducing confusion here.  Phrasing suggestions very welcome.  I suppose we need to resolve the title of this section first though.

doc/tutorial/intro-to-autoinstall.rst
dbungert: Right, I agree and think this is poorly placed in the tutorial doc as is but wanted to give it a home before we moved stuff around.

We need a proper tutorial, which I intend to make qemu based.  As part of that we can show the process for kernel command line overrides.  This note in particular more feels like a reference bit but lacks a good spot as of yet.  I'm thinking some sort of secondary reference document alongside but different to the "Autoinstall configuration reference manual", as this isn't a directive.

doc/tutorial/intro-to-autoinstall.rst
dbungert: "directives" might be a better word - it's meant to refer to the top level items shown in "Autoinstall configuration reference manual"

doc/tutorial/intro-to-autoinstall.rst
dbungert: Yes, I'm erring on the side of foolproof.  We have adequate user evidence that there is a lot of user confusion on this point.  I'd normally prefer the shorter code sample.

Something I've debated doing is to allow direct specification of the `autoinstall.yaml` file to be in Cloud config data format - if `autoinstall` is the only top level directive, then this is simply handled subiquity side.  But what if there are other cloud-config top level keys?  I think Subiquity should warn or error in that case, as those aren't going to be processed.

If we did that, then 100% of the time it would be possible to just copy-paste the code sample in Cloud config data format and get a working result, regardless of if providing the data with the help of cloud-init or with modified install media.

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,239 @@
+.. _tutorial_intro-to-autoinstall:
+
+Introduction to Autoinstall
+***************************
+
+Ubuntu installation automation is performed with the autoinstall format.
```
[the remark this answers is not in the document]
dbungert: Done.

doc/tutorial/intro-to-autoinstall.rst
dbungert: Note added.  I absolutely got burned multiple times by this detail and think elaborating on it is sensible.

doc/tutorial/intro-to-autoinstall.rst
dbungert: Done

doc/conf.py
```diff
@@ -22,7 +22,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'Ubuntu Install Guide'
-copyright = f'Canonical Group Ltd, {datetime.date.today().year}'
+copyright = f'Canonical Group Ltd, 2020 - {datetime.date.today().year}'
```
blackboxsw, line 25: https://hynek.me/til/copyright-years/   ?? can we drop copyright years altogether now that git history is reasonable enough substrate to read and defend copyright?

doc/tutorial/intro-to-autoinstall.rst
s-makin: I might wait to see what Chad throws into the mix from the other direction, and have a few possible suggestions for you on Monday :)

doc/tutorial/intro-to-autoinstall.rst
s-makin: That would work!

doc/tutorial/intro-to-autoinstall.rst
s-makin: That sounds great! Since it's on the to-do list, it's fine to mark this as resolved.

doc/tutorial/intro-to-autoinstall.rst
s-makin: I don't have very strong opinions about what word is used, as long as we define somewhere how we're using the term. So, whether we use features or directives, it should be stated up at the top that that's how we're going to refer to them (to make it completely clear what our terminology is)

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,250 @@
...
+======================================
+
+While cloud-init may provide the autoinstall configuration to the
+Ubuntu installer, it does not process the autoinstall itself.
+
+If cloud-init directives are intended to modify the ephemeral system, they
+must appear at the top level of the cloud-config. If instead
+cloud-init directives are intended to modify the system being installed, they
+must appear under a :code:`user-data` section in :code:`autoinstall`.
+
```
blackboxsw, line 108: Good intro, I think this section is where we get most rub. Let's add a bit more context around where cloud-init runs relative to the ephemeral and installer environment here as cloud-init will point that this section too.

```suggestion
Cloud-init runs in both the ephemeral system, during install, and in the target system during first boot. Cloud-init then becomes inert for every subsequent
reboot.

While cloud-init may provide the autoinstall configuration to the
Ubuntu installer, it does not process the autoinstall directives
itself.

To modify the ephemeral system with cloud-init, any :external+cloud-init:ref:`#cloud-config module schema keys<modules>` can
be provided. If instead cloud-init directives are intended to modify the system
being installed, they must appear under a :code:`user-data:` section under
:code:`autoinstall:`.
```

blackboxsw (changes requested):
@dbungert this looks really good. I think the only thing I'd like to see is a bit more explicit content describing that cloud-init runs in both ephmeral and at first boot and a pointer to the honored cloud-init cloud-conifg module schema directives 

 Given that this is a tutorial, do you plan on having a section in a later pull request that actually walks someone through the cmdline to download live ISO, launch python3 -m http.server with nocloud user-data containing autoinstall: #cloud-config directives via kernel cmdline?  Maybe that's the quickstart you referred to as a followup (out of scope for this PR).

dbungert:
> Given that this is a tutorial, do you plan on having a section in a later pull request that actually walks someone through the cmdline to download live ISO, launch python3 -m http.server with nocloud user-data containing autoinstall: #cloud-config directives via kernel cmdline? Maybe that's the quickstart you referred to as a followup (out of scope for this PR).

Yes, I see what you mention as a proper tutorial and what I'm referring to as a quickstart above.

When I sat down to bring https://ubuntu.com/server/docs/install/autoinstall into rtd/diataxis, I read the entire doc attempting to view it like a first time reader, and realized that much of it didn't really fit an introduction or tutorial in the diataxis sense.  I was a bit paralyzed wondering what to do about it.

To split the problem a bit, I first wanted to just get the document kind-of as-is into rtd/diataxis, and sort out the tutorial versus other 3 categories problem later.  Also, if we can cut over the existing docs to diataxis, it will be mentally easier to add new content.

So I really wanted to import the introduction, with most of the warts in place, rather than trying to build out a tutorial I liked better and deal with moving all the rest of the content to their proper locations.

I started the reference document largely as an exercise in learning how the ref links work.

doc/tutorial/intro-to-autoinstall.rst
dbungert: Going with "directives" for now, +1 to consistent terminology.

doc/tutorial/intro-to-autoinstall.rst
dbungert: Done

blackboxsw (approved):
LGTM! thanks for the text and context around this iteration of the refactor. +1

doc/tutorial/intro-to-autoinstall.rst
dbungert: Chad did have a suggestion, which is now incorporated in this PR.

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,258 @@
...
+Differences from debian-installer preseeding
+============================================
+
+*preseeds* are the way to automate an installer based on debian-installer
+(also known as d-i).
+
+Autoinstalls differ from preseeds in the following ways:
+ * The format is different: autoinstalls use YAML instead of the preseed
+   debconf-set-selections
```
s-makin, line 27:
```suggestion
   debconf-set-selections.
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,258 @@
...
+   file.
+
+Autoinstall on the install media
+--------------------------------
+
+Another option for supplying autoinstall to the Ubuntu installer is to place a
+file named :code:`autoinstall.yaml` on the install media itself.
+
+There are two potential locations for the :code:`autoinstall.yaml` file:
+ * At the root of the "cdrom". When you write the installation ISO to a USB
```
s-makin, line 82:
```suggestion
 * At the root of the "CD-ROM". When you write the installation ISO to a USB
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,258 @@
...
+   partition containing the contents of the ISO - i.e.,
+   in the directory containing the ``casper`` sub-directory.
+ * On the rootfs of the installation system - this option will typically
+   require modifying the installation ISO and is not suggested, but is
+   supported.
+
+Directly specifying autoinstall as a :code:`autoinstall.yaml` file does not
+require a :code:`#cloud-config` header, and does not use a top level
+``autoinstall:`` key. The autoinstall directives are placed at the top
+level.
```
s-makin, line 93:
```suggestion
level. For example:
```

doc/tutorial/intro-to-autoinstall.rst
```diff
@@ -0,0 +1,258 @@
...
+
+    version: 1
+    ....
+
+
+Cloud-init and autoinstall interaction
+======================================
+
+Cloud-init runs in both the ephemeral system, during install, and in the target
+system during first boot. Cloud-init then becomes inert for every subsequent
```
s-makin, line 105:
```suggestion
Cloud-init runs in both the ephemeral system (during install) and in the target
system during first boot. Cloud-init then becomes inert for every subsequent
```

s-makin: The "both" was confusing to me because it looked like three things were listed with the placing of the commas. I think this would make it clearer.

s-makin (approved):
Good changes! Found a few teeny nits, but otherwise LGTM :)

PR #1710, opened 2023-07-09
add annotations to storage model objects

mwhudson:
I think this will make something else I am working on easier... maybe... it's a nice cleanup in any case I think.


dbungert (approved):
LGTM, thanks

PR #1713, opened 2023-07-11
tidy up examples/ directory

mwhudson:
Move machine configs, sources, autoinstall files and dry-run configs into their own subdirectories.


mwhudson:
This is obviously very important!!

ogayot:
Thanks for doing some cleanup. Should we remove the `source`, `answers`, `autoinstall`, `dr-cfg` from the filenames now that they're in the relevant directories?

mwhudson:
yes i think that looks better.

ogayot (approved):
Thanks!

PR #1716, opened 2023-07-12
factor out code to install a package from the pool in the live session

mwhudson:
I need efibootmgr to configure the rp and it is in the pool but not the live session by default.


ogayot (approved):
LGTM!

I would suggest a docstring mentioning that the pkghelper / PackageInstaller stuff only installs packages in the live system ; as opposed to what's in subiquity/server/controllers/install.py - which installs in the target system.

On that note, we could perhaps make the PackageInstaller take an optional rootfs - so that we can have a single implementation for installing in the live environment or the target environment. But this is for another day :)

PR #1717, opened 2023-07-12
Use true/false instead of yes/no for bools

lindhe:
Before, the reference documentation was inconsistent with boolean values. It sometimes used true/false and other times yes/no.

This changes makes it so that the reference documentation is consistently using true/false.

### DIFF
--- a/documentation/autoinstall-reference.md
+++ b/documentation/autoinstall-reference.md
@@ -84,7 +84,7 @@ The mapping contains keys:
 #### update
 
 **type:** boolean
-**default:** `no`
+**default:** `false`
 
 Whether to update or not.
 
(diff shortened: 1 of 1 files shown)


lindhe:
How do I sign the CLA? I can't find any documentation on how to do it; you have no CONTRIBUTING file...

lindhe:
Found it here: https://ubuntu.com/legal/contributors/agreement

Should be signed now! :)

dbungert (approved):
Thanks for hunting this down.  I think it's worth being consistent in the values here (even if the yaml doesn't have to be that strict.)

I opened a [PR](https://github.com/canonical/subiquity/pull/1722) with a starter contributing guide, thanks also for the reminder there.

PR #1718, opened 2023-07-13
Fix typo in reference: minimized --> minimal

lindhe:
There is no such thing as `ubuntu-server-minimized`. It should be `ubuntu-server-minimal`.

See for example here:
https://github.com/canonical/subiquity/blob/f8bf1105bf96e044b013e7c30b688e4c89de2f47/examples/sources/install.yaml#L4

Another piece of evidence is found in the Ubuntu ISO:

image


ogayot (approved):
LGTM. Thanks for this contribution!

PR #1723, opened 2023-07-18
offer guided options for the install media

mwhudson:
Also include the install media in responses to client requests for disks.


mwhudson:
This is quite a subtle one. I haven't tested in a VM yet, or at least not a very recent version of this.

dbungert (approved):

PR #1725, opened 2023-07-18
oem: check variant before waiting for apt configuration

mwhudson:
because if we add support for variant == 'core' (spoilers!), then apt may never be configured.


mwhudson:
Well that was confusing but it turns out that it was the answers support calling API methods in a surprising order.

dbungert (approved):

PR #1726, opened 2023-07-18
generalize the way screens are skipped when installing desktop

mwhudson:
Also skip identity screen when installing neither server nor desktop (i.e. core).

### DIFF
--- a/subiquity/server/controllers/identity.py
+++ b/subiquity/server/controllers/identity.py
@@ -62,6 +62,8 @@ class IdentityController(SubiquityController):
         'additionalProperties': False,
         }
 
+    interactive_for_variants = {'desktop', 'server'}
+
     def __init__(self, app):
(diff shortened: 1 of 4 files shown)


subiquity/server/controllers/identity.py
```diff
@@ -62,6 +62,8 @@ class IdentityController(SubiquityController):
         'additionalProperties': False,
         }
 
+    interactive_for_variants = {'desktop', 'server'}
```
dbungert, line 65: I think desktop wants to not be on this list.

dbungert (approved):
Really nice generalization.  Thanks!

subiquity/server/controllers/identity.py
mwhudson: yeah but not yet

PR #1729, opened 2023-07-19
filesystem: fix ~6GB extra pad on classic min size

dbungert:
For https://github.com/canonical/ubuntu-desktop-installer/issues/2252

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -383,13 +383,8 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
                 if info is not None:
                     self._variation_info[name] = info
             else:
-                # This calculation is pretty much a hack and we should
-                # actually think about it at some point (like: maybe the
-                # source catalog should directly specify the minimum suitable
-                # size?)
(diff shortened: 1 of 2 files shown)


mwhudson (approved):
Ah oops, thanks for catching my mistake here.

It's sort of annoying that calculate_suggested_install_min has to take a part_align argument. I wonder if we could just make it assume the largest alignment of any ptable type or something, I don't think that would make much difference in practice?

dbungert:
> It's sort of annoying that calculate_suggested_install_min has to take a part_align argument. I wonder if we could just make it assume the largest alignment of any ptable type or something, I don't think that would make much difference in practice?

I would be ok with that.  It's meant to be an estimate, but I do like that it give an aligned answer.

mwhudson:
> > It's sort of annoying that calculate_suggested_install_min has to take a part_align argument. I wonder if we could just make it assume the largest alignment of any ptable type or something, I don't think that would make much difference in practice?
> 
> I would be ok with that. It's meant to be an estimate, but I do like that it give an aligned answer.

I have a better idea! https://github.com/canonical/subiquity/pull/1732

PR #1730, opened 2023-07-19
Zfs action ordering

dbungert:
Address action ordering with zpool/zfs objects - these objects are 'Mountlike' objects and may represent a viable mountpoint, which affects rootfs detection and the required ordering of curtin storage actions.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1929,21 +1929,14 @@ class FilesystemModel(object):
                 return False
         return True
 
-    def add_zpool(self, device, pool, mountpoint):
-        fs_properties = dict(
-            acltype='posixacl',
-            relatime='on',
-            canmount='on',
-            compression='gzip',
-            devices='off',
-            xattr='sa',
-        )
+    def add_zpool(self, device, pool, mountpoint, *,
+                  fs_properties=None, pool_properties=None):
         zpool = ZPool(
             m=self,
             vdevs=[device],
             pool=pool,
             mountpoint=mountpoint,
-            pool_properties=dict(ashift=12),
+            pool_properties=pool_properties,
             fs_properties=fs_properties)
         self._actions.append(zpool)
         return zpool
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1623,7 +1623,7 @@ class FilesystemModel(object):
                         if dep.type in ['disk', 'raid']:
                             ensure_partitions(dep)
                     return False
-            if obj.type in MountlikeNames:
+            if obj.type in MountlikeNames and obj.path is not None:
                 # Any mount actions for a parent of this one have to be emitted
                 # first.
                 for parent in pathlib.Path(obj.path).parents:
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1132,6 +1132,11 @@ class ZFS:
 
 ConstructedDevice = Union[Raid, LVM_VolGroup, ZPool]
 
+# A Mountlike is a literal Mount object, or one similar enough in behavior.
+# Mountlikes have a path property that may return None if that given object is
+# not actually mountable.
+MountlikeNames = ('mount', 'zpool', 'zfs')
+
 
 def align_up(size, block_size=1 << 20):
     return (size + block_size - 1) & ~(block_size - 1)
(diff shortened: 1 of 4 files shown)


subiquity/models/filesystem.py
```diff
@@ -1106,9 +1110,29 @@ class ZFS:
...
 
+    @property
+    def fstype(self):
+        return 'zfs'
+
+    @property
+    def path(self):
+        if self.properties is None:
+            return self.volume
+        if not yaml.safe_load(self.properties.get('canmount', 'on')):
```
mwhudson, line 1121: Ditto

mwhudson (approved):
the safe_loads make me twitch, otherwise ok

subiquity/models/filesystem.py
```diff
@@ -1092,7 +1093,10 @@ def name(self):
         return self.pool
 
     @property
-    def mount(self):
+    def path(self):
+        if self.fs_properties is not None:
+            if not yaml.safe_load(self.fs_properties.get('canmount', 'off')):
```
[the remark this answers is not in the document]
mwhudson: Oh right this is partly about what curtin accepts for this field too isn't it. It looks like curtin accepts booleans in this field as well as 'on' and 'off' strings (because lets face it, noone is going to remember to write `canmount: 'on'` rather than `canmount: on` in their yaml) -- but of course "safe_load(True)" explodes. So I think code interpreting these fields needs to accept: True, False, "on", "off" and "noauto". But no others.

Does that make sense? I'm not sure my brain is working at "making statements about technology I don't really understand" levels today.

subiquity/models/filesystem.py
dbungert: Done

subiquity/models/filesystem.py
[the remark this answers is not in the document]
dbungert: Done

subiquity/models/filesystem.py
```diff
@@ -1711,6 +1756,12 @@ def _all(self, **kw):
     def all_mounts(self):
         return self._all(type='mount')
 
+    def all_mountlikes(self):
```
mwhudson, line 1759: I almost wonder if you should change all_mounts to do this. It would break the server tui in a more obvious way than the slightly subtle way leaving it alone does :-)

subiquity/models/filesystem.py
```diff
@@ -1110,9 +1135,29 @@ class ZFS:
     # options to pass to zfs dataset creation
     properties: Optional[dict] = None
 
+    @property
+    def fstype(self):
+        return 'zfs'
+
+    @property
+    def canmount(self):
+        return get_canmount(self.properties, False)
```
mwhudson, line 1144: The previous version sure looked like it had different defaults for canmount for zfs and zpool objects. Want to check that detail? (I certainly don't know what is correct here!)

mwhudson (approved):
Hi, thanks for cleaning that up, much nicer now.

I have some questions but don't need to review again once you've answered them (unless you want me to!)

subiquity/models/filesystem.py
dbungert: Intentional!  I have found it helpful to think of a ZPool as implicitly creating a ZFS object, and the fs_properties on the zpool should be equivalent to the properties on a zfs.

Then I made sure the above matched curtin's default behavior, and I think we should revisit that, as curtin defaults canmount='off' while zfsutils-linux defaults to canmount='on', for both a zfs dataset and the implicit zfs dataset created by the zpool.

So for right now I think this part is correct, and we should ponder https://git.launchpad.net/curtin/tree/curtin/block/zfs.py#n19 .

subiquity/models/filesystem.py
```diff
@@ -1074,6 +1074,25 @@ def can_delete(self):
         return True
 
 
+def get_canmount(properties: Optional[dict], default: bool) -> bool:
+    """Handle the possible values of the zfs canmount property, which should be
+    on/off/noauto.  Due to yaml handling, on/off may be turned into a bool, so
+    handle that also."""
+    if properties is None:
+        return default
+    vals = {
```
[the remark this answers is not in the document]
dbungert: It does.  I'm going to reluctantly decline at this point, as I do think running the integration tests on focal has some value.

PR #1735, opened 2023-07-20
Obtain autoinstall from combined-cloud-config

dbungert:


mwhudson (commented):
lgtm but will review properly when it's marked ready for that :-)

subiquity/cloudinit.py
```diff
@@ -0,0 +1,12 @@
+"""Shared cloudinit utility functions"""
+
+import json
+
+
+def get_host_combined_cloud_config() -> dict:
+    """Return the host system /run/cloud-init/combined-cloud-config.json"""
+    try:
+        with open("/run/cloud-init/combined-cloud-config.json") as fp:
+            return json.load(fp)
```
blackboxsw, line 10: Do we want to log success here so we can track any call-sites since models/network.py also re-reads this file. If we grow lots of readers, we may look toward `lru_cache` of this function in the future even though it's not very expensive.
```suggestion
            config = json.load(fp)
        log.debug("loading cloud config from /run/cloud-init/combined-cloud-config.json")
```

subiquity/server/server.py
```diff
@@ -533,6 +532,48 @@ async def start_site(self, runner: web.AppRunner):
...
+        # data) is newer than the one inside the snap (which reads the pickle
+        # data if we do stages.Init()).  LP: #2022102
+
+        # The stages.Init() code path should be retained until we can assume a
+        # minimum cloud-init version of 23.3 (when Subiquity drops support for
+        # Ubuntu 22.04.2 LTS and earlier, presumably)
+
+        cloud_cfg = get_host_combined_cloud_config()
+        if len(cloud_cfg) > 0:
+            log.debug("loading cloud config from combined-cloud-config")
```
blackboxsw, line 549: Do we want to push this message into the get_host_combined_cloud_config() so we can track all callsites that successfully pull this data.

subiquity/server/server.py
```diff
@@ -533,6 +532,48 @@ async def start_site(self, runner: web.AppRunner):
...
+        # occur when the cloud-init outside the snap (which writes the pickle
+        # data) is newer than the one inside the snap (which reads the pickle
+        # data if we do stages.Init()).  LP: #2022102
+
+        # The stages.Init() code path should be retained until we can assume a
+        # minimum cloud-init version of 23.3 (when Subiquity drops support for
+        # Ubuntu 22.04.2 LTS and earlier, presumably)
+
+        cloud_cfg = get_host_combined_cloud_config()
+        if len(cloud_cfg) > 0:
```
blackboxsw, line 548: could just use truthy value of dicts to avoid calculating length of keys and then comparing to 0
```suggestion
        if cloud_cfg:
```

subiquity/server/server.py
```diff
@@ -533,6 +532,48 @@ async def start_site(self, runner: web.AppRunner):
...
+            log.debug("loading cloud config from combined-cloud-config")
+            system_info = cloud_cfg.get('system_info', {})
+            default_user = system_info.get('default_user', {})
+            self.installer_user_name = default_user.get('name')
```
blackboxsw, line 552: I think this makes sense, the only thing I can think is a little advertised user-data can manipulate what the default user is using deprecated keys that could override the configuration present in system_info. 

Our example doc on this is below and [in doc examples for users and groups](https://cloudinit.readthedocs.io/en/latest/reference/modules.html#users-and-groups)
```
        user:
          name: mynewdefault
          sudo: null
```

I'll think through this path when the PR is up as this #cloud-config override probably also affects the other logical path you have with stages.Init() too.

blackboxsw (commented):
looks good, one question generally about what impact `#cloud-config` `user:` overrides or renames the `default_user` from `system_info`

subiquity/server/server.py
dbungert: If this is the combined-cloud-config version of `ug_util.extract_default`, then I think we're set.

The `None` [default user](https://github.com/canonical/subiquity/pull/1687) one was rough because it could be encountered with certain cloud-config before we even got as far as showing the UI.

My goal isn't some sort of comprehensive support for weird corner cases here - more along the lines of "don't crash".

subiquity/server/server.py
dbungert: Yes, but we tend to be more explicit on this point in Subiquity, so I'm matching codebase convention.

subiquity/server/server.py
dbungert: Done

subiquity/cloudinit.py
dbungert: Done

subiquity/cloudinit.py
```diff
@@ -1,12 +1,20 @@
...
     """Return the host system /run/cloud-init/combined-cloud-config.json"""
     try:
         with open("/run/cloud-init/combined-cloud-config.json") as fp:
-            return json.load(fp)
+            config = json.load(fp)
+            log.debug(
+                "Loaded cloud config from "
+                "/run/cloud-init/combined-cloud-config.json"
+            )
+            return config
```
mwhudson, line 18: Is it worth logging something in the "we did not use combined cloud config" case?

subiquity/server/server.py
```diff
@@ -626,13 +656,9 @@ def use_passwd(passwd):
...
 
-        (users, _groups) = ug_util.normalize_users_groups(
-            self.cloud.cfg, self.cloud.distro)
-        (username, _user_config) = ug_util.extract_default(users)
+        username = self.installer_user_name
 
-        self.installer_user_name = username
-
-        if username is None:
+        if self.installer_user_name is None:
```
mwhudson, line 661: nitpick level extreme but why change this to refer to self.installer_user_name when all the other references are still to username?

mwhudson (approved):
just some nitpicking

subiquity/cloudinit.py
```diff
@@ -1,12 +1,20 @@
...
     try:
         with open("/run/cloud-init/combined-cloud-config.json") as fp:
-            return json.load(fp)
+            config = json.load(fp)
+            log.debug(
+                "Loaded cloud config from /run/cloud-init/combined-cloud-config.json"
+            )
+            return config
     except (IOError, OSError, AttributeError, json.decoder.JSONDecodeError):
+        log.debug("Failed to load combined-cloud-config")
```
mwhudson, line 19:
```suggestion
        log.debug("Failed to load combined-cloud-config")
```
```suggestion
    except (IOError, OSError, AttributeError, json.decoder.JSONDecodeError) as e:
        log.debug("Failed to load combined-cloud-config: %s", e)
```

mwhudson (approved):
One more but feel free to not bother

subiquity/cloudinit.py
dbungert: I think that's fine at debug level, wouldn't want anything too scary looking as this is expected behavior for existing install ISOs to not have the file.
I'll follow up.

mwhudson: good point!

dbungert: https://github.com/canonical/subiquity/pull/1743

PR #1737, opened 2023-07-21
mount rp at predictable path and leave it mounted

mwhudson:
I have a use case for wanting to manipulate the rp contents in a late-command.

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -395,13 +395,14 @@ class InstallController(SubiquityController):
         rp = fs_controller.reset_partition
         if rp is not None:
             mounter = Mounter(self.app)
-            async with mounter.mounted(rp.path) as mp:
-                await run_curtin_step(
-                    name="populate recovery",
-                    stages=["extract"],
-                    step_config=self.rp_config(logs_dir, mp.p()),
-                    source="cp:///cdrom",
-                )
+            rp_target = os.path.join(self.app.root, "factory-reset")
+            mp = await mounter.mount(rp.path, mountpoint=rp_target)
+            await run_curtin_step(
+                name="populate recovery",
+                stages=["extract"],
+                step_config=self.rp_config(logs_dir, mp.p()),
+                source="cp:///cdrom",
+            )
             await self.create_rp_boot_entry(context=context, rp=rp)
 
     @with_context(description="creating boot entry for reset partition")


subiquity/server/controllers/install.py
```diff
@@ -389,12 +389,13 @@ async def run_curtin_step(name, stages, step_config, source=None):
         rp = fs_controller.reset_partition
         if rp is not None:
             mounter = Mounter(self.app)
-            async with mounter.mounted(rp.path) as mp:
-                await run_curtin_step(
-                    name="populate recovery", stages=["extract"],
-                    step_config=self.rp_config(logs_dir, mp.p()),
-                    source='cp:///cdrom',
-                    )
+            rp_target = os.path.join(self.app.root, 'rp')
```
ogayot, line 392: how about /recovery maybe?

ogayot (commented):
in documentation/autoinstall-reference.md#late-commands, we document that the installed system is mounted at /target.

Would you also mention where the recovery partition (assuming there is one) is expected to be mounted - unless we want to keep this info private-ish for now.

subiquity/server/controllers/install.py
mwhudson: Well. The terminology gets a bit confusing around this. "recovery" in some context means things like recovering your bitlocker / luks passphrase, and I've been trying to not talk about the work we are doing for OEM as a recovery partition. I think talking about it as the "factory reset" or "restore" partition is a bit less ambiguous. Fortunately (?) recovery, reset and restore all start with 'r' so talking about it as 'rp' avoids making any decisions :-)

If /rp is too obscure (which I am sympathetic with!), how about "/factory-reset"?

mwhudson:
> Would you also mention where the recovery partition (assuming there is one) is expected to be mounted - unless we want to keep this info private-ish for now.

Yeah I should document the new reset / recovery stuff. It's a bit tempting to hold off until it has been validated as being useful but I'm not sure that's a helpful impulse.

subiquity/server/controllers/install.py
dbungert: Yes, I think `/factory-reset` best captures the spirit of what this partition does, I suggest going with that.

ogayot: Sorry for the delay - yes /factory-reset seems right :)

dbungert (approved):

PR #1740, opened 2023-07-24
do not require identity data when reset_partition_only

mwhudson:


subiquity/server/controllers/identity.py
```diff
@@ -85,9 +85,13 @@ def load_autoinstall_data(self, data):
...
     @with_context()
     async def apply_autoinstall_config(self, context=None):
-        if not self.model.user:
-            if 'user-data' not in self.app.autoinstall_config:
-                raise Exception("no identity data provided")
+        if self.model.user:
+            return
+        if 'user-data' in self.app.autoinstall_config:
+            return
+        if self.app.controllers.Filesystem.reset_partition_only:
```
dbungert, line 92: This is fine for now, but I have a suspicion we'll need to make this bit more extensible in the future - identity to filesystem coupling is a bit surprising.

dbungert (approved):

subiquity/server/controllers/identity.py
mwhudson: There is another spelling we can use which looks a bit less arbitrary, take a look?

dbungert (approved):

PR #1741, opened 2023-07-24
shutdown: move to command runner

dbungert:
The app command runner already knows if we're dry-run or not, so move to that.  This allows us to skip more command runs, in particular `journalctl -b`.  For my current reboot the integration tests run in half the time!


mwhudson (approved):
ah ahaha I think the journal in the crash log is why it takes so long for crash reports to parse as well sometimes?

PR #1744, opened 2023-07-26
tox: use skip_install to speed up black/isort

dbungert:
Skips the install when it's already there, does the right thing if .tox not present / old dependency / missing dependency.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,17 +2,18 @@
 envlist =
     black
     isort
-recreate = True
 
 [testenv]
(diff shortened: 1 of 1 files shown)


mwhudson (commented):
This is a lot better -- 1.2 vs 7.5s on my system. Running the commands directly is about 0.6s (isort is way slower than black!) so there's still some overhead but way less. Thanks!

mwhudson (approved):

PR #1745, opened 2023-07-26
Permit ESP on Raid if PowerEdge

dbungert:
Per discussion in LP: #[1961079](https://bugs.launchpad.net/ubuntu/+source/grub2/+bug/1961079), permit bootable raid if the system appears to be a "PowerEdge".

Tested only by way of CI.

### DIFF
--- a/subiquity/common/dmidecode.py
+++ b/subiquity/common/dmidecode.py
@@ -1,0 +1,27 @@
+# Copyright 2023 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


mwhudson (approved):
This looks fine and is a nice refactoring anyway, thanks! It would be nice to get confirmation that it works before merging I guess.

subiquity/common/dmidecode.py
```diff
@@ -0,0 +1,27 @@
...
+
+from subiquitycore.utils import run_command
+
+
+def dmidecode_get(key: str) -> str:
+    """Retrieve a value from `dmidecode -s`
+    :param key: dmidecode(8) KEYWORD
+    :returns: stripped string output, or empty string on failure
+    """
+    sp = run_command(["dmidecode", "-s", key], check=False, text=True)
```
ogayot, line 24: should we add dmidecode to the apt dependencies list ? 

If the command is not found, this will raise a `FileNotFoundError`.

PR #1748, opened 2023-08-02
network: fix NETPLAN_CONFIG_ROOT_READ_ONLY tests

dbungert:


subiquity/models/tests/test_subiquity.py
```diff
@@ -185,18 +187,31 @@ def test_storage_version(self):
...
+
+            for fspec in config["write_files"].values():
+                if fspec["path"].startswith("etc/netplan"):
+                    if netplan_content is not None:
+                        self.fail("writing two files to netplan?")
+                    netplan_content = fspec["content"]
+            self.assertEqual(np_content_expected, netplan_content is not None)
+            if np_content_expected:
+                netplan = yaml.safe_load(netplan_content)
+                self.assertConfigHasVal(netplan, "network.version", 2)
```
mwhudson, line 214: should we have another case here that asserts netplan_content is None if np_content_expected is false?

mwhudson (approved):

subiquity/models/tests/test_subiquity.py
dbungert: I think that's covered already, pushed a tweak that is a bit more readible.  If this isn't what you meant please clarify.

mwhudson: thanks!

PR #1751, opened 2023-08-02
move oem configuration stuff to a separate method

mwhudson:
and do not call it when installing core

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -224,6 +224,51 @@ class InstallController(SubiquityController):
                 device_map = json.load(fp)
             self.app.controllers.Filesystem.update_devices(device_map)
 
+    @with_context(description="configuring OEM packages")
+    async def pre_curthooks_oem_configuration(self, context):
+        # For OEM, we basically mimic what ubuntu-drivers does:
+        # 1. Install each package with apt-get install
+        # 2. For each package, run apt-get update using only the source
+        # installed by said package.
+        # 3. Run apt-get install again for each package. This will upgrade
+        # them to the version found in the OEM archive.
+
+        # NOTE In ubuntu-drivers, this is done in a single call to apt-get
+        # install.
+        for pkg in self.model.oem.metapkgs:
+            await self.install_package(package=pkg.name)
+
+        if self.model.network.has_network:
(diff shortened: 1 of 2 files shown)


mwhudson:
not at all confident in my choice of method name or user facing language here btw

subiquity/server/controllers/install.py
```diff
@@ -222,6 +222,51 @@ async def run_curtin_step(
...
+            # NOTE In ubuntu-drivers, this is done in a single call to
+            # apt-get install.
+            for pkg in self.model.oem.metapkgs:
+                await self.install_package(package=pkg.name)
+
+        # If we already have a kernel installed, don't bother requesting
+        # curthooks to install it again or we might end up with two
+        # kernels.
+        if await list_installed_kernels(Path(self.tpath())):
+            self.model.kernel.curthooks_no_install = True
```
ogayot, line 268: I think these lines should go into a separate method too.

ogayot (commented):
I think the method name is fine. I don't know if I want to show the "configuring OEM package" message when there is no applicable package though.

Consider doing something like the below maybe?

```python
async def pre_curthooks_oem_configuration(self):
    @with_context(description="installing applicable OEM metapackages")
    async def install_oem_metapackages(context):
         ...

    if not self.model.oem.metapkgs:
        return
    await install_oem_metapackages()

mwhudson:
> Consider doing something like the below maybe?

Yeah that makes sense I think. Pushed something like your suggestion

subiquity/server/controllers/install.py
ogayot: Because it can also have an effect outside the OEM scenario - when the source squashfs contains a kernel package (it was the case until recently on desktop). See https://bugs.launchpad.net/subiquity/+bug/2026225

Whether it's a behavior we want or not is a different question. But I'd rather keep the current behavior (i.e., run it even if no OEM metapackages are applicable) at least for this PR - if that makes sense?

mwhudson: Oh I see. I pushed something like that.

ogayot (approved):
Thanks for the updates, looks good to me now !

dbungert (approved):

PR #1752, opened 2023-08-02
add support for 'dd' image sources

mwhudson:
I think only a core dd image source will work for now. Probably.

### DIFF
--- a/examples/sources/core.yaml
+++ b/examples/sources/core.yaml
@@ -4,7 +4,7 @@
   locale_support: none
   name:
     en: Ubuntu Server
-  path: ubuntu-core.squashfs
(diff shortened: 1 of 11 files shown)


mwhudson:
With this and the other just-proposed PRs I can install core desktop \o/

Please feel free to suggest ways to split this commit up!

dbungert (commented):
Integration tests are mad, and a integration test (or pseudo-integration test in the unit tests) would be nice.

otherwise makes sense and LGTM.

mwhudson:
Can you take a look at what I did to the integration tests?

dbungert (approved):

PR #1753, opened 2023-08-02
shutdown: use direct run_command

dbungert:


dbungert:
On s390x, it has been observed by @frank-heimes that recently the reboot button isn't actually rebooting.  I'm assuming that something is going wrong with process shutdown that we aren't actually completing the reboot, and Frank reports that a test snap that I sent him with this change improves things.

mwhudson (approved):
I don't see why this would help really? But it doesn't look harmful

PR #1755, opened 2023-08-03
Make sure journalctl -b is properly captured on shutdown

ogayot:
With recent changes, the `journalctl -b` command would fail to run:

```python
      File "subiquity/server/controllers/shutdown.py", line 118, in copy_logs_to_target
        await self.app.command_runner.run(
      File "subiquity/server/runner.py", line 98, in run
        proc = await self.start(cmd, **opts)
                     ^^^^^^^^^^^^^^^^^^^^^^^
    TypeError: DryRunCommandRunner.start() got an unexpected keyword argument 'stdout'
```

Fixed by making sure that:
* the output of `journalctl -b` is captured instead of going into the journal (which in itself feels like a bug).
* the command runner expects the `stdout`, `stderr` and other possible keyword arguments.

### DIFF
--- a/subiquity/server/controllers/shutdown.py
+++ b/subiquity/server/controllers/shutdown.py
@@ -120,7 +120,10 @@ class ShutdownController(SubiquityController):
         try:
             with open_perms(journal_txt) as output:
                 await self.app.command_runner.run(
-                    ["journalctl", "-b"], stdout=output, stderr=subprocess.STDOUT
+                    ["journalctl", "-b"],
+                    capture=True,
+                    stdout=output,
+                    stderr=subprocess.STDOUT,
                 )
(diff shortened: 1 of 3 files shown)


ogayot:
With this change, the `installer-journal.txt` file will contain some boilerplate data as shown in the example above:

```
Running as unit: run-u1060.service
not running: journalctl -b      <---- This is the output of the command
Finished with result: success
Main processes terminated with: code=exited/status=0
Service runtime: 4ms
CPU time consumed: 3ms
```

If this isn't ok, we should use something that does not rely on systemd-run (or try the `--quiet` option.

dbungert (approved):

mwhudson (approved):

PR #1758, opened 2023-08-08
preserve old loaders, and set bootnext to rp

mwhudson:


dbungert (approved):
I'm not suggesting we do it today, but we probably want to sort this out in curtin.  efibootmgr calls in two places sounds fragile.

mwhudson:
> I'm not suggesting we do it today, but we probably want to sort this out in curtin. efibootmgr calls in two places sounds fragile.

Yeah it's not great. Currently we don't communicate the fact that there is anything special about the recovery partition to curtin at all, which we'd have to do to get it to treat it specially. There's also the issue that efibootmgr just doesn't provide the information we need to do the things I want to do -- maybe we'll end up having to write a python wrapper for libefivar or something...

PR #1761, opened 2023-08-10
several fixes around reset partition handling

mwhudson:
1. adjust the RP to use a different casper uuid, to avoid confusing when install media is still present
2. record the partuuid of the RP on the kernel command line when booted from there so subiquity knows to add a grub boot entry pointing to it in this case
3. fix generating the grub boot entry pointing to the RP as this was broken in a few ways
4. get the logic the right way around to only set BootNext in reset-partition-only case


mwhudson:
Apologies for doing this in one commit. I can probably recreate a more detailed commit history if you want.

medicalwei:
The EFI boot entry for reset partition is not created after running "reset-partition-only" autoinstall config.

subiquity/server/controllers/tests/test_install.py
```diff
@@ -221,23 +226,56 @@ async def test_install_package(self, m_sleep):
...
+        m_get_efibootmgr.side_effect = iter([efi_state_no_rp, efi_state_with_rp])
+        self.setup_rp_test()
+        await self.controller.configure_rp_boot_uefi(rp=self.part)
+        calls = [
+            call(
+                [
+                    "efibootmgr",
+                    "--create",
+                    "--loader",
+                    "\\EFI\\boot\\shimx64.efi",
```
medicalwei, line 254: According to https://cdimage.ubuntu.com/daily-live/current/mantic-desktop-amd64.list, there's no shimx64.efi in the recovery partition (which is a copy of the disk image content)

My test suggests that this boot entry won't appear in the boot selection.

medicalwei (changes requested):

subiquity/server/controllers/tests/test_install.py
mwhudson: Oops, you're right of course, fixed hopefully. Tianocore's habit of booting anything that looks plausible strikes again :(

subiquity/server/controllers/install.py
```diff
@@ -421,22 +424,81 @@ async def run_curtin_step(name, stages, step_config, source=None):
...
-                os.chmod(fp.fileno(), 0o755)
-                fp.write(conf)
+        else:
+            casper_uuid = None
+            for casper_uuid_file in glob.glob("/cdrom/.disk/casper-uuid-*"):
+                with open(casper_uuid_file) as fp:
+                    casper_uuid = fp.read().strip()
+            rp_partuuid = self.app.kernel_cmdline.get("rp-partuuid")
+            if casper_uuid is not None and rp_partuuid is not None:
+                rp = self.app.base_model.partition_by_partuuid(rp_partuuid)
```
medicalwei, line 438: Should it be
```
rp = self.app.base_model.filesystem.partition_by_partuuid(rp_partuuid)
```

medicalwei (changes requested):

subiquity/server/controllers/install.py
mwhudson: Yes. Sorry.

subiquity/server/controllers/install.py
```diff
@@ -421,22 +424,81 @@ async def run_curtin_step(name, stages, step_config, source=None):
...
+        )
+        fs_uuid, rp_uuid = cp.stdout.decode("ascii").strip().split()
+        conf = grub_reset_conf.format(
+            HEADER=generate_timestamped_header(),
+            PARTITION=rp.number,
+            FS_UUID=fs_uuid,
+            CASPER_UUID=casper_uuid,
+            RP_UUID=rp_uuid,
+        )
+        with open(self.tpath("etc/grub.d/99_reset"), "w") as fp:
```
medicalwei, line 499: The following `update-grub` will need to write to the exact file.

I got a failure saying:
```
/usr/sbin/grub-mkconfig: 301: /etc/grub.d/99_reset: Text file busy
```

dbungert (approved):
one nitpick

subiquity/server/controllers/install.py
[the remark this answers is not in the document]
mwhudson: This is confusing and should probably be rewritten to be less confusing. There are two cases though:

1. we are creating and populating the RP (this is when fs_controller.reset_partition is set) 
2. we are running from the RP (this is when rp-partuuid= is set on the kernel command line)

In both these cases, if we are also installing the system in a more regular way as well, we want to add the rp as a menu option in grub.

subiquity/server/controllers/install.py
```diff
@@ -489,18 +557,28 @@ async def create_rp_boot_entry(self, context, rp):
...
-        if not fs_controller.reset_partition_only:
+        if self.model.target is None:
             cmd = [
                 "efibootmgr",
                 "--bootnext",
                 rp_bootnum,
             ]
             await self.app.command_runner.run(cmd)
 
+    async def configure_rp_boot(self, context, rp: Partition, casper_uuid: str):
```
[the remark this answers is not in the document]
mwhudson: Is it? It's definitely not valid to call this function with casper_uuid = None -- I think anyway. Will have a look and get back to you :-)

PR #1763, opened 2023-08-14
Add content and change to reST

s-makin:
- Add content from /documentation folder into /doc (with Diataxis)
- Change all content from .md to .rst
- Add some tutorial content from Server docs
- Move introduction to top-level navigation to make it easier to find
- Update README.md with instructions for building local docs

This is a large "first step", but there's a lot of content to be moved. At
least it populates all sections with something. I have a few questions,
which I'll add as comments directly in this PR.


doc/reference/autoinstall-reference.rst
```diff
@@ -3,184 +3,1013 @@
...
-.. _ai-late-commands:
+The layout of any attached keyboard. Often systems being automatically
+installed will not have a keyboard at all in which case the value used here
+does not matter.
 
-late-commands
-=============
+The mapping's keys correspond to settings in the :file:`/etc/default/keyboard`
+configuration file. See
+`its manual page <http://manpages.ubuntu.com/manpages/bionic/en/man5/keyboard.5.html>`_
```
s-makin, line 148: This link needs to be updated, not sure what to update it to. I recommend a version that is somehow agnostic to release if possible, to reduce maintenance burden going forward.

doc/reference/autoinstall-reference.rst
```diff
@@ -3,184 +3,1013 @@
...
-.. _ai-packages:
+    network:
+      version: 2
+      ethernets:
+        enp0s31f6:
+          dhcp6: true
 
-packages
-========
+Note that because of a bug, the version of Subiquity released with 20.04 GA
```
s-makin, line 237: Should we link to the bug here?

doc/reference/autoinstall-reference.rst
```diff
@@ -3,184 +3,1013 @@
...
 
-.. _ai-snaps:
+The default is:
 
-snaps
-=====
+.. code-block:: yaml
 
-stub
+    apt:
```
s-makin, line 296: This code block has different spacing/tabbing than some of the earlier code blocks. Are they incorrect? (I'm not familiar with yaml, but the discrepancy looks odd to me).

doc/reference/autoinstall-reference.rst
```diff
@@ -3,184 +3,1013 @@
...
+Configure whether common restricted packages (including codecs) from
+[multiverse] should be installed.
+
+install
+~~~~~~~
+
+* **type:** boolean
+* **default:** ``false``
+
+Whether to install the ubuntu-restricted-addons package.
```
s-makin, line 747: Should we link to the package? (and/or, say what it's for)

doc/howto/autoinstall-quickstart.rst
```diff
@@ -0,0 +1,183 @@
...
+
+    kvm -no-reboot -m 2048 \
+        -drive file=image.img,format=raw,cache=none,if=virtio
+
+This will boot into the freshly installed system and you should be able to log
+in as ``ubuntu/ubuntu``.
+
+.. LINKS
+
+.. _22.10 ISO download page: https://releases.ubuntu.com/22.10/
```
s-makin, line 183: I'm a big believer in making links version-agnostic if possible. Would it harm the guide if we just used releases.ubuntu.com? Or do we need to use 22.10 specifically?

doc/howto/report-bugs.rst
```diff
@@ -0,0 +1,58 @@
...
+up to date as the installer changes.
+
+Update Subiquity
+================
+
+The first thing to do is to update your Subiquity snap. Not only because we
+fix issues that cause failures over time but also because we've been working on
+features to make failure reporting easier.
+
+<command here? What if the snap isn't installed?>
```
s-makin, line 19: Can we include the relevant commands here?

doc/reference/autoinstall-reference.rst
dbungert: I love the idea of a release agnotic version but not sure where that would be.  Let's bump the URL to https://manpages.ubuntu.com/manpages/focal/en/man5/keyboard.5.html for now.

doc/howto/autoinstall-quickstart.rst
dbungert: yes, this is a great place for a generic link for releases.ubuntu.com

doc/reference/autoinstall-reference.rst
mwhudson: Somehow I recently learned that https://manpages.ubuntu.com/manpages/keyboard.5.html works as a release agnostic version. It would be nice to have a "latest LTS" link but I don't know how any of this works!

doc/explanation/configure-storage.rst
```diff
@@ -0,0 +1,177 @@
...
+disk, which will automatically be created when a disk is selected as a boot
+device (a disk with an existing GPT partition table can only be used as a boot
+device if it has this partition).
+
+UEFI mode
+---------
+
+In UEFI mode, the bootloader loaded from a "EFI System Partition" (ESP), which
+is a partition with a particular type GUID. The installer automatically creates
+a 512MiB ESP on a disk when it is selected as a boot device and will install
```
dbungert, line 143: I'd like to omit the size default, it's subject to frequent debate and also scales to available space.
If we do leave a number in, the current minimum is 538 MiB.

doc/explanation/configure-storage.rst
```diff
@@ -0,0 +1,177 @@
...
+all logical volumes are linear and all MD raid devices use the default metadata
+format (1.2).
+
+These limits can both be worked around in the same way: drop to a shell and use
+the usual shell commands to edit the partition table or create the LV or RAID
+with desired parameters, and then select these partitions or devices as mount
+points in the installer. Any changes you make while the installer is running
+but before altering the storage configuration will reflected in the installer.
+
+The installer cannot yet configure iSCSI mounts, ZFS at all, or btrfs
```
dbungert, line 176: Let's remove the mention of ZFS, we'll be able to offer a limited form in 23.10 (but not through the TUI)

doc/howto/report-bugs.rst
dbungert: My sense for the docs that the only case we should talk about updating snaps is if there is a new stable version released, which means the snap will be offered anyhow.

The only cases why a stable snap would not be offered are:
1) offline installs or network issues
2) old install ISOs that no longer receive newer subiquity snaps being offered (bionic)

My sense is that we can remind people here of the purpose of the refresh page, and if they are offline to try again online and see if a newer snap is offered.

doc/howto/autoinstall-quickstart.rst
s-makin: I have taken the liberty of making the entire page release-agnostic. Please take a look and let me know if I've changed anything I shouldn't (or missed anything I should have!)

doc/explanation/configure-storage.rst
s-makin: So can it configure ZFS now? Or not yet?

doc/explanation/configure-storage.rst
s-makin: Could you suggest a wording you'd rather use instead, and I'll change it to that? (I think it'd be better to leave the minimum number in there if possible, it's always helpful to give people even an order-of-magnitude estimate of how much space will be required)

doc/reference/autoinstall-reference.rst
dbungert: That sounds like a great idea but I'm not sure what that would be!  The best I've come up with is the description in the control file but that's not much better:
https://git.launchpad.net/ubuntu/+source/ubuntu-restricted-addons/tree/debian/control#n13

@local-optimum - do you have a suggestion here for how we describe what's going on with this "codecs" functionality?

doc/reference/autoinstall-reference.rst
dbungert: Looking at our examples autoinstall files, we mostly use a 2 space indent for yaml, so I propose that we standardize on that.

doc/explanation/configure-storage.rst
dbungert: yes, but the 23.10 answer will be:
desktop -> a checkbox in advanced
desktop and server -> an option in autoinstall (https://github.com/canonical/subiquity/pull/1763/files#diff-d2236521d63f00962c3248afbee5d33b032b8dfd54c29e5310fe492655fa8267R409)

doc/explanation/configure-storage.rst
dbungert: maybe just the word "approximately" thrown in there?

> The installer automatically creates an ESP on a disk, approximately 512MiB in size, when it is selected as a boot device ...

doc/reference/autoinstall-reference.rst
dbungert: I found the commit in question, but not a bug number.  I'm not sure this has a specific bug number, maybe we focus less on the bug part and just say "20.04 GA behavior is ..."

dbungert: Long term I want to do the versioned docs thing so someone can pull up subiquity version X and see a version of the autoinstall reference correct for that release.  That may be difficult for old releases but it would be nice to make new ones look like that.

dbungert: confirmed https://github.com/canonical/subiquity/pull/739 - the discourse report is good but doesn't offer much additional context, I suggest omitting a link.

s-makin: Understood. Thanks for clarifying!

doc/reference/autoinstall-reference.rst
s-makin: It shall be done!

doc/explanation/configure-storage.rst
s-makin: Understood, thanks for the clarification

doc/explanation/configure-storage.rst
s-makin: I've made a slight tweak. I've gone with the "minimum 538" option as I think it's clearer for users.

dbungert (approved):

PR #1764, opened 2023-08-16
filesystem: Use the first fitting gap for ESP, instead of the largest gap

medicalwei:
This is to handle the issue, that when using recovery (reset) partition to install to the same disk, it:

  1. removes all partitions except recovery partition on the target disk, then
  2. picks the largest gap to install ESP, usually after the recovery partition.

This should not affect disk installation where the disk partitions are all removed, but affects installations where there are some partitions exist but not removed.

### DIFF
--- a/subiquity/common/filesystem/boot.py
+++ b/subiquity/common/filesystem/boot.py
@@ -224,8 +224,22 @@ def get_add_part_plan(device, *, spec, args, resize_partition=None):
     # is a bad idea.  So avoid putting any sort of boot stuff on a logical -
     # it's probably a bad idea for all cases.
 
-    gap = gaps.largest_gap(device, in_extended=False)
-    if gap is not None and gap.size >= size and gap.is_usable:
+    # find the first available gap (not always the largest one)
+    gap = next(
+        iter(
+            [
+                gap
+                for gap in gaps.parts_and_gaps(device)
+                if isinstance(gap, gaps.Gap)
+                and not gap.in_extended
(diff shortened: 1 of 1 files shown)


subiquity/common/filesystem/boot.py
```diff
@@ -224,8 +224,22 @@ def get_add_part_plan(device, *, spec, args, resize_partition=None):
     # is a bad idea.  So avoid putting any sort of boot stuff on a logical -
     # it's probably a bad idea for all cases.
 
-    gap = gaps.largest_gap(device, in_extended=False)
-    if gap is not None and gap.size >= size and gap.is_usable:
+    # find the first available gap (not always the largest one)
+    gap = next(
```
mwhudson, line 228: I generally prefer explicit for loops to this kind of thing but well

mwhudson (commented):
Thanks for this, it looks ok on my phone. Will look properly tomorrow

dbungert (approved):
This is an improvement but I think we should relocate this to gaps.py at some point.  Can easily imagine other cases of wanting a gap that supports a certain size instead of always demanding the largest.

Note that this code doesn't check the criteria where a gap would be unusable (TOO_MANY_PRIMARY_PARTS) but I guess that's not a scenario here.

dbungert:
@medicalwei Thanks for the PR, but I think it's obsolete given the implementation in #1765

medicalwei:
Closing this in favor of #1765, thanks!

PR #1765, opened 2023-08-17
Use the first fitting gap for ESP, instead of the largest gap

mwhudson:
This is basically @medicalwei's #1764 with potentially useful functionality factored into a method.

### DIFF
--- a/subiquity/common/filesystem/boot.py
+++ b/subiquity/common/filesystem/boot.py
@@ -224,7 +224,7 @@ def get_add_part_plan(device, *, spec, args, resize_partition=None):
     # is a bad idea.  So avoid putting any sort of boot stuff on a logical -
     # it's probably a bad idea for all cases.
 
-    gap = gaps.largest_gap(device, in_extended=False)
+    gap = gaps.gap_with_size(device, size, in_extended=False)
     if gap is not None and gap.size >= size and gap.is_usable:
         create_part_plan.gap = gap.split(size)[0]
         return create_part_plan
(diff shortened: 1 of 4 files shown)


subiquity/common/filesystem/gaps.py
```diff
@@ -252,6 +252,14 @@ def largest_gap_size(device, in_extended=None):
     return 0
 
 
+def gap_with_size(device, size, *, in_extended=None):
+    for pg in parts_and_gaps(device):
```
dbungert, line 256: `gap_with_size` sounds like any gap will do that fits the requirement, or maybe even the closest gap (first gap is large, second gap is small but large enough, choose the second).  Would you add a docstring clarifying intent?

subiquity/common/filesystem/boot.py
```diff
@@ -224,7 +224,7 @@ def get_add_part_plan(device, *, spec, args, resize_partition=None):
     # is a bad idea.  So avoid putting any sort of boot stuff on a logical -
     # it's probably a bad idea for all cases.
 
-    gap = gaps.largest_gap(device, in_extended=False)
+    gap = gaps.gap_with_size(device, size, in_extended=False)
     if gap is not None and gap.size >= size and gap.is_usable:
```
dbungert, line 228: Some of the logic is redundant now

dbungert (approved):

subiquity/common/filesystem/gaps.py
mwhudson: I changed the method name to 'first_gap_with_size'. For this use case, finding the smallest suitable gap might actually make more sense but well. Minor stuff I think.

dbungert (approved):

PR #1766, opened 2023-08-17
Add API and TUI support for LUKS recovery key

ogayot:
This is basically working and ready to be reviewed IMO. Some more automated / units tests would be beneficial.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1281,7 +1281,7 @@ class ActionRenderMode(enum.Enum):
     FORMAT_MOUNT = enum.auto()
 
 
-class FilesystemModel(object):
+class FilesystemModel:
     target = None
 
     _partition_alignment_data = {
@@ -1330,10 +1330,11 @@ class FilesystemModel(object):
         else:
             return Bootloader.BIOS
 
-    def __init__(self, bootloader=None):
+    def __init__(self, bootloader=None, *, root: str):
         if bootloader is None:
             bootloader = self._probe_bootloader()
         self.bootloader = bootloader
+        self.root = root
         self.storage_version = 1
         self._probe_data = None
         self.dd_target: Optional[Disk] = None
@@ -1356,7 +1357,7 @@ class FilesystemModel(object):
         # the original state.  _orig_config plays a similar role, but is
         # expressed in terms of curtin actions, which are not what we want to
         # use on the V2 storage API.
-        orig_model = FilesystemModel(self.bootloader)
+        orig_model = FilesystemModel(self.bootloader, root=self.root)
         orig_model.target = self.target
         orig_model.load_probe_data(self._probe_data)
         return orig_model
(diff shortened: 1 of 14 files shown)


subiquity/models/filesystem.py
```diff
@@ -47,6 +48,84 @@ class NotFinalPartitionError(Exception):
...
+    live_location: Optional[pathlib.Path]
+    # Where to store the key in the target system. /target will automatically
+    # be prefixed.
+    backup_location: pathlib.Path
+
+    _key: Optional[str] = attr.ib(repr=False, default=None)
+
+    @classmethod
+    def from_spec(
+        cls, spec: Optional[Dict[str, Any]]
```
dbungert, line 63: We do use these "spec" dictionaries for filesystem stuff, but want to move away from them.  Would you create a new attrs object containing your desired fields?

subiquity/models/subiquity.py
```diff
@@ -176,11 +176,13 @@ def __init__(self, root, hub, install_model_names, postinstall_model_names):
             self.target = root
             self.chroot_prefix = []
 
+        log.debug("root is %s", root)
```
dbungert, line 179: debug stuff, is this meant for inclusion in the PR?

subiquity/ui/views/filesystem/lvm.py
```diff
@@ -205,6 +221,17 @@ def done(self, sender):
...
         mdc = self.form.devices.widget
         result["devices"] = mdc.active_devices
+        if "create_recovery_key" in result:
+            if result["create_recovery_key"]:
+                backup_prefix = pathlib.Path("/var/log/installer")
+                backup_filename = pathlib.Path(f"recovery-key-{result['name']}.txt")
+                result["recovery-key"] = {
+                    "backup-location": str(backup_prefix / backup_filename),
+                    "live-location": str(
+                        pathlib.Path("~").expanduser() / backup_filename
```
dbungert, line 231: the use of the term "backup" sounds like a copy-paste error but really it's fine, so maybe rename backup_filename to a term that users neither the word "backup" nor "live".

subiquity/server/controllers/install.py
```diff
@@ -688,6 +688,7 @@ async def postinstall(self, *, context):
                     hostname = f.read().strip()
 
             await self.app.controllers.Ad.join_domain(hostname, context)
+        self.model.filesystem.copy_artifacts_to_target()
```
dbungert, line 691: user hasn't been created yet

mwhudson: but the key is getting copied to /var/log/installer right?

mwhudson (commented):
I agree with Dan :-)

subiquity/models/subiquity.py
ogayot: Nope! thanks for noticing!

subiquity/server/controllers/install.py
ogayot: Yes, it's exposed in $HOME in the live session, but then ends up copied to /var/log/installer in the target.

subiquity/ui/views/filesystem/lvm.py
ogayot: Good point! Renaming to "filename".

dbungert (approved):

PR #1767, opened 2023-08-17
oem: fix IndexError when using OEMController.wants_oem_kernel

ogayot:
```python
    File "subiquity/server/controllers/oem.py", line 161, in load_metapackages_list
      self.model.metapackages = [
    File "subiquity/server/controllers/oem.py", line 164, in <listcomp>
      wants_oem_kernel=await self.wants_oem_kernel(
    File "subiquity/server/controllers/oem.py", line 115, in wants_oem_kernel
      flavor = line.split("=", maxsplit=1)[1].strip()
  IndexError: list index out of range
```


mwhudson (approved):
Oops. Black seems sad?

PR #1770, opened 2023-08-24
Create zpools / zfs datasets like ubiquity does

dbungert:


subiquity/models/filesystem.py
```diff
@@ -1115,6 +1115,8 @@ class ZPool:
     # default dataset options for the zfses in the pool
     fs_properties: Optional[dict] = None
 
+    default_features: Optional[bool] = True
```
mwhudson, line 1118: Er this doesn't seem to match the change to made to curtin in the end. Did you forget to push an update to this branch?

dbungert: No, I think this is correct, just confusing (and this is the less confusing version!)
See also the matching curtin side, https://git.launchpad.net/curtin/commit/?id=4195867e55953afd8c2481f813db3cda17791ead

in the simplest `zpool create $poolname $device` case, no `-d` arg is present and we include the default features.
If we want maximum control, we can pass `-d`, default_features=False, and then supply feature flags that we want.

mwhudson: oh ok

mwhudson (approved):
I have not checked the consistency with ubiquity line by line. I can if you want!

PR #1771, opened 2023-08-25
several: add --no-enhanced-secureboot flag

dbungert:
It disables all core boot types really.  But this is the only planned core boot type at this point.

LP: #2030852

### DIFF
--- a/subiquity/cmd/server.py
+++ b/subiquity/cmd/server.py
@@ -129,6 +129,13 @@ def make_server_args_parser():
         default=".subiquity",
         help="in dryrun, control basedir of files",
     )
+    parser.add_argument(
+        "--no-enhanced-secureboot",
+        dest="enhanced_secureboot",
+        action="store_false",
+        default=True,
+    )
+
     parser.add_argument("--storage-version", action="store", type=int)
     parser.add_argument("--use-os-prober", action="store_true", default=False)
     parser.add_argument(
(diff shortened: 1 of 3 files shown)


subiquity/cmd/server.py
```diff
@@ -129,6 +129,12 @@ def make_server_args_parser():
         default=".subiquity",
         help="in dryrun, control basedir of files",
     )
+    parser.add_argument(
+        "--no-enhanced-secureboot",
+        dest="enhanced_secureboot",
+        action="store_false",
+        default=True,
+    )
```
mwhudson, line 137: Can't you say `parser.add_argument("--enhanced-secureboot", action="store_true", default=True)` to get the same effect? (i.e. doesn't argparse add --no-$blah options automatically for boolean options?)

subiquity/server/controllers/filesystem.py
```diff
@@ -294,6 +294,9 @@ async def _unmount_systems_dir(self):
             self._source_handler = None
 
     async def _get_system(self, variation_name, label):
+        if not self.app.opts.enhanced_secureboot:
```
mwhudson, line 297: It probably doesn't make any difference in practice, but I think we should continue to offer options that have a snapd_system_label but the model thus referenced does not include a gadget. I think 

```
             if system is not None and len(system.volumes) > 0:
+                if not self.app.opts.enhanced_secureboot:
+                    continue
                 info = self.info_for_system(name, label, system)
```

in _examine_systems would be better.

mwhudson (approved):
fine but i think it could be even better!

subiquity/cmd/server.py
dbungert: `action=argparse.BooleanOptionalAction` is the magic

subiquity/server/controllers/filesystem.py

subiquity/cmd/server.py
dbungert: py3.9.  I'm going to use focal-compatible syntax rather than mess with integration tests for this.

PR #1773, opened 2023-08-25
install zfsutils-linux if needed

dbungert:


subiquity/server/controllers/install.py
```diff
@@ -627,8 +632,7 @@ async def install(self, *, context):
                 fsc = self.app.controllers.Filesystem
                 for_install_path = self.model.source.get_source(fsc._info.name)
 
-            if self.app.controllers.Filesystem.reset_partition:
-                self.app.package_installer.start_installing_pkg("efibootmgr")
+            await self.install_live_packages()
```
mwhudson, line 635: It's probably not really important, but I was letting the install in the live environment run in parallel with the other bits here, whereas this will wait for the installs to be done. Was that intended?

subiquity/server/controllers/install.py
```diff
@@ -597,6 +597,11 @@ async def create_core_boot_classic_fstab(self, *, context):
         with open(self.tpath("etc/fstab"), "w") as fp:
             fp.write("/run/mnt/ubuntu-boot/EFI/ubuntu /boot/grub none bind\n")
 
+    @with_context(description="installing packages to live system")
+    async def install_live_packages(self, *, context):
```
mwhudson, line 601: This will show "installing packages to live system" to the user even if there are no packages being installed. It's annoying to fix but that's kind of bad UX.

mwhudson (approved):
Approved but a couple of comments/questions

subiquity/server/controllers/install.py
dbungert: having zfsutils-linux up front is intended, we'll need them fairly early (before rsync).
efibootmgr is safe with the later call waiting for task completion.
Updated to make this more nuanced, by allowing zfsutils-linux to be required up front and efibootmgr can run in the background.

subiquity/server/controllers/install.py
dbungert: good point, fixed

dbungert:
The installdeps change is to address the CI failure on mantic due to conffile prompt.  

Prompt issue reported at https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/2033308

mwhudson (approved):
Feels like this might be a bit overengineered now but well whatever, you've done all the required typing!

PR #1774, opened 2023-08-29
filesystem: capability flags in a specific order

dbungert:
Show DIRECT first, list CORE_BOOT ones later.
LP: #2033244


mwhudson (approved):
I see get_classic_capabilities orders by name (I wonder why!), I guess that should be fixed? It only matters for the use_gap targets I think.

PR #1777, opened 2023-08-30
refresh: fix crash during autoinstall

dbungert:
Controllers have started, but we have decided that no refresh check is needed, so no check_task was started (or assigned).

GET /refresh is called, resulting in:
DEBUG subiquity.server.server:446 request to /refresh?wait=true crashed Traceback (most recent call last):
  File "subiquity/server/controllers/refresh.py", line 233, in GET
    await self.check_task.wait()
AttributeError: 'NoneType' object has no attribute 'wait'


mwhudson (approved):
Makes sense. Not sure why this endpoint would be called in this case? But anyway it shouldn't crash, that's for sure.

PR #1779, opened 2023-08-31
disable deb822 apt sources during offline install

mwhudson:


mwhudson:
random thought: we should be using pathlib a lot more

ogayot:
Would you explain why this change is needed?

subiquity/server/apt.py
```diff
@@ -242,6 +243,10 @@ async def configure_for_install(self, context):
             proxy_path = self.install_tree.p("etc/apt/apt.conf.d/90curtin-aptproxy")
             if os.path.exists(proxy_path):
                 os.unlink(proxy_path)
+            for sources_file in glob.glob(
+                self.install_tree.p("etc/apt/sources.list.d/*.sources")
```
dbungert, line 247: there could be legacy-style sources here in .list format, so I think we want to hit every file in that directory.
also, up above in the has_network == True case, the sources may now only be at a file called etc/apt/sources.list.d/ubuntu.sources (so the os.rename may fail on a non-existent sources.list file), or we can have both, or the files in sources.list.d may have different names.

mwhudson: ok i went to town a little more, let me know what you think. we should probably add an integration test that exercises the no-network path somehow...

dbungert (approved):
This looks correct.  +1 to test automation around it.
I'm merging now before test automation is sorted out.

PR #1780, opened 2023-08-31
oem+drivers: fix wrong order or lower layers when building overlays

ogayot:
When doing an offline install, ubuntu-drivers would sometimes list a package that is available in the archive but not present in the pool. This would lead to a failed install when trying to download/install such package.

This is not something we would expect since we run apt-get update (with only the pool configured when offline) in the install tree.

However, it turned out that we create the overlay with the lower layers specified in the wrong order - which essentially makes APT indexes visible in the source tree also visible in the OEM/third-party driver overlay.

When calling `setup_overlay(lowers=[a, b, c])`, Subiquity invokes mount with `lowerdir=c:b:a` (in the reverse order).

This means that c is top, b is middle and a is bottom.

For the OEM and third-party drivers, we build overlays that are based on:

 * the source tree
 * the configured tree
 * the install tree

Unfortunately, we were doing the opposite. Fixed by reversing the order of the lower layers.

### DIFF
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -268,9 +268,9 @@ class AptConfigurer:
     async def overlay(self):
         overlay = await self.mounter.setup_overlay(
             [
-                self.install_tree.upperdir,
-                self.configured_tree.upperdir,
                 self.source_path,
+                self.configured_tree.upperdir,
+                self.install_tree.upperdir,
             ]
         )
         try:
--- a/subiquity/server/mounter.py
+++ b/subiquity/server/mounter.py
@@ -155,6 +155,15 @@ class Mounter:
                 path.unlink(missing_ok=True)
 
     async def setup_overlay(self, lowers: List[Lower]) -> OverlayMountpoint:
+        """Setup a RW overlay FS over one or more lower layers.
+        Be careful, when multiple lower layers are specified, they are stacked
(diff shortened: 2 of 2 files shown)


ogayot:
With this patch, I am able to make an offline install on certified hardware when the OEM meta-packages are not in the pool. Essentially, it treats the hardware as if it weren't certified. 

Online installs still behave as expected (the OEM meta-package + kernel gets installed).

mwhudson (approved):
Your logic seems sound and the change correct. The question I have is how on earth did we get away with this for so long???

dbungert (approved):
Fine to take the fix, but I think we soon want to have setup_overlay() take the list of lowers in the same order that the mount option expects.  That reversed order detail would be quite the surprise to someone who hadn't read this comment yet.

ogayot:
> The question I have is how on earth did we get away with this for so long???

Yeah that's difficult to imagine :D I guess nobody ever tested an offline install with an nvidia chip + "Search drivers [x]" selected? I can't test that scenario myself but I expect it would have failed too..

> I think we soon want to have setup_overlay() take the list of lowers in the same order that the mount option expects. That reversed order detail would be quite the surprise to someone who hadn't read this comment yet.

Yes, I would not mind removing the reversed logic either. I agree that mount stacking lower layers from right to left can look strange though.

PR #1781, opened 2023-09-01
Min size revisited

dbungert:
* if a source entry has variations, use the install_min of the largest of the variations
* several tweaks to suggested min install size
* clearer logging of this outcome
* refresh the desktop install sources yaml

analysis of the above

```
10G drive tests
no case here creates a swapfile

CORE_BOOT_ENCRYPTED, canary iso build 20230831.7
  $ df -h -x tmpfs
  Filesystem      Size  Used Avail Use% Mounted on
  /dev/vda2       721M   61M  609M   9% /run/mnt/ubuntu-boot
  /dev/vda1       749M  3.7M  745M   1% /run/mnt/ubuntu-seed
  /dev/dm-0       8.3G  4.4G  3.5G  56% /
  /dev/dm-1        20M  120K   18M   1% /var/lib/snapd/save

  suggested install minimum size: 8.339G
  catalog size: 4.042G
  calculated min_size (end offset of last partition): 5.497G

LVM_LUKS, canary iso build 20230831.7
  $ df -h -x tmpfs
  Filesystem                         Size  Used Avail Use% Mounted on
  /dev/mapper/ubuntu--vg-ubuntu--lv  7.5G  4.9G  2.2G  69% /
  /dev/vda2                          1.7G  137M  1.5G   9% /boot
  /dev/vda1                          537M  6.1M  531M   2% /boot/efi

  suggested install minimum size: 8.339G
  catalog size: 3.77G

LVM_LUKS, desktop iso build 20230831.4
  $ df -h -x tmpfs
  Filesystem                         Size  Used Avail Use% Mounted on
  /dev/mapper/ubuntu--vg-ubuntu--lv  7.5G  4.9G  2.3G  68% /
  /dev/vda2                          1.7G  137M  1.5G   9% /boot
  /dev/vda1                          537M  6.1M  531M   2% /boot/efi

  suggested install minimum size: 9.644G
  catalog size: 4.91G

LVM_LUKS, live-server iso build 20230831
  $ df -h -x tmpfs
  Filesystem                         Size  Used Avail Use% Mounted on
  /dev/mapper/ubuntu--vg-ubuntu--lv  7.5G  2.5G  4.7G  36% /
  /dev/vda2                          1.7G  142M  1.5G   9% /boot
  /dev/vda1                          537M  6.1M  531M   2% /boot/efi

  using source: ubuntu-server
  suggested install minimum size: 5.530G
  catalog size: 1.25G
```

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -157,13 +157,8 @@ class VariationInfo:
     def capability_info_for_gap(
         self,
         gap: gaps.Gap,
+        install_min: int,
     ) -> CapabilityInfo:
-        if self.label is None:
-            install_min = sizes.calculate_suggested_install_min(
-                self.min_size, gap.device.alignment_data().part_align
-            )
-        else:
-            install_min = self.min_size
         r = CapabilityInfo()
         r.disallowed = list(self.capability_info.disallowed)
         if self.capability_info.allowed and gap.size < install_min:
@@ -1015,7 +1010,9 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             capability_info = CapabilityInfo()
             for variation in self._variation_info.values():
                 gap = gaps.largest_gap(disk._reformatted())
-                capability_info.combine(variation.capability_info_for_gap(gap))
+                capability_info.combine(
+                    variation.capability_info_for_gap(gap, install_min)
+                )
             reformat = GuidedStorageTargetReformat(
                 disk_id=disk.id,
                 allowed=capability_info.allowed,
@@ -1042,7 +1039,9 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             for variation in self._variation_info.values():
                 if variation.is_core_boot_classic():
                     continue
-                capability_info.combine(variation.capability_info_for_gap(gap))
+                capability_info.combine(
+                    variation.capability_info_for_gap(gap, install_min)
+                )
             api_gap = labels.for_client(gap)
             use_gap = GuidedStorageTargetUseGap(
                 disk_id=disk.id,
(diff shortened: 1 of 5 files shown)


subiquity/tests/api/test_api.py
```diff
@@ -386,7 +386,7 @@ async def test_v2_flow(self):
         cfg = self.machineConfig("examples/machines/simple.json")
         with cfg.edit() as data:
             attrs = data["storage"]["blockdev"]["/dev/sda"]["attrs"]
-            attrs["size"] = str(25 << 30)
+            attrs["size"] = str(10 << 30)
```
mwhudson, line 389: Is that not the size in the json already?

subiquity/server/controllers/filesystem.py
```diff
@@ -157,13 +158,8 @@ def is_valid(self) -> bool:
     def capability_info_for_gap(
         self,
         gap: gaps.Gap,
+        install_min: int,
```
mwhudson, line 161: This is fine for now, but it strikes me that the only use this methods makes of its arguments now is "gap.size < install_min", which I don't know feels a bit strange, like maybe callers should be checking that and there should be a method on CapabilityInfo that just returns a "too small" variant of the info. Or something.

mwhudson (approved):
Looks OK although I think things have ended up structured a little strangely. But this should make the behaviour much less surprising.

subiquity/tests/api/test_api.py
dbungert: It used to be, we added a zero to the default size of this file since it's so high traffic for dry-run testing and allows for more stuff.
Also I like the idea of at least one API test having a low number like this, so that it's independent of whatever other tweaks simple.json might see.

PR #1782, opened 2023-09-01
api: encode x-error-msg as JSON - so it does not contain <CR> or <LF>

ogayot:
When the server raises an exception in a HTTP request handler context, more often than not, the exception is sent back to the client in the body.

Additionally, the message of the exception (if any), is also copied as is in a `x-error-msg` HTTP header.

That said, HTTP headers must obey strict rules. The `"\r\n"` sequence indicate the end of the current HTTP header. When using aiohttp, the library rejects any header that has a `"\r"` or `"\n"` in its value:

```python
ValueError: Newline or carriage return character detected in HTTP status message or header. This is a potential security issue.
```

As an example, any `curtin.util.ProcessExecutionError` exception will contain `"\n"` characters when converted into a string.

We now encode the error message as JSON before copying it in the HTTP header.

### DIFF
--- a/subiquity/common/api/server.py
+++ b/subiquity/common/api/server.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import inspect
+import json
 import logging
 import os
 import traceback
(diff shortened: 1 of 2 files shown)


mwhudson (approved):
Oops

PR #1784, opened 2023-09-04
Disallow core boot if searching for third party drivers

mwhudson:
This disables core boot options if the user selected "search for third party drivers".

It also adapts the recent fix to not offer core boot options on non-UEFI systems to indicate that this is happening via a disallowed guided option.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -419,7 +419,23 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
                     info.capability_info.disallow_if(
                         lambda cap: cap.is_core_boot(),
                         GuidedDisallowedCapabilityReason.NOT_UEFI,
-                        "Enhanced secure boot options only available on UEFI systems.",
+                        _(
+                            "Enhanced secure boot options only available on UEFI "
+                            "systems."
+                        ),
+                    )
+                if self.app.base_model.source.search_drivers:
+                    log.debug(
+                        "Disabling core boot based install options as third-party "
+                        "drivers selected"
+                    )
+                    info.capability_info.disallow_if(
+                        lambda cap: cap.is_core_boot(),
+                        GuidedDisallowedCapabilityReason.THIRD_PARTY_DRIVERS,
+                        _(
+                            "Enhanced secure boot options cannot currently install "
+                            "third party drivers."
+                        ),
                     )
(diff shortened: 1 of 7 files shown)


mwhudson:
Hmm this has a poor interaction with the autoinstall default for search_drivers -- basically, if any autoinstall file is passed at all, search_drivers defaults to True. Maybe we should think of a path towards changing that, or otherwise we need to be more cautious and not apply this as a default until we have checked to see if the user has asked for a core boot install or not (which is a bit of a layering violation, sigh)

subiquity/server/controllers/filesystem.py
```diff
@@ -288,6 +289,11 @@ async def configured(self):
         self._configured = True
         if self._info is None:
             self.set_info_for_capability(GuidedCapability.DIRECT)
+        if (
+            self.app.base_model.source.search_drivers
+            is SEARCH_DRIVERS_AUTOINSTALL_DEFAULT
+        ):
+            self.app.base_model.source.search_drivers = not self.is_core_boot_classic()
```
dbungert, line 296: This ends up kind of a weird default.  I think we should note it in the autoinstall reference.
The reference will need update anyhow for core boot but I don't want to forget this part.

dbungert (approved):

subiquity/server/controllers/filesystem.py
mwhudson: Yeah this behaviour is a bit strange. I've added a note.

dbungert: Thank you, I like the doc blurb

dbungert (approved):

subiquity/server/controllers/filesystem.py
```diff
@@ -160,18 +193,11 @@ def capability_info_for_gap(
...
-                r.disallowed.append(
-                    GuidedDisallowedCapability(
-                        capability=capability,
-                        reason=GuidedDisallowedCapabilityReason.TOO_SMALL,
-                    )
-                )
-        else:
-            r.allowed = list(self.capability_info.allowed)
+        r = self.capability_info.copy()
+        if gap.size < install_min:
```
dbungert, line 197:
```suggestion
        if gap is None or gap.size < install_min:
```

dbungert: Minor conflict to resolve, something like this should do it

PR #1785, opened 2023-09-04
network: fix Wi-Fi interfaces not listed in dry-run

ogayot:
When a Wi-Fi interface is present in the machine configuration (e.g., mwhudson.json), the GUI seemingly ignores it. This happens because there is a filter on the server side which only returns Wi-Fi interfaces if the wlan_support_install_state() function returns
PackageInstallState.DONE.

However, calling the /network endpoint shows that the state is set to the wrong value:

```json
 {"wlan_support_install_state": "NOT_NEEDED"}
```

This turns out to be inconsistent because:
 * we lean on a PackageInstaller instance to tell if wpasupplicant is installed (this is what the wlan_support_install_state() function reflects) ; but
 * in dry-run mode, we pretend to install wpasupplicant without actually relying on the PackageInstaller instance.

Fixed by using the PackageInstaller instance to install the wpasupplicant package - with a special implementation that only pretends to install it. This is enough to make the PackageInstaller instance think the package is installed.

### DIFF
--- a/subiquity/server/controllers/network.py
+++ b/subiquity/server/controllers/network.py
@@ -109,14 +109,22 @@ class NetworkController(BaseNetworkController, SubiquityController):
 
     async def _install_wpasupplicant(self):
         if self.opts.dry_run:
-            await asyncio.sleep(10 / self.app.scale_factor)
-            a = "DONE"
-            for k in self.app.debug_flags:
-                if k.startswith("wlan_install="):
-                    a = k.split("=", 2)[1]
-            r = getattr(PackageInstallState, a)
+
+            async def fake_install(pkgname: str) -> PackageInstallState:
+                await asyncio.sleep(10 / self.app.scale_factor)
+                a = "DONE"
+                for k in self.app.debug_flags:
+                    if k.startswith("wlan_install="):
+                        a = k.split("=", 2)[1]
+                return getattr(PackageInstallState, a)
+
+            install_coro = fake_install
         else:
-            r = await self.app.package_installer.install_pkg("wpasupplicant")
+            install_coro = None
+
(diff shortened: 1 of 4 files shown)


mwhudson (changes requested):
I think I would rather have a DryRunPackageInstaller class, like we do for CommandRunner. In fact, if we change PackageInstaller to use app.command_runner can we delete a bunch of code??

ogayot:
> Thanks for this, but I think I would rather have a DryRunPackageInstaller class, like we do for CommandRunner. In fact, if we change PackageInstaller to use app.command_runner can we delete a bunch of code??

Updated. I must admit I don't find these DryRun classes very practical. But I do like that it makes the "main" code cleaner (less cluttered with dry-run specific things).

I had to introduce a little bit of code duplication here, hope it's okay..

mwhudson (approved):
Sorry for not reviewing yesterday, I did read the code but somehow forgot to write anything down!

This looks OK. I don't think we have to regard the interpretation of SUBIQUITY_DEBUG_FLAGS with any kind of stability though -- seems like it would make things easier and more flexible to switch to using some syntax like SUBIQUITY_DEBUG_FLAGS=pkginstall=wpasupplicant=FAILED or something?

ogayot:
> Sorry for not reviewing yesterday, I did read the code but somehow forgot to write anything down!
> 
> This looks OK. I don't think we have to regard the interpretation of SUBIQUITY_DEBUG_FLAGS with any kind of stability though -- seems like it would make things easier and more flexible to switch to using some syntax like SUBIQUITY_DEBUG_FLAGS=pkginstall=wpasupplicant=FAILED or something?

Yes that sounds like good improvements to implement going forward :)

PR #1786, opened 2023-09-05
network: fix crash upon accessing Help menu after creating a bond

ogayot:
Steps to reproduce (works in dry-run mode too):

1. Navigate to the network page.
2. Create a bond
3. Go to the help menu

```python
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "subiquity/server/server.py", line 117, in ssh_info_GET
     ips.extend(map(str, dev.actual_global_ip_addresses))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "subiquitycore/models/network.py", line 394, in actual_global_ip_addresses
     for _, addr in sorted(self.info.addresses.items())
                           ^^^^^^^^^^^^^^^^^^^
 AttributeError: 'NoneType' object has no attribute 'addresses'
```

### DIFF
--- a/subiquity/server/controllers/network.py
+++ b/subiquity/server/controllers/network.py
@@ -255,6 +255,8 @@ class NetworkController(BaseNetworkController, SubiquityController):
     async def global_addresses_GET(self) -> List[str]:
         ips: List[str] = []
         for dev in self.model.get_all_netdevs():
+            if dev.info is None:
+                continue
             ips.extend(map(str, dev.actual_global_ip_addresses))
         return ips
 
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -114,6 +114,8 @@ class MetaController:
         ips: List[str] = []
         if self.app.base_model.network:
             for dev in self.app.base_model.network.get_all_netdevs():
+                if dev.info is None:
+                    continue
                 ips.extend(map(str, dev.actual_global_ip_addresses))
         if not ips:
             return None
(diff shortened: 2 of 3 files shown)


mwhudson (approved):
Hm finding bugs in code I wrote in like 2017, rude!

Would this be better fixed in the device object, like the various _address properties should be empty in this case?

ogayot:
> Hm finding bugs in code I wrote in like 2017, rude!
> 
> Would this be better fixed in the device object, like the various _address properties should be empty in this case?

Yes, I imagine it would be a better fix for it. Actually, I am trying to debug this code (because of another crash occurring when an interface is removed) and the `NetworkDev.info` instance variable is puzzling me.

When creating a `NetworkDev` instance, the `info` variable gets set to `None`. IIUC, we expect it to hold a `NetDevInfo` instance at a later stage. To create such instance, we can call the `dev.netdev_info()` method ... but it is implemented in a way that dereferences `dev.info` for specific types of interfaces.

```python
dev = NetworkDev(...)
dev.info = dev.netdev_info()  #  <-- This will crash for some types of interfaces (e.g., "eth", "wlan")
```

`dev.info` can be assigned by the `NetworkModel.new_link` method, but I don't know where the `NetDevInfo` instances comes from..

dbungert (approved):

mwhudson (approved):
Yes, self.info is None for devices that exist in the UI but not yet on the system (or at least that's the idea -- like I said this is very old code!)

PR #1787, opened 2023-09-05
storage: fix crash when guided and a small disk

dbungert:
In LP: #2034270, a 1MiB disk is present.  This is triggering a crash while attempting to decide if we can do a guided install.


mwhudson (approved):
Hm I wonder if we should filter disks like this at some higher level? But well, this is better than what went before :-)

PR #1790, opened 2023-09-07
Ensure the security archive is set to the proper URL

ogayot:
A regression was introduced as part of mirror testing.

On new Ubuntu installations, the -security pocket is accessed through the primary mirror instead of being accessed through security.ubuntu.com/ubuntu (or ports.ubuntu.com/ubuntu-ports)

This happens because when no `security` section is provided to curtin, the value configured in the `primary` section is reused.

We now ensure that the right URI is used for the security repository.

It has a slight impact on the behavior of mirror testing though. Instead of downloading content from the primary repository only, we would download content from the primary+security repository. Therefore, we would try all the candidate mirrors without success is the security repository is unreachable. To work around this issue, we now disable the -security suite when running mirror testing. This requires a fix in curtin for releases that use deb822 sources: https://code.launchpad.net/~ogayot/curtin/+git/curtin/+merge/450893

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -54,6 +54,18 @@ validate () {
                 ;;
             *)
                 python3 scripts/validate-autoinstall-user-data.py < $tmpdir/var/log/installer/autoinstall-user-data
+                # After the lunar release and the introduction of mirror testing, it
+                # came to our attention that new Ubuntu installations have the security
+                # repository configured with the primary mirror URL (i.e.,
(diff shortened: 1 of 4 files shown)


dbungert (commented):
Functionally this sounds fine but I really want test automation here, to attempt to avoid a repeat.    Integration tests can examine var/log/installer/subiquity-curtin-apt.conf.    Let's think about the unit test angle as well.  I appreciate it.

dbungert (approved):
Good for merge when the curtin part is ready.

PR #1792, opened 2023-09-08
Fix snap build, and CI enforce it

dbungert:
The snap build is broken at the moment.
Fix the build, and add CI showing the build working.

Samples of current build failure:

https://github.com/dbungert/subiquity/actions/runs/6126545979/job/16630745991#step:3:89
https://launchpadlibrarian.net/686174599/buildlog_snap_ubuntu_jammy_amd64_subiquity_BUILDING.txt.gz

### DIFF
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -1,0 +1,10 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  snap-build:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+    - uses: snapcore/action-build@v1
--- a/subiquitycore/models/network.py
+++ b/subiquitycore/models/network.py
@@ -21,7 +21,6 @@ from socket import AF_INET, AF_INET6
 from typing import Dict, List, Optional
 
 import attr
-import probert.network
 import yaml
 
 from subiquitycore import netplan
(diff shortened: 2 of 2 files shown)


.github/workflows/snap.yaml
```diff
@@ -0,0 +1,10 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  pack:
```
mwhudson, line 6: Can you change the name here to "build-snap" or something?

mwhudson (approved):

subiquitycore/models/network.py
```diff
@@ -196,10 +195,15 @@ def __init__(self, model, name, typ):
         self._name = name
         self.type = typ
         self.config = {}
+
+        # import done here to break a chain where anybody importing
```
ogayot, line 199: Oops! Thanks for fixing my mistake.

BTW, I learned about this syntax, not sure it would be suitable here?

```python
if typing.TYPE_CHECKING:
    import expensive_mod
```

ogayot (approved):
Thanks!

subiquitycore/models/network.py
dbungert: It was a bit of a surprise that this change broke the build.  Having the snap build logs in CI should help - the github action build log for this failure is able to show a very short log but point right to the backtrace.

We could do the TYPE_CHECKING check, I'll let you play with it after I land the snap build fix.  I guess if we're not TYPE_CHECKING we go back to `self.info = None`.

PR #1793, opened 2023-09-08
workflows: to checkout@v4

dbungert:
Remove remaining Checkout v2 usage to address the nodejs12 deprecation warnings


mwhudson (approved):

ogayot (approved):
I think you could have updated static-typing as well - not sure anybody will notice it if it causes a regression though :upside_down_face:

PR #1795, opened 2023-09-14
codecs: attempt install on first boot if install fails

ogayot:
Instead of giving up installing ubuntu-drivers-addons if we are not online (or if another error occurs), we now fall back to asking cloud-init to install the package on first boot.

If we were doing an offline install, it is highly likely that cloud-init will also fail to install the package on first boot. Furthermore, it will only try to install it once before disabling itself.

So, this change would only be beneficial in scenarios where a temporary network failure is encountered.

### DIFF
--- a/subiquity/common/pkg.py
+++ b/subiquity/common/pkg.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Canonical, Ltd.
+# Copyright 2023 Canonical, Ltd.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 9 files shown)


dbungert (commented):
This looks good.  This is marked draft - what needs to happen next to land this?

ogayot:
> This looks good. This is marked draft - what needs to happen next to land this?

Sadly, I have yet to witness a scenario where it really helps in practice - so I don't think we should merge for mantic. Maybe we should apply this to other packages as well - and somehow have a way to inform the users that said packages are going to be installed on first boot. Errors on first boot look pretty bad too ..

PR #1796, opened 2023-09-15
include api-only data when sending storage objects from client to server

mwhudson:
Making an install that used an existing RAID failed because of an attempt to log the size of the RAID when rendering the curtin config.

This turns out to be because when the client sends the storage objects back to the server it loses all the "api only" data including the udev data that is needed to display the size.

In some sense this is a bit silly, we could just drop the log statement and it would be fine but I think it's probably better to always have the full storage objects in the server (until we can get away from this hackish API anyway).

### DIFF
--- a/subiquity/client/controllers/filesystem.py
+++ b/subiquity/client/controllers/filesystem.py
@@ -28,7 +28,12 @@ from subiquity.common.types import (
     ProbeStatus,
     StorageResponseV2,
 )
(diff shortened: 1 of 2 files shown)


mwhudson:
It would be good to have an integration test for this i guess. Not sure I'm going to get around to writing one.

dbungert (approved):

PR #1799, opened 2023-09-19
s390x: ensure chreipl is called before unmounting /target

ogayot:
For ZFS, we recently introduced a call to `umount --recursive /target` slighly before shutting down or rebooting. Unfortunately, on s390x, we also had a very late call to `chreipl` to make the firmware boot from the installed system.

The call to `chreipl` reads data from `/target/boot`, and it fails if the filesystem is no longer mounted.

Fixed by calling `chreipl` earlier in the installation, during the postinst phase rather than after the user clicks "reboot".

[LP: #2029479](https://bugs.launchpad.net/subiquity/+bug/2029479)

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -19,6 +19,7 @@ import glob
 import json
 import logging
 import os
+import platform
 import re
 import shlex
 import shutil
@@ -666,6 +667,17 @@ class InstallController(SubiquityController):
             )
             raise
 
+    async def platform_postinstall(self):
+        """Run architecture specific commands/quirks"""
+        if platform.machine() == "s390x":
+            try:
+                # Ensure we boot from the installed system.
+                await arun_command(["chreipl", "/target/boot"])
+            except subprocess.CalledProcessError as cpe:
+                if cpe.stderr is not None:
+                    log.warning("chreipl stderr:\n%s", cpe.stderr)
+                raise
+
     @with_context(
         description="final system configuration", level="INFO", childlevel="DEBUG"
     )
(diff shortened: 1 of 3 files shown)


dbungert (approved):
Very nice work.

ogayot:
Tested on z/VM, After reboot, the system booted the newly installed system.

Furthermore, the logs on the target system show the successful call to `chreipl`.
```
2023-09-19 15:35:51,905 DEBUG subiquitycore.utils:118 arun_command called: ['chreipl', '/target/boot']
2023-09-19 15:35:51,907 DEBUG subiquitycore.utils:132 arun_command ['chreipl', '/target/boot'] exited with code 0
2023-09-19 15:35:51,907 INFO root:30 finish: subiquity/Install/install/postinstall: SUCCESS: final system configuration
2023-09-19 15:35:51,908 DEBUG root:30 finish: subiquity/Install/install: SUCCESS: 
```

subiquity/server/controllers/install.py
```diff
@@ -666,6 +667,17 @@ async def install(self, *, context):
             )
             raise
 
+    async def platform_postinstall(self):
+        """Run architecture specific commands/quirks"""
```
mwhudson, line 671: genuine question do we want to use a context here to inform the user this is happening?

mwhudson (approved):

dbungert:
Merging - Olivier, please consider the context change for a follow-on.

subiquity/server/controllers/install.py
ogayot: I think that would be a good idea. At least for platforms that don't just noop.

PR #1802, opened 2023-09-21
ui: have a distinct state file for rich mode over serial

ogayot:
We recently made sure that after doing a snap refresh, the rich mode (i.e., either rich or basic) is preserved. This was implemented by storing the rich mode in a state file. When the client starts, it loads the rich mode from said state file if it exists.

Unfortunately, on s390x, it causes installs to default to basic mode. This happens because on this architecture, a subiquity install consists of:

 * a first client (over serial) showing the SSH password
 * a second client (logging over SSH) actually going through the installation UI.

Since the first client uses a serial connection, the state file is created with rich-mode set to basic. Upon connecting using SSH, the state file is read and the rich-mode is set to basic as well.

Fixed by storing the rich-mode in two separate files, one for clients over serial and one for other clients.

[LP: #2036096](https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/2036096)

### DIFF
--- a/subiquitycore/tests/test_tui.py
+++ b/subiquitycore/tests/test_tui.py
@@ -1,0 +1,76 @@
+# Copyright 2023 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import json
+import pathlib
+from unittest.mock import Mock, patch
+
+from subiquitycore.tests import SubiTestCase
+from subiquitycore.tui import TuiApplication
+
+
+class TestTuiApplication(SubiTestCase):
+    def setUp(self):
+        with patch("subiquitycore.tui.Application.__init__", return_value=None):
+            opts = Mock()
+            opts.answers = None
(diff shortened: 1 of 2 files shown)


ogayot:
> Oh grunk. Would it be better to append ttyname or something to the state filename?

Oh that could be an option as well. I just renamed I just renamed the `rich-mode` file to `rich-mode-tty` and added logic to handle the former file-name (so that we don't lose the state on snap refresh).

But let me know if you think we should do `rich-mode-tty-dev-pts-2` or something. I can do that.

subiquitycore/tests/test_tui.py
```diff
@@ -0,0 +1,79 @@
...
+class TestTuiApplication(unittest.TestCase):
+    def setUp(self):
+        with patch("subiquitycore.tui.Application.__init__", return_value=None):
+            opts = Mock()
+            opts.answers = None
+            opts.dry_run = True
+            self.tui = TuiApplication(opts)
+            # Usually, the below are assigned by Application.__init__()
+            self.tui.opts = opts
+            tmpdir = tempfile.TemporaryDirectory()
```
dbungert, line 34: SubiTestCase has a tmp_dir() function which would help simplify this a bit.

dbungert (approved):

subiquitycore/tests/test_tui.py
ogayot: Thanks! Yes that looks simpler.

subiquitycore/tui.py
```diff
@@ -286,6 +290,37 @@ def extra_urwid_loop_args(self):
...
+        else:
+            with fp:
+                return json.load(fp)
+
+        try:
+            # During the 23.10 development cycle, there was only one rich-mode
+            # state file. Let's handle the scenario where we just snap
+            # refresh-ed from a pre 23.10 release.
+            # Once mantic is EOL, let's remove this code.
+            fp = open(self.state_path("rich-mode"))
```
mwhudson, line 314: honestly my opinion is that we don't need to bother with this. but you've done the work now so fine!

mwhudson (approved):

PR #1803, opened 2023-09-21
Allow for specifying the autoinstall file over the kernel command line

Chris-Peterson444:
This change allows for specifying the location of the autoinstall.yaml file in the kernel command line via `subiquity.autoinstallpath`. So, something like `subiquity.autoinstallpath=path/to/autoinstall.yaml`.

I have additionally changed the precedence for the various ways the autoinstall file may be passed from:
```
1. root of the drive
2. subiquity cmdline
3. cloud config
4. root of cdrom
```

To:
```
1. subiquity cmdline
2. kernel cmdline
3. root of the drive
4. cloud config
5. root of the cdrom
```

For the tests:
1. I've added a test for the kernel command line case.
1. Reordered them to align with precedence changes
1. Used explicit "missing" values where possible (e.g., `None` for the subiquity arg when the case is no arg is passed), instead of simply leaving them undefined.

### DIFF
--- a/doc/intro-to-autoinstall.rst
+++ b/doc/intro-to-autoinstall.rst
@@ -78,24 +78,47 @@ Autoinstall on the install media
 Another option for supplying autoinstall to the Ubuntu installer is to place a
 file named :code:`autoinstall.yaml` on the install media itself.
 
-There are two potential locations for the :code:`autoinstall.yaml` file:
- * At the root of the "CD-ROM". When you write the installation ISO to a USB
-   Flash Drive, this can be done by copying the :code:`autoinstall.yaml` to the
-   partition containing the contents of the ISO - i.e.,
-   in the directory containing the ``casper`` sub-directory.
- * On the rootfs of the installation system - this option will typically
-   require modifying the installation ISO and is not suggested, but is
-   supported.
+There are two potential locations that subiquity will check for the
+:code:`autoinstall.yaml` file:
 
-Directly specifying autoinstall as a :code:`autoinstall.yaml` file does not
-require a :code:`#cloud-config` header, and does not use a top level
-``autoinstall:`` key. The autoinstall directives are placed at the top
-level. For example:
+* At the root of the "CD-ROM". When you write the installation ISO to a USB
+  Flash Drive, this can be done by copying the :code:`autoinstall.yaml` to the
+  partition containing the contents of the ISO - i.e.,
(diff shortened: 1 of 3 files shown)


subiquity/server/server.py
```diff
@@ -574,10 +574,11 @@ async def wait_for_cloudinit(self):
 
     def select_autoinstall(self):
         # precedence
-        # 1. autoinstall at root of drive
-        # 2. command line argument autoinstall
-        # 3. autoinstall supplied by cloud config
-        # 4. autoinstall baked into the iso, found at /cdrom/autoinstall.yaml
+        # 1. command line argument autoinstall
```
dbungert, line 577: Please update doc/intro-to-autoinstall.rst with a new section talking about this order

dbungert (commented):
Doc update please, and please rebase the commits down to one or two.  Doc update could arguably be a distinct commit.  We don't use the squash merge feature so the commits should be long-term interesting.

Take a peek at recent commit history for the convention being used, for instance "initial fix" is more generic than I would like.

Thanks!

Chris-Peterson444:
Docs updated and commits rebased. Let me know if you would like the section in the doc to be elaborated further.

doc/intro-to-autoinstall.rst
```diff
@@ -78,24 +78,52 @@ Autoinstall on the install media
...
+Order precedence of the autoinstall locations
+======================================
+
+Since there are many ways to specify the autoinstall file, it may happen that
+multiple locations are specified at once. Subiquity will look for the
+autoinstall file in the following order and pick the first existing one:
 
-    version: 1
-    ....
+1. Subiquity command line
```
dbungert, line 122: While this is correct, I feel reluctant to document it here.  Noting this here raises the question "and how do I pass a custom subiquity command line?" which we have no plan for.  I'd rather drop the reference to this "Subiquity command line" item entirely.

dbungert (approved):
Thanks.  One doc tweak and this is done.

doc/intro-to-autoinstall.rst
Chris-Peterson444: Fair enough. I will remove this line and the reference above.

dbungert (approved):

subiquity/server/server.py
ogayot, line 577: Thanks for move CLI argument up the chain. It feels more natural.

PR #1804, opened 2023-09-21
snapd api: wait longer

dbungert:
While these changes are not supposed to take nearly this long, per LP: #[2034715](https://bugs.launchpad.net/snapd/+bug/2034715) we know that they are, and that some systems will correctly perform the finish_install() step if just given more time.

### DIFF
--- a/subiquitycore/snapd.py
+++ b/subiquitycore/snapd.py
@@ -33,6 +33,11 @@ log = logging.getLogger("subiquitycore.snapd")
 
 
 class SnapdConnection:
+    # In LP: #2034715, we found that while some requests should be
+    # non-blocking, they are actually blocking and exceeding one minute.
+    # Extending the timeout helps.
+    default_timeout_seconds = 600
+
     def __init__(self, root, sock):
         self.root = root
         self.url_base = "http+unix://{}/".format(quote_plus(sock))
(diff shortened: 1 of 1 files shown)


Chris-Peterson444 (approved):
This looks straightforward to me. We may want to reconsider reverting this once the issue in snapd is resolved.

subiquitycore/snapd.py
```diff
@@ -41,13 +41,15 @@ def get(self, path, **args):
         if args:
             path += "?" + urlencode(args)
         with requests_unixsocket.Session() as session:
-            return session.get(self.url_base + path, timeout=60)
+            return session.get(self.url_base + path, timeout=600)
```
ogayot, line 44: Please consider adding a comment linking to the LP bug.

ogayot (approved):
LGTM

subiquitycore/snapd.py
ogayot: I like that it's present in the commit message though, so feel free to disregard if you think it's enoug.

mwhudson: Yeah I'd like a comment near the code too please

mwhudson (approved):

PR #1806, opened 2023-09-25
filesystem: revamp udev handling

dbungert:
In LP: #2009141, we are hitting kernel limits and pyudev buffer limits.
We don't care about specific events, so much as getting one event,
waiting for things to calm down, then reprobing.

Outright disable the event monitor, and re-enable later.  If there is a
storm of events, testing has shown that stopping the listener is not
enough.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -20,7 +20,6 @@ import json
 import logging
 import os
 import pathlib
-import select
 import time
 from typing import Any, Callable, Dict, List, Optional, Union
 
@@ -273,6 +272,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         self._system_mounter: Optional[Mounter] = None
         self._role_to_device: Dict[str, _Device] = {}
         self._device_to_structure: Dict[_Device, snapdapi.OnVolume] = {}
+        self._pyudev_context: Optional[pyudev.Context] = None
         self.use_tpm: bool = False
         self.locked_probe_data: bool = False
         # If probe data come in while we are doing partitioning, store it in
@@ -299,7 +299,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         ):
             self.app.base_model.source.search_drivers = not self.is_core_boot_classic()
         await super().configured()
-        self.stop_listening_udev()
+        self.stop_monitor()
 
     async def _mount_systems_dir(self, variation_name):
         self._source_handler = self.app.controllers.Source.get_handler(variation_name)
@@ -1293,6 +1293,13 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             finally:
                 elapsed = time.time() - start
                 log.debug(f"{short_label} probing took {elapsed:.1f} seconds")
+                # In the past, this start_monitor() equivalent was much sooner.
+                # We don't actually need the information it provides though
+                # until a probe has finished, so the start_monitor() is delayed
+                # to here. start_monitor() is allowed after a failed probe, in
+                # case of a hotplug event, perhaps to remove a problematic
+                # device.
+                self.start_monitor()
             break
 
     async def run_autoinstall_guided(self, layout):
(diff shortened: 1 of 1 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -1495,7 +1488,20 @@ def _udev_event(self):
         while select.select([self._monitor.fileno()], [], [], 0)[0]:
             action, dev = self._monitor.receive_device()
             log.debug("_udev_event %s %s", action, dev)
-        self.ensure_probing()
```
dbungert, line 1498: well we probably want that bit somewhere

subiquity/server/controllers/filesystem.py
```diff
@@ -1293,6 +1292,7 @@ async def _probe(self, *, context=None):
             finally:
                 elapsed = time.time() - start
                 log.debug(f"{short_label} probing took {elapsed:.1f} seconds")
+            self.start_monitor()
```
ogayot, line 1295: If there is a probing operation running when storage is marked configured, won't this call restart the monitoring despite the call to `self.stop_monitor()` in `configured()` ?

ogayot: Also, this call will not be executed in case in exception is raised, it this intended?

subiquity/server/controllers/filesystem.py
```diff
@@ -1456,21 +1456,28 @@ def start(self):
         self._start_task = schedule_task(self._start())
 
     async def _start(self):
-        context = pyudev.Context()
-        self._monitor = pyudev.Monitor.from_netlink(context)
-        self._monitor.filter_by(subsystem="block")
-        self._monitor.enable_receiving()
-        self.start_listening_udev()
+        self.pyudev_context = pyudev.Context()
```
ogayot, line 1459: Could you declare this in the initializer as an `Optional[pyudev.Context]` ?

subiquity/server/controllers/filesystem.py
ogayot: Maybe a `add_done_callback(self.start_monitor)` would be more suitable. And that would leave the option to call `remove_done_callback(self.start_monitor)` in the `configured` method.

subiquity/server/controllers/filesystem.py
```diff
@@ -1456,21 +1456,28 @@ def start(self):
...
-    def stop_listening_udev(self):
+    def stop_monitor(self):
+        if self._monitor is None:
+            return
+
+        log.debug("stop_monitor")
         loop = asyncio.get_running_loop()
         loop.remove_reader(self._monitor.fileno())
 
+        del self._monitor
```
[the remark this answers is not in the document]
dbungert: Dropped.  I guess if we need predictable behavior there then it's an upstream change.  I considered this more nice to have, if it does some cleanup.

subiquity/server/controllers/filesystem.py
dbungert: It's intentional to not monitor on the exception case, because today we have no plan forward after the probing failure due to timeout.  Perhaps if probing timeout failure were recoverable.

Also the restricted probe is going to go away, so either our probe is passing and everything is great or we have no interest in listening for more events.

subiquity/server/controllers/filesystem.py

subiquity/server/controllers/filesystem.py
dbungert: > add_done_callback

An interesting idea, but this would be called on timeout, which I don't think we want.

ogayot: > It's intentional to not monitor on the exception case, because today we have no plan forward after the probing failure due to timeout. Perhaps if probing timeout failure were recoverable.
> 
> Also the restricted probe is going to go away, so either our probe is passing and everything is great or we have no interest in listening for more events.

So you're suggesting that we don't do retries in case of probing failure? I'm not 100% convinced that we should consider retries useless. If you hit a timeout and then eject a device (that has multiple partitions?) for instance, there is a chance that you will get a successful probing the second time. But if you think the change does more good than bad as is, then it's probably acceptable.

> If there is a probing operation running when storage is marked configured, won't this call restart the monitoring despite the call to self.stop_monitor() in configured() ?

I still think this is something we want to fix though. Thoughts ?

ogayot (approved):
LGTM, thanks for the changes!

subiquity/server/controllers/filesystem.py
dbungert: > If you hit a timeout and then eject a device (that has multiple partitions?) for instance, there is a chance that you will get a successful probing the second time.

This sounds plausible.  I'm moving start_monitor() to the `finally` block.  Nice suggestion.

PR #1808, opened 2023-09-26
filesystem: don't crash if v2/orig_config is called early

ogayot:
When v2/orig_config is called too early, the load_probe_data function will fail because probe_data is None:

```python
    File "subiquity/server/controllers/filesystem.py", line 1029, in v2_orig_config_GET
      model = self.model.get_orig_model()
    File "subiquity/models/filesystem.py", line 1428, in get_orig_model
      orig_model.load_probe_data(self._probe_data)
    File "subiquity/models/filesystem.py", line 1894, in load_probe_data
      for devname, devdata in probe_data["blockdev"].items():
  TypeError: 'NoneType' object is not subscriptable
```

Make sure we don't dereference model._probe_data if it is None.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1435,7 +1435,8 @@ class FilesystemModel:
         # use on the V2 storage API.
         orig_model = FilesystemModel(self.bootloader, root=self.root)
         orig_model.target = self.target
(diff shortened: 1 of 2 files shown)


dbungert (approved):
Thanks!

PR #1811, opened 2023-09-28
autoinstall: Don't use snap env when invoking early and late commands

Chris-Peterson444:
LP [#2032961](https://pad.lv/2032961): We are not cleaning the environment provided to subprocesses correctly. When trying to invoke a program that should run outside of the snap context, the `LD_LIBRARY_PATH` variable is pointing to the linker on the snap, which causes errors like:
```
curl: /lib/x86_64-linux-gnu/libc.so/6: version `GLIBC_2.33` not found (required by /snap/subiquity/x1/usr/lib/x86_64-linux-gnu/libpsl.so.5)
```

We had a similar issue solved in #1300. This fix similarly uses `orig_environ` to clean the environment before running commands in the respective command controllers. 

I would especially appreciate some input if the license header on the new test file looks correct. It's an exact copy and paste from one of the other new tests files from this year, but I just want to be sure.

### DIFF
--- a/subiquity/server/controllers/cmdlist.py
+++ b/subiquity/server/controllers/cmdlist.py
@@ -16,6 +16,7 @@
 import asyncio
 import os
 import shlex
+import shutil
 from typing import List, Sequence, Union
 
 import attr
@@ -25,7 +26,7 @@ from subiquity.common.types import ApplicationState
 from subiquity.server.controller import NonInteractiveController
 from subiquitycore.async_helpers import run_bg_task
 from subiquitycore.context import with_context
-from subiquitycore.utils import arun_command
+from subiquitycore.utils import arun_command, orig_environ
 
 
 @attr.s(auto_attribs=True)
@@ -78,6 +79,15 @@ class CmdListController(NonInteractiveController):
         env = self.env()
         for i, cmd in enumerate(tuple(self.builtin_cmds) + tuple(self.cmds)):
             desc = cmd.desc()
+
+            # If the path to the command isn't found on the snap we should
+            # drop the snap specific environment variables.
+            command = shlex.split(desc)[0]
+            path = shutil.which(command)
+            if path is not None:
+                if not path.startswith("/snap"):
+                    env = orig_environ(env)
+
             with context.child("command_{}".format(i), desc):
                 args = cmd.as_args_list()
                 if self.syslog_id:
--- a/subiquity/server/controllers/tests/test_cmdlist.py
+++ b/subiquity/server/controllers/tests/test_cmdlist.py
@@ -1,0 +1,84 @@
+# Copyright 2023 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from unittest import IsolatedAsyncioTestCase, mock
+
+from subiquity.server.controllers import cmdlist
+from subiquity.server.controllers.cmdlist import (
+    CmdListController,
+    Command,
+    EarlyController,
+    LateController,
+)
+from subiquitycore.tests.mocks import make_app
+from subiquitycore.utils import orig_environ
+
+
+@mock.patch.object(cmdlist, "orig_environ", side_effect=orig_environ)
+@mock.patch.object(cmdlist, "arun_command")
+class TestCmdListController(IsolatedAsyncioTestCase):
+    controller_type = CmdListController
+
+    def setUp(self):
+        self.controller = self.controller_type(make_app())
+        self.controller.cmds = [Command(args="some-command", check=False)]
+        snap_env = {
+            "LD_LIBRARY_PATH": "/var/lib/snapd/lib/gl",
+        }
+        self.mocked_os_environ = mock.patch.dict("os.environ", snap_env)
+
+    @mock.patch("shutil.which", return_value="/usr/bin/path/to/bin")
+    async def test_no_snap_env_on_call(
+        self,
+        mocked_shutil,
+        mocked_arun,
+        mocked_orig_environ,
+    ):
+        with self.mocked_os_environ:
+            await self.controller.run()
+            args, kwargs = mocked_arun.call_args
+            call_env = kwargs["env"]
+
+            mocked_orig_environ.assert_called()
+            self.assertNotIn("LD_LIBRARY_PATH", call_env)
+
+    @mock.patch("shutil.which", return_value="/snap/path/to/bin")
+    async def test_with_snap_env_on_call(
+        self,
+        mocked_shutil,
+        mocked_arun,
+        mocked_orig_environ,
+    ):
+        with self.mocked_os_environ:
+            await self.controller.run()
+            args, kwargs = mocked_arun.call_args
+            call_env = kwargs["env"]
+
+            mocked_orig_environ.assert_not_called()
+            self.assertIn("LD_LIBRARY_PATH", call_env)
+
+
+class TestEarlyController(TestCmdListController):
+    controller_type = EarlyController
+
+    def setUp(self):
+        super().setUp()
+
+
+class TestLateController(TestCmdListController):
+    controller_type = LateController
+
+    def setUp(self):
+        super().setUp()
--- a/subiquitycore/tests/mocks.py
+++ b/subiquitycore/tests/mocks.py
@@ -43,4 +43,9 @@ def make_app(model=None):
     app.opts = mock.Mock()
     app.opts.dry_run = True
     app.scale_factor = 1000
+    app.echo_syslog_id = None
+    app.log_syslog_id = None
+    app.report_start_event = mock.Mock()
+    app.report_finish_event = mock.Mock()
+
     return app
--- a/subiquitycore/utils.py
+++ b/subiquitycore/utils.py
@@ -19,7 +19,7 @@ import logging
 import os
 import random
 import subprocess
-from typing import Any, Dict, List, Sequence
+from typing import Any, Dict, List, Optional, Sequence
 
 log = logging.getLogger("subiquitycore.utils")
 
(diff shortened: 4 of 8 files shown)


subiquity/server/controllers/cmdlist.py
```diff
@@ -124,7 +127,7 @@ def __init__(self, app):
                 ]
 
     def env(self):
-        env = super().env()
+        env = orig_environ(None)
```
ogayot, line 130: Our [docs](https://github.com/canonical/subiquity/blob/9bf0a50a24b08ba7f80277212acf10a380ef03b0/documentation/autoinstall-reference.md#late-commands) indicate:

> You can run curtin in-target -- $shell_command [...] to run in the target system

and I think sadly your change will break the ability to do that. Maybe we need to update the docs, or add support for specifying if we want a clean env or not, as in:

```yaml
late-commands:
    # specifies a command directly (runs with a clean env)
    - echo "hello world"
    # Run `cat /etc/machine-id` in the target system
    - cmd: curtin in-target -- cat /etc/machine-id
      clean-env: false  # defaults to true? (true feels more natural but false might be better for backward compatibility)
``` 

or maybe we add the ability to run a command in the target system:

```yaml
late-commands:
    # specifies a command directly (runs in the installer environment)
    - echo "hello world"
    # Run `cat /etc/machine-id` in the target system
    - cmd: cat /etc/machine-id
      in-target: true    # defaults to false
```

in any case, I think we need to think about what it means in terms of backward compatibility

subiquity/server/controllers/tests/test_cmdlist.py
```diff
@@ -0,0 +1,69 @@
...
+
+class TestEarlyController(TestCmdListController):
+    def setUp(self):
+        super().setUp()
+        self.controller = EarlyController(make_app())
+        self.controller.cmds = [Command(args="some-command", check=False)]
+
+    async def test_no_library_path_on_run(self):
+        with (
+            self.mocked_arun as faked_process_call,
```
ogayot, line 44: python3.8 is being annoying (hopefully we will get rid of it soon !) but you can work around the issue by dropping the parentheses :+1:

subiquitycore/tests/mocks.py
```diff
@@ -43,4 +43,9 @@ def make_app(model=None):
     app.opts = mock.Mock()
     app.opts.dry_run = True
     app.scale_factor = 1000
+    app.echo_syslog_id = None
+    app.log_syslog_id = None
+    app.report_start_event = mock.Mock()
+    app.report_finish_event = mock.Mock()
```
ogayot, line 49: minor:

I think it's a useful change, thanks! Would you remove those now obsolete instructions (ideally in a separate commit along)?
```
subiquity/server/controllers/tests/test_filesystem.py:        self.app.report_start_event = mock.Mock()
subiquity/server/controllers/tests/test_filesystem.py:        self.app.report_start_event = mock.Mock()
subiquity/server/controllers/tests/test_install.py:        self.controller.app.report_start_event = Mock()
subiquity/server/controllers/tests/test_install.py:        self.controller.app.report_start_event = Mock()
subiquity/server/controllers/tests/test_refresh.py:        self.app.report_start_event = mock.Mock()
subiquity/server/controllers/tests/test_snaplist.py:        self.app.report_start_event = Mock()
```

ogayot (changes requested):
LGTM from the code standpoint, thanks!

I am concerned about breaking the ability to run  `curtin in-target [...]` in late-commands though. I think we need to think about what it means for autoinstall users who are going to take a new subiquity snap.

subiquity/server/controllers/cmdlist.py
ogayot: Another option (which is not my favorite!) would be to check if the command starts with "curtin":

```python
if shlex.split(cmd.desc())[0] == "curtin":
    clean_env = False
```

dbungert: I do think `shlex.split()` is where this needs to go.  Maybe we also pair that with `shutil.which()` and check if that resolves to something in the snap.

Chris-Peterson444: Thanks for pointing this out. For the general case, I like the idea of inspecting the path from `shutil.which()` and determining if we should drop the snap environment variables or not.

Eventually adding support in the autoinstall config to specify the command environment seems like a good call, although perhaps we could encourage the use of `env` until we decide if want to and how that would look.

subiquitycore/tests/mocks.py
Chris-Peterson444: Done! I went ahead and double checked any other instances of these instructions and it looks these were it.

subiquitycore/utils.py
```diff
@@ -35,21 +35,24 @@ def _clean_env(env, *, locale=True):
...
         if key.endswith("_ORIG"):
-            key_to_restore = key[: -len("_ORIG")]
+            key_to_restore: str = key[: -len("_ORIG")]
             if val:
                 ret[key_to_restore] = val
             else:
                 del ret[key_to_restore]
             del ret[key]
-    ret.pop("LD_LIBRARY_PATH", None)
+    _ = ret.pop("LD_LIBRARY_PATH", None)
```
dbungert, line 55: The convention we're using in Subiquity is to not assign to underscore, as it's used for translation strings.  Do we need to assign to something?

subiquity/server/controllers/cmdlist.py
```diff
@@ -78,6 +79,16 @@ async def run(self, context):
...
         for i, cmd in enumerate(tuple(self.builtin_cmds) + tuple(self.cmds)):
             desc = cmd.desc()
+
+            # If the path to the command isn't found on the snap we should
+            # drop the snap specific environment variables. If the command
+            # is `env`, we should let the user manage the environment
+            command = shlex.split(desc)[0]
+            path = shutil.which(command)
+            if path is not None:
+                if not path.startswith("/snap") and command != "env":
```
dbungert, line 89: If we're special casing `env`, someone might have specified the full path, which means all of `("env", "/bin/env", "/usr/bin/env")` are possible.

subiquity/server/controllers/tests/test_cmdlist.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @mock.patch("shutil.which", return_value="/usr/bin/path/to/bin")
+    async def test_no_snap_env_on_call(self, mocked_shutil):
+        with self.mocked_arun as faked_process_call, self.mocked_os_environ, self.mocked_orig_environ as mocked_orig_environ:
+            await self.controller.run()
+            _, kwargs = faked_process_call.call_args
+            call_env = kwargs["env"]
+
+            mocked_orig_environ.assert_called()
+            assert "LD_LIBRARY_PATH" not in call_env
```
dbungert, line 53: We're using the asserts that are available as test class members, generally.

```suggestion
            self.assertNotIn("LD_LIBRARY_PATH", call_env)
```

subiquitycore/utils.py
Chris-Peterson444: Nope, this was just a personal preference when I was making edits.If that's convention I'll remove it.

subiquity/server/controllers/cmdlist.py
Chris-Peterson444: Hmmm, that's a good point. My thought was if someone passed their command via `env` they probably don't want us stripping everything for them (unless of course they're passing `env -i`). Should we special case on all three of these options, or just not special case at all?

dbungert: /usr/bin/env does link to libc6, so the better answer is orig_env, I think.
for `env -i`  I think it doesn't matter what is done here.

> if someone passed their command via env they probably don't want us stripping everything for them

I would also prefer to not break someone's existing autoinstall, but if they are setting environment variables today I think the common scenarios are:
* they have a special case that is not covered by the above
* they already ran into this problem and are working around it, and giving a clean environment should still be ok

So I think the answer is to not special case env.  If you can think of a corner case not covered in this plan please point it out.

subiquity/server/controllers/tests/test_cmdlist.py
```diff
@@ -0,0 +1,81 @@
...
+        }
+        self.mocked_os_environ = mock.patch.dict("os.environ", snap_env_vars)
+
+    @mock.patch("shutil.which", return_value="/usr/bin/path/to/bin")
+    async def test_no_snap_env_on_call(self, mocked_shutil):
+        with self.mocked_arun as faked_process_call:
+            with self.mocked_orig_environ as mocked_orig_environ:
+                with self.mocked_os_environ:
+                    await self.controller.run()
+                    _, kwargs = faked_process_call.call_args
```
dbungert, line 51: more underscores

subiquity/server/controllers/tests/test_cmdlist.py
```diff
@@ -0,0 +1,81 @@
...
+                    mocked_orig_environ.assert_called()
+                    self.assertNotIn("LD_LIBRARY_PATH", call_env)
+
+    @mock.patch("shutil.which", return_value="/snap/path/to/bin")
+    async def test_with_snap_env_on_call(self, mocked_shutil):
+        with self.mocked_arun as faked_process_call:
+            with self.mocked_orig_environ as mocked_orig_environ:
+                with self.mocked_os_environ:
+                    await self.controller.run()
+                    _, kwargs = faked_process_call.call_args
```
dbungert, line 63: same

subiquity/server/controllers/cmdlist.py
Chris-Peterson444: I can't think of any other cases. That covers my concerns, so I'll remove the special casing. Thanks!

subiquity/server/controllers/tests/test_cmdlist.py
Chris-Peterson444: Thanks for the catch, fixed.

subiquity/server/controllers/tests/test_cmdlist.py
Chris-Peterson444: Also fixed.

subiquity/server/controllers/tests/test_cmdlist.py
```diff
@@ -0,0 +1,79 @@
...
+    CmdListController,
+    Command,
+    EarlyController,
+    LateController,
+)
+from subiquitycore.tests.mocks import make_app
+from subiquitycore.utils import orig_environ
+
+
+@mock.patch.dict("os.environ", {"LD_LIBRARY_PATH": "/var/lib/snapd/lib/gl"})
```
dbungert, line 29: I played with this a bit, is this just bugged on python3.8?

dbungert (approved):

Chris-Peterson444:
I tried `curtin in-target -- touch /etc/test-file-here` and I can verify it shows up under `/target/etc/test-file-here`

subiquity/server/controllers/tests/test_cmdlist.py
dbungert: I think this line is obsolete now?

PR #1813, opened 2023-10-02
shutdown: do not try to unmount /target if install was not started

ogayot:
If we ask for reboot before the installation has started (i.e., if curtin install was not invoked at least once), the following call fails and prevents the system from rebooting.

 $ umount --recursive /target

Make sure we check that /target exists and is mounted before calling umount.

Another approach would be to check the return value of umount but the values are not documented.

`LP #2037539`


dbungert (approved):
Thanks!

PR #1815, opened 2023-10-02
workflows: differentiate CI and Snap workflows

ogayot:
The workflows defined respectively in build.yaml and snap.yaml were both called "CI". On the Github web interface, it resulted in two menus called "CI" with no easy way to know which is which.

To make things clearer, we now:
 * rename build.yaml -> ci.yaml
 * call "Snap" the workflow defined by snap.yaml


dbungert (approved):
Good idea.

PR #1819, opened 2023-10-05
Cloud init log

dbungert:
The cloud init log files are originally owned by syslog:adm, but that user ID is not necessarily the same on the target system.  Normalize that owning user id of the file to root.

### DIFF
--- a/subiquity/server/controllers/shutdown.py
+++ b/subiquity/server/controllers/shutdown.py
@@ -92,6 +92,7 @@ class ShutdownController(SubiquityController):
         for logfile in cloudinit_logs:
             if not os.path.exists(logfile):
                 continue
+            set_log_perms(logfile)
             await self.app.command_runner.run(
                 ["cp", "-a", logfile, "/var/log/installer"]
             )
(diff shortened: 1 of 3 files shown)


subiquity/server/controllers/shutdown.py
```diff
@@ -83,6 +83,20 @@ async def _run(self):
...
+        cloudinit_logs = (
+            "/var/log/cloud-init.log",
+            "/var/log/cloud-init-output.log",
+        )
+        for logfile in cloudinit_logs:
+            if not os.path.exists(logfile):
+                continue
+            set_log_perms(logfile)
+            await self.app.command_runner.run(
+                ["cp", "-a", logfile, "/var/log/installer"]
```
Chris-Peterson444, line 97: Why change the owner of the original as well?

Chris-Peterson444 (approved):
One non-blocking question, but looks good to me.

subiquity/server/controllers/shutdown.py
dbungert: Simplicity.  We're in the shutdown controller and we are about to shutdown, we're doing the final final steps before unmount of /target.  The scenario where the ephemeral system log files are presumed to not matter, but maybe there is an obscure case where that is still a thing.

PR #1823, opened 2023-10-05
po: merge from translations-export

dbungert:


po/ca.po
```diff
@@ -1917,9 +1917,8 @@ msgid ""
 msgstr ""
 
 #: ../subiquity/ui/views/identity.py:212
-#, fuzzy
 msgid "Passwords"
-msgstr "contrasenyes"
+msgstr ""
```
Chris-Peterson444, line 1921: Are we fine deleting translations?

liushuyu: I think the script is removing all the fuzzy translations.
Those are not used by the `gettext` during the runtime as far as I am aware.

Chris-Peterson444: Thanks @liushuyu

Chris-Peterson444 (approved):

PR #1828, opened 2023-10-09
apt: fix source.list mode

dbungert:
logfile hardening changes have resulted in a warning appearing when running c-n-f about not being able to read sources.list.  Write this file as 644.


sil2100 (approved):
644 seems like the correct permissions for sources.list.

PR #1832, opened 2023-10-10
Fix crash when switching sources having different variations

ogayot:
When the source changes, the available variations should change as well. If we keep the old variations in the FilesystemController._variations_info dictionary, we end up with a crash later in the install, showing something like:

```
2023-09-29 12:10:36,383 INFO subiquity.common.errorreport:415 saving crash report 'install failed crashed with KeyError' to /var/crash/1695989436.283925295.install_fail.crash
2023-09-29 12:10:36,383 ERROR root:30 finish: subiquity/Install/install: FAIL: 'minimal'
```

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -404,6 +404,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         return info
 
     async def _examine_systems(self):
+        self._variation_info.clear()
         catalog_entry = self.app.base_model.source.current
         for name, variation in catalog_entry.variations.items():
             system = None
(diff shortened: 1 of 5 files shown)


ogayot:
I just tested the fix in the desktop installer - and I confirm the crash does not occur anymore

subiquity/server/controllers/filesystem.py
```diff
@@ -404,6 +404,7 @@ def disallowed_encryption(msg):
         return info
 
     async def _examine_systems(self):
+        self._variation_info.clear()
```
dbungert, line 407: do we also need to reset `self._info`?

ogayot: That's a fair point. `self._info` is set through a call to `set_info_for_capability` - which only happens after a POST to /storage/guided (or /storage). I would need a build a test-case for that as well.

dbungert: OK, not blocking for this

dbungert (approved):

mwhudson:
Oops. I'm sorry all this ended up so complicated :(

PR #1838, opened 2023-10-12
Fixes invalid links reported by linkcheck in docs

rkratky:
Fixes [FR-5664](https://warthogs.atlassian.net/browse/FR-5664)

See $subj.

[FR-5664]: https://warthogs.atlassian.net/browse/FR-5664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/custom_conf.py
+++ b/doc/custom_conf.py
@@ -122,6 +122,7 @@ custom_linkcheck_anchors_ignore_for_url = [
 # Add extensions
 custom_extensions = [
     'sphinx.ext.intersphinx',
+    'sphinx.ext.extlinks'
 ]
 
 # Add files or directories that should be excluded from processing.
@@ -172,3 +173,7 @@ root_doc = 'index'
 # can be copied even if they don't contain an EOF line.
 copybutton_prompt_text = '$ '
 copybutton_only_copy_prompt_lines = False
+
+extlinks = {
+    'manualpage': ( 'https://manpages.ubuntu.com/manpages/lunar/en/%s', '' )
+}
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -144,8 +144,7 @@ installed will not have a keyboard at all in which case the value used here
 does not matter.
 
 The mapping's keys correspond to settings in the :file:`/etc/default/keyboard`
-configuration file. See
-`its manual page <http://manpages.ubuntu.com/manpages/keyboard.5.html>`_
+configuration file. See the :manualpage:`keyboard(5) manual page <man5/keyboard.5.html>`
 for more details.
 
 The mapping contains keys:
@@ -635,10 +634,9 @@ password
 The password for the new user, encrypted. This is required for use with
 ``sudo``, even if SSH access is configured.
 
-The crypted password string must conform to what
-`passwd <https://manpages.ubuntu.com/manpages/passwd.1.html>`_
-expects. Depending on the special characters in the password hash, quoting may
-be required, so it's safest to just always include the quotes around the hash.
+The crypted password string must conform to what the
+``passwd`` command requires. See the :manualpage:`passwd(1) manual page <man1/passwd.1.html>`
+for details. Quote the password hash to ensure correct treatment of any special characters.
 
 Several tools can generate the crypted password, such as ``mkpasswd`` from the
 ``whois`` package, or ``openssl passwd``.
--- a/doc/explanation/operate-server-installer.rst
+++ b/doc/explanation/operate-server-installer.rst
@@ -5,7 +5,7 @@ Operating the Server installer
 
 This document explains how to use the installer in general terms. For a
 step-by-step guide through the screens of the installer, you can use our
-`screen-by-screen reference guide <https://discourse.ubuntu.com/t/draft-using-the-server-installer-step-by-step/16690>`_.
+`screen-by-screen reference guide <https://discourse.ubuntu.com/t/screen-by-screen-installer-guide/16690>`_.
 
 Get the installer
 =================
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -83,7 +83,7 @@ constructive feedback.
    development/index
 
 .. Links:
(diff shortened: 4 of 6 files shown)


Chris-Peterson444 (approved):
So the link checker doesn't follow redirects? Good to know!

doc/reference/autoinstall-reference.rst
```diff
@@ -144,8 +144,8 @@ installed will not have a keyboard at all in which case the value used here
 does not matter.
 
 The mapping's keys correspond to settings in the :file:`/etc/default/keyboard`
-configuration file. See
-`its manual page <http://manpages.ubuntu.com/manpages/keyboard.5.html>`_
+configuration file. See the
+`keyboard(5) manual page <https://manpages.ubuntu.com/manpages/lunar/en/man5/keyboard.5.html>`_
```
ogayot, line 148: Changing http for https is a great change. However, I feel a bit uncomfortable about hardcoding lunar here. The redirects are expected to change over time to point to newer Ubuntu releases ; which means we would need to manually update the links.

doc/reference/autoinstall-reference.rst
```diff
@@ -635,10 +635,9 @@ password
 The password for the new user, encrypted. This is required for use with
 ``sudo``, even if SSH access is configured.
 
-The crypted password string must conform to what
-`passwd <https://manpages.ubuntu.com/manpages/passwd.1.html>`_
-expects. Depending on the special characters in the password hash, quoting may
-be required, so it's safest to just always include the quotes around the hash.
+The crypted password string must conform to what the
+``passwd`` command requires. See the `passwd(1) manual page <https://manpages.ubuntu.com/manpages/lunar/en/man1/passwd.1ssl.html>`_
```
ogayot, line 639: This is more of a problem for this link because the redirect from https://manpages.ubuntu.com/manpages/passwd.1.html is currently wrong (I read a discussion about it IIRC). What it should point to is the passwd(1) manpage - which can be found here for instance: https://manpages.ubuntu.com/manpages/lunar/en/man1/passwd.1.html

If we're hardcoding the links to lunar, https://manpages.ubuntu.com/manpages/lunar/en/man1/passwd.1.html is the one we should use instead.

ogayot (changes requested):

doc/reference/autoinstall-reference.rst
ogayot: On second thought, maybe we want passwd(5) instead, I'm not sure.

rkratky:
> So the link checker doesn't follow redirects? Good to know!

@Chris-Peterson444  It does. The man pages at manpages.ubuntu.com don't use HTTP redirects. AFAICT, they're JS redirects, which aren't recognized by most agents. See:
```
$ curl --head --location https://manpages.ubuntu.com/manpages/passwd.1.html
HTTP/2 404 
server: nginx/1.14.0 (Ubuntu)
date: Fri, 13 Oct 2023 10:04:34 GMT
content-type: text/html
```

rkratky:
> AFAICT, they're JS redirects

Scratch that. They pbbly use nginx rewrite.

doc/reference/autoinstall-reference.rst
rkratky: > I feel a bit uncomfortable about hardcoding lunar here

Good point. Unfortunately, I don't know a good workaround (yet). I'll be looking into a Sphinx extension that allows the use of variables (rST substitutions) within URLs, which would allow us to have something like `https://manpages.ubuntu.com/manpages/|distro_version|/en/man5/keyboard.5.html`. Until then, I'm afraid it needs to be manual.

doc/reference/autoinstall-reference.rst
rkratky: > what it should point to is the passwd(1) manpage

Ugh, yes, I saw that but then forgot about it. Fixing it now.

doc/reference/autoinstall-reference.rst
ogayot: Okay, I guess that's acceptable.

ogayot (approved):
Thanks for the update!

dbungert (changes requested):
I would like you to explore linkcheck_allowed_redirects (if it works, I assume it won't) and linkcheck_ignore for the manpage problem.  Or if that's not an option, we might as well bump the series to mantic at this point.

dbungert:
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_ignore

rkratky:
> I would like you to explore linkcheck_allowed_redirects (if it works, I assume it won't) and linkcheck_ignore for the manpage problem.

Thanks for the suggestion. Looking into it.

rkratky:
Introduced the `extlinks` extension to handle manpage links.

dbungert (commented):
Very happy for the linkcheck test.  Please update series, RTD doesn't like extlinks at the moment, and rebase to drop the merge commit.  Thanks!

doc/custom_conf.py
```diff
@@ -172,3 +173,7 @@
 # can be copied even if they don't contain an EOF line.
 copybutton_prompt_text = '$ '
 copybutton_only_copy_prompt_lines = False
+
+extlinks = {
+    'manualpage': ( 'https://manpages.ubuntu.com/manpages/lunar/en/%s', '' )
```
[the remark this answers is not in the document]
rkratky: Ah, yeah :)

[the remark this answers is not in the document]
rkratky: Hmm, the thing is... those pages are not yet up, so the linkcheck would fail ;) (The internal nginx redirect handles it by redirecting back to the Lunar pages :)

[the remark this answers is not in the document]
dbungert: OK

dbungert (approved):

PR #1844, opened 2023-10-19
Merge discourse 2023 10 18

dbungert:
Merge updates on discourse to the locations here, mostly as a method of checking to see if there are updates needed for the RTD version.   The local version here is intended to go away ASAP once the RTD version is live and in use.

### DIFF
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -195,10 +195,14 @@ search_drivers
 ~~~~~~~~~~~~~~
 
 * **type:** boolean
-* **default:** ``true``
+* **default:** ``true`` (mostly, see below)
 
 Whether the installer should search for available third-party drivers. When
 set to ``false``, it disables the drivers :ref:`screen and section<ai-drivers>`.
+
+The default is true for most installs but false when a "core boot" or
+"enhanced secure boot" method is selected (where third-party drivers
+cannot currently be installed).
 
 id
 ~~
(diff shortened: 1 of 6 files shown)


ogayot (approved):

doc/reference/autoinstall-reference.rst
```diff
@@ -189,17 +189,21 @@ source
...
 ~~~~~~~~~~~~~~
 
 * **type:** boolean
-* **default:** ``true``
+* **default:** ``true`` (mostly, see below)
 
 Whether the installer should search for available third-party drivers. When
 set to ``false``, it disables the drivers :ref:`screen and section<ai-drivers>`.
 
+The default is true for most installs but false when a "core boot" or
```
rkratky, line 203:
```suggestion
The default is true for most installations but false when a "core boot" or
```

doc/reference/autoinstall-reference.rst
```diff
@@ -215,9 +219,9 @@ network
 
 * **type:** Netplan-format mapping, see below
 * **default:** DHCP on interfaces named ``eth*`` or ``en*``
-* **can be interactive:** yes
+* **can be interactive:** true
 
-`Netplan-formatted <https://netplan.io/reference>`_ network configuration.
+`Netplan-formatted <https://netplan.readthedocs.io/en/stable/netplan-yaml/>`_ network configuration.
```
rkratky, line 224: I think it's sensible to leave this redirect in. Their RTD URL may change, but `netplan.io/reference` will likely be always point to the right thing.
```suggestion
`Netplan-formatted <https://netplan.io/reference>`_ network configuration.
```

rkratky (commented):
A couple of nitpicks.

PR #1851, opened 2023-10-23
Update apport reporting 

Chris-Peterson444:
Some name and file location changes to be consistent with the (to be upload) updated subiquity hook in apport.

The only source of the traceback - if any - is the one in memory that we grab and manually attach to the report. We also need to explicitly write out the traceback information when we crash, otherwise a call to `ubuntu-bug` or similar won't be able to pick it up.

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -62,7 +62,7 @@ validate () {
                 # not to do that.
                 # If we run an autoinstall that customizes the security section as part
                 # of the test-suite, we will need to adapt this test.
-                python3 scripts/check-yaml-fields.py $tmpdir/var/log/installer/subiquity-curtin-apt.conf \
+                python3 scripts/check-yaml-fields.py $tmpdir/var/log/installer/curtin-install/subiquity-curtin-apt.conf \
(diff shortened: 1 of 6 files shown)


Chris-Peterson444:
The not-yet proposed subiquity hook in apport:
https://git.launchpad.net/~cpete/ubuntu/+source/apport/commit/?id=a48837b0c161afae1b0e889ece318a07b6b7a2c6

Chris-Peterson444:
Here are some example crash reports for these changes: 

Subiquity's geneated crash report: https://bugs.qastaging.launchpad.net/subiquity/+bug/1743387
Apport's generated crash report: https://bugs.qastaging.launchpad.net/subiquity/+bug/1743388

Since these links will be dead after some time, here are the corresponding crash files (renamed to .log so github will let me post them):
[apport-crash.log](https://github.com/canonical/subiquity/files/13079449/apport-crash.log)
[subiquity-crash.log](https://github.com/canonical/subiquity/files/13079452/subiquity-crash.log)

dbungert (approved):

PR #1853, opened 2023-10-25
Review reference docs

rkratky:
Ton of fixes.

* Heading markup made consistent with starter pack
* Proper inline markup used; semantic where possible
* Language scrubbed


doc/reuse/substitutions.txt
```diff
@@ -1,2 +1,3 @@
 .. |ubuntu-latest-version| replace:: 23.04
 .. |ubuntu-latest-codename| replace:: Lunar Lobster
+.. role:: command(literal)
```
dbungert, line 3: Is anything blocking updating the ubuntu-latest values?

doc/reference/autoinstall-reference.rst
```diff
@@ -133,79 +116,82 @@ The channel to check for updates.
...
+~~~~~~~~
 
 * **type:** mapping, see below
 * **default:** US English keyboard
 * **can be interactive:** true
 
-The layout of any attached keyboard. Often systems being automatically
-installed will not have a keyboard at all in which case the value used here
-does not matter.
+The layout of any attached keyboard. Often systems being automatically installed do have a keyboard. In such cases, the setting is not used.
```
dbungert, line 125: The meaning has been inverted here.  Maybe we just drop all the text starting with "Often systems  ..."

doc/reference/autoinstall-reference.rst
```diff
@@ -215,17 +201,13 @@ Identifier of the source to install (e.g., ``ubuntu-server-minimal``).
...
 * **type:** Netplan-format mapping, see below
 * **default:** DHCP on interfaces named ``eth*`` or ``en*``
 * **can be interactive:** true
 
-`Netplan-formatted <https://netplan.readthedocs.io/en/stable/netplan-yaml/>`_ network configuration.
-This will be applied during installation as well as in the installed system.
-The default is to interpret the configuration for the installation media, which runs
-DHCP version 4 on any interface with a name matching ``eth*`` or ``en*`` but then
-disables any interface that does not receive an address.
+`Netplan-formatted <https://netplan.io/reference>`_ network configuration. This is applied during installation as well as in the installed system. The default is to interpret the configuration for the installation media, which runs DHCP version 4 on any interface with a name matching ``eth*`` or ``en*`` but then disables any interface that does not receive an address.
```
dbungert, line 210: (not blocking) The counter-argument here is that I'd like to use intersphinx to cross-reference to netplan, as it is one of several related yaml formats (with cloud-init and curtin, among others if we really build an extensive configuration) that can be part of a valid autoinstall.  In that case, where would we want to link in that document set?

doc/reference/autoinstall-reference.rst
rkratky: I just forgot the "not" :)

doc/reuse/substitutions.txt
rkratky: Just my memory :)

doc/reference/autoinstall-reference.rst
rkratky: > Maybe we just drop all the text starting with "Often systems ..."

True. Redundant.

dbungert (approved):

doc/reference/autoinstall-reference.rst
rkratky: The redirects goes to https://netplan.readthedocs.io/en/latest/netplan-yaml/ now, which I believe is the right place -- the top of the `network` config reference. If ever they change the redirect (why?), we would need to reconsider. But until then, the redirect serves its ourpose I think, so we can use the "pretty" URL.

PR #1855, opened 2023-10-25
Revert orig_environ fix in command controllers

Chris-Peterson444:
This reverts commit 39f1ea9cb679b6e5c5cd098a342e1f80cb5e3901. The fix proposed in this patch caused more issues than it fixed. We will have to revisit this in a more nuanced way in the future. In the meantime users can make use of env directly to strip/modify the subcommand environment.


dbungert:
Reverting the merge also reverts 8f28063b9235dc93018ee419c1a62f692befce8e, some unrelated test-only changes.  I'd rather we revert just 39f1ea9cb679b6e5c5cd098a342e1f80cb5e3901

Chris-Peterson444:
Good catch. Fixed to only revert https://github.com/canonical/subiquity/commit/39f1ea9cb679b6e5c5cd098a342e1f80cb5e3901, thanks!

dbungert (approved):

PR #1866, opened 2023-11-02
doc: fix some link headings

dbungert:
The use of spellexception means we have some links that look like: intro-to-autoinstall.html#autoinstall-by-way-of-cloud-spellexception-config-spellexception

Also these are already wordlisted, so we can fix by just dropping the inline spellexception.

Also fix the `debian-installer` wordlist, just `debian` is correct.

### DIFF
--- a/doc/.custom_wordlist.txt
+++ b/doc/.custom_wordlist.txt
@@ -49,7 +49,7 @@ codecs
 config
 curtin
 debconf
-debian-installer
+debian
 el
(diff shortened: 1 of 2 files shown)


ogayot (approved):

rkratky:
Oops, I didn't realize the URLs would be formed using mark-up :)

rkratky (approved):

PR #1868, opened 2023-11-07
kvm-test: have different profiles for server and desktop

ogayot:
Booting Desktop and server live installer ISOs comes with different platform requirements. When running kvm-test, we now accept the name of a profile via the --profile option.

Profiles provide default settings for memory, disk size and extra QEMU options. For now, two profiles are hard-coded: "server" - which is the default and "desktop".

For desktop, we use two vCPU, 8 GiB of RAM, a 20 GiB disk and pass the -device qxl option.

### DIFF
--- a/scripts/kvm-test.py
+++ b/scripts/kvm-test.py
@@ -39,6 +39,15 @@ iso:
         focal: focal/ubuntu-20.04.3-live-server-amd64.iso
         bionic: bionic/bionic-live-server-amd64.iso
     default: edge
+profiles:
+    server:
+        memory: 2G
+        disk-size: 12G
+        extra-qemu-options: []
+    desktop:
+        memory: 8G
+        disk-size: 20G
+        extra-qemu-options: [-device, qxl, -smp, "2"]
 '''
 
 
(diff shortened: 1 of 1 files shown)


Chris-Peterson444 (approved):

scripts/kvm-test.py
```diff
@@ -43,6 +42,22 @@
...
+class Profile:
+    name: str
+    memory: str
+    disk_size: str
+    extra_qemu_options: list[str]
+
+
+profiles = {
+
+    'server': Profile(name='server', memory='2G', disk_size='12G', extra_qemu_options=[]),
```
ogayot, line 55: @dbungert you had previously set the default memory to 8 GiB. Would 2 GiB work for you? I can bump it up if needed.

dbungert: 2GB is fine for a default, but this breaks the config setup, so it'd be nice to be able to specify the profiles in the yaml cfg.  Do you have thoughts on that part?

ogayot: I expected that this would be a request, yes :)

I've pushed a change allowing to override a profile (or to create a new one) using:

```yaml
iso:
   # [...]
profiles:
  server:
    # server already exists in the default config, so we do not need to specify all the settings
    memory: 4G

  server-cpu4-ram16-disk20
    # this will create a new profile, all the properties must be specified
    memory: 16G
    qemu-extra-options: [-smp, "4"]
    disk-size: 20G
```

does it help?

ogayot:
I'm merging but feel free to reply here if there are any adjustments to make

PR #1869, opened 2023-11-07
console_conf: use the correct /run directory

kubiko:
We should use one `/run/console_conf` directory.
As this is defined dynamically based on the package name, align
other places where `/run/console-conf` is used and use `/run/console_conf` instead, to avoid cluttering of `/run` directory.
This also complicated strict confinement.

### DIFF
--- a/console_conf/core.py
+++ b/console_conf/core.py
@@ -26,7 +26,7 @@ log = logging.getLogger("console_conf.core")
 class ConsoleConf(TuiApplication):
     from console_conf import controllers as controllers_mod
 
-    project = "console_conf"
+    project = "console-conf"
 
     make_model = ConsoleConfModel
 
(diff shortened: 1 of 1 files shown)


dbungert (commented):
Can we look this up from SNAP_NAME or something?

kubiko:
@dbungert I have found from where `/run/console_conf` was coming from
this is set here: [core.py](https://github.com/canonical/subiquity/blob/main/subiquitycore/core.py#L64)
So the culprit was actually how the `project` name was set.
So for now, to keep consistency I would propose to align the project name here with what we use in other parts of the code as a name to form `/run` dir.

kubiko:
rebasing on main, to make clean diff

bboozzoo:
@dbungert I've opened https://github.com/canonical/subiquity/pull/1904 which includes the patch from @kubiko with some tweaks and a unit test for the identity controller. Please have a look.

kubiko:
closing as this has been now superseded by https://github.com/canonical/subiquity/pull/1904

PR #1871, opened 2023-11-07
console_conf: identity: handle os.ttyname() error

kubiko:
In strict snap confinement getting tty name is not permitted and calling `os.ttyname()` fails.
Handle the error and fall back to the use of the value from env variable, this is already set by the invoking process if needed.

### DIFF
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -102,11 +102,22 @@ Personalize your account at https://login.ubuntu.com.
 """
 
 
+def get_tty_name():
+    # handle error if we cannot get tty, e.g. in snap confinement
+    # try to get correct tty from env, it should be ready there
+    try:
+        tty_name = os.ttyname(0)[5:]  # strip off the /dev/
+    except OSError:
+        tty_name = os.getenv("tty", "")
+    return tty_name
+
+
 def write_login_details(fp, username, ips):
     sshcommands = "\n"
     for ip in ips:
         sshcommands += "    ssh %s@%s\n" % (username, ip)
-    tty_name = os.ttyname(0)[5:]  # strip off the /dev/
+
+    tty_name = get_tty_name()
     version = get_core_version() or "16"
     if len(ips) == 0:
         fp.write(
@@ -135,7 +146,7 @@ def write_login_details_standalone():
             print("device managed without user")
             return 2
         else:
-            tty_name = os.ttyname(0)[5:]
+            tty_name = get_tty_name()
             version = get_core_version() or "16"
             print(login_details_tmpl_no_ip.format(tty_name=tty_name, version=version))
             return 2


console_conf/controllers/identity.py
```diff
@@ -106,7 +106,12 @@ def write_login_details(fp, username, ips):
     sshcommands = "\n"
     for ip in ips:
         sshcommands += "    ssh %s@%s\n" % (username, ip)
-    tty_name = os.ttyname(0)[5:]  # strip off the /dev/
+    # handle error if we cannot get tty, e.g. in snap confinement
+    # try to get correct tty from env, it should be ready there
+    try:
+        tty_name = os.ttyname(0)[5:]  # strip off the /dev/
+    except OSError:
+        tty_name = os.getenv("tty", "")
```
dbungert, line 114: 1. Are we sure it's lower case?  If it's upper case, this will not obtain the correct value.
2. if we don't get a tty_name, we're going to print something like
```Ubuntu Core 24 on 1.2.3.4 ()```
3. While this sequence is short, it's also duplicated, so I'd like to see it extracted out to a utility function.

kubiko: env variable is set by the supporting script which is invoked by the console-conf wrapper, so we can use here what we prefer
Original script is setting lower case so I maintained that

kubiko (commented):
@dbungert I will prepare helper function for this then

console_conf/controllers/identity.py
dbungert: if tty is set, is that the value intended to be used in preference to what os.ttyname() reports?

dbungert (changes requested):
I want a unit test for this change.

console_conf/controllers/identity.py
kubiko: Good question. TBH I have no strong opinion here.
Favouring `os.ttyname` means preserving current behaviour when possible, which was my goal, to have the least obstructive change and only introduce fallback.

But now when you ask, I suppose setting `tty` env would provide the possibility to override default behaviour, which might be sometimes useful.
Happy to change the logic if you prefer it that way.

bboozzoo:
I did a little investigation here and I think this PR is no longer needed. I reverted the patch locally and can complete the device setup through console-conf. What I understand happened here, is that console-conf was run under ssh, hence tty was on /dev/pts/*, which will get blocked in the snap because snaps get their private /dev/pts. However, in case of console-conf this isn't the case and we always have a real tty.

kubiko:
After more testing, it seems like this has been the only problem when `console-conf` is invoked over ssh session, which contradicts itself.
`console-conf` should only run on unprovisioned systems, hence it should never run over ssh session, as the sole purpose of `console-conf` is to set up the device so ssh can be used to access the device securely.

PR #1872, opened 2023-11-07
console_conf: identity: use snapd unix socket

kubiko:
When running in strict snap confinement `snap` client binary is not accessible.
Additionally, the output format of `snap` client binary is not guaranteed not to change.
snapd REST API should be used instead when communicating to the snapd and expecting a parsable response.
This also improves error handling for this call.

### DIFF
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -13,7 +13,6 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import json
 import logging
 import os
 import pwd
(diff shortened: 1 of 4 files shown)


mwhudson (commented):
Is this still the one synchronous snapd api call? 

It would be good to change console-conf to the asynchronous snapd api and support fake connections like subiquity does but one thing at a time I guess.

kubiko:
@mwhudson yes this is still a synchronous call. This is also synchronous on the snapd REST API level, so this would need to be supported as async call there first,

dbungert (commented):
Conceptually this is fine, but I consider a change like this high risk because I'm not clear on the plan to test this and ensure that it does not regress.

I would like to hear on that detail before we proceed.

kubiko:
@dbungert I think risk-wise, this change does the exact opposite.
[snapd REST API](https://snapcraft.io/docs/snapd-api) on unix socket is stable and snapd team is committed to keeping it stable/compatible.


While parsing of the `snap` client stdout  is something which is not guaranteed and can change without prior notice.

@pedronis would you be able to comment on risk of moving from parsing `snap xxx` stdout  to using snapd REST API?

kubiko:
@dbungert if you have existing tests for the use of the existing `snap create-user` option, can you please point me to those? I can have a look and see how to adapt it to test REST API case.

mwhudson:
In general I prefer to talk to the API than invoke the snap client. Is the create user API ever going to get fixed to be async though (i.e. have the POST return a change id) It's currently very much a special snowflake among snapd apis. I guess that would be done in a backwards compatible way...

It would also be nice to have console-conf use the richer approach to talking to snapd subiquity does but well. One thing at a time!

kubiko:
@mwhudson no argument about async option here. At least REST API gives you this option if API exist. Unlike parsing stdout of the client invocation.
Unfortunately, you cannot use a trick to spool another thread to turn this sync call into async, as python uses shared memory to sync threads and this fails in the strict confinement. I do have a patch for this, but it essentially requires [patching libpthread](https://github.com/kubiko/glibc-shm-shim/blob/core22/shm-snap.patch) shared memory path generator. And I'd avoid this in the official Canonical snap...

Currently, this is rarely used API. Essentially it's only down to `console-conf` use-case, so it is probably not that high on the priority list for snapd to provide async option for the call.

bboozzoo:
@dbungert I've updated the PR along with some unit tests, please have a look.

dbungert (approved):

PR #1873, opened 2023-11-07
console_conf: identity: use prepared fingerprints

kubiko:
In strict snap confinement, sshd config or host keys are not accessible.
If strict confinement is detected, instead allow reuse of the host keys fingerprints already prepared by invoking process. Prepared fingerprints are stored in: `/run/console-conf/host-fingerprints.txt`

Additionally, in strict snap confinement, snap can not access the other user's ssh keys or authorised keys.
Do not read those, it does not seem like we are displaying them at the moment anyway.

### DIFF
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -19,7 +19,6 @@ import os
 import pwd
 import shlex
 import sys
-from pathlib import Path
 
 from console_conf.ui.views import IdentityView, LoginView
 from subiquitycore.snapd import SnapdConnection
@@ -117,17 +116,7 @@ def write_login_details(fp, username, ips):
         )
     else:
         first_ip = ips[0]
-        key_info = None
-        if os.getenv("SNAP_CONFINEMENT", "classic") == "strict":
-            # if we run in confinement, we have no direct accesss to host
-            # keys info use prepared finger prints if exist
-            host_fingerprints_path = "/run/console-conf/host-fingerprints.txt"
-            host_fingerprints = Path(host_fingerprints_path)
-            if host_fingerprints.is_file():
-                fingerprints = open(host_fingerprints_path, "r")
-                key_info = fingerprints.read()
-        else:
-            key_info = host_key_info()
+        key_info = host_key_info()
         fp.write(
             login_details_tmpl.format(
                 sshcommands=sshcommands,
--- a/console_conf/controllers/identity.py
+++ b/console_conf/controllers/identity.py
@@ -19,6 +19,7 @@ import os
 import pwd
 import shlex
 import sys
+from pathlib import Path
 
 from console_conf.ui.views import IdentityView, LoginView
 from subiquitycore.snapd import SnapdConnection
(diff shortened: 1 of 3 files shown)


console_conf/controllers/identity.py
```diff
@@ -116,10 +117,21 @@ def write_login_details(fp, username, ips):
...
     else:
         first_ip = ips[0]
+        key_info = None
+        if os.getenv("SNAP_CONFINEMENT", "classic") == "strict":
+            # if we run in confinement, we have no direct accesss to host
+            # keys info use prepared finger prints if exist
+            host_fingerprints_path = "/run/console-conf/host-fingerprints.txt"
+            host_fingerprints = Path(host_fingerprints_path)
+            if host_fingerprints.is_file():
+                fingerprints = open(host_fingerprints_path, "r")
```
dbungert, line 127: `open` is perfectly content to open a Pathlib object, so we might as well drop the host_fingerprints_path value.

console_conf/controllers/identity.py
```diff
@@ -116,10 +117,21 @@ def write_login_details(fp, username, ips):
         )
     else:
         first_ip = ips[0]
+        key_info = None
+        if os.getenv("SNAP_CONFINEMENT", "classic") == "strict":
+            # if we run in confinement, we have no direct accesss to host
+            # keys info use prepared finger prints if exist
+            host_fingerprints_path = "/run/console-conf/host-fingerprints.txt"
```
dbungert, line 124: Let's use `SNAP_NAME` instead of hardcoding console-conf, this will help future confined Subiquity efforts.

console_conf/controllers/identity.py
```diff
@@ -116,10 +117,21 @@ def write_login_details(fp, username, ips):
         )
     else:
         first_ip = ips[0]
+        key_info = None
```
dbungert, line 120: Other code calls `host_key_info()`, so the logic here should be moved to there instead of only handling the confined version in some cases.

console_conf/controllers/identity.py
```diff
@@ -156,9 +168,21 @@ def make_ui(self):
             device_owner = get_device_owner()
             if device_owner:
                 self.model.add_user(device_owner)
-                key_file = os.path.join(device_owner["homedir"], ".ssh/authorized_keys")
-                cp = run_command(["ssh-keygen", "-lf", key_file])
-                self.model.user.fingerprints = cp.stdout.replace("\r", "").splitlines()
+                # if we are in strict snap confinement we cannot
+                # access random user's .ssh dir
+                if os.getenv("SNAP_CONFINEMENT", "classic") == "strict":
```
dbungert, line 173: What should the login view look like in this case?  It seems to quite strongly assume that we can show fingerprints, just omitting them isn't great.

dbungert (changes requested):

kubiko:
@dbungert things are now moved around

kubiko (commented):
@dbungert things are now moved around

console_conf/controllers/identity.py
bboozzoo: I think this isn't a problem anymore. I reviewed the code and how User is used throughout the places here and in the login view and none of the places reference `fingerprints`, which means they were never shown and were collected for no clear reason. At this point I think we can just drop it and we'll have one less problem to address. I'll propose a separate branch with the cleanup.

bboozzoo: opened https://github.com/canonical/subiquity/pull/1905

bboozzoo:
@dbungert I have pushed some tweaks, the snap helpers bits are proposed as a separate PR right here https://github.com/anonical/subiquity/pull/1906 to make the review a little easier. The user's key fingerprints are dropped in https://github.com/canonical/subiquity/pull/1905. The PRs can land separately and then I'll rebase this one. What's left is basically just getting the pre-made host fingerprints if the file is present.

bboozzoo:
@dbungert I've updated the PR, please have a look.

dbungert (approved):
LGTM, merging.  I took the liberty of fixing a file leak, which is already included in the commit set.

bboozzoo:
Good catch, I missed that. Thank you!

PR #1874, opened 2023-11-07
subiquitycore: do not set uid and gid in snap strict confinement

kubiko:
Snap strict confinement does not permit uid and gid alteration.
This is also not necessary, we should be running as root anyway. If strict snap confinement is detected, skip this step.

### DIFF
--- a/subiquitycore/file_util.py
+++ b/subiquitycore/file_util.py
@@ -30,6 +30,8 @@ log = logging.getLogger("subiquitycore.file_util")
 
 
 def set_log_perms(target, *, group_write=False, mode=None, group=_DEF_GROUP):
+    if os.getenv("SNAP_CONFINEMENT", "classic") == "strict":
+        return
     if os.getuid() != 0:
         log.warning(
             "set_log_perms: running as non-root - not adjusting"


dbungert (commented):
So what's the log situation during strict confinement?

kubiko:
@dbungert it logs as expected. As `console-conf` already runs as root in strict confinement, there is no need to change did/gid.
 Strict confinement just explicitly blocks those system calls as they are known tools used for privilege escalations.

subiquitycore/file_util.py
```diff
@@ -30,6 +30,8 @@
 
 
 def set_log_perms(target, *, group_write=False, mode=None, group=_DEF_GROUP):
+    if os.getenv("SNAP_CONFINEMENT", "classic") == "strict":
```
bboozzoo, line 33: @dbungert to be precise, the problem is with chown() further down. The seccomp template for strict snaps prevents a process from executing chown() with target user/group other than -1 or root. In this case https://github.com/canonical/subiquity/blob/53180bbc355f91e569427b97d73ac6d4825e1407/subiquitycore/log.py#L26 is being called which attempts to chown with group `adm`. I can push a patch to skip the chown call when executing under a strict confinement. Though I think it may be preferrable to update `log.setup_logger()` to use `root` in such scenario, wdyt?

bboozzoo: I've opened https://github.com/canonical/subiquity/pull/1903 please have a look

PR #1875, opened 2023-11-08
console_conf: modify wrapper helper to consider consol-conf as snap

kubiko:
Modify the wrapper to consider consol-conf installed as snap.
Prepare sshd finger prints to be used by consol-conf snap.

### DIFF
--- a/bin/console-conf-wrapper
+++ b/bin/console-conf-wrapper
@@ -84,7 +84,13 @@ if [ "$(snap managed)" = "true" ]; then
     exit 0
 fi
 
-cat /usr/share/subiquity/console-conf-wait
+# if there is NOT console-conf snap, exit gracefully
+if [ ! -e "/snap/console-conf" ]; then
+  touch /var/lib/console-conf/complete
+  exit 0
+fi
+
+snap run console-conf.wait
 read REPLY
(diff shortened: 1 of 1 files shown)


bin/console-conf-wrapper
```diff
@@ -84,12 +84,26 @@ if [ "$(snap managed)" = "true" ]; then
     exit 0
 fi
 
-cat /usr/share/subiquity/console-conf-wait
+# if there is console-conf snap, exit gracefully
+if [ ! -e "/snap/console-conf" ]; then
```
dbungert, line 88: The comment and code don't seem to match, so I don't understand the intent.  Please clarify.

dbungert (changes requested):

bin/console-conf-wrapper
kubiko: Ups this is an inverted comment, it should be
`# if there is NO console-conf snap, exit gracefully`

kubiko: BTW this entire file will be likely removed as we won't be pulling subiquity to `core-base` build anymore, so this is to keep things in sync.

dbungert (approved):

PR #1880, opened 2023-12-18
Accept a fixed partition size for reset partition in autoinstall.yaml

medicalwei:
This changeset will allow reset_partition to accept int or human-readable disk size (e.g. "12G"), so that the size of reset partition can be fixed.  This will be used for some OEM deployments and automated testing where overwriting or updating reset partition is necessary.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -29,6 +29,7 @@ import pyudev
 from curtin import swap
 from curtin.commands.extract import AbstractSourceHandler
 from curtin.storage_config import ptable_part_type_to_flag
+from curtin.util import human2bytes
 
 from subiquity.common.apidef import API
 from subiquity.common.errorreport import ErrorReportKind
@@ -707,7 +708,10 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             raise Exception("failed to locate gap after adding boot")
 
         if choice.reset_partition:
-            if self.app.opts.dry_run:
+            if choice.reset_partition_size is not None:
+                part_align = disk.alignment_data().part_align
+                reset_size = align_up(choice.reset_partition_size, part_align)
+            elif self.app.opts.dry_run:
                 reset_size = DRY_RUN_RESET_SIZE
             else:
                 cp = await arun_command(["du", "-sb", "/cdrom"])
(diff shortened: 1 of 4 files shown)


medicalwei:
There's some underlying serialization issues since attempting to union primitives upsets serialize.py as it seems only taking unions of non-primitive types. Is this intentional?

subiquity/server/controllers/filesystem.py
```diff
@@ -1405,6 +1409,20 @@ async def run_autoinstall_guided(self, layout):
...
+        reset_partition = False
+        reset_partition_size = None
+        rp_input = layout.get("reset-partition", None)
+        if rp_input:
+            reset_partition = True
+            if type(rp_input) in (str, int):
+                reset_partition_size = int(human2bytes(rp_input))
+                log.info(
+                    "autoinstall: will install reset partition"
+                    f"of size {reset_partition_size}"
```
ogayot, line 1421: looks like we're missing a space :)

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -459,6 +459,26 @@ async def test_guided_reset_partition(self):
...
+            GuidedChoiceV2(
+                target=target,
+                capability=GuidedCapability.DIRECT,
+                reset_partition=True,
+                reset_partition_size=fixed_reset_size,
+            )
+        )
+        [d1p1, d1p2, d1p3] = self.d1.partitions()
+        self.assertEqual("/boot/efi", d1p1.mount)
+        self.assertEqual(None, d1p2.mount)
```
ogayot, line 478:
```python
self.assertIsNone(d1p2.mount)
```

subiquity/server/controllers/filesystem.py
```diff
@@ -1405,6 +1409,20 @@ async def run_autoinstall_guided(self, layout):
                 disk_id=gap.device.id, gap=gap, allowed=[]
             )
 
+        reset_partition = False
+        reset_partition_size = None
+        rp_input = layout.get("reset-partition", None)
+        if rp_input:
+            reset_partition = True
+            if type(rp_input) in (str, int):
```
ogayot, line 1417:
```python
if isinstance(rp_input, (str, int)):
```

ogayot (approved):
Thanks, LGTM! Just a few minor tweaks

subiquity/server/controllers/filesystem.py
medicalwei: Added a space in the end of the previous line, thanks!

dbungert (commented):
I would like to see some doc/reference/autoinstall-reference.rst material for this, but then I see now that reset-partition isn't documented yet.  Unfortunate.  @medicalwei would you add some content for this in the reference and ignore the bit where the reset_partition bool isn't documented yet?

medicalwei:
@dbungert Reference added, and I hope my writing style doesn't look strange

dbungert (approved):

PR #1882, opened 2024-01-04
ai: include link to docs in post-install user data

Chris-Peterson444:
We write out the autoinstall data to make the install repeatable but this should also include a reference to the autoinstall documentation to increase usability.

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -166,6 +166,20 @@ class InstallController(SubiquityController):
             and self.model.source.current.variant != "core"
         )
 
+    def write_autoinstall_config(self) -> None:
+        autoinstall_path = os.path.join(
+            self.app.root, "var/log/installer/autoinstall-user-data"
+        )
+        autoinstall_config = (
+            "#cloud-config\n"
+            "# See the auto install documentation at:"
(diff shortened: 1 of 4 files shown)


subiquity/server/controllers/install.py
```diff
@@ -166,6 +166,20 @@ def supports_apt(self) -> bool:
...
         )
 
+    def write_autoinstall_config(self) -> None:
+        autoinstall_path = os.path.join(
+            self.app.root, "var/log/installer/autoinstall-user-data"
+        )
+        autoinstall_config = (
+            "#cloud-config\n"
+            "# See the auto install documentation at:"
+            "# https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html\n"  # noqa: E501
```
ogayot, line 176: Not willing to block your PR on this but do we expect this URL to be valid going forward?

Chris-Peterson444: For now, but I'm pretty sure the `canonical-subiquity.readthedocs-host.com` part will change at some point as the docs evolve. I wanted to put this somewhere intelligent to only have to change it in one spot, but hadn't come up with anything nice.

ogayot: OK, I don't think it's a big deal.

ogayot (approved):

PR #1884, opened 2024-01-12
Nvme o tcp poc

ogayot:
This PR adds basic support for NVMe over TCP, with the following limits:

* only partitions that are not critical for booting (I hard-coded `/home` currently) can be placed on remote storage

This PR depends on https://code.launchpad.net/~ogayot/curtin/+git/curtin/+merge/458446 for proper NVMe-over-TCP detection.

For testing, you can `s/simple/nvme-over-tcp/` in the makefile and then use `make dryrun`

Some things to address before merging:

- [x] Add tests
- [x] Ensure we do not attempt to create an efistub on a remote drive
- [x] Ensure that NVMe drives are gray-ed out in the guided screen

I also wanted to ensure that API endpoints would enforce restrictions on remote storage, but since we are going to change the restrictions in the near future, I don't think it is worth the effort here. Desktop is also not the main use-case for NVMe over TCP AFAICT.


subiquity/models/filesystem.py
```diff
@@ -704,12 +707,21 @@ class Dasd:
     preserve: bool = False
 
 
+@fsobj("nvme_controller")
+class NVMeController:
+    transport: str
```
dbungert, line 712: This is a pretty finite list, correct?  An Enum would be nice.

dbungert (approved):

subiquity/ui/mount.py
```diff
@@ -79,6 +79,16 @@ def enable_common_mountpoints(self):
             if isinstance(opt.value, str):
                 opt.enabled = opt.value in common_mountpoints
 
+    def disable_all_mountpoints_but_home(self):
+        """Currently, filesystems of non-local disks can only be mounted at
+        /home."""
```
mwhudson, line 84: The comment isn't quite right as users can type whatever they like into the "other" box, including things like / and /boot. Is it worth displaying a warning like we do when the user mounts a partition at say / without reformatting it?

subiquity/ui/views/filesystem/partition.py
```diff
@@ -161,15 +161,27 @@ def valid_char(self, ch):
...
         self.device = device
         self.existing_fs_type = None
+        self.remote_storage: bool = remote_storage
         if device:
             ofstype = device.original_fstype()
             if ofstype:
                 self.existing_fs_type = ofstype
         initial_path = initial.get("mount")
+        if not initial_path and remote_storage:
+            initial["mount"] = "/home"
```
mwhudson, line 184: Probably shouldn't do this if there is already a mount for /home?

subiquity/common/filesystem/labels.py
```diff
@@ -117,6 +117,8 @@ def desc(device):
 def _desc_disk(disk):
     if disk.multipath:
         return _("multipath device")
+    if disk.on_remote_storage():
+        return _("NVMe/TCP drive")
```
mwhudson, line 121: This looks a bit like an easy thing to forget if we add iscsi or nvme-oF or whatever; I think you should poke under the covers a bit more here.

mwhudson (changes requested):
Hi this is looking really good. A few niggles though!

subiquity/models/filesystem.py
mwhudson: I guess we should consider what happens if there is a NVMe-oF devices of some other kind on the system.

subiquity/ui/views/filesystem/partition.py
ogayot: Yeah, it's not ideal. But since we show an error message in the form when a partition already exists, I do think that it's not that bad.

2024-01-26T14:24:13,147115123+01:00

subiquity/ui/mount.py
ogayot: Good point, updated the comment and added a warning message:

2024-01-26T14:52:04,155473089+01:00

subiquity/common/filesystem/labels.py
ogayot: Updated with a default value of "remote drive". Currently no scenario should lead to this value being displayed since only NVMe over TCP drives are considered remote.

subiquity/models/filesystem.py
ogayot: Yes, I'm not really sure what the other values could be. "rdma"  would probably be one of them but I haven't tested it.

Curtin's storage config will not complain if it encounters a value that it does not know about (it only knows about "tcp" really, not even "pcie"). Instead of raising a `ValueError` if the value for transport is unknown, I'd rather atempt the install. wdyt?

mwhudson: I think I agree with your plan here

subiquity/ui/views/filesystem/partition.py
mwhudson: I guess that's OK

mwhudson (commented):
I think I'm happy with this now. Obviously can't land until curtin changes land.

ogayot:
The curtin bits are merged now. @mwhudson it's just missing your approval (since you requested changes earlier) before I can  merge. Thanks!

mwhudson (approved):
Sorry forgot to set the radio button earlier apparently!

PR #1885, opened 2024-01-15
apt: Only backup/restore sources.list if it exists

julian-klode:
We backup the sources.list to a file in sources.list.d to be able to insert our source first; if we do not have a sources.list this is not necessary.

If an image uses ubuntu.sources, this logic also works fine as sources.list that we write will be sourced before the ubuntu.sources.

### DIFF
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -327,8 +327,10 @@ class AptConfigurer:
         # The file only exists if we are online
         with contextlib.suppress(FileNotFoundError):
             os.unlink(target_mnt.p("etc/apt/sources.list.d/original.list"))
-        with contextlib.suppress(FileNotFoundError):
+        try:
             _restore_file("etc/apt/sources.list")
+        except FileNotFoundError:
+            os.unlink(target_mnt.p("etc/apt/sources.list"))
 
         with contextlib.suppress(FileNotFoundError):
             _restore_file("etc/apt/apt.conf.d/90curtin-aptproxy")
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -243,10 +243,11 @@ class AptConfigurer:
         await self.mounter.mount("/cdrom", self.install_tree.p("cdrom"), options="bind")
 
         if self.app.base_model.network.has_network:
-            os.rename(
-                self.install_tree.p("etc/apt/sources.list"),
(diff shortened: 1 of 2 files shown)


julian-klode:
Reading this more I'm not sure if this actually deletes the /etc/apt/sources.list with the `cdrom` line, I guess the `_restore_file` is weird and should always succeed or something.

ogayot:
Related to LP:2026647

subiquity/server/apt.py
```diff
@@ -326,7 +327,8 @@ def _restore_file(path: str) -> None:
         # The file only exists if we are online
         with contextlib.suppress(FileNotFoundError):
             os.unlink(target_mnt.p("etc/apt/sources.list.d/original.list"))
-        _restore_file("etc/apt/sources.list")
+        with contextlib.suppress(FileNotFoundError):
+            _restore_file("etc/apt/sources.list")
```
ogayot, line 331: I think we don't want to keep this file on the target system if it wasn't present in the source overlay. I'd suggest something along these lines:

```python
try:
    _restore_file("etc/apt/sources.list")
except FileNotFoundError:
    os.unlink(target_mnt.p("etc/apt/sources.list"))
```

ogayot (changes requested):
Thanks Julian!

> Reading this more I'm not sure if this actually deletes the /etc/apt/sources.list with the `cdrom` line, I guess the `_restore_file` is weird and should always succeed or something.

Yes, the presence of `etc/apt/sources.lists` containing the cdrom entry is unconditional https://github.com/canonical/subiquity/blob/753ce9f7012a526df6deda80ac6a91315b332078/subiquity/server/apt.py#L259-L262

I'm not sure if it would be a huge deal to leave this file on the target system, but for now, I think it's better to keep the current behavior - which means removing the file if it does not exist in the source overlay.

ogayot (approved):
I pushed a change to drop sources.list from the target if it wasn't present in the configured tree.

Also added a set of tests for the .deconfigure method.

Marking "Approved" since I requested changes earlier. But it would be fair to get a second approval before merging since I authored some of the changes.

Chris-Peterson444 (approved):
LGTM!

mwhudson (approved):
Thanks for this, and for adding tests!

PR #1886, opened 2024-01-17
filesystem: double probert timeout when running os-prober

ogayot:
We know that running os-prober on some systems is slow, potentially very slow, especially when multiple removable devices are present.

This results in many bug reports showing "block probing crashed with TimeoutError".

For now, I suggest we double the probert timeout when os-prober is involved.


Chris-Peterson444 (approved):
This makes sense to me. In the future I wonder if we could use [shield](https://docs.python.org/3/library/asyncio-task.html#asyncio.shield) to conditionally quit and/or report it's a hardware / probert issue.

dbungert (approved):

PR #1888, opened 2024-01-18
apport: show name of curtin stage when an error occurs

ogayot:
When an error occurs during the execution of a "$ curtin install" invocation, the associated bug report would be titled:

  "install failed crashed with CalledProcessError"

This is not very user friendly and it does not immediately give any insight on:

With the number of reports that we receive these days, I think classifying these reports would make our life a bit easier, and possibly make the users understand better what's going on.

This change adds a wat to detect if a bug report is the result of a "$ curtin install" invocation failing. If it is, subiquity sets the title of the bug report accordingly. Examples:

 * "partitioning crashed with CurtinInstallError"
 * "extract crashed with CurtinInstallError"
 * "curthooks crashed with CurtinInstallError"

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -55,6 +55,12 @@ from subiquitycore.utils import arun_command, log_process_streams
 log = logging.getLogger("subiquity.server.controllers.install")
 
 
+class CurtinInstallError(Exception):
+    def __init__(self, *, stages: List[str]) -> None:
+        super().__init__()
+        self.stages = stages
+
+
 class TracebackExtractor:
     start_marker = re.compile(r"^Traceback \(most recent call last\):")
     end_marker = re.compile(r"\S")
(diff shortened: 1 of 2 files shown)


Chris-Peterson444 (approved):
Very nice!

dbungert:
I think this implementation is OK, but I'm looking at subiquity/server/curtin.py `run_curtin_command` and wondering if we should look at this problem there.  In that case, it's not just a CalledProcessError, we have more context.  What do you think about wrapping CalledProcessError with a Curtin-specific error that contains this information?  That would seem to allow us to avoid fullmatch() tricks and pass this info around more explicitly - perhaps the name would need to be passed onward from subiquity/server/controllers/install.py `run_curtin_step`.

ogayot:
> I think this implementation is OK, but I'm looking at subiquity/server/curtin.py `run_curtin_command` and wondering if we should look at this problem there. In that case, it's not just a CalledProcessError, we have more context. What do you think about wrapping CalledProcessError with a Curtin-specific error that contains this information? That would seem to allow us to avoid fullmatch() tricks and pass this info around more explicitly - perhaps the name would need to be passed onward from subiquity/server/controllers/install.py `run_curtin_step`.

Thanks for the idea!
I came up with an in between implementation, I hope it's ok.
It removes all the ugly fullmatch logic but covers only the `curtin install` use-cases. I think we can iterate if we want to catch more use cases that are not part of the `curtin install` stuff. (e.g., `curtin apt-config`).

dbungert (approved):

PR #1892, opened 2024-01-24
ubuntu-pro: enable on noble, although it's not yet marked LTS

ogayot:
The /etc/lsb-release file currently on noble shows:

```
  DISTRIB_ID=Ubuntu
  DISTRIB_RELEASE=24.04
  DISTRIB_CODENAME=noble
  DISTRIB_DESCRIPTION="Ubuntu Noble Numbat (development branch)"
```

Therefore, the Ubuntu Pro pages get skipped. This makes testing on noble difficult, so make an exception for noble now and we'll drop it once it's marked LTS.

### DIFF
--- a/subiquity/client/controllers/tests/test_ubuntu_pro.py
+++ b/subiquity/client/controllers/tests/test_ubuntu_pro.py
@@ -78,3 +78,25 @@ class TestUbuntuProController(unittest.IsolatedAsyncioTestCase):
         view.assert_called_once_with(
             ctrler, token="", has_network=False, pre_release=True
         )
+
+    @patch("subiquity.client.controllers.ubuntu_pro.UbuntuProView")
+    @patch(
+        "subiquity.client.controllers.ubuntu_pro.lsb_release",
+        return_value={
+            "description": "Ubuntu R R (development branch)",
+            "release": "26.04",
+        },
+    )
+    async def test_make_ui__26_04_future(self, release, view):
+        ctrler = self.ctrler
+
+        rv = UbuntuProResponse(token="", has_network=False)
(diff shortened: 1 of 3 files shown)


Chris-Peterson444 (approved):
Makes sense and looks good to me!

mwhudson (commented):
Do we display any wording to users when they are installing a pre-release version already? We don't want to imply that a user can actually get support via pro using pre-release media do we?

ogayot:
> Do we display any wording to users when they are installing a pre-release version already? We don't want to imply that a user can actually get support via pro using pre-release media do we?

Good point. And also the text needs a refresh (and would ideally depend on the version of Ubuntu that we install ...).

I've added a warning message

2024-01-26T10:50:54,005138624+01:00

subiquity/client/controllers/tests/test_ubuntu_pro.py
```diff
@@ -0,0 +1,80 @@
...
+            await self.ctrler.make_ui()
+
+        view.assert_not_called()
+
+    @patch("subiquity.client.controllers.ubuntu_pro.UbuntuProView")
+    @patch(
+        "subiquity.client.controllers.ubuntu_pro.lsb_release",
+        return_value={
+            "description": "Ubuntu Noble Numbat (development branch)",
+            "release": "24.04",
```
dbungert, line 65: I'm sure the same problem will happen in 2 years, do you mind if we generalize this for any 2x.04 ?

ogayot: Done, thanks!

PR #1893, opened 2024-01-24
Fix emtpy list of authorized_keys in /ssh endpoint

ogayot:
The /ssh endpoint was, I think, meant to return a list of authorized_keys but it didn't. 

Fixed the issue and also removed a variable that probably has been unused for years now.


mwhudson (approved):
Eh well. The return type of the GET is only the same as POST out of laziness really I think. This is more correct though and would help enable a better interface on the client side (really the client UI should let you download as many keys as you want to and then allow you to proceed rather than fetching as part of moving to the next screen...)

PR #1894, opened 2024-01-25
factory-reset: remove EFI boot entry, retarget GRUB entry to chainload Reset Partition GRUB

medicalwei:
This resolves https://pad.lv/2051099

The reason removing EFI boot entry is because any of our known factory workflow is not able to create EFI boot entry on the target machine.

GRUB boot entry is that we would like to adjust kernel boot commands in the reset partition.  Do we have any concern using chainloading?

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -31,7 +31,6 @@ from typing import Any, Dict, List, Optional
 
 import yaml
 from curtin.config import merge_config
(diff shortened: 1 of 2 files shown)


mwhudson (approved):
I think this broadly makes sense (and hey, it deletes code overall, a nice bonus).

Is there a reason it's still draft?

medicalwei:
No, was fixing code so I set it to draft. I am setting that back.

mwhudson (approved):

PR #1897, opened 2024-01-29
AutoinstallError exception  

Chris-Peterson444:
Adds a special exception for Autoinstall related failures and a
specific implementation for autoinstall validation failures.

When a user passes incorrect autoinstall data, the generated
bug report will be titled:

    "Autoinstall crashed with AutoinstallValidationError"

Where the traceback will start with a hint on the offending
section. For example:

    "Malformed autoinstall in 'apt' section"

Information from the original jsonschema ValidationError is
also available in the traceback to point to specific issues
with the provided data.

The section reporting is a little crude in the validation
of the base schema done by SubiquityServer, which can't discern
between the 'interative-sessions' and 'version' keys, but for now
the scope is pretty limited and can be fixed up at a later time.

### DIFF
--- a/subiquity/server/autoinstall.py
+++ b/subiquity/server/autoinstall.py
@@ -1,0 +1,31 @@
+# Copyright 2024 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import logging
+
+log = logging.getLogger("subiquity.server.autoinstall")
+
+
+class AutoinstallError(Exception):
+    pass
+
+
+class AutoinstallValidationError(AutoinstallError):
+    def __init__(
+        self,
+        owner: str,
+    ):
+        self.message = f"Malformed autoinstall in {owner!r} section"
+        self.owner = owner
+        super().__init__(self.message)
--- a/subiquity/server/controller.py
+++ b/subiquity/server/controller.py
@@ -19,8 +19,10 @@ import os
 from typing import Any, Optional
 
 import jsonschema
+from jsonschema.exceptions import ValidationError
 
 from subiquity.common.api.server import bind
+from subiquity.server.autoinstall import AutoinstallValidationError
 from subiquity.server.types import InstallerChannels
 from subiquitycore.context import with_context
 from subiquitycore.controller import BaseController
@@ -54,6 +56,19 @@ class SubiquityController(BaseController):
             await self.configured()
             self._active = False
 
+    def validate_autoinstall(self, ai_data: dict) -> None:
+        try:
+            jsonschema.validate(ai_data, self.autoinstall_schema)
+
+        except ValidationError as original_exception:
+            section = self.autoinstall_key
+
+            new_exception: AutoinstallValidationError = AutoinstallValidationError(
+                section,
+            )
+
+            raise new_exception from original_exception
+
     def setup_autoinstall(self):
         if not self.app.autoinstall_config:
             return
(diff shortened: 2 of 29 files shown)


Chris-Peterson444:
Commit f6a7819ef9932ff45edbe6338f38137ff51fbb31 is a little big and I'm tempted to create a separate PR for it. Although it's mostly repeated lines to just test the the json schemas defined by the controllers are valid. 

Testing this is generally important, but `jsonschema.validate` can throw an error from either non-conforming data or bad json schemas. We should catch the latter at test time so only validation errors happen at run time.

Chris-Peterson444:
I updated the commit adding `interactive-sections` in the base schema to also include it in the documentation's schema

Chris-Peterson444:
Rebased to not conflict with https://github.com/canonical/subiquity/pull/1898

subiquity/server/autoinstall.py
```diff
@@ -0,0 +1,50 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import logging
+
+from jsonschema.exceptions import ValidationError
+
+log = logging.getLogger("subiquity.server.autoinstall")
+
+
+class AutoinstallError(Exception):
```
dbungert, line 22: Do we need to build out this functionality?  The result is that AutoinstallError doesn't really behave like other exceptions - str() and repr() look very different, we can get to the message on an exception with str(exception), etc.  Basically, what are the advantage of this approach over an declaration like:
```
class AutoinstallError(Exception):
    pass
```

subiquity/server/controller.py
```diff
@@ -54,6 +57,26 @@ async def _confirmed(self):
...
+
+        except ValidationError as original_exception:
+            section = self.autoinstall_key
+
+            new_exception: AutoinstallValidationError = AutoinstallValidationError(
+                section,
+                original_exception,
+            )
+
+            self.app.make_apport_report(
```
dbungert, line 72: By default, autoinstall errors should not be apport reports.  This sort of flaw should move to manual reporting - we have this problem today were invalid data can trigger the bug flow and we don't want that.  Most invalid autoinstall is a user configuration error.  Select "invalid" autoinstall is really a bug, so we want to rely on manual user reporting for that case.

subiquity/server/autoinstall.py
Chris-Peterson444: You're right. I don't think any of the logic here in `AutoinstallError` is necessary. Mostly remnant of old ideas. Any special handling should be done by the subclass (like `__str__` in `AutoinstallValidationError).

subiquity/server/autoinstall.py
```diff
@@ -0,0 +1,50 @@
...
+        self,
+        message: str,
+    ):
+        super().__init__(message)
+        self.message = message
+
+    def __repr__(self):
+        return f"<{self.__class__.__name__}: {self.message!r}>"
+
+    def __str__(self):
```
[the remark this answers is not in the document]
Chris-Peterson444: Agreed, as above.

subiquity/server/controller.py
Chris-Peterson444: This is true. I'll flip this around and add some special handling in the `SubiquityServer` exception handler to not write out apport reports on instances of `AutoinstallError`.

Chris-Peterson444:
I simplified the new exception types a bit and added handling to explicitly not generate an apport report when failing on an instance of `AutoinstallError`.

subiquity/server/server.py
```diff
@@ -396,7 +404,8 @@ def make_apport_report(self, kind, thing, *, wait=False, **kw):
         return self.error_reporter.make_apport_report(kind, thing, wait=wait, **kw)
 
     async def _run_error_cmds(self, report):
-        await report._info_task
+        if hasattr(report, "_info_task"):
```
dbungert, line 407: I'd like you to double-check this line.
`ErrorReport` unconditionally creates the _info_task attribute, so it should either be None or an asyncio task.   So it's always there, but may be None, and `await None` is bad.  Should this be `if report._info_task is not None` instead?

dbungert (commented):
One item to review then I think this is ready.

subiquity/server/server.py
Chris-Peterson444: I didn't see the default value for the attribute, so your recommendation is right if `report` is of type `ErrorReport`. Although in the case of `AutoinstallError` exceptions I'm passing `report=None`. I was trying to catch both cases here, but I should have been more explicit.

My assumption is that the `ErrorReport` class is about generating apport reports based on the exception, but here we don't want to do that.

What do you think of rewriting as:
```python
async def _run_error_cmds(self, report: Optional[ErrorReport] = None):
    if report is not None and report._info_task is not None:
        await report._info_task
     ...
```

dbungert: Ah, report=None would also have been hosed (`getattr(None, ...)` is also bad.)
I expect that ErrorReport is the correct type here, so your latest suggestion looks correct.

dbungert (approved):

PR #1898, opened 2024-01-29
doc: use literalinclude for autoinstall-schema

Chris-Peterson444:
Instead of manually changing the autoinstall-schema reference, use literalinclude to insert the autoinstall-schema.json file from the root of the repository.


dbungert (approved):
great find

PR #1899, opened 2024-01-29
Enforce top level keys in autoinstall user data

Chris-Peterson444:
Subiquity has, until now, silently ignored invalid top-level keys.
This has likely been a major source of confusion of autoinstall
capabilities.

The following autoinstall config:

    version: 1
    interative-sections:
        - identity
    literally-anything: lmao

will now throw an AutoinstallValidationError.

### DIFF
--- a/subiquity/server/autoinstall.py
+++ b/subiquity/server/autoinstall.py
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import logging
+from typing import Optional
 
 log = logging.getLogger("subiquity.server.autoinstall")
 
@@ -24,8 +25,13 @@ class AutoinstallError(Exception):
 class AutoinstallValidationError(AutoinstallError):
     def __init__(
         self,
-        owner: str,
+        section: str,
+        message: Optional[str] = None,
     ):
-        self.message = f"Malformed autoinstall in {owner!r} section"
-        self.owner = owner
+        if not message:
+            self.message: str = f"Malformed autoinstall in {section!r} section"
+        else:
+            self.message: str = message
+
+        self.section: str = section
         super().__init__(self.message)
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import asyncio
+import copy
 import logging
 import os
 import sys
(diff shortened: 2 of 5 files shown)


Chris-Peterson444:
https://github.com/canonical/subiquity/pull/1897 is a prerequisite for this PR. Once that is resolved, the only commit of interest here should be the latest (9375b1fe42e76e0d36e302937ba2dc04df578da2)

Chris-Peterson444:
Additionally, should we consider this behavior change a breaking change? Since the interface hasn't changed, and I think the silent failures are more a bug than a feature (even though it was explicit), I think not. Opinions?

subiquity/server/server.py
```diff
@@ -466,6 +476,79 @@ async def apply_autoinstall_config(self, context):
...
+                message="top-level key 'version' is missing",
+            )
+
+        if version <= 1:
+            for key in bad_keys:
+                log.warning(f"Unrecognized top-level autoinstall key {key!r}")
+
+            log.warning(
+                "Unrecognized keys may cause autoinstall to crash in future versions"
+            )
```
Chris-Peterson444, line 514: Currently all this will do is print with `WARNING` to the log file. Do we see a need to use something like `warning.warn` to print to stderr instead? (Or stdout where the main server process is run)

Chris-Peterson444:
Force pushed w/ updates to change from a run time error to a warning for now.

subiquity/server/autoinstall.py
```diff
@@ -24,8 +25,13 @@ class AutoinstallError(Exception):
 class AutoinstallValidationError(AutoinstallError):
     def __init__(
         self,
-        owner: str,
+        section: str,
+        message: Optional[str] = None,
     ):
-        self.message = f"Malformed autoinstall in {owner!r} section"
-        self.owner = owner
+        if not message:
```
dbungert, line 31: A standard across Subiquity is to prefer "is not None" over this case.  I suspect that's a little better here.

subiquity/server/server.py
```diff
@@ -475,14 +476,73 @@ async def apply_autoinstall_config(self, context):
...
+        # If version early enough, only warn
+        version: int = base_config.get("version", None)
+
+        if not version:
+            raise AutoinstallValidationError(
+                section="top-level keys",
+                message="top-level key 'version' is missing",
+            )
+
+        if version <= 1:
```
dbungert, line 508: if we're adding v1/v2 behavior in this PR, then I want to see the autoinstall-reference updated at the same time.

subiquity/server/server.py
dbungert: > crash

I don't love this word, as the word `crash` will be seen to mean a defect. 

> stderr

No, that's not quite right.  We need two cases to show the fault
* interactive - we need to show the results in a dialog that indicates that the autoinstall is bad.  The existing crash dialog is a good clue, but we won't reuse it directly because this is a non-defect case.
* non-interactive - output needs to go to the same place as the rest of the UI, which may be on a serial console etc.  Follow the existing context logging for that.

Also warn is the wrong mindset in that if we're supposed to be autoinstalling and we have stopped, then that is fatal to the program.

> may cause

We have a distinct plan, it will cause, so let's set that text accordingly.

Chris-Peterson444:
Closing, superseded by https://github.com/canonical/subiquity/pull/1947

PR #1901, opened 2024-01-30
Update validate-autoinstall-user-data script

Chris-Peterson444:
./scripts/validate-autoinstall-user-data is used by the integration tests to verify the written user data is correct, but we can additionally advertise this script as something for users to pre-validate their autoinstall data. 

Changes include some refactoring and trying to catch some likely common mistakes.

Todo:
- [x] Write how-to page in documentation
- ~~[ ] Tests~~

### DIFF
--- a/doc/howto/autoinstall-validation.rst
+++ b/doc/howto/autoinstall-validation.rst
@@ -1,0 +1,249 @@
+.. _autoinstall_validation:
+
+Autoinstall Validation
+=====================================
+
+The following how-to guide demonstrates how to perform pre-validation of a autoinstall config.
+
+Autoinstall config is validated against a :doc:`JSON schema <../reference/autoinstall-schema>` during runtime before it is applied. This check ensures existence of required keys and their data types, but does not guarantee the validity of the data provided (e.g., a bad :ref:`match directive <disk_selection_extensions>`). Additionally, the following validation script is unable to replicate some of the cloud-config based :ref:`delivery checks <how_the_delivery_is_verified>`. There are some basic checks performed to catch simple delivery-related errors, which you can read more about in the examples section, but the focus of the validation is on the Autoinstall configuration *after* it has been delivered to the installer.
+
+.. note::
+   See the cloud-init documentation for `how to validate your cloud-config`_.
+
+
+Pre-validating the Autoinstall configuration
+--------------------------------------------
+
+You can validate autoinstall config prior to install time by using the `validate-autoinstall-user-data script <https://github.com/canonical/subiquity/blob/main/scripts/validate-autoinstall-user-data.py>`_ in the Subiquity GitHub repository.
+
+Getting Started
+^^^^^^^^^^^^^^^
+
+Running the validation script requires downloading the Subiquity source code and installing the development dependencies. First, clone the Subiquity repository and ``cd`` into the root of the repository:
+
+.. code:: none
+
+   git clone https://github.com/canonical/subiquity.git && cd subiquity/
+
+Then the required dependencies can be installed by running:
+
+.. code:: none
+
+   make install_deps
+
+
+Now you can invoke the validation script with:
+
+.. code:: none
+
+   ./scripts/validate-autoinstall-user-data <path-to-config>
+
+
+or you can feed the configuration data via stdin:
+
+
+.. code:: none
+
+   # a trivial example
+   cat <config> | ./scripts/validate-autoinstall-user-data
+
+.. warning::
+
+   Never run the validation script as ``sudo``.
+
+Finally, after running the validation script it will report the result of the validation attempt:
+
+.. code:: none
+
+   $ ./scripts/validate-autoinstall-user-data.py <path-to-config>
+   Success: The provided autoinstall config validated successfully
+
+You can also use the exit codes to determine the result: 0 (success) or 1 (failure).
+
+
+Choice of Delivery Method
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default the validation script will expect your autoinstall configuration to be passed via cloud-config and expects a valid cloud-config file containing an ``autoinstall`` section:
+
+.. code:: none
+
+   #cloud-config
+
+   # some cloud-init directives
+
+   autoinstall:
+      # autoinstall directives
+
+This allows you to use the script directly on your cloud-config data. The validation script will extract the autoinstall configuration from the provided cloud-config data and perform the validation on the extracted autoinstall section directly.
+
+
+If you want to validate autoinstall configurations which will be delivered via the installation media, like the following example:
+
+.. code:: none
+
+   autoinstall:
+      # autoinstall directives
+
+then this can be signalled by passing the ``--no-expect-cloudconfig`` flag. Both formats in this delivery method, with or without a top-level ``autoinstall`` keyword, are supported in this mode.
+
+
+------------
+
+Examples
+--------
+
+Common mistake #1
+^^^^^^^^^^^^^^^^^
+
+If a top level ``autoinstall`` keyword is not found in the provided cloud-config during runtime then the installer will miss the autoinstall config and present an interactive session. To prevent occurrences of this issue, the validation script will report a failure if the provided cloud-config does not contain an autoinstall section. *This does not indicate a crash at runtime*, as you can definitely provide cloud-config without autoinstall, but it is a useful result for checking a common formatting mistake.
+
+.. tabs::
+
+   .. tab:: Validation output
+
+
+      Validating cloud-config which is missing the ``autoinstall`` keyword:
+
+      .. code:: none
+
+         $ ./scripts/validate-autoinstall-user-data.py <path-to-config>
+         AssertionError: Expected data to be wrapped in cloud-config but could not find top level 'autoinstall' key.
+         Failure: The provided autoinstall config did not validate successfully
+
+   .. tab:: Faulty config
+
+      As an example, the following cloud-config contains an autoinstall section but has misspelled the ``autoinstall`` keyword:
+
+      .. code:: none
+
+         #cloud-config
+         autoinstll:
+            # autoinstall directives
+
+
+Common Mistake #2
+^^^^^^^^^^^^^^^^^
+
+Another common mistake is to forget the ``#cloud-config`` header in the cloud-config file, which will result in the installer "missing" the autoinstall configuration.
+
+.. tabs::
+
+   .. tab:: Validation output
+
+      The validator will fail the provided cloud-config data if it does not contain the right header:
+
+
+      .. code:: none
+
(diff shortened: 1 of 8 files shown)


scripts/validate-autoinstall-user-data.py
```diff
@@ -29,56 +29,117 @@
...
+        dest="expect_cloudconfig",
+        action="store_false",
+        help="Assume the data is not wrapped in cloud-config.",
+        default=True,
+    )
+    parser.add_argument(
+        "--check-link",
+        dest="check_link",
+        action="store_true",
+        help="Assert the documentation link is in the user data",
```
dbungert, line 86: I suggest help=argparse.SUPPRESS, as this is not something I expect users to be concerned about and may instead just be confusing.  I wouldn't want people to think that this is a requirement of their autoinstall data.  Please replace with a comment explaining that we just want the link in autoinstall-user-data.

Chris-Peterson444: Good point, done. Thanks!

Chris-Peterson444:
I think this is ready. It's not perfect but it's a good start. I think I would like to punt writing tests for this since it uses the dry-run logic that is already tested. Eventually I think we could use this in CI for some more testing, which at that point we could write some more tests for this particular script if we wanted. 

Additionally, I would like to package this (and subiquity as a whole) in a way that users could just install the subiquity snap and invoke the validator from there, which would reduce the setup here.

scripts/validate-autoinstall-user-data.py
```diff
@@ -27,64 +27,224 @@
...
+        ai_data: str = yaml.dump(ai_dict)
+    else:
+        ai_data = str_data
 
-    # Verify autoinstall schema
-    user_data.seek(stream_pos)
+    with tempfile.TemporaryDirectory() as td:
+        path = Path(td) / "autoinstall.yaml"
+        with open(path, "w") as tf:
+            tf.write(ai_data)
```
ogayot, line 240: suggestion: 

```python
path.write_text(ai_data)
```

https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text

scripts/validate-autoinstall-user-data.py
```diff
@@ -27,64 +27,224 @@
...
+    # the documentation in the auto-generated autoinstall file post-install
+    parser.add_argument(
+        "--check-link",
+        dest="check_link",
+        action="store_true",
+        help=argparse.SUPPRESS,
+        default=False,
+    )
+
+    args: Namespace = parser.parse_args()
```
ogayot, line 118: return parser.parse_args() ?

scripts/validate-autoinstall-user-data.py
```diff
@@ -27,64 +27,224 @@
...
 
-    # Verify autoinstall schema
-    user_data.seek(stream_pos)
+    with tempfile.TemporaryDirectory() as td:
+        path = Path(td) / "autoinstall.yaml"
+        with open(path, "w") as tf:
+            tf.write(ai_data)
 
-    data = yaml.safe_load(user_data)
+        ret_code = asyncio.run(
```
ogayot, line 242: return asyncio.run(...) ?

scripts/validate-autoinstall-user-data.py
```diff
@@ -27,64 +27,224 @@
...
+            raise AssertionError("Documentation link missing from user data")
+
+    # Parse out the autoinstall if expected within cloud-config
+    if args.expect_cloudconfig:
+        try:
+            ai_dict: dict[str, Any] = parse_cloud_config(str_data)
+        except Exception as exc:
+            print(f"{type(exc).__name__}: {exc}")
+            print(FAILURE_MSG)
+            sys.exit(1)
```
ogayot, line 231: return 1?

doc/howto/autoinstall-validation.rst
```diff
@@ -0,0 +1,249 @@
...
+You can validate autoinstall config prior to install time by using the `validate-autoinstall-user-data script <https://github.com/canonical/subiquity/blob/main/scripts/validate-autoinstall-user-data.py>`_ in the Subiquity GitHub repository.
+
+Getting Started
+^^^^^^^^^^^^^^^
+
+Running the validation script requires downloading the Subiquity source code and installing the development dependencies. First, clone the Subiquity repository and ``cd`` into the root of the repository:
+
+.. code:: none
+
+   git clone git@github.com:canonical/subiquity.git && cd subiquity/
```
ogayot, line 26: I'd recommend cloning with the https protocol, rather than SSH ; so it works for everyone.

doc/howto/autoinstall-validation.rst
```diff
@@ -0,0 +1,249 @@
...
+   start: subiquity/Reporting/load_autoinstall_data:
+   finish: subiquity/Reporting/load_autoinstall_data: SUCCESS:
+   start: subiquity/Error/load_autoinstall_data:
+   finish: subiquity/Error/load_autoinstall_data: SUCCESS:
+   start: subiquity/core_validation:
+   finish: subiquity/core_validation: FAIL: Malformed autoinstall in 'version or interactive-sections' section
+   finish: subiquity/load_autoinstall_config: FAIL: Malformed autoinstall in 'version or interactive-sections' section
+   Malformed autoinstall in 'version or interactive-sections' section
+   Traceback (most recent call last):
+     File "/home/cpete/software/subiquity/scripts/../subiquity/server/server.py", line 654, in validate_autoinstall
```
ogayot, line 218: I'd typically drop the "/home/cpete/software/subiquity" prefix from the tracebacks but it's up to you

scripts/validate-autoinstall-user-data.py
```diff
@@ -27,64 +27,224 @@
...
+    # Do both validation phases
+    try:
+        app.load_autoinstall_config(only_early=True, context=None)
+        app.load_autoinstall_config(only_early=False, context=None)
+    except Exception as exc:
 
-    # Verify autoinstall doc link is in the file
+        print(exc)  # Has the useful error message
+
+        # Print the full traceback if verbosity >=2
```
ogayot, line 196: > 2

scripts/validate-autoinstall-user-data.py
```diff
@@ -27,64 +27,224 @@
...
+sys.path.insert(1, str(curtin_root))
+sys.path.insert(2, str(probert_root))
+
+from subiquity.cmd.server import make_server_args_parser  # noqa: E402
+from subiquity.server.dryrun import DRConfig  # noqa: E402
+from subiquity.server.server import SubiquityServer  # noqa: E402
+
+
+SUCCESS_MSG = "Success: The provided autoinstall config validated succesfully"
+FAILURE_MSG = "Failure: The provided autoinstall config did not validate succesfully"
```
ogayot, line 59: nitpick: I'd reword just a bit so that the last words written are not "validate successfully". I'm a lazy reader :laughing: 

The provided autoinstall config failed validation?

scripts/validate-autoinstall-user-data.py
```diff
@@ -27,64 +27,224 @@
...
+# Python path trickery so we can import subiquity code and still call this
+# script without using the makefile
+scripts_dir = sys.path[0]
+subiquity_root = Path(scripts_dir) / ".."
+curtin_root = subiquity_root / "curtin"
+probert_root = subiquity_root / "probert"
+# At the very least, local curtin needs to be in the front of the python path
+sys.path.insert(0, str(subiquity_root))
+sys.path.insert(1, str(curtin_root))
+sys.path.insert(2, str(probert_root))
```
ogayot, line 51: I don't love this but I don't have any better suggestions, so :laughing:

scripts/validate-autoinstall-user-data.py
ogayot: Also, I'm tempted to suggest `rich` (https://rich.readthedocs.io/en/stable/introduction.html) but it's probably difficult to justify for only one message.

ogayot (commented):
Looks nice overall. A few comments though.

scripts/validate-autoinstall-user-data.py
Chris-Peterson444: TIL. Thanks for pointing that out!

doc/howto/autoinstall-validation.rst
Chris-Peterson444: Good catch, I hadn't thought about that. I replaced the real prefix with ellipses. Does that seem better?

scripts/validate-autoinstall-user-data.py
Chris-Peterson444: > lazy reader

Me too. I couldn't come up with something at the time but I like it. I used your suggested text. 

> I'm tempted to suggest `rich`

I do like the idea. And it's in Main, which is good. Although maybe once (if) we start having more error messaging we could use it?

scripts/validate-autoinstall-user-data.py
Chris-Peterson444: I agree, it's not great. I can't yet think of any less worse alternatives, except for packaging it into the snap somehow.

Chris-Peterson444:
Updated + rebased onto main.

Thanks for the feedback @ogayot

scripts/validate-autoinstall-user-data.py
ogayot: yeah, let's think about it later. This is not blocking.

ogayot (commented):
LGTM once we can make the CI happy.

Chris-Peterson444:
Finally coming back to this. A couple of things:

1. I've refactored the code a bit and split the changes into smaller logical
commits. Hopefully this makes re-review a little easier if necessary.

2. Changing the CI to rely on the new behavior isn't going to work right away.
It's uncovering quite a few things we should eventually fix, but I don't think
we need to do that all in this PR. I'm proposing a hidden `--legacy` flag we
can use in CI to rely on the old JSON schema direct validation behavior and
over time we can work out the fixes to make CI happy with the dry-run validation.
What do you think?

dbungert (approved):
An interesting challenge with the run-time verifier is that it has assumptions that may not match the target installation ISO.
Examples, which can be valid but will fail validation:
* missing identity / user-data - would be valid on Desktop
* source asking for anything but `id: synthesized` - might be valid, if the id matches install-sources.yaml on the ISO

I feel like the validator is still useful but I think it would be worth mentioning these limitations, if we haven't already.

Chris-Peterson444:
Great point. This is one of the issues I was trying to avoid in CI for now and we should definitely document it. How does this look? I added a new section "Validator Limitations" and tried to address all of the things I might consider as limitations of the validator (mostly rearranging the disclaimers I had put in the intro previously), including the points you provided.

dbungert (approved):

PR #1902, opened 2024-02-02
cloudinit: handle format=json status

dbungert:
Add version checks for cloud-init to know if we can read status in JSON format or not.  If so, use that for a superior answer to the legacy format.  Handle legacy code paths also.

### DIFF
--- a/subiquity/cloudinit.py
+++ b/subiquity/cloudinit.py
@@ -1,7 +1,12 @@
 """Shared cloudinit utility functions"""
 
+import asyncio
 import json
 import logging
+import re
+from typing import Optional
+
+from subiquitycore.utils import arun_command, run_command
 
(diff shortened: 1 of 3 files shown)


Chris-Peterson444 (approved):
LGTM. I would complain about the comparing strings of numbers but I don't think we'll get to triple digits for a long time!

dbungert:
> LGTM. I would complain about the comparing strings of numbers but I don't think we'll get to triple digits for a long time!

I have added a test of the sort I have been alluding to recently, I'm going to call it a "canary" test.  If the cloud-init version style changed - which I think unlikely, given the history, it should trip.

Chris-Peterson444 (approved):
> "canary" test

Seems like a pretty good canary to me. Thanks for the extra test!

PR #1903, opened 2024-02-05
subiquitycore/log: use 'root' as group for strictly confined snaps

bboozzoo:
When setting up the logging in a strictly confined snap, use the 'root' group, rather than 'adm'. This will not interfere with the sandbox's policy but also does not result in providing wider access to the logs.

Supersedes #1874


dbungert (approved):
LGTM, but I'm going to wait for the tests to be happy before I press the merge button (even though I expect no issues).

kubiko:
@bboozzoo nice one! way more elegant than my version :) thank you!

dbungert:
Merging, thanks!

bboozzoo:
Thanks for landing it!

PR #1904, opened 2024-02-06
console-conf: set project name, consistently use as state directory

bboozzoo:
Set a console-conf as project name, which ensures that the early app setup creates a relevant state directory under /run. Update code where /run/console-conf was hardcoded to use the app level helpers to derive the correct state directory location.

The intention is to supersede #1869

### DIFF
--- a/console_conf/controllers/tests/test_identity.py
+++ b/console_conf/controllers/tests/test_identity.py
@@ -1,0 +1,58 @@
+# Copyright 2024 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


console_conf/controllers/tests/test_identity.py
```diff
@@ -0,0 +1,58 @@
...
+
+            c = IdentityController(app)
+            c.identity_done("foo@bar.com")
+            run_command.assert_called_with(
+                ["snap", "create-user", "--sudoer", "--json", "foo@bar.com"]
+            )
+
+            with open(os.path.join(statedir, "login-details.txt")) as inf:
+                data = inf.read()
+            assert "Ubuntu Core 24 on 1.2.3.4 (tty1)\n" in data
```
dbungert, line 58: Minor detail, but I will enforce for consistency.  We prefer the methods like `self.assertIn`.

dbungert (commented):
One tweak then this will be good for merge.

console_conf/controllers/tests/test_identity.py
bboozzoo: Thanks, I've pushed an update.

dbungert (approved):

PR #1910, opened 2024-02-13
Do not use hard-coded EOL year for ESM updates on the ubuntu-pro screen

ogayot:
The Ubuntu Pro screens showed 2032 as the EOL year of ESM updates. This is only true when installing jammy. On other LTSes, this value is different.

Showing the wrong value is misleading for users and give them a bad impression.

We now lean on distro-info to find the EOL year of ESM updates. If we cannot find out, we write 20xx instead of hardcoding the value.

The number of packages in the archive is still hardcoded (2300 for main ; 23000 for universe) but this is less of a problem.


mwhudson (approved):
Nice, thanks

PR #1911, opened 2024-02-14
permissions: update netplan configs to 0600

Chris-Peterson444:
Since netplan 0.106.1, Netplan YAMLs should have file permissions with mode 0o600 (owner RW only) and root owner.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,7 +31,7 @@ jobs:
       matrix:
         image:
           - ubuntu-daily:jammy   # match the core snap we're running against
-          - ubuntu-daily:noble   # latest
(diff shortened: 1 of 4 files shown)


Chris-Peterson444:
These configs are already locked down when written by [`_write_config`](https://github.com/canonical/subiquity/blob/ce938e6d033a2b019056c0d5ff6a1ef93c513609/subiquitycore/controllers/network.py#L262) due to the changes in `write_file` in https://github.com/canonical/subiquity/commit/80b144f220fc874263313328ab0e68462beb7b48 but the permissions in the files returned by the NetworkController's `render` method also need to be updated.

dbungert (approved):
Some sort of setup problem for the lint check?  LGTM otherwise.

PR #1914, opened 2024-02-15
networking: fix typo in BondConfig to_config method

Chris-Peterson444:
A mismatch between the keynames in BondConfig's to_config method and NetworkDev's netdev_info function was causing subiquity to crash when creating a bond with a valid transmit hash policy and then later trying to edit it. (LP: #2051586)

### DIFF
--- a/examples/answers/bond.yaml
+++ b/examples/answers/bond.yaml
@@ -9,6 +9,18 @@ Keyboard:
   layout: us
 Network:
   actions:
+    # LP: #2051586 the UI would crash when editing a 802.3ad bond
+    - action: create-bond
+      data:
+        name: bond10
+        devices:
+          - [interface index 0]
+        mode: 802.3ad
+    - obj: [interface name bond10]
+      action: EDIT_BOND
+    - obj: [interface name bond10]
+      action: DELETE
+
     - action: create-bond
       data:
         name: bond10
--- a/subiquitycore/models/network.py
+++ b/subiquitycore/models/network.py
@@ -234,7 +234,7 @@ class NetworkDev:
             bond = BondConfig(
                 interfaces=self.config["interfaces"],
                 mode=params["mode"],
-                xmit_hash_policy=params.get("xmit-hash-policy"),
+                xmit_hash_policy=params.get("transmit-hash-policy"),
                 lacp_rate=params.get("lacp-rate"),
             )
         vlan: Optional[VLANConfig] = None
--- a/subiquitycore/models/tests/test_network.py
+++ b/subiquitycore/models/tests/test_network.py
@@ -15,8 +15,9 @@
 
 from unittest.mock import Mock
 
-from subiquitycore.models.network import NetworkDev
+from subiquitycore.models.network import BondConfig, BondParameters, NetworkDev
 from subiquitycore.tests import SubiTestCase
+from subiquitycore.tests.parameterized import parameterized
 
 
 class TestRouteManagement(SubiTestCase):
@@ -44,6 +45,9 @@ class TestRouteManagement(SubiTestCase):
 
 
 class TestNetworkDev(SubiTestCase):
+    def setUp(self):
+        self.model = Mock(get_all_netdevs=Mock(return_value=[]))
+
     def test_netdev_info_eth_inexistent(self):
         # LP: #2012659 - just after physically removing an Ethernet interface
         # from the system, Subiquity tries to collect information via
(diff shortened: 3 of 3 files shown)


ogayot (approved):
LGTM! Good work

I suggest adding a test for that in `examples/answers/bond.yaml` since we have support for it.

Something like that:
```diff
--- a/examples/answers/bond.yaml
+++ b/examples/answers/bond.yaml
@@ -9,6 +9,16 @@ Keyboard:
   layout: us
 Network:
   actions:
+ # LP: #2051586 the UI would crash when editing a 802.3ad bond
+    - action: create-bond
+      data:
+        name: bond10
+        devices:
+          - [interface index 0]
+        mode: 802.3ad
+    - obj: [interface name bond10]
+      action: EDIT_BOND
+    - obj: [interface name bond10]
+      action: DELETE
+
```

Chris-Peterson444:
Thanks and good idea! I'm working on this now, but I'm hitting [a crash](https://github.com/Chris-Peterson444/subiquity/actions/runs/7933072816/job/21661007889) in the jammy image tests. Perhaps this caught another bug!

dbungert (approved):

dbungert:
Uh, let's get CI happy but otherwise LGTM

Chris-Peterson444:
Thank you both!

@ogayot So the [jammy image doesn't like these types of bonds](https://github.com/Chris-Peterson444/subiquity/actions/runs/7937514236/job/21674836980), but the others images are fine? I printed the logs in the CI runner and it looks like the call to `netplan generate --root <root>` isn't returning correctly. I can't reproduce this locally on the same image.

What do you think of waiting to add the CI tests until we can figure out what's breaking?

dbungert:
> What do you think of waiting to add the CI tests until we can figure out what's breaking?

That sounds like a bug that the tests have helped to find.
Letting this PR through still sounds like improvement?
Can we skiptest based on series?

Chris-Peterson444:
>Letting this PR through still sounds like improvement?
Agreed.

>That sounds like a bug that the tests have helped to find.
Potentially, but I can't tell why it's not reproducible on my local lxc runner. Could you test it for me?

If you don't see anything then yeah we could just skiptest it

ogayot (changes requested):
I had a look at the CI failure on jammy. `netplan generate` failed with:

```
2024-02-19 08:39:46,082 WARNING subiquitycore.controllers.network:334 netplan generate stderr: /tmp/tmp.EY5ZztTuwE/etc/netplan/00-installer-config.yaml:10:9:
Error in network definition: unknown key 'xmit-hash-policy'
        xmit-hash-policy: layer2
```

and I can't find any reference to `xmlt-hash-policy` in netplan's source code. There is `xmit_hash_policy` (with underscores) and `transmit-hash-policy`. It makes me question the fix is correct but I might be overlooking something.

Can you double check? If it's another bug and not a regression, I'm happy for us to click merge.

BTW, on my end I could reproduce the error easily. I just went through dry-run and created the bond interface with 802.3ad. However, you won't get a crash dialog. You have to look into the subiquity server logs to see it.

ogayot:
I looked more into why the CI only fails in jammy. I believe it's because:
* we only run `netplan generate` in the network screen if `/lib/netplan/generate` exists
* on mantic/noble, one should use `/usr/libexec/netplan/generate` instead of `/lib/netplan/generate` (the latter does not exist anymore). We could also use `/usr/libexec/netplan/generate` on jammy but not on focal.

https://github.com/canonical/subiquity/blob/477fef959095151a2beffd4a4eb3b6b252e768b1/subiquitycore/controllers/network.py#L318-L330

Nowadays, 
canonical/netplan@0e2a8ffc787e4649f21618939607b1649a1279de

Chris-Peterson444:
Thanks for looking into this @ogayot! I was able to reproduce this thanks to your help. 

Since the key name conflict was only between the one pair, I just chose one to stick with and rename the other. It seems I just so happened to rename it the wrong way to produce the error on Jammy! 

I looked into which of `transmit-hash-policy` and `xmit_hash_policy` that netplan is happy with and `transmit-hash-policy` seems to be the only one accepted in pure netplan config. (It seems `xmit_hash_policy` is acceptable, but just not in the format that we write out).

I rebased onto your commit from #1915 to show CI is happy with the combined fixes, but I can rebase onto main once it's merged.

ogayot (approved):
This seems much better now. Thanks!

Feel free to rebase and merge.

dbungert:
We (myself included) really should use that "Rebase and merge" option more often in the merge dropdown.

PR #1915, opened 2024-02-19
network: fix detection of netplan on recent Ubuntu releases

ogayot:
Subiquity has a mechanism to detect the presence of netplan. It does so by checking the existence of the file `/lib/netplan/generate`. This mechanism is used in the network screen to validate the YAML configuration.

However, since netplan 0.107 (present in mantic and noble), the file `/lib/netplan/generate` is no longer present. It used to be provided as an alias for `/usr/libexec/netplan/generate` ; starting in jammy.

This made Subiquity unable to detect that netplan is running ; and therefore skip the YAML validation against netplan.

Since we support focal in Subiquity, let's change the detection code so that we look for both locations. When we stop supporting Focal in the future, we can drop the reference to `/lib/netplan/generate`.

### DIFF
--- a/subiquitycore/controllers/network.py
+++ b/subiquitycore/controllers/network.py
@@ -18,6 +18,7 @@ import asyncio
 import contextlib
 import logging
 import os
+import pathlib
 import subprocess
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


dbungert (approved):

Chris-Peterson444 (approved):
Great find!

PR #1917, opened 2024-02-21
types: rename ZFS_LUKS -> ZFS_LUKS_KEYSTORE

dbungert:
Clarify that the implemented mechanism is the keystore setup.  An implementation that doesn't use native ZFS encryption is plausibly interesting, which is a better use of the term ZFS_LUKS.

### DIFF
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -347,7 +347,7 @@ class GuidedCapability(enum.Enum):
     LVM = enum.auto()
     LVM_LUKS = enum.auto()
     ZFS = enum.auto()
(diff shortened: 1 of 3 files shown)


ogayot (approved):
This looks harmless. Does this change mandate an update in ubuntu desktop installer?

dbungert:
> This looks harmless. Does this change mandate an update in ubuntu desktop installer?

Doesn't seem so yet.  I just pushed this flag recently before realizing I wanted to rename it.  The backend for this is in flight so using the flag isn't super helpful yet, aside for my own testing purposes and landing parts of this as I can.

PR #1921, opened 2024-02-29
filesystem: accept to place the rootfs on remote storage for NVMe-o-TCP

ogayot:
Now that curtin more or less supports NVMe over TCP with the rootfs on remote storage, relax the contraints set by subiquity.

We still need the /boot (and /boot/efi) partitions on local storage though.

We probably want to display a warning somewhere "Please note that NVMe over TCP is experimental and in active development" for instance.

Depends on https://code.launchpad.net/~curtin-dev/curtin/+git/curtin/+merge/461452

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -70,7 +70,7 @@ parts:
 
     source: https://git.launchpad.net/curtin
     source-type: git
-    source-commit: 237053d9d18916dd72cf861280474d4df0e9fd24
+    source-commit: 5094d95fa5b8506412979e21548069bc12681dc7
 
     override-pull: |
       craftctl default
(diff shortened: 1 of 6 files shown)


snapcraft.yaml
```diff
@@ -70,7 +70,7 @@ parts:
 
     source: https://git.launchpad.net/curtin
     source-type: git
-    source-commit: 237053d9d18916dd72cf861280474d4df0e9fd24
+    source-commit: 79d831871d32e396326106b5f05ba144e61badaa
```
ogayot, line 73: Update when the relevant curtin MP gets merged.

dbungert (approved):

snapcraft.yaml
dbungert: Is what you need merged at this point?

ogayot: I just hit merge for the curtin part. I originally wanted to wait for the FFe decision but we can still exclude the changes from the stable branch in case it gets rejected.

PR #1923, opened 2024-03-01
bug(cloudinit): do not call cloudinit stages.Init when disabled

blackboxsw:
Inspect cloud-init status for disabled. When disabled do not call stages.Init due to a side-effect that cloud-init tries to rediscover any viable datasource when none are previously discovered.

LP: #2055077


dbungert (approved):
I had expected that unit test to be less painful, thanks for doing so.

PR #1925, opened 2024-03-04
Support for Non Reportable Errors

Chris-Peterson444:
The original attempt to make a new class of non apport reportable errors in #1897 didn't quite cover everything needed, especially in regards to the changes needed in the API to inform the client of the nuance of the error state. These changes include:

- A refactor to make the apport generation suppression generic and tested
- Changes to the /meta/status API response object (`ApplicationStatus`) to include a new field `nonreportable_error`
  - Reportable errors are still available under the `error` field, so disambiguating the error difference requires checking which of `error` and `nonreportable_error` isn't `None`
- The endpoints `/meta/restart` (POST) and `/meta/ssh_info` (GET) are now available before the server has finished loading all of its controllers. 
- The server now sets the `interactive` state as the first step in loading the autoinstall data, not the last
- The subiquity TUI can discern between reportable and non-reportable errors and will show a different overlay accordingly

### DIFF
--- a/subiquity/client/controllers/progress.py
+++ b/subiquity/client/controllers/progress.py
@@ -15,11 +15,12 @@
 
 import asyncio
 import logging
+from typing import Optional
 
 import aiohttp
 
 from subiquity.client.controller import SubiquityTuiController
-from subiquity.common.types import ApplicationState, ShutdownMode
+from subiquity.common.types import ApplicationState, ApplicationStatus, ShutdownMode
 from subiquity.ui.views.installprogress import InstallRunning, ProgressView
 from subiquitycore.async_helpers import run_bg_task
 from subiquitycore.context import with_context
@@ -32,6 +33,7 @@ class ProgressController(SubiquityTuiController):
         super().__init__(app)
         self.progress_view = ProgressView(self)
         self.app_state = None
+        self.has_nonreportable_error: Optional[bool] = None
         self.crash_report_ref = None
         self.answers = app.answers.get("InstallProgress", {})
 
@@ -75,16 +77,14 @@ class ProgressController(SubiquityTuiController):
                 await asyncio.sleep(1)
                 continue
             self.app_state = app_status.state
+            self.has_nonreportable_error = app_status.nonreportable_error is not None
 
             self.progress_view.update_for_state(self.app_state)
             if self.ui.body is self.progress_view:
                 self.ui.set_header(self.progress_view.title)
 
-            if app_status.error is not None:
-                if self.crash_report_ref is None:
-                    self.crash_report_ref = app_status.error
-                    self.ui.set_body(self.progress_view)
-                    self.app.show_error_report(self.crash_report_ref)
+            if self.app_state == ApplicationState.ERROR:
+                self._handle_error_state(app_status)
 
             if self.app_state == ApplicationState.NEEDS_CONFIRMATION:
                 if self.showing:
@@ -104,6 +104,22 @@ class ProgressController(SubiquityTuiController):
             if self.app_state == ApplicationState.DONE:
                 if self.answers.get("reboot", False):
                     self.click_reboot()
+
+    def _handle_error_state(self, app_status: ApplicationStatus):
+        if (error := app_status.error) is not None:
+            if self.crash_report_ref is None:
+                self.crash_report_ref = error
+                self.ui.set_body(self.progress_view)
+                self.app.show_error_report(error)
+
+        elif (error := app_status.nonreportable_error) is not None:
+            self.ui.set_body(self.progress_view)
+            self.app.show_nonreportable_error(error)
+
+        else:
+            # There is the case that both are None but still in an error state
+            # but this is likely a bug so raise an Exception
+            raise Exception("Server in ERROR state but no error received")
 
     def make_ui(self):
         if self.app_state == ApplicationState.NEEDS_CONFIRMATION:
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -121,11 +121,11 @@ class API:
                 """Confirm that the installation should proceed."""
 
         class restart:
+            @allowed_before_start
             def POST() -> None:
                 """Restart the server process."""
 
         class ssh_info:
-            @allowed_before_start
             def GET() -> Optional[LiveSessionSSHInfo]:
                 ...
 
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -125,6 +125,7 @@ class API:
                 """Restart the server process."""
 
         class ssh_info:
+            @allowed_before_start
             def GET() -> Optional[LiveSessionSSHInfo]:
                 ...
 
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -24,6 +24,7 @@ from typing import Any, Dict, List, Optional, Union
 
 import attr
 
+from subiquity.server.nonreportable import NonReportableException
 from subiquitycore.models.network import NetDevInfo
 
 
@@ -53,6 +54,21 @@ class ErrorReportRef:
     kind: ErrorReportKind
     seen: bool
     oops_id: Optional[str]
+
+
+@attr.s(auto_attribs=True)
+class NonReportableError:
+    cause: str
+    message: str
+    details: Optional[str]
+
+    @classmethod
+    def from_exception(cls, exc: NonReportableException):
+        return cls(
+            cause=type(exc).__name__,
+            message=str(exc),
+            details=exc.details,
+        )
 
 
 class ApplicationState(enum.Enum):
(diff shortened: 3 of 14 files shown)


Chris-Peterson444:
>The endpoints /meta/restart (POST) and /meta/ssh_info (GET) are now available before the server has finished loading all of its controllers.

I wanted to expand on this a bit but without cluttering up the summary. This is motivated by the fact that exceptions thrown during the autoinstall loading/validation phase necessarily means the controllers won't be fully loaded. For the restart endpoint I think this is probably fine (you should be able to restart whenever you want I suppose), but the ssh_info endpoint is a little hairy. As long as the client waits until after the `CLOUD_INIT_WAIT` state, then this change should be fine. In the case of autoinstall errors, this is fine since we don't start loading the autoinstall until after we've got the password information from cloud-init. 

I suppose one way to ensure this behavior would be to create some state variable to make the ssh_info function blocking until cloudinit has finished?

subiquity/common/types.py
```diff
@@ -87,6 +102,7 @@ class ApplicationStatus:
     state: ApplicationState
     confirming_tty: str
     error: Optional[ErrorReportRef]
+    nonreportable_error: Optional[NonReportableError]
     cloud_init_ok: Optional[bool]
     interactive: Optional[bool]
     echo_syslog_id: str
```
Chris-Peterson444, line 108: @d-loose I am proposing to change the response object for `/meta/status` to include a new field `nonreportable_error`, to represent errors subiquity doesn't automatically generate apport reports for. In the case that `state` is `ApplicationState.ERROR`, if the error results in an apport report then `error` will be populated with the reportref as before; otherwise, it will be null/none and `nonreportable_error` will be filled in with a `NonReportableError`. Does this seem okay from the u-d-b side of things? The one thing I'm concerned about is if there's any spots that always expect `error` to be not empty when `state == ApplicationState.ERROR`.

d-loose: Thanks for the heads-up! I don't see any problems with this on our side. Currently any `ApplicationState.ERROR` is treated as fatal to the installation process. The `error` isn't being used anywhere so far.

@spydon FYI

subiquity/client/controllers/progress.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, app):
         super().__init__(app)
         self.progress_view = ProgressView(self)
         self.app_state = None
+        self.error_is_reportable: bool = True
```
ogayot, line 35: Should this be None instead of True when there is no error?

ogayot (commented):
Good work!

subiquity/server/server.py
```diff
@@ -272,6 +274,9 @@ class SubiquityServer(Application):
 
     supported_variants = ["server", "desktop"]
 
+    # Classes of exceptions to suppress creation of apport reports for
+    apport_suppressed_exceptions: tuple[Exception] = (AutoinstallError,)
```
dbungert, line 278: What's your thoughts of handling this through the inheritance chain?  So adding a `NonReportableException` or similar and have `AutoinstallError` inherit that and then we can just `isinstance(foo, NonReportableException)` ?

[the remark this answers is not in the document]
Chris-Peterson444: Agreed, this always gets me. Thanks!

subiquity/client/controllers/progress.py
Chris-Peterson444: When I added this, my thought process was that the behavior before was effectively if this was always True so I wanted to avoid any issues where ProgressView.update_for_state is called and error_is_reportable wasn't set correctly.

Although now I'm thinking you're right, setting it to None seems better. I think we could change the name to error_is_not_reportable, which will be slightly harder to reason about, but at least None will evaluate to False. What do you think?

examples/autoinstall/bad-early-command.yaml
dbungert, line 1: As discussed elsewhere, just adding a note here to ask that we start moving this sort of data out of the examples directory.

subiquity/ui/views/error.py
```diff
@@ -421,3 +421,90 @@ def _report_changed(self, report):
...
+        self.error: NonReportableError = error
+
+        self.btns: dict[str, AttrMap] = {
+            "close": close_btn(self, _("Close")),
+            "debug_shell": other_btn(_("Switch to a shell"), on_press=self.debug_shell),
+            "restart": other_btn(_("Restart the installer"), on_press=self.restart),
+        }
+        # Get max button width and create even button sizes
+        width: int = 0
+        for _name, button in self.btns.items():
```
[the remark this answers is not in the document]
dbungert:
```
width = max((widget_width(button) for button in self.btns.values()))
```

subiquity/server/server.py
```diff
@@ -495,9 +514,11 @@ def load_autoinstall_config(self, *, only_early):
             self.autoinstall,
         )
         if not self.autoinstall:
+            self.interactive = True
```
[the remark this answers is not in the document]
Chris-Peterson444: Yeah I would be surprised if the state of `interactive` changed between invocations of the function, but I think it's better to be flexible about it than not. I'll update with a comment.

subiquity/ui/views/error.py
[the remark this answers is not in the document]
Chris-Peterson444: Even better, thanks!

subiquity/common/apidef.py
```diff
@@ -121,10 +121,12 @@ def POST(tty: str) -> None:
                 """Confirm that the installation should proceed."""
 
         class restart:
+            @allowed_before_start
```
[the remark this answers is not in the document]
Chris-Peterson444: Yeah to enable the new dialog flow. Crashing before finishing controller setup means we'll be blocked on most endpoints. 

The only instance of `meta.restart.POST` in the client is the one this change is fixing, and yes no other dependencies.

There's a couple points where `meta.ssh_info.GET` is called, which does the require two things:
- The cloud init stage to finish. 
  -  Which it looks like the clients, the server client at least, wait to fully start until after this stage. So password info from cloud init should be set
- The network model to be configured. 
  - Looking into this one actually it looks like the network model won't be fully configured by time it's called in the error state. Essentially making it useless I think. What if we add some logic to detect the early fail case and load the network controller, otherwise block until it's configured?

subiquity/server/server.py
Chris-Peterson444: I think it's better, but early on I wasn't sure about it yet because I was still playing with the API structure. I'm a little more confident in it now so I'll make the changes.

subiquity/client/controllers/progress.py
Chris-Peterson444: I went ahead and changed it to `error_is_not_reportable`, let me know what you think.

Chris-Peterson444:
Updated to address the feedback so far. The only thing left to do is figure out what to do about `ssh_info` during an early fail. Does it make sense to manually start up the network controller during an early fail? Or should the client detect `ssh=True` and `ApplicationState=ERROR` and print out that the installer stopped before the network could be configured?

Chris-Peterson444:
Updated to remove the special handling of the the `ssh_info` endpoint. The subiquity client just now skips getting the info if the server is in an error state.

dbungert (approved):
LGTM but please "s/interative-sessions/interactive-sections/" which was in a previous PR.

Chris-Peterson444:
Done, thanks!

subiquity/client/controllers/progress.py
ogayot: I think that's better yes. May I suggest `has_reportable_error`? Wdyt?

subiquity/client/controllers/progress.py
```diff
@@ -32,6 +33,7 @@ def __init__(self, app):
         super().__init__(app)
         self.progress_view = ProgressView(self)
         self.app_state = None
+        self.error_is_not_reportable: bool = Optional[None]
```
ogayot, line 36: uh-oh

```diff
-     self.error_is_not_reportable: bool = Optional[None]
+     self.error_is_not_reportable: Optional[bool] = None
```

ogayot (changes requested):
I'm happy with the changes now. Please fix the assignment of `error_is_not_reportable` and I will approve.

subiquity/client/controllers/progress.py
Chris-Peterson444: That reads much better! I slightly tweaked it to be `has_nonreportable_error` to keep the same truthiness though.

subiquity/client/controllers/progress.py
Chris-Peterson444: Heh whoops. Thanks for catching that!

subiquity/client/controllers/progress.py
ogayot: Excellent, thanks!

ogayot (approved):
LGTM!

PR #1927, opened 2024-03-06
debconf-selections: work around grub2 bug until it is fixed

ogayot:
Currently, installations of Ubuntu Server and Ubuntu Desktop 24.04 both fail on UEFI-based systems because of a grub2 bug. This is a workaround to avoid failing installs until grub2 migrates to the release pocket.


dbungert (approved):
Great idea.

ogayot:
@d-loose All 24.04 installs of Ubuntu Server & Desktop fail on UEFI based systems ATM. Do you think you could pick this workaround up in `ubuntu-desktop-bootstrap` so we can stop the bleeding?

d-loose:
Thanks, will do!

PR #1929, opened 2024-03-07
Add ubuntu-image reference

rkratky:
## Background

In order to make `ubuntu-image` documentation available to users (considering that a) `ubuntu-image` has no online presence other than its GH repo, and b) following the transition to packaging `ubuntu-image` only as a Snap, it doesn't have a man page), I'd like to add basic reference info about it to the Installation Guide.

While at this point, the IG contains only Autoinstall/Subiquity docs, the plan is to expand it to include other kinds of installation-related content. This makes `ubuntu-image` docs a good candidate for addition.

## Content

This PR only adds a single reference page with basic info about the `ubuntu-image` tool and its command-line options. It's derived from the erstwhile `ubuntu-image` man page, but the plan is to add other reference and how-to content relating to the tool (mainly its YAML config file schema reference and some practical examples).

Fixes [FR-6783](https://warthogs.atlassian.net/browse/FR-6783).

[FR-6783]: https://warthogs.atlassian.net/browse/FR-6783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/reference/ubuntu-image.rst
+++ b/doc/reference/ubuntu-image.rst
@@ -1,0 +1,246 @@
+ubuntu-image syntax and options
+===============================
+
+``ubuntu-image`` is a program for generating bootable disk images. It supports building snap_-based and classical preinstalled Ubuntu images.
+
+
+Snap-based images
+-----------------
+
+Snap-based images are built from a *model assertion*, which is a YAML file describing a particular combination of core, kernel, and gadget snaps, along with other declarations, signed with a digital signature asserting its authenticity.  The assets defined in the model assertion uniquely describe the device for which the image is built.
+
+
+Classical images
+----------------
+
+Classical images are built from a local `gadget tree`_ path.  The `gadget tree`_ is a primed `gadget snap`_ containing a ``gadget.yaml`` file in the ``meta`` directory and all the necessary bootloader gadget bits built. For instance, a `gadget tree`_ can be prepared by fetching a specially tailored `gadget snap`_ source and running ``snapcraft prime`` on it, with the resulting tree being stored in the ``prime/`` directory.
+
+
+The ``gadget.yaml`` file
+------------------------
+
+As part of the model assertion, a `gadget snap`_ is specified.  The gadget contains a ``gadget.yaml`` file, which contains the exact description of the disk-image contents in the YAML format.  The ``gadget.yaml`` file describes, among other things:
+
+* Names of all the volumes to be produced (volumes are roughly analogous to disk images).
+
+* Structures (structures define the layout of the volume, including partitions, Master Boot Records, or any other relevant content.) within the volume. :none:`wokeignore:rule=master,`
+
+* Whether the volume contains a bootloader and if so, what kind of bootloader.
+
+Note that ``ubuntu-image`` communicates with the Snap Store using the ``snap prepare-image`` sub-command.  The model-assertion file is passed to ``snap prepare-image``, which handles downloading the appropriate gadget and any extra snaps.  See the :manualpage:`snap(8) manual page <man8/snap.8.html>` for additional details.
+
+
+Basic syntax
+------------
+
+.. code-block:: yaml
+
+    ubuntu-image snap [options] model.assertion
+
+    ubuntu-image classic [options] image_definition.yaml
+
+
+General options
+~~~~~~~~~~~~~~~
+
+-h, --help
+    Show the help message and exit.
+
+--version
+    Show the program version number and exit.
+
+
+Options of the ``snap`` command
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+These are the options for defining the contents of snap-based images.  They can only be used when the ``ubuntu-image snap`` command is used.
+
+``model_assertion``
+    Path to the model assertion file.  This positional argument must be given for this mode of operation.
+
+--cloud-init USER-DATA-FILE
+    ``cloud-config`` data to be copied to the image.
+
+--disable-console-conf
+    Disable ``console-conf`` on the resulting image.
+
+--factory-image
+    Hint that the image is meant to boot in a device factory.
+
+--validation=<ignore|enforce>
+    Control whether validations should be ignored or enforced.
+
+--snap SNAP
+    Install an extra snap.  This is passed through to ``snap prepare-image``. The snap argument can include additional information about the channel and/or risk with the following syntax: ``<snap>=<channel|risk>``. Note that this flag will cause an error if the model assertion has a grade higher than dangerous.
+
+--revision <SNAP_NAME:REVISION>
+    Install a specific revision of a snap rather than the latest available in a particular channel. The snap specified with ``SNAP_NAME`` must be included either in the model assertion or as an argument to ``--snap``. If both a revision and channel are provided, the revision specified is installed in the image, and updates come from the specified channel.
+
+--preseed
+    Preseed the image (Ubuntu Core 20 and higher only).
+
+--preseed-sign-key=<key>
+    Name of the key to use to sign the preseed assertion, otherwise use the default key.
+
+--sysfs-overlay=<path to directory that contains sysfs overlay>
+    Specify the directory that contains the sysfs overlay. This options also requires the ``--preseed`` and ``--preseed-sign-key`` options.
+
+Options of the ``classic`` command
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+These are the options for defining the contents of classical preinstalled Ubuntu images. They can only be used when the ``ubuntu-image classic`` command is used.
+
+``image_definition``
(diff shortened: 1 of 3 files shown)


dbungert (approved):
LGTM

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
...
+Snap-based images
+-----------------
+
+Snap-based images are built from a *model assertion*, which is a YAML file describing a particular combination of core, kernel, and gadget snaps, along with other declarations, signed with a digital signature asserting its authenticity.  The assets defined in the model assertion uniquely describe the device for which the image is built.
+
+
+Classical images
+----------------
+
+Classical images are built from a local `gadget tree`_ path.  The `gadget tree`_ is a primed `gadget snap`_ containing a ``gadget.yaml`` file in the ``meta`` directory and all the necessary bootloader gadget bits built. For instance, a `gadget tree`_ can be prepared by fetching a specially tailored `gadget snap`_ source and running ``snapcraft prime`` on it, with the resulting tree being stored in the ``prime/`` directory.
```
upils, line 16: We still can build the image this way (by providing a pre-built gadget tree) but now the favored way is to set the URL and branch of the gadget in the image definition and let ubuntu-image fetch and build it. So now it feels like "advanced" instructions that may let users think this is the default way to do things. What do you think about moving this at the bottom and replace with:


Classical images are built from an image definition YAML file, defining its content and resulting output format. Key elements of this configuration are:
- the gadget section, defining how to get the gadget required to build boot assets and partition layout.
- the rootfs section, defining parameters to build the content of the rootfs
- the artifacts section, defining the types of artifacts to create, including the actual images, manifest files, etc.

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
...
+--cloud-init USER-DATA-FILE
+    ``cloud-config`` data to be copied to the image.
+
+--disable-console-conf
+    Disable ``console-conf`` on the resulting image.
+
+--factory-image
+    Hint that the image is meant to boot in a device factory.
+
+--validation=<ignore|enforce>
```
upils, line 70: This is now a common option, so I think it can move to the common section. I will rework this when I have some time because this is confusing. The help message does not let the user know what validation this is about and what is the default value.

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
...
+    Control whether validations should be ignored or enforced.
+
+--snap SNAP
+    Install an extra snap.  This is passed through to ``snap prepare-image``. The snap argument can include additional information about the channel and/or risk with the following syntax: ``<snap>=<channel|risk>``. Note that this flag will cause an error if the model assertion has a grade higher than dangerous.
```
upils, line 74: This flag accepts a list of snaps.

And here also the `snap prepare-image` is now replaced.

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
...
+
+
+State-machine options
+~~~~~~~~~~~~~~~~~~~~~
+
+.. caution:: The options described here are primarily for debugging and testing purposes and should not be considered part of the stable, public API.  State-machine step numbers and names can change between releases.
+
+``ubuntu-image`` internally runs a state machine to create the disk image. These are some options for controlling this state machine.  Other than ``--workdir``, these options are mutually exclusive.  When ``--until`` or ``--thru`` is given, the state machine can be resumed later with ``--resume``, but ``--workdir`` must be given in that case since the state is saved in a ``ubuntu-image.json`` file in the working directory.
+
+-w DIRECTORY, --workdir DIRECTORY
```
upils, line 142: I think we could move this option to the common ones. Even though I guess it was initially created to enable the "resume" feature, it can be used for other purposes.

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
...
+
+Environment variables
+---------------------
+
+The following environment variables are recognised by ``ubuntu-image``.
+
+``UBUNTU_IMAGE_PRESERVE_UNPACK``
+    When set, the variable specifies the directory for preserving a pristine copy of the unpacked gadget contents.  The directory must exist, and an ``unpack`` directory will be created under this directory.  The full contents of the ``<workdir>/unpack`` directory after the ``snap prepare-image`` sub-command has run is copied here.
+
+There are a few other environment variables used for building and testing only.
```
upils, line 172: I checked and `UBUNTU_IMAGE_PRESERVE_UNPACK` is now the only one. So this can be removed.

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
+ubuntu-image syntax and options
+===============================
+
+``ubuntu-image`` is a program for generating bootable disk images. It supports building snap_-based and classical preinstalled Ubuntu images.
```
upils, line 4: Should we add a link somewhere on this page to the snap store page of the tool? Or to instructions on how to install it?

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
...
+* Names of all the volumes to be produced (volumes are roughly analogous to disk images).
+
+* Structures (structures define the layout of the volume, including partitions, Master Boot Records, or any other relevant content.) within the volume. :none:`wokeignore:rule=master,`
+
+* Whether the volume contains a bootloader and if so, what kind of bootloader.
+
+Note that ``ubuntu-image`` communicates with the Snap Store using the ``snap prepare-image`` sub-command.  The model-assertion file is passed to ``snap prepare-image``, which handles downloading the appropriate gadget and any extra snaps.  See the :manualpage:`snap(8) manual page <man8/snap.8.html>` for additional details.
+
+
+Basic syntax
```
upils, line 33: I do not know if this considered basic syntax but here we could probably mention that the tool must be run with sudo (or as root) to work properly. We do not check it when the execution starts but at some point in the build process, some underlying tools (debootstrap, etc.) will fail if not run as root.

upils (commented):
Thanks for this PR. Most of my comments are things you could not know about because the current doc is outdated.

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
...
+
+``UBUNTU_IMAGE_PRESERVE_UNPACK``
+    When set, the variable specifies the directory for preserving a pristine copy of the unpacked gadget contents.  The directory must exist, and an ``unpack`` directory will be created under this directory.  The full contents of the ``<workdir>/unpack`` directory after the ``snap prepare-image`` sub-command has run is copied here.
+
+There are a few other environment variables used for building and testing only.
+
+
+.. _list of steps:
+
+List of steps
```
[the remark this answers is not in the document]
rkratky: I'm removing the list of steps. When the `--dry-run` option is implemented, we can mention it in appropriate places.

Wouldn't it be useful to have an option to list all potential steps for reference?

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
...
+#. ``populate_bootfs_contents``
+#. ``populate_prepare_partitions``
+#. ``make_disk``
+#. ``generate_snap_manifest``
+
+
+Additional resources
+--------------------
+
+* :manualpage:`snap(8) manual page <man8/snap.8.html>`
```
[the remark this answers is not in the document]
rkratky: Good point.

doc/reference/ubuntu-image.rst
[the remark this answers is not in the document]
upils: I am not sure about that. Because I do not want users thinking "backward" and trying to hack their image definition to end up executing a specific step. For now we do not really make any promise regarding the order of steps or even the name and existence of steps between versions. This is an advanced feature and we do not expect most users using it. We are thinking about creating "meta-states" representing a high-level state (and regrouping several steps) that we could keep stable between versions, but we are not there yet.

So for now I think only listing steps that will execute for a given configuration is fine.

doc/reference/ubuntu-image.rst
```diff
@@ -0,0 +1,246 @@
+ubuntu-image syntax and options
+===============================
+
+``ubuntu-image`` is a program for generating bootable disk images. It supports building snap_-based and classical preinstalled Ubuntu images.
+
+
+Snap-based images
+-----------------
+
+Snap-based images are built from a *model assertion*, which is a YAML file describing a particular combination of core, kernel, and gadget snaps, along with other declarations, signed with a digital signature asserting its authenticity.  The assets defined in the model assertion uniquely describe the device for which the image is built.
```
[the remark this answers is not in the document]
rkratky: I'm adding the link to the paragraph, but I think it's useful to have a short description as an intro, too.

PR #1930, opened 2024-03-08
ubuntu-image docs fixes & updates

rkratky:
Follow up to #1929. Minor fixes, updates, and additions to the `ubuntu-image` reference doc.

### DIFF
--- a/doc/reference/ubuntu-image.rst
+++ b/doc/reference/ubuntu-image.rst
@@ -4,16 +4,37 @@ ubuntu-image syntax and options
 ``ubuntu-image`` is a program for generating bootable disk images. It supports building snap_-based and classical preinstalled Ubuntu images.
 
 
+Installing ``ubuntu-image``
+---------------------------
+
+The ``ubuntu-image`` tool is distributed as a snap package through the Snap Store: `ubuntu-image <https://snapcraft.io/ubuntu-image>`_. Install it either from a Snap-enabled software store provided by your distribution, or by running this command with ``sudo`` or as the ``root`` user:
+
+.. code-block::
+
+    snap install ubuntu-image --classic
+
+Note that you need have the snap daemon installed. See `Installing the daemon <https://snapcraft.io/docs/installing-snapd>`_ for instructions for your distribution.
+
+
 Snap-based images
 -----------------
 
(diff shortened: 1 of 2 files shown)


dbungert (approved):

rkratky:
> LGTM but it looks like you missed 3 comments from the previous review:

Not on purpose -- I didn't see them. I pbbly opened the review before you were finished, and I didn't reload. My bad. I'll add fixes for these, too.

rkratky:
@upils I added the remaining fixes based on yyour review. Apologies for having missed those before.

doc/reference/ubuntu-image.rst
```diff
@@ -4,16 +4,39 @@ ubuntu-image syntax and options
...
+``gadget``
+    Defines how to get the gadget required to build boot assets and the partition layout.
+
+``rootfs``
+    Defines the parameters to build the content of the rootfs.
+
+``artifacts``
+    Defines the types of artefacts to create, including the actual images, manifest files, and others.
+
+Alternatively, classical images can also be built from a local `gadget tree`_ path.  The `gadget tree`_ is a primed `gadget snap`_ containing a ``gadget.yaml`` file in the ``meta`` directory and all the necessary bootloader gadget bits built. For instance, a `gadget tree`_ can be prepared by fetching a specially tailored `gadget snap`_ source and running ``snapcraft prime`` on it, with the resulting tree being stored in the ``prime/`` directory.
```
upils, line 39: My previous comment on this was confusing. That is not an alternative: the image definition YAML file is mandatory to build a classic image. This paragraph describes how to build the gadget assets from a "gadget tree". This is only a part of what is built by ubuntu-image.

Maybe this paragraph would make more sense in the `gadget` section in the documentation of the image definition. As you can see in https://github.com/canonical/ubuntu-image/blob/main/internal/imagedefinition/README.rst, the user can choose a `type`. This paragraph is a semi-detailed explanation of what could be done to get a valid gadget tree if the user selected `type: prebuilt`. But since ubuntu-image now handles this process pretty well, most of our users should probably use `type: git` and let ubuntu-image do the work.

upils (changes requested):
LGTM aside from the small bit about the classic image build explanation.

doc/reference/ubuntu-image.rst
rkratky: > My previous comment on this was confusing. That is not an alternative: the image definition YAML file is mandatory to build a classic image.

Thank you. I need to find the time to actually use the tool, so that I don't just blindly blunder about... :/

I'll add the paragraph to the image-definition doc.

upils (approved):
LGTM

PR #1933, opened 2024-03-11
Update autoinstall-reference.rst

ben-ba:
Missing hypen for uri will result in an error, if it is the only value in list.

error: { 'uri': 'http://archive.ubuntu.com/ubuntu/' } is not of type 'array'

### DIFF
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -270,9 +270,9 @@ The default is:
         primary:
           - country-mirror
           - arches: [i386, amd64]
-            uri: "http://archive.ubuntu.com/ubuntu"
+          - uri: "http://archive.ubuntu.com/ubuntu"
           - arches: [s390x, arm64, armhf, powerpc, ppc64el, riscv64]
-            uri: "http://ports.ubuntu.com/ubuntu-ports"
+          - uri: "http://ports.ubuntu.com/ubuntu-ports"
       fallback: abort
       geoip: true
 


doc/reference/autoinstall-reference.rst
```diff
@@ -270,9 +270,9 @@ The default is:
         primary:
           - country-mirror
           - arches: [i386, amd64]
-            uri: "http://archive.ubuntu.com/ubuntu"
+          - uri: "http://archive.ubuntu.com/ubuntu"
```
ogayot, line 273: The change does not look right to me. The key called "primary" expects an array of elements, where each element can either be "country-mirror" or an object with a URI + optionally a set of architectures if applies to.

If we translate the YAML into JSON, we have:

```yaml
primary:
  - arches: [i386, amd64]
    uri: http://archive.ubuntu.com/ubuntu
```
equivalent to
```json
{
   "primary": [
        {
            "arches": ["i386", "amd64"],
            "uri": "http://archive.ubuntu.com/ubuntu"
        }
    ]
}
```
and
```yaml
primary:
  - arches: [i386, amd64]
  - uri: "http://archive.ubuntu.com/ubuntu"
```
equivalent to
```json
{
   "primary": [
        {
            "arches": ["i386", "amd64"]
        }, {
            "uri": "http://archive.ubuntu.com/ubuntu"
        }
    ]
}
```

ogayot (commented):
Hello and thank you for your contribution!

However, I think the change would introduce invalid autoinstall directives. A `arches` directive (without a corresponding `uri`) would be rejected by Subiquity.

Maybe there is something incorrect somewhere in the documentation?

doc/reference/autoinstall-reference.rst
ben-ba: 1. if you scroll down in the documentation you will see the uri already with an hyphen. -> https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#geoip
2. If i use arches + uri the hyphen for uri isn't necessary, but if i use only uri, the error mentioned get thrown.

my user-data looks like following for apt;
working one
`  apt:
    mirror-selection:
      primary:
         - uri: "http://archive.ubuntu.com/ubuntu/"
    fallback: offline-install`

failed one, error see 1 comment
``  apt:
    mirror-selection:
      primary:
         uri: "http://archive.ubuntu.com/ubuntu/"
    fallback: offline-install``

ben-ba: btw primary is an array of country-mirror and an object with two properties; uri and arches

so each list needs an hyphen, so when uri is needed, the dokumentation should switch to the following;

`primary:
          - country-mirror
          - uri: "http://archive.ubuntu.com/ubuntu"
            arches: [i386, amd64]
`

Chris-Peterson444: I hadn't refreshed and didn't see your comment, apologies! I also had to look at the code for this one. Perhaps we should update the docs to better reflect what's expected.

ogayot: @ben-ba I'd be happy to accept this as a doc improvement. Would you like to update this PR with the switched `arches` / `uri` statements?

ogayot: We'd also need you to sign the SLA before merging the change. We have some instructions on how to do so:

https://github.com/canonical/subiquity/blob/main/CONTRIBUTING.md#contributor-license-agreement

ben-ba: PR updated, CLA signed.

ben-ba: I would be really helpfully to read something about the validate of a autoinstall configuration with 
`./scripts/validate-autoinstall-user-data.py ../autoinstall/simple/user-data --no-expect-cloudconfig`
(comment out the autoinstall in user-data before)

Chris-Peterson444 (commented):
I'm okay with the change, but I'll wait for another agree to mark approved. 

Also, it looks like the CLA checker isn't happy. Could you confirm the email / github account you used to sign it is the same as this one? If it is, it may just be a little slow to update so we can try again later.

ben-ba:
E-Mail address is the same, but maybe it's an issue, because it is hidden in my github account?!

ben-ba:
its weird or confusing... 

Let me explane what my goal is.
My goal is to have a simple autoinstall setup with all (that means include security <- important) updates installed from an local apt mirror.
To reach this, i started with autoinstall or be exact, with subiquity.
After testing some configs, my finale experience is the following.

1. do never ever mix up curtin style and subiquity style of config for apt. 
2. an optional subiquity parameter like arches isn't an optional parameter for curtin
3. docs are never perfect like https://github.com/canonical/curtin/blob/54cc7d980b9300339e773dc579daa5f11fd12fd0/examples/apt-source.yaml#L100 (missing hypen, missing arches key...)

whats the final suggestions from an 'user' perspective;
make it clear in the docs, that the key security won't work if you use mirror-selection or maybe other keys (not tested)

finally here an overview of different configs which are not working - not really clear by documentation or/and no error with the validation script (./scripts/validate-autoinstall-user-data.py ../autoinstall/simple/user-data)

------attempt 1------
no validation error, error when cloud-init is loading -> finish: subiquity/Mirror/load_autoinstall_data: argument of type 'NoneType' is not iterable.
```
  apt:
    preserve_sources_list: false
    primary:
      - uri: http://linuxmirror.local/ubuntu
```

------attempt 2------
no validation error, subiquity follows the own recommendation (see (https://wiki.ubuntu.com/SecurityTeam/FAQ#What_repositories_and_pockets_should_I_use_to_make_sure_my_systems_are_up_to_date.3F and (https://www.phoronix.com/news/Ubuntu-Security-Pocket-Issue) to not edit the security update source, but remember, we use curtin configuration inside subiquity, subiquity never overwrite security update source, but curtin says, 
>  # security is optional, if not defined it is set to the same value as primary
so my expectation was, security would be also linuxmirror...

```
  apt:
    preserve_sources_list: false
    primary:
      - uri: http://linuxmirror.local/ubuntu
        arches: [ default ]
```

------attempt 3------
curtin key security mixing with subiquity mirror-selection, no verification error, not working. but wait, arrrr we forget the arches key for curtins security... but we get a new error, autoinstall process starts, and then the installer crashed... 


```
  apt:
    preserve_sources_list: false
    mirror-selection:
      primary:
      - uri: http://linuxmirror.local/ubuntu
    security:
      - uri: http://linuxmirror.local/ubuntu
        arches: [ default ]
    fallback: offline-install
```

------attempt 4------
curtin key security mixing with subiquity mirror-selection, no verification error, working

```
  apt:
    preserve_sources_list: false
    mirror-selection:
      primary:
      - uri: http://linuxmirror.local/ubuntu
    security:
      - uri: http://linuxmirror.local/ubuntu
        arches: [ default ]
    fallback: offline-install
```

------attempt 5------
plain curtin config, no verification error, working
```
  apt:
    preserve_sources_list: false
    primary:
      - uri: http://linuxmirror.local/ubuntu
        arches: [ default ]
    security:
      - uri: http://linuxmirror.local/ubuntu
        arches: [ default ]
    fallback: offline-install
```

Chris-Peterson444:
A few things:

- For the CLA, I read through the checker code again it's trying to match your membership to https://github.com/CanonicalContributorAgreement. There may be an issue the automation has with adding you to the organization if the email address you provided is hidden. Perhaps the simplest way to to get around this it to make a launchpad account and resubmit the form with your launchpad ID, which is checked as a backup. The launchpad account has to be associated with the email you provide in the CLA form and also has to be the same as the commit author.

- I see there are two commits. One with the original suggestion and one with the new. Could you squash the commits down to one with just the new effective changes?

- For the rest of your comments: Looking at the existing code, it doesn't look like we support `security` under the mirror-selection key. Perhaps we can be more explicit on what the behavior difference is from when `primary` is and isn't under that key. Please feel free to provide suggestions if you have any to make the docs clearer. Additionally, I am unsure why attempt 1 and 3 are failing. I would need to dig into the code a little more but in general if you believe you are running into an error where you shouldn't be, please feel free to submit a bug here https://bugs.launchpad.net/subiquity/+filebug and we can try to diagnose the issue a little better there with logs from `/var/log/installer/`.

PR #1940, opened 2024-03-14
kvm-test: add --secure-boot option

Chris-Peterson444:
Adds an option to configure the VM to load SecureBoot enabled firmware. SecureBoot is a feature of UEFI, which means --boot (for legacy BIOS boot) is incompatible with this flag. Providing both will result in an error.

This will be useful for testing MOK enrollment behavior, as on reboot you can confirm if you are presented with mokmanager or regular first boot.

### DIFF
--- a/scripts/kvm-test.py
+++ b/scripts/kvm-test.py
@@ -17,6 +17,7 @@ import dataclasses
 import os
 import pathlib
 import shlex
+import shutil
 import socket
 import subprocess
 import sys
@@ -99,6 +100,11 @@ class Context:
         self.iso = f'/tmp/kvm-test/{self.release}-test.iso'
         self.hostname = f'{self.release}-test'
         self.target = f'/tmp/kvm-test/{self.hostname}.img'
+        parent = pathlib.Path(self.target).parent
+        self.ovmf = {
+                'CODE': parent / f'{self.hostname}_OVMF_CODE_4M_ms.fd',
+                'VARS': parent / f'{self.hostname}_OVMF_VARS_4M_ms.fd'
+        }
         self.password = salted_crypt('ubuntu')
         self.cloudconfig = f'''\
 #cloud-config
@@ -168,8 +174,13 @@ parser.add_argument('--basesnap', default=None, action='store',
                     help='use slimy-update-snap on this snap')
 parser.add_argument('--snap', default=None, action='store',
                     help='inject this snap into the ISO')
-parser.add_argument('-B', '--bios', action='store_true', default=False,
+# Boot Options
+boot_group = parser.add_mutually_exclusive_group()
+boot_group.add_argument('-B', '--bios', action='store_true', default=False,
                     help='boot in BIOS mode (default mode is UEFI)')
+boot_group.add_argument('--secure-boot', action='store_true', default=False,
+                    help='Use SecureBoot', dest="secureboot")
+
 parser.add_argument('-c', '--channel', action='store',
                     help='build iso with snap from channel')
 parser.add_argument('-d', '--disksize', help='size of disk to create')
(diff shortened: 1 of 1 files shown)


Chris-Peterson444:
These are the changes I made that I've been using for testing out the MOK enrollment things. This only provides the secureboot bit. I don't think there's a way (or at least an easy one) for providing a fake device to the VM for tricking ubunut-drivers.

I imagine this will be useful for testing mok enrollment via Autoinstall where it technically doesn't necessarily matter if we install the drivers or not, just if a MOK was successfully enrolled. Although I've been using the following late-commands snippet to install broadcom drivers to verify dkms behavior

```yaml
  late-commands:
    - curtin in-target -- sh -c "DEBIAN_FRONTEND=noninteractive DEBIAN_HAS_FRONTEND= apt-get --no-upgrade install broadcom-sta-dkms -y </dev/null"
```

scripts/kvm-test.py
```diff
@@ -618,6 +647,11 @@ def boot(ctx):
 
     with kvm_prepare_common(ctx) as kvm:
         kvm.extend(drive(target))
+        if ctx.args.secureboot:
+            if not ctx.ovmf["VARS"].exists():
+                raise Exception(f"Couldn't find firmware variables file {str(ctx.ovmf['VARS'])!r}")
+            if not ctx.ovmf["CODE"].exists():
+                raise Exception(f"Couldn't find firmware code file {str(ctx.ovmf['VARS'])!r}")
```
ogayot, line 654: s/VARS/CODE/

scripts/kvm-test.py
```diff
@@ -252,6 +260,9 @@ def parse_args():
                         'LIVEFS_EDITOR to it\n'
                         'https://github.com/mwhudson/livefs-editor')
 
+    if ctx.args.bios and ctx.args.secureboot:
+        raise Exception("--secure-boot and --bios are mutually exclusive")
```
ogayot, line 264: Consider using argparse's native handling of exclusive groups?

```diff
- parser.add_argument('-B', '--bios', action='store_true', default=False,
-                    help='boot in BIOS mode (default mode is UEFI)')
- parser.add_argument('--secure-boot', action='store_true', default=False,
-                     help='Use SecureBoot', dest="secureboot")
+ boot_group = parser.add_mutually_exclusive_group()
+ boot_group.add_argument('-B', '--bios', action='store_true', default=False,
+                        help='boot in BIOS mode (default mode is UEFI)')
+ boot_group.add_argument('--secure-boot', action='store_true', default=False,
+                         help='Use SecureBoot', dest="secureboot")

File: scripts/kvm-test.py [hunk 3]
```diff
@@ -99,6 +100,11 @@ def __init__(self, args):
         self.iso = f'/tmp/kvm-test/{self.release}-test.iso'
         self.hostname = f'{self.release}-test'
         self.target = f'/tmp/kvm-test/{self.hostname}.img'
+        parent = pathlib.Path(self.target).parent
+        self.ovmf = {
+                'CODE': parent / 'OVMF_CODE_4M_ms.fd',
```
ogayot, line 105: Maybe prefix with the hostname in case we're spinning two vms at the same time?

scripts/kvm-test.py
```diff
@@ -475,7 +486,17 @@ def tpm(emulator: Optional[TPMEmulator]) -> List[str]:
...
+        ret = []
+    elif ctx.args.secureboot:
+        # Speical setup for a secureboot virtual machine
+        # https://wiki.debian.org/SecureBoot/VirtualMachine
+        ret = ['-machine',  'q35,smm=on',
+               '-global', 'driver=cfi.pflash01,property=secure,value=on',
+               '-drive', f'if=pflash,format=raw,unit=0,file={ctx.ovmf["CODE"]}',
+               '-drive', f'if=pflash,format=raw,unit=1,file={ctx.ovmf["VARS"]}'
+               ]
+    elif not ctx.args.bios:
```
ogayot, line 499: `else`?

scripts/kvm-test.py
```diff
@@ -475,7 +486,17 @@ def tpm(emulator: Optional[TPMEmulator]) -> List[str]:
 def bios(ctx):
     ret = []
```
ogayot, line 487: You can drop this line (or use `pass` / `return []` in the body of the `if`)

ogayot (commented):
LGTM I few minor suggestions.

scripts/kvm-test.py
Chris-Peterson444: Ah good catch, thanks!

scripts/kvm-test.py
Chris-Peterson444: TIL another thing about argparse. Thanks for this!

scripts/kvm-test.py
Chris-Peterson444: Ah good point. I hadn't thought of doing that. I wonder how confusing that would have been.

Chris-Peterson444:
Updated with feedback

ogayot (approved):

PR #1947, opened 2024-03-19
Enforce top-level keys

Chris-Peterson444:
Subiquity currently ignores invalid top-level keys, but this has likely been a major source of confusion of autoinstall capabilities.

In future versions, the following autoinstall config will throw an AutoinstallValidationError:
```yaml
version: 2
interactive-sections:
  - identity
 literally-anything: lmao
```
 This patch adds warnings for version 1 and will begin to throw an AutoinstallValidationError on these instances in version 2 once it has been enabled.

### DIFF
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -14,11 +14,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import asyncio
+import copy
 import logging
 import os
 import sys
 import time
-from typing import List, Optional
+from typing import Any, List, Optional
 
 import jsonschema
 import yaml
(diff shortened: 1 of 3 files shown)


Chris-Peterson444:
Supersedes https://github.com/canonical/subiquity/pull/1899 and requires  https://github.com/canonical/subiquity/pull/1945

subiquity/server/server.py
```diff
@@ -552,8 +553,93 @@ async def apply_autoinstall_config(self, context):
...
+        # If the version is early enough, only warn
+        version: int = autoinstall_config["version"]
+
+        ctx = context
+        if version == 1:
+            # Warn then clean out bad keys and return
+
+            for key in bad_keys:
+                warning = f"Unrecognized top-level key {key!r}"
+                log.warning(warning)
```
dbungert, line 608: as mentioned elsewhere the double `log` / `ctx` feels a bit tedious, please refactor a bit in a future PR.

subiquity/server/server.py
```diff
@@ -552,8 +553,93 @@ async def apply_autoinstall_config(self, context):
             await controller.apply_autoinstall_config()
             await controller.configured()
 
+    def _strip_controller_keys(
+        self,
+        autoinstall_config: dict[str, Any],
+    ) -> dict[str, Any]:
+        """Removes all unkown keys from autoinstall config."""
```
dbungert, line 560: Comment is inverted, since we're returning autoinstall with known controllers removed, we're actually returning the unknowns

subiquity/server/server.py
```diff
@@ -552,8 +553,93 @@ async def apply_autoinstall_config(self, context):
...
+        Only checks for unrecognized keys, doesn't validate them.
+        Requires a version number, so this should be called after
+        validating against the schema.
+
+        """
+
+        server_config: dict[str, Any] = self._strip_controller_keys(autoinstall_config)
+        valid_keys: set[str] = set(self.base_schema["properties"].keys())
+        top_level_keys: set[str] = set(server_config.keys())
+        bad_keys: set[str] = top_level_keys - valid_keys
```
dbungert, line 593: And because `server_config` is the unknowns, that means that this calculation of `bad_keys` works because it already has the unknowns stored in `top_level_keys` and we subtract any `valid_keys` (which won't be there) so we get the correct answer but in a bit of an obfuscated way.

subiquity/server/server.py
```diff
@@ -552,8 +553,93 @@ async def apply_autoinstall_config(self, context):
             await controller.apply_autoinstall_config()
             await controller.configured()
 
+    def _strip_controller_keys(
```
dbungert, line 556: I kind of want you to change how this function works, to have it return two dicts.
One dict of the known-good keys, one of the known-bad keys.
I think it would simplify the later operations, as
* you won't have to separately calculate bad_keys (this will tell you)
* the calculation of the good keys will be in the same spot (right now there are a few similar-looking things and the logic is spread over this function and the caller), and the calling function can just take action

subiquity/server/server.py
```diff
@@ -552,8 +553,93 @@ async def apply_autoinstall_config(self, context):
...
+        In Autoinstall v2, unknown top-level keys result in a fatal
+        AutoinstallValidationError
+
+        Only checks for unrecognized keys, doesn't validate them.
+        Requires a version number, so this should be called after
+        validating against the schema.
+
+        """
+
+        server_config: dict[str, Any] = self._strip_controller_keys(autoinstall_config)
```
dbungert, line 590: `server_config` is a bit generic, following the restructure of `_strip_controller_keys` maybe it could be `good_keys` / `bad_keys`?

subiquity/server/tests/test_server.py
```diff
@@ -232,6 +245,68 @@ async def test_autoinstall_validation__no_error_report(self):
...
+        self.pseudo_load_controllers()
+
+        # OK in Version 1 but ensure warnings and stripped config
+        self.server.validate_autoinstall()
+        log_mock.warning.assert_called()
+        log_mock.error.assert_not_called()
+        self.assertEqual(self.server.autoinstall_config, stripped_ai_data)
+
+        # Not OK in Version >= 2
+        log_mock.reset_mock()
```
dbungert, line 275: I guess I kind of don't trust `reset_mock` - it's tough to convince myself that everything that matters has been reset - are the later tests only passing because of retained state from earlier?  What do you think about splitting this test in two parts to cover your version=1 and version=2 cases?  Common setup logic could be refactored.

dbungert (changes requested):
Generally the plan is good but I'd like some restructuring.

Chris-Peterson444:
I've made updates per your suggestions @dbungert. I really liked your recommendation for the changes to the stripping function. The filtering function is really nice and made testing pretty simple. Thanks!

subiquity/server/server.py
```diff
@@ -491,8 +553,97 @@ async def apply_autoinstall_config(self, context):
             await controller.apply_autoinstall_config()
             await controller.configured()
 
+    def filter_autoinstall(
+        self,
+        autoinstall_config: dict[str, Any],
+    ) -> tuple[dict[str, Any], dict[str, Any]]:
+        """Separates autoinstall config by known and uknown keys"""
```
dbungert, line 560: typo

dbungert (commented):
LGTM with one typo fix when underlying PRs are resolved

subiquity/server/server.py
Chris-Peterson444: Fixed, thanks!

subiquity/server/tests/test_server.py
```diff
@@ -232,6 +262,70 @@ async def test_autoinstall_validation__no_error_report(self):
...
+        }
+
+        self.server.base_schema = SubiquityServer.base_schema
+        self.pseudo_load_controllers()
+
+        valid, invalid = self.server.filter_autoinstall(autoinstall_config)
+
+        real_valid_keys = set(("version", "interactive-sections", "apt"))
+        self.assertEqual(set(valid.keys()), real_valid_keys)
+        self.assertEqual(set(invalid.keys()), set(("some-bad-key",)))
```
dbungert, line 327: nitpick: 
```suggestion
        self.assertEqual(set(invalid.keys()), {"some-bad-key"})
```

subiquity/server/tests/test_server.py
```diff
@@ -232,6 +262,70 @@ async def test_autoinstall_validation__no_error_report(self):
...
+            "literally-anything": "lmao",
+        }
+
+        good, bad = self.load_config_and_controllers(bad_ai_data)
+
+        # OK in Version 1 but ensure warnings and stripped config
+        self.server.validate_autoinstall()
+        log_mock.warning.assert_called()
+        log_mock.error.assert_not_called()
+        self.assertEqual(self.server.autoinstall_config, good)
```
dbungert, line 282: minor: maybe assert empty here on bad, more for documentation sake than anything else

dbungert (approved):
I have two minor notes, please consider then it's good for merge I think.

subiquity/server/tests/test_server.py
Chris-Peterson444: Not quite empty in this case but I get what you mean. I added the check for the bad config.

dbungert (approved):

PR #1954, opened 2024-03-25
Cloud init schema failures

Chris-Peterson444:
Users attempting to do autoinstall may incorrectly send autoinstall directives as cloud-config, which will result in cloud-init schema validation errors. When loading autoinstall from cloud-config, we now check to see if there are any cloud-init schema validation errors and warn the user. Additionally, if the source of the error is from a known autoinstall error, we inform the user and halt the installation with a nonreportable AutoinstallError.

Requires #1945 and #1947.

### DIFF
--- a/subiquity/cloudinit.py
+++ b/subiquity/cloudinit.py
@@ -4,11 +4,30 @@ import asyncio
 import json
 import logging
 import re
+from collections.abc import Awaitable
+from subprocess import CompletedProcess
 from typing import Optional
 
+from subiquity.server.nonreportable import NonReportableException
 from subiquitycore.utils import arun_command, run_command
 
 log = logging.getLogger("subiquity.cloudinit")
+
+
+class CloudInitSchemaValidationError(NonReportableException):
+    """Exception for cloud config schema validation failure.
+
+    Attributes:
+        keys -- List of keys which are the cause of the failure
+    """
+
+    def __init__(
+        self,
+        keys: list[str],
+        message: str = "Cloud config schema failed to validate.",
+    ) -> None:
+        super().__init__(message)
+        self.keys = keys
 
 
 def get_host_combined_cloud_config() -> dict:
@@ -67,6 +86,34 @@ def read_legacy_status(stream):
     return None
 
 
+async def get_schema_failure_keys() -> list[str]:
+    """Retrieve the keys causing schema failure."""
+
+    cmd: list[str] = ["cloud-init", "schema", "--system"]
+    status_coro: Awaitable = arun_command(cmd, clean_locale=True)
+    try:
+        sp: CompletedProcess = await asyncio.wait_for(status_coro, 10)
+    except asyncio.TimeoutError:
+        log.warning("cloud-init schema --system timed out")
+        return []
+
+    error: str = sp.stderr  # Relies on arun_command decoding to utf-8 str by default
+
+    # Matches:
+    # ('some-key' was unexpected)
+    # ('some-key', 'another-key' were unexpected)
+    pattern = r"\((?P<args>'[^']+'(,\s'[^']+')*) (?:was|were) unexpected\)"
+    search_result = re.search(pattern, error)
+
+    if search_result is None:
+        return []
+
+    args_list: list[str] = search_result.group("args").split(", ")
+    no_quotes: list[str] = [arg.strip("'") for arg in args_list]
+
+    return no_quotes
+
+
 async def cloud_init_status_wait() -> (bool, Optional[str]):
     """Wait for cloud-init completion, and return if timeout ocurred and best
     available status information.
@@ -86,3 +133,23 @@ async def cloud_init_status_wait() -> (bool, Optional[str]):
     else:
         status = read_legacy_status(sp.stdout)
     return (True, status)
+
+
+async def validate_cloud_init_schema() -> None:
+    """Check for cloud-init schema errors.
+    Returns (None) if the cloud-config schmea validated OK according to
+    cloud-init. Otherwise, a CloudInitSchemaValidationError is thrown
+    which contains a list of the keys which failed to validate.
+    Requires cloud-init supporting recoverable errors and extended status.
+
+    :return: None if cloud-init schema validated succesfully.
+    :rtype: None
+    :raises CloudInitSchemaValidationError: If cloud-init schema did not validate
+            succesfully.
+    """
+    causes: list[str] = await get_schema_failure_keys()
+
+    if causes:
+        raise CloudInitSchemaValidationError(keys=causes)
+
+    return None
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -28,7 +28,12 @@ from cloudinit.config.cc_set_passwords import rand_user_password
 from jsonschema.exceptions import ValidationError
 from systemd import journal
 
-from subiquity.cloudinit import cloud_init_status_wait, get_host_combined_cloud_config
+from subiquity.cloudinit import (
+    CloudInitSchemaValidationError,
+    cloud_init_status_wait,
+    get_host_combined_cloud_config,
+    validate_cloud_init_schema,
+)
 from subiquity.common.api.server import bind, controller_for_request
 from subiquity.common.apidef import API
 from subiquity.common.errorreport import ErrorReport, ErrorReporter, ErrorReportKind
@@ -43,7 +48,7 @@ from subiquity.common.types import (
     PasswordKind,
 )
 from subiquity.models.subiquity import ModelNames, SubiquityModel
-from subiquity.server.autoinstall import AutoinstallValidationError
+from subiquity.server.autoinstall import AutoinstallError, AutoinstallValidationError
 from subiquity.server.controller import SubiquityController
 from subiquity.server.dryrun import DRConfig
 from subiquity.server.errors import ErrorController
@@ -723,7 +728,66 @@ class SubiquityServer(Application):
     def base_relative(self, path):
         return os.path.join(self.base_model.root, path)
 
-    def load_cloud_config(self):
+    @with_context(name="extract_autoinstall")
+    async def _extract_autoinstall_from_cloud_config(
+        self,
+        *,
+        cloud_cfg: dict[str, Any],
+        context: Context,
+    ) -> dict[str, Any]:
+        """Extract autoinstall passed via cloud config."""
+
+        # Not really is-install-context but set to force event reporting
+        context.set("is-install-context", True)
+        context.enter()  # publish start event
+
+        try:
+            await validate_cloud_init_schema()
+        except CloudInitSchemaValidationError as exc:
+            bad_keys: list[str] = exc.keys
+            raw_keys: list[str] = [f"{key!r}" for key in bad_keys]
+            context.warning(
+                f"cloud-init schema validation failure for: {', '.join(raw_keys)}",
+                log=log,
+            )
+
+            # Raise AutoinstallError if we found any autoinstall as a cause
+            # of the schema validation error, otherwise continue
+
+            # Filter only the bad keys
+            potential_autoinstall: dict[str, Any] = dict(
+                ((key, cloud_cfg[key]) for key in bad_keys)
+            )
+            autoinstall, other = self.filter_autoinstall(potential_autoinstall)
+
+            if len(autoinstall) != 0:
+                for key in autoinstall:
+                    context.error(
+                        message=(
+                            f"{key!r} is valid autoinstall but not "
+                            "found under 'autoinstall'."
+                        ),
+                        log=log,
+                    )
+
+                raise AutoinstallError(
+                    (
+                        "Misplaced autoinstall directives resulted in a cloud-init "
+                        "schema validation failure."
+                    )
+                ) from exc
+
+            else:
+                log.debug(
+                    "No autoinstall keys found among bad cloud config. Continuing."
+                )
+
+        cfg: dict[str, Any] = cloud_cfg.get("autoinstall", {})
+
+        return cfg
+
+    @with_context()
+    async def load_cloud_config(self, *, context: Context):
         # cloud-init 23.3 introduced combined-cloud-config, which helps to
         # prevent subiquity from having to go load cloudinit modules.
         # This matters because a downgrade pickle deserialization issue may
@@ -755,13 +819,16 @@ class SubiquityServer(Application):
             users = ug_util.normalize_users_groups(cloud_cfg, cloud.distro)[0]
             self.installer_user_name = ug_util.extract_default(users)[0]
 
-        if "autoinstall" in cloud_cfg:
+        autoinstall = await self._extract_autoinstall_from_cloud_config(
+            cloud_cfg=cloud_cfg, context=context
+        )
+
+        if autoinstall != {}:
             log.debug("autoinstall found in cloud-config")
-            cfg = cloud_cfg["autoinstall"]
             target = self.base_relative(cloud_autoinstall_path)
             from cloudinit import safeyaml
 
-            write_file(target, safeyaml.dumps(cfg))
+            write_file(target, safeyaml.dumps(autoinstall))
         else:
             log.debug("no autoinstall found in cloud-config")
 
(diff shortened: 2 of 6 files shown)


subiquity/cloudinit.py
```diff
@@ -67,6 +69,56 @@ def read_legacy_status(stream):
...
+    return all_errors
+
+
+async def get_schema_failure_sources() -> list[str]:
+    """Retrieve the keys causing schema failure."""
+
+    cmd: list[str] = ["cloud-init", "schema", "--system"]
+    status_coro: Awaitable = arun_command(cmd)
+    try:
+        sp: CompletedProcess = await asyncio.wait_for(status_coro, 600)
```
dbungert, line 101: So while the big wait on `cloud-init status --wait` is allowed to be 10 minutes, most of the others should be instantaneous.  Can you do some real vm testing on the typical response times for these calls?  I'm thinking `timeout=10` might be more realistic.  Part of that 10 minutes is the expectation that the system may still be starting, cloud-init is fetching the data from the datasource, etc.  Whereas the calls here should be much quicker.

subiquity/cloudinit.py
```diff
@@ -86,3 +138,31 @@ async def cloud_init_status_wait() -> (bool, Optional[str]):
...
+    """
+
+    if not supports_recoverable_errors():
+        return (True, None)
+
+    # Recoverable errors support implies extended status support
+    cmd: list[str] = ["cloud-init", "status", "--format=json"]
+    status_coro: Awaitable = arun_command(cmd)
+    try:
+        sp: CompletedProcess = await asyncio.wait_for(status_coro, 600)
```
dbungert, line 159: see above

dbungert (commented):
A comment for now, more review tomorrow.

subiquity/cloudinit.py
Chris-Peterson444: Good point. In my tests they seem pretty instantaneous, even with a large config with many incorrect keys. I've updated the timeouts to the suggested 10 seconds.

Chris-Peterson444:
@blackboxsw could you please take a look at this as well?

subiquity/cloudinit.py
```diff
@@ -86,3 +138,31 @@ async def cloud_init_status_wait() -> (bool, Optional[str]):
...
+    """Check for cloud-init schema errors.
+
+    Returns True if validated OK or can't tell. Otherwise, False and
+    a list of bad keys found.
+    Requires cloud-init supporting recoverable errors and extended status.
+    :return: tuple of (ok, list of bad keys or None)
+    """
+
+    if not supports_recoverable_errors():
+        return (True, None)
```
dbungert, line 153: I guess it would be more pythonic to have this raise with an exception with the list of bad keys, which could in turn be re-raised as that `AutoinstallError`.  What do you think?

subiquity/server/server.py
```diff
@@ -723,7 +727,66 @@ async def start_site(self, runner: web.AppRunner):
...
+        cloud_cfg: dict[str, Any],
+        context: Context,
+    ) -> dict[str, Any]:
+        """Extract autoinstall passed via cloud config."""
+
+        # Not really is-install-context but set to force event reporting
+        context.set("is-install-context", True)
+        context.enter()  # publish start event
+
+        if "autoinstall" not in cloud_cfg:
```
dbungert, line 743: Well, part of the problem is that people aren't supplying an autoinstall key, but supplying autoinstall directives directly to cloud-init  (or could be doing both, but I'm guessing that's not common).  So I think we need to revisit this part.  retrieving `cloud_cfg["autoinstall"]` could be after `validate_cloud_init_schema()` in the current setup.

dbungert (commented):
Chris, I'm really glad to see this work done.  I think it's going to help reduce user confusion.  I have some items to follow-up on.

subiquity/cloudinit.py
Chris-Peterson444: Yeah I think that makes sense. I'll rework the logic. 

Since that if section is highlighted: Part of my assumption here was that we needed recoverable error support to get the bad keys. i.e. we need to ask cloud-init if invalid cloud-config was something it encountered. I suppose this isn't necessarily true, and we could just query the schema always. This has the benefit of cutting out one of the subprocess calls. What do you think?

subiquity/server/server.py
Chris-Peterson444: That makes sense to me. This probably flows better with the throws-on-failure logic.

subiquity/cloudinit.py
dbungert: That makes sense to me but let's VM confirm it with older images, where `supports_recoverable_errors() == False`

Chris-Peterson444:
Thanks for the review @dbungert. I've gone ahead and implemented your suggestions. I added an extra test case to capture the scenario in which autoinstall keys are present in the cloud-config and the `autoinstall` key itself is not.

Also, by removing the dependence on the recoverable errors I was able to test this successfully on a 20.04.6 image which has cloud-init version `22.4.2-0ubuntu0~20.04.2`

subiquity/cloudinit.py
```diff
@@ -67,6 +85,56 @@ def read_legacy_status(stream):
...
+
+async def get_schema_failure_sources() -> list[str]:
+    """Retrieve the keys causing schema failure."""
+
+    cmd: list[str] = ["cloud-init", "schema", "--system"]
+    status_coro: Awaitable = arun_command(cmd)
+    try:
+        sp: CompletedProcess = await asyncio.wait_for(status_coro, 10)
+    except asyncio.TimeoutError:
+        return []
```
ogayot, line 119: ditto, a warning would be good I think :)

subiquity/cloudinit.py
```diff
@@ -67,6 +85,56 @@ def read_legacy_status(stream):
...
+        all_errors.extend(errors_by_level[_level])
+
+    return all_errors
+
+
+async def get_schema_failure_sources() -> list[str]:
+    """Retrieve the keys causing schema failure."""
+
+    cmd: list[str] = ["cloud-init", "schema", "--system"]
+    status_coro: Awaitable = arun_command(cmd)
```
ogayot, line 115: I think it's worth passing `clean_locale=True` explicitly (even though it's the default) since we're potentially parsing locale dependent output.

subiquity/cloudinit.py
```diff
@@ -86,3 +154,23 @@ async def cloud_init_status_wait() -> (bool, Optional[str]):
     else:
         status = read_legacy_status(sp.stdout)
     return (True, status)
+
+
+async def validate_cloud_init_schema() -> bool:
```
ogayot, line 159: Maybe make the function return None since we're going to ignore the return value anyway?

subiquity/cloudinit.py
```diff
@@ -86,3 +154,23 @@ async def cloud_init_status_wait() -> (bool, Optional[str]):
...
+    :rtype: bool
+    :raises CloudInitSchemaValidationError: If cloud-init schema did not validate
+            succesfully.
+    """
+    causes: list[str] = await get_schema_failure_sources()
+
+    if causes == []:
+        return True
+    else:
+        raise CloudInitSchemaValidationError(keys=causes)
```
ogayot, line 176: suggestion:

```python
causes: list[str] = await get_schema_failure_sources()

if causes:
    raise CloudInitSchemaValidationError(keys=causes)
```

subiquity/server/tests/test_server.py
```diff
@@ -357,6 +359,92 @@ async def test_autoinstall_validation__filter_autoinstall(self, config, good, ba
...
+        cloud_data = copy.copy(cloud_cfg)
+        cloud_data.pop("valid-cloud", None)
+        cloud_data.pop("autoinstall", None)
+
+        with patch("subiquity.server.server.validate_cloud_init_schema") as val_mock:
+            if len(cloud_data) == 0:
+                val_mock.return_value = True
+            else:
+                val_mock.side_effect = CloudInitSchemaValidationError(
+                    keys=[key for key in cloud_data.keys()]
```
ogayot, line 433: suggestion

keys=list(cloud_data.keys())

subiquity/cloudinit.py
```diff
@@ -67,6 +85,56 @@ def read_legacy_status(stream):
...
+def read_recoverable_errors(
+    stream: bytes | str,
+    level: Optional[str] = None,
+) -> list[str]:
+    """Retrieve recoverable errors from cloud-init status."""
+
+    try:
+        status: dict[str, Any] = json.loads(stream)
+    except json.JSONDecodeError:
+        return []
```
ogayot, line 97: I think a warning would be good here. Not sure if we should try to display the contents of `stream` in the warning message though...

ogayot (approved):
Good work! A few minor comments but nothing blocking

subiquity/cloudinit.py
Chris-Peterson444: I think that's a good idea. It has to expect the output from `cloud-init status --format=json` so I think something like "Failed to decode cloud-init status output" is informative enough.

Chris-Peterson444:
Thanks @ogayot I've implemented all of your suggestions

dbungert (approved):
One tweak please then LGTM.

subiquity/cloudinit.py
```diff
@@ -4,13 +4,31 @@
...
+from collections.abc import Awaitable
+from subprocess import CompletedProcess
+from typing import Any, Optional
 
 from subiquitycore.utils import arun_command, run_command
 
 log = logging.getLogger("subiquity.cloudinit")
 
 
+class CloudInitSchemaValidationError(Exception):
```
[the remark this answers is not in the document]
Chris-Peterson444: I think that's reasonable, let's do that. I'll update shortly.

Chris-Peterson444:
Thanks! Since I changed CloudInitSchemaValidationError to a NonReportableException, I also added some lines to the client code so the error overlay works for those error types. Could you give that a look over too @dbungert ?

blackboxsw (commented):
Thanks for the ping on this for our awareness.

While I think consuming the CLI and reading stderr is potentially fragile, I get that subiquity may not want to tightly couple to internal cloudinit python libraries and functions to process and iterate over SchemaValidationErrors 
raised by [cloudinit.config.schema.validate_cloudconfig_schema](
https://github.com/canonical/cloud-init/blob/main/cloudinit/config/schema.py#L685) or [cloudnit.config.schema.validate_cloudconfig_file)[https://github.com/canonical/cloud-init/blob/main/cloudinit/config/schema.py#L1037]. 

To use these functions, subiquity would have to likely have to do something like the following:
```
from cloudinit.config.schema import validate_cloudconfig_file, get_schema, SchemaValidationError
cloudconfig_schema = get_schema()
try:
    validate_cloudconfig_file('/var/lib/cloud/instance/user-data.txt', schema=cloudconfig_schema)
except SchemaValidationError as e:
    return [schema_problem.path for schema_problem in e.schema_errors if "unexpected" in schema_problem.message]
```
Note that even the structured SchemaValidationError will not set a 'path' attribute if multiple unexpected keys exist on the object. So, you'd still need your pattern matching to `schema_problem.message` to extract the separate keys so it's nearly the same logic you have for parsing the stderr of `cloud-init schema --system`.

The fragility in CLI approach in this PR will be due to cloud-init relying on `jsonschema` for that error string as cloud-init just presents that error message directly without modification. If `jsonschema` module changes their error messaging format this could break subiquity parsing.

That said, I don't see this error output format from jsonschema  being any different between jammy and noble and I think cloud-init would like to work on a machine-readable representation of `cloud-init schema --system --format=yaml` per your feature request https://github.com/canonical/cloud-init/issues/5100 that will make this easier to process in the future.

Minor changes requested that you can take or leave as you see fit
- better regex pattern match, splitting of the parsed jsonschema error messages
- dropping unused functions/tests
- 
We'll make sure we keep you informed when we start tackling https://github.com/canonical/cloud-init/issues/5100. So this code can consume more friendly structured content when available.

subiquity/cloudinit.py
```diff
@@ -67,6 +86,58 @@ def read_legacy_status(stream):
     return None
 
 
+def read_recoverable_errors(
```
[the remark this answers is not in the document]
Chris-Peterson444: Yeah this was part of an earlier implementation. Sorry about that. I thought to keep it around but I'll be removing it for now since it's not in use.

subiquity/cloudinit.py
```diff
@@ -67,6 +86,58 @@ def read_legacy_status(stream):
...
+    stream: bytes | str,
+    level: Optional[str] = None,
+) -> list[str]:
+    """Retrieve recoverable errors from cloud-init status."""
+
+    try:
+        status: dict[str, Any] = json.loads(stream)
```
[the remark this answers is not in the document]
Chris-Peterson444: As above, dropped for now. Thanks!

subiquity/cloudinit.py
```diff
@@ -67,6 +85,58 @@ def read_legacy_status(stream):
...
+    errors_by_level: dict[str, list[str]] = status.get("recoverable_errors", {})
+
+    if level is not None:
+        return errors_by_level.get(level, [])
+
+    all_errors: list[str] = []
+    for _level in errors_by_level:
+        all_errors.extend(errors_by_level[_level])
+
+    return all_errors
```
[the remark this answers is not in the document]
Chris-Peterson444: Ah TIL. Thanks for the recommendation

subiquity/cloudinit.py
```diff
@@ -67,6 +86,58 @@ def read_legacy_status(stream):
...
+
+
+async def get_schema_failure_sources() -> list[str]:
```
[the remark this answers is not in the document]
Chris-Peterson444: Agreed, changing. Thanks

subiquity/cloudinit.py
```diff
@@ -67,6 +86,58 @@ def read_legacy_status(stream):
...
+    """Retrieve the keys causing schema failure."""
+
+    cmd: list[str] = ["cloud-init", "schema", "--system"]
```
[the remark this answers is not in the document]
Chris-Peterson444: I think in general we want to capture errors from all sources since we read the combined config when extracting the autoinstall config (e.g. is it possible someone sends autoinstall config in their network-config?), but capturing the file source would improve our error messaging for sure. What do you think about leaving this to future improvements?

subiquity/cloudinit.py
```diff
@@ -67,6 +86,58 @@ def read_legacy_status(stream):
...
+    except asyncio.TimeoutError:
+        log.warning("cloud-init schema --system timed out")
+        return []
+
+    error: str = sp.stderr  # Relies on arun_command decoding to utf-8 str by default
+
+    # Matches:
+    # ('some-key' was unexpected)
+    # ('some-key', 'another-key' were unexpected)
+    pattern = r"\((?P<args>'[\w\-]+'(,\s'[\w\-]+')*)+ (?:was|were) unexpected\)"
```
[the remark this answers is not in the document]
Chris-Peterson444: Thanks for the review on this espeically (Regex is hard!). I think we can definitely (1) replace `[\w\-]` with `[^']` to be more flexible and (3) remove the trailing `+`, but (2) isolating the key names without the quotes is difficult. The suggested regex doesn't quite work and my attempts thus far have been insufficient.

I think I'm okay with another round of processing to strip the quotes and this is still an improvement on the parsing.

subiquity/tests/test_cloudinit.py
```diff
@@ -132,3 +137,114 @@ async def test_cloud_init_status_wait_legacy(self, m_wait_for):
             args=[], returncode=0, stdout="status: done\n"
         )
         self.assertEqual((True, "done"), await cloud_init_status_wait())
+
+
+class TestCloudInitUtil(SubiTestCase):
```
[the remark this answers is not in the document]
Chris-Peterson444: As above, correct. removing!

Chris-Peterson444:
@blackboxsw Thanks a lot for the extensive review! I went ahead and implemented most of your suggested changes (all but two, I left my reasoning in the comments and kept them unresolved). Based on your analysis I think we're okay with the potential fragility in the error parsing until we move to the structured output when it's available. I look forward to seeing more on https://github.com/canonical/cloud-init/issues/5100!

dbungert (approved):

dbungert:
OK to merge, despite Noble test failures (unrelated archive problems)

PR #1956, opened 2024-03-26
Allow top-level autoinstall in all delivery methods

Chris-Peterson444:
Allows autoinstall to be passed with a top level `autoinstall` key like the following:

```yaml
autoinstall:
  version: 1
  identity:
    ...
```
in all autoinstall delivery methods. The old format, with no top level `autoinstall`, is still supported. 

I've updated the docs to reflect this change by adding the top-level `autoinstall` to most of the examples, adjusted the directions to default use this structure, and removed the note about not using it in non-cloud-config cases. I also included a small note in the reference section about this structure not technically being a strict requirement in non-cloud-config scenarios and that the old way is still supported, but it also encourages users to use the new format.

In the examples, I think there are still cases where we don't necessarily need to include the top-level keyword. For example, in the more focused `storage` sub-sections where the focus is on sub-key behavior, I purposefully left the top-level `autoinstall` keyword out since it is already focused on a dependent sub-section.

### DIFF
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -3,10 +3,40 @@
 Autoinstall configuration reference manual
 ==========================================
 
-The autoinstall file uses the YAML format. At the top level, it must be a
-mapping containing the keys described in this document. Unrecognised keys
-are ignored in version 1, but will cause a fatal validation error in future
-versions.
+The autoinstall file uses the YAML format. At the top level is a
+single key ``autoinstall`` which contains a mapping of the keys described in
+this document. Unrecognised keys are ignored in version 1, but will cause a
+fatal validation error in future versions.
+
+Here is an example of a minimal autoinstall configuration:
+
+.. code-block:: yaml
+
+    autoinstall:
+      version: 1
+      identity:
+       ...
+
+
+At the top level is the ``autoinstall`` keyword, which contains a version section
+and an (incomplete) identity section which are explained in more detail below.
+Any other key at the level of ``autoinstall``, will result in an autoinstall
+validation error at runtime.
+
+.. warning::
+
+   This behaviour was first introduced during 24.04 (Noble). On any ISOs built
+   before this, you will need to refresh the installer to see this behaviour.
+   Please the note below about the old format.
+
+.. note::
+
+  Technically, in all but one case the top level ``autoinstall`` keyword is
+  strictly unnecessary. This keyword is only necessary when serving autoinstall
+  via cloud-config. For backwards compatibility this format is still supported
+  for non-cloud-config based delivery methods; however, it is
+  **highly recommended** to use the format with a top-level ``autoinstall``
+  keyword as mistakes in this formatting are a common source of confusion.
 
 
 .. _ai-schema:
@@ -29,6 +59,10 @@ Several configuration keys are lists of commands to be executed. Each command ca
 Top-level keys
 --------------
 
+The following keys can be used to configure various aspects of the installation.
+If the global ``autoinstall`` key is provided, then all "top-level keys" must
+be provided underneath it and "top-level" refers to this sub-level. The
+examples below demonstrate this structure.
 
 .. warning::
   In version 1, Subiquity will emit warnings when encountering unrecognised
@@ -57,12 +91,13 @@ A list of configuration keys to still show in the user interface (UI). For examp
 
 .. code-block:: yaml
 
-    version: 1
-    interactive-sections:
-      - network
-    identity:
-      username: ubuntu
-      password: $crypted_pass
+    autoinstall:
+      version: 1
+      interactive-sections:
+        - network
+      identity:
+        username: ubuntu
+        password: $crypted_pass
 
 This example stops on the network screen and allows the user to change the defaults. If a value is provided for an interactive section, it is used as the default.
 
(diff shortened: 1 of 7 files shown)


scripts/validate-autoinstall-user-data.py
```diff
@@ -60,7 +60,13 @@ def main() -> None:
         assert user_data.readline() == "#cloud-config\n"
         def get_autoinstall_data(data): return data["autoinstall"]
     else:
-        def get_autoinstall_data(data): return data
+        def get_autoinstall_data(data):
+            try:
+                cfg = data.pop("autoinstall")
```
ogayot, line 65: I guess it's better not to mutate the original object if we return the data we're interested in? I'd suggest

```python
cfg = data["autoinstall"]
```

Chris-Peterson444: Ah good point.  I think in this script the way we consume the config it wouldn't be a deal breaker but I agree in principle. I get the KeyError I wanted using `[]` anyways. Updated, thanks!

doc/reference/autoinstall-reference.rst
```diff
@@ -4,11 +4,35 @@ Autoinstall configuration reference manual
 ==========================================
 
 The autoinstall file uses the YAML format. At the top level, it must be a
```
dbungert, line 6: This part needs rephrased slightly, as while we strongly suggest using the top-level `autoinstall` and putting stuff under it, I plan to continue to support the existing format.  So the word 'must' here is not quite correct.

subiquity/server/server.py
```diff
@@ -663,7 +663,38 @@ def validate_autoinstall(self):
...
 
-    def load_autoinstall_config(self, *, only_early):
+    @with_context(name="read_config")
+    def _read_config(self, *, cfg_path: str, context: Context) -> dict[str, Any]:
+        with open(cfg_path) as fp:
+            config: dict[str, Any] = yaml.safe_load(fp)
+
+        autoinstall_config: dict[str, Any] = dict()
+
+        # Support "Autoinstall" as a top-level key
```
dbungert, line 673:
```suggestion
        # Support "autoinstall" as a top-level key
```

subiquity/server/server.py
```diff
@@ -663,7 +663,38 @@ def validate_autoinstall(self):
                 context=ctx,
             )
 
-    def load_autoinstall_config(self, *, only_early):
+    @with_context(name="read_config")
+    def _read_config(self, *, cfg_path: str, context: Context) -> dict[str, Any]:
+        with open(cfg_path) as fp:
+            config: dict[str, Any] = yaml.safe_load(fp)
+
+        autoinstall_config: dict[str, Any] = dict()
```
dbungert, line 671:
```suggestion
        autoinstall_config: dict[str, Any] = None
```
All code paths set `autoinstall_config`, so we can skip the unused `dict()`

subiquity/server/server.py
```diff
@@ -663,7 +663,38 @@ def validate_autoinstall(self):
...
+
+        # Support "Autoinstall" as a top-level key
+        if "autoinstall" in config:
+            autoinstall_config = config.pop("autoinstall")
+
+            # but the only top level key
+            if len(config) != 0:
+                self.interactive = bool(autoinstall_config.get("interactive-sections"))
+                msg: str = (
+                    "autoinstall is not a valid cloud config datasource.\n"
```
dbungert, line 681:
```suggestion
                    "autoinstall.yaml on the ISO is not a valid cloud config datasource.\n"
```
or something like that

subiquity/server/server.py
Chris-Peterson444: Fair, although really I was just trying to annotate the type once and avoid an `| None`. What if we just do 

 `autoinstall_config: dict[str, Any]`

and defer assignment?

dbungert: works for me

doc/reference/autoinstall-reference.rst
Chris-Peterson444: Makes sense. I was trying to be a bit too forceful in moving users to the new format. I took out the `must` but left it unchanged as I think it still reads fine otherwise. Do you think this is sufficient with the note below that paragraph?

.github/workflows/ci.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
         image:
           - ubuntu-daily:jammy
           - ubuntu-daily:mantic
-          - ubuntu-daily:noble
+          # - ubuntu-daily:noble
```
dbungert, line 20: I don't usually like approving red CI builds, but with current noble archive state I'll OK it.  Let's leave this check in place please.

Chris-Peterson444: Sounds good. Commit dropped!

subiquity/server/server.py
Chris-Peterson444: I think just "autoinstall.yaml is not..." is sufficient. I don't think it's likely but I don't want users to think autoinstall elsewhere is a valid cloud config datasource. What do you think?

Chris-Peterson444:
Rebased on top of main to fix merge conflict

Chris-Peterson444:
Pushed with slight changes to the docs:
1. Remove verbiage about sub-keys in the first section. This is already handled in the top-level keys section below.
2. Add a note about needing to do a snap-refresh to see this behavior on ISOs built before Noble

Chris-Peterson444:
Apologies. One more push for lint + spelling

dbungert (approved):
LGTM when the checks (excluding Noble) are happy
@rkratky - When you get a moment take a look at the doc portions please.  Also, we have a larger problem here about how we want to show things that are supported only at certain Subiquity versions, we're just handling one new case of that in this PR.

rkratky:
> LGTM when the checks (excluding Noble) are happy @rkratky - When you get a moment take a look at the doc portions please. Also, we have a larger problem here about how we want to show things that are supported only at certain Subiquity versions, we're just handling one new case of that in this PR.

Apologies -- I only got to this today, so I submitted https://github.com/canonical/subiquity/pull/1963 with some fixes.

PR #1957, opened 2024-03-27
filesystem: don't suggest target resize if part is mounted

ogayot:
When a partition is mounted in the live environment (either by casper or another process) and is selected for resize, curtin will fail when running e2fsck (or equivalent) with an error such as:

```
Resize requested for format ext4
Resizing /dev/sdb4 of type ext4 down to 12546211840
Running command ['e2fsck', '-p', '-f', '/dev/sdb4'] with allowed return codes [0] (capture=False)
/dev/sdb4 is mounted.
e2fsck: Cannot continue, aborting.

An error occured handling 'disk-sdb': ProcessExecutionError - Unexpected error while running command.
Command: ['e2fsck', '-p', '-f', '/dev/sdb4']
Exit code: 8
```
This is what happens when casper creates a partition on the installer media to store /var/log and /var/crash. Usually the partition is large enough to install Ubuntu so a GuidedStorageTargetResize scenario gets emitted.

This patch prevents the emission of a GuidedStorageTargetResize scenario if the partition is marked as mounted already.

LP:#2059389

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -1141,6 +1141,8 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         for disk in self.potential_boot_disks(check_boot=False):
             part_align = disk.alignment_data().part_align
             for partition in disk.partitions():
+                if partition._is_in_use:
+                    continue
                 vals = sizes.calculate_guided_resize(
                     partition.estimated_min_size,
                     partition.size,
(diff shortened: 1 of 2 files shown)


ogayot:
I think the change makes sense. It will prevent partitions that are mounted in the live environment from appearing in a GuidedStorageTargetResize scenario. In the desktop installer. the user can mount partitions by opening the file explorer and clicking on a the device under "other locations". This means that Subiquity will no longer consider a resize scenario involving that partition. This should be fine as long as something else in the UI does not automount partitions accidentally. In any case, the current behavior in such an event would be a failed install so I wouldn't call it a regression.

dbungert (approved):

Chris-Peterson444 (approved):
LGTM. If I remember right, there was at least one bug I looked at where a user tried to install onto an external ssd that was already mounted in the live environment and resulted in an error, which I think this would also solve.

PR #1960, opened 2024-04-02
Update intro-to-autoinstall.rst

VariableDeclared:
The installer will throw an error if the `version` key is not on the top level. This patch addresses this in our documentation.


dbungert (changes requested):
Unfortunately that's not quite correct.  Either all the keys for Subiquity, including version, would be under autoinstall if we're using cloud-init to deliver the config to the installation system, or all the keys would be at the top level for the on-ISO case.

We're discussing this in other channels, so we can update this PR when that's done.

VariableDeclared:
After discussing with Dan, we've narrowed down the cause of the confusion! I will close my PR

PR #1964, opened 2024-04-05
pre-commit: add spelling check

Chris-Peterson444:
Tired of forgetting to run `make spelling` before committing changes to the documentation? This is why we have pre-commit!

The downside is that the spell check can take a little while, but I think it's better than waiting for CI to yell at you. I recently had a PR that was unnecessarily noisy due to fixing spelling errors.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,20 @@
-files: "(console_conf|subiquity|subiquitycore|system_setup)"
+files: "(console_conf|subiquity|subiquitycore|system_setup|doc)"
 repos:
   - repo: https://github.com/psf/black
     rev: 23.7.0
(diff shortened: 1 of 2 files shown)


rkratky:
We're working on getting away from `pyspelling` in order to be able to run spellcheck on source files (that way, it would be possible to check only modified files). So, this should soon (hopefully) be much faster.

dbungert (approved):
This is a good idea.  I have suggestions about improving this so that we don't need the external script, but I'll open that PR once this is resolved.

Chris-Peterson444:
Thanks! I suppose the entry point could have just been `bash -c "cd doc/ && make spelling"` but it felt right to do the environment check. I look forward to any improvements!

PR #1965, opened 2024-04-09
identity: fail on missing section sooner

Chris-Peterson444:
Moves the logic for checking if user-data or identity section is provided (on server) in the autoinstall config to the load_autoinstall_data function. Without this change, the exception thrown in apply_autoinstall_config won't halt the installer until
the postinstall steps (LP: 2060547).

The change is ready but I am keeping this as a draft PR until I can test this on Desktop.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -294,6 +294,16 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         # log.debug("self.ai_data = %s", data)
         self.ai_data = data
 
+    # The identity and user-data section are optional if we are only installing
+    # the reset partition, however the identity controller needs to know this
+    # before the filesystem controller naturally sets this. So this is a
+    # function to inspect the outcome early.
+    # See: https://github.com/canonical/subiquity/pull/1965
+    def is_reset_partition_only(self):
+        storage_config = self.app.autoinstall_config.get(self.autoinstall_key, {})
+        layout = storage_config.get("layout", {})
+        return layout.get("reset-partition-only", False)
+
     async def configured(self):
         self._configured = True
         if self._info is None:
--- a/subiquity/server/controllers/identity.py
+++ b/subiquity/server/controllers/identity.py
@@ -89,14 +89,7 @@ class IdentityController(SubiquityController):
         if self.app.base_model.source.current.variant != "server":
             return
         # 3. we are only refreshing the reset partition
-        # (The identity controller doesn't figure this out until the apply
-        # step, so we are going to cheat and inspect the situation here)
-        storage_config = self.app.autoinstall_config.get("storage")
-        if (
-            storage_config is not None
-            and storage_config.get("layout") is not None
-            and storage_config["layout"].get("reset-partition-only")
-        ):
+        if self.app.controllers.Filesystem.is_reset_partition_only():
             return
         # 4. identity section is interactive
         if self.interactive():
--- a/subiquity/server/controllers/identity.py
+++ b/subiquity/server/controllers/identity.py
@@ -22,8 +22,8 @@ import attr
 from subiquity.common.apidef import API
 from subiquity.common.resources import resource_path
 from subiquity.common.types import IdentityData, UsernameValidation
+from subiquity.server.autoinstall import AutoinstallError
 from subiquity.server.controller import SubiquityController
-from subiquitycore.context import with_context
 
 log = logging.getLogger("subiquity.server.controllers.identity")
 
@@ -79,18 +79,29 @@ class IdentityController(SubiquityController):
                 crypted_password=data["password"],
             )
             self.model.add_user(identity_data)
-
-    @with_context()
-    async def apply_autoinstall_config(self, context=None):
-        if self.model.user:
             return
+
+        # The identity section is required except if (any):
+        # 1. a user-data section is provided
         if "user-data" in self.app.autoinstall_config:
             return
-        if self.app.base_model.target is None:
-            return
+        # 2. we are installing not-Server (Desktop)
         if self.app.base_model.source.current.variant != "server":
             return
-        raise Exception("neither identity nor user-data provided")
+        # 3. we are only refreshing the reset partition
+        # (The identity controller doesn't figure this out until the apply
+        # step, so we are going to cheat and inspect the situation here)
+        storage_config = self.app.autoinstall_config.get("storage")
+        if (
+            storage_config is not None
+            and storage_config.get("layout") is not None
+            and storage_config["layout"].get("reset-partition-only")
+        ):
+            return
+        # 4. identity section is interactive
+        if self.interactive():
+            return
+        raise AutoinstallError("neither identity nor user-data provided")
 
     def make_autoinstall(self):
         if self.model.user is None:
--- a/subiquity/server/controllers/source.py
+++ b/subiquity/server/controllers/source.py
@@ -29,6 +29,8 @@ from subiquity.common.types import SourceSelection, SourceSelectionAndSetting
 from subiquity.server.controller import SubiquityController
 from subiquity.server.types import InstallerChannels
 
+SOURCES_DEFAULT_PATH = "/cdrom/casper/install-sources.yaml"
+
 log = logging.getLogger("subiquity.server.controllers.source")
 
 
@@ -77,8 +79,14 @@ class SourceController(SubiquityController):
         super().__init__(app)
         self._handler = None
         self.source_path: Optional[str] = None
-        self.ai_source_id: Optional[str] = None
         self._configured: bool = False
+
+        path = SOURCES_DEFAULT_PATH
+        if self.app.opts.source_catalog is not None:
+            path = self.app.opts.source_catalog
+        if os.path.exists(path):
+            with open(path) as fp:
+                self.model.load_from_file(fp)
 
     def make_autoinstall(self):
         return {
@@ -86,7 +94,7 @@ class SourceController(SubiquityController):
             "id": self.model.current.id,
         }
 
-    def load_autoinstall_data(self, data: Any) -> None:
+    def load_autoinstall_data(self, data: Optional[dict[str, Any]]) -> None:
         if data is None:
             data = {}
 
@@ -98,22 +106,11 @@ class SourceController(SubiquityController):
             "search_drivers", SEARCH_DRIVERS_AUTOINSTALL_DEFAULT
         )
 
-        # At this point, the model has not yet loaded the sources from the
-        # catalog. So we store the ID and lean on self.start to select the
-        # current source accordingly.
-        self.ai_source_id = data.get("id")
+        # Assign the current source if hinted by autoinstall.
+        if id := data.get("id"):
+            self.model.current = self.model.get_matching_source(id)
 
     def start(self):
-        path = "/cdrom/casper/install-sources.yaml"
-        if self.app.opts.source_catalog is not None:
-            path = self.app.opts.source_catalog
-        if not os.path.exists(path):
-            return
-        with open(path) as fp:
-            self.model.load_from_file(fp)
-        # Assign the current source if hinted by autoinstall.
-        if self.ai_source_id is not None:
-            self.model.current = self.model.get_matching_source(self.ai_source_id)
         self.app.hub.subscribe(
             (InstallerChannels.CONFIGURED, "locale"), self._set_locale
         )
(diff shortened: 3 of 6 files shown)


Chris-Peterson444:
This doesn't quite work for desktop because `app.base_model.source.current` isn't set correctly until the controllers are started, which only happens after loading the autoinstall data. I think the quickest (and hackiest) way to solve this is to inspect the sources information in `/cdrom/casper/install-sources.yaml` from the identity controller itself.

ogayot:
I've only tested in dry-run mode but I'm not seeing an obvious difference between the previous and the new behavior. What is it that we are trying to achieve? Exit the subiquity server process before partitioning starts?

Chris-Peterson444:
Yes. On server, if you omit both the identity section and the user-data section then an error will be emitted when the identity controller loads the autoinstall data, but the installation with proceed until the postinstall stage where it hangs. We want to make sure this error halts the installer immediately.

dbungert:
To elaborate on what Chris said, the failure mode here is really best seen to be believed.

So you supply a trivial cloud-config like so to live-server:
```
#cloud-config
autoinstall:
  version: 1
```

An exception is seen in the logs complaining about the lack of `user-data` or `identity`, but the entire install section continues, including partitioning of disks and copying to the target system.

THEN, at this point, the install hangs, as the post-install models are not configured, and there isn't even a visual indication on why this is other than that Exception shown all the way before partitioning even started.

My goal here was to move that check earlier, since the time between the problem report and when the install stops is measured in how long it takes for partitioning + rsync to happen.

Chris-Peterson444:
Yeah, sorry for not being more explicit. So the way I see it this could have regressed the following conditions:
- Server autoinstall:
  - Includes only Identity section: OK
  - Includes only user-data section: OK
  - Identity section is explicitly interactive: OK
  - Identity section is implicitly interactive (i.e. * ): OK
  - reset-partition-only is true: OK
  - none of the above scenarios and missing both identity and user-data sections: Error
- Desktop autoinstall:
  - Provides neither identity or userdata: OK

I was able to confirm all of these cases pass in VM tests. I did find a separate bug in the reset-partition-only handling but I'll file a separate bug about that.

subiquity/server/controllers/identity.py
```diff
@@ -79,18 +84,36 @@ def load_autoinstall_data(self, data):
...
+        if self._install_source_is_desktop():
+            return
+        # 3. we are only refreshing the reset partition
+        # (The identity controller doesn't figure this out until the apply
+        # step, so we are going to cheat and inspect the situation here)
+        storage_config = self.app.autoinstall_config.get("storage")
+        if (
+            storage_config is not None
+            and storage_config.get("layout") is not None
+            and storage_config["layout"].get("reset-partition-only")
```
dbungert, line 106: Thanks!  Can we add a utility function in `server/controllers/filesystem.py` and stuff this check there?

subiquity/server/controllers/identity.py
```diff
@@ -79,18 +84,36 @@ def load_autoinstall_data(self, data):
...
+        ):
+            return
+        # 4. identity section is interactive
+        if self.interactive():
+            return
+        raise AutoinstallError("neither identity nor user-data provided")
+
+    @with_context()
+    async def apply_autoinstall_config(self, context=None):
+        pass
```
dbungert, line 116: I think this matches the default implementation provided in `SubiquityController` and can be removed.

subiquity/server/controllers/identity.py
```diff
@@ -135,3 +158,49 @@ async def validate_username_GET(self, username: str) -> UsernameValidation:
...
 
         return UsernameValidation.OK
+
+    # The identity section (and user-data) section are entirely optional
+    # on Desktop, but setting the variant doesn't happen until controllers
+    # are started. For the identity controller to properly validate
+    # this behavior, we need to do look-ahead on the source controller
+    # logic for loading the available sources.
+    # See: https://github.com/canonical/subiquity/pull/1965
+    def _install_source_is_desktop(self) -> bool:
```
dbungert, line 168: similar feedback here, I'd like to lift this to a utility function in `subiquity/server/controllers/identity.py`

subiquity/server/controllers/identity.py
```diff
@@ -135,3 +158,49 @@ async def validate_username_GET(self, username: str) -> UsernameValidation:
...
+        # Check the autoinstall config first
+        source_config = self.app.autoinstall_config.get("source")
+        if (
+            source_config is not None
+            and (id := source_config.get("id")) is not None
+            and "server" not in id
```
dbungert, line 174: I'm wary of checking the string set on the id, it's semi free form.
What's the motivation for this section?  It looks like we can drop it and get the same outcome?  If I'm missing something let me know.

dbungert (commented):
Well that was a bit more involved than I would have guessed.  Nice work!

Conceptually this is fine, my major feedback is to just relocate some parts to a place where we might remove the duplication, or at least make it more greppable.  I would be quite surprised to find some of this in `identity.py`.

subiquity/server/controllers/identity.py
Chris-Peterson444: We could, I was mostly putting this stuff in the Identity controller since the other controllers don't necessarily need to know that the Identity controller has quirks depending on their data, but it does make more sense for the filesystem controller to know how to handle its own autoinstall data. The other concern though is that we would have to ensure the identity controller comes after the filesystem controller in the server's controller list since we would need to have the filesystem controller's data loaded before the identity controller loads and calls this check. I guess we could have the filesystem controller cheat and read the autoinstall config from the server but it feels a little wrong. What do you think?

subiquity/server/controllers/identity.py
Chris-Peterson444: Ah good point, thanks.

subiquity/server/controllers/identity.py
Chris-Peterson444: in `subiquity/server/controllers/source.py` in `load_autoinstall_data` we get the source id with a `.get("id")` on the autoinstall data and do an exact match on the IDs in the catalog via `get_matching_source` in `subiquity/models/source.py` so I felt that it wasn't that bad of a comparison to just check that `server` wasn't in the name. 

In the autoinstall case if the id the user provides doesn't match any sources in the catalog then we get a `KeyError` anyways so I wasn't super concerned with weird outcomes if the user provides something unrealistic.

Chris-Peterson444: *but this section matches the source controller/model logic of letting autoinstall inform us of the source id

Chris-Peterson444:
>Well that was a bit more involved than I would have guessed. Nice work!

Right? And thanks!  

Another thing I had in mind about keeping the logic in the identity controller: we probably want to remove this logic at some point in the future and put it back into the load_autoinstall function when we can just throw from there and expect an immediate halt to the overall install.

Chris-Peterson444:
Moved to separate utility functions owned by the respective controllers, as requested. There is still the open question about the source controller logic.

subiquity/server/controllers/filesystem.py
```diff
@@ -293,6 +293,24 @@ def load_autoinstall_data(self, data):
...
+        # not dependent on the server's controller load order
+        storage_config = self.app.autoinstall_config.get(self.autoinstall_key)
+        if (
+            storage_config is not None
+            and storage_config.get("layout") is not None
+            and storage_config["layout"].get("reset-partition-only")
+        ):
+            return True
+        else:
+            return False
```
dbungert, line 312:
```suggestion
        storage_config = self.app.autoinstall_config.get(self.autoinstall_key, {})
        layout = storage_config.get("layout", {})
        return layout.get("reset_partition_only", False)
```

subiquity/server/controllers/source.py
```diff
@@ -176,3 +179,49 @@ async def POST(self, source_id: str, search_drivers: bool = False) -> None:
...
+
+        # In dryrun we are on server, unless you spoof with:
+        #   source:
+        #    id: desktop
+        if self.app.opts.dry_run:
+            return False
+
+        # Check the sources available the same way the source controller does
+        # but return a bool for if the variant is desktop
+        path = "/cdrom/casper/install-sources.yaml"
```
dbungert, line 207: let's move the existing magic path and reference that common definition, please, or maybe this is obsolete given the previous comment

subiquity/server/controllers/source.py
```diff
@@ -176,3 +179,49 @@ async def POST(self, source_id: str, search_drivers: bool = False) -> None:
...
+    # are started. For the identity controller to properly validate
+    # this behavior, we need to do look-ahead on the source controller
+    # logic for loading the available sources.
+    # See: https://github.com/canonical/subiquity/pull/1965
+    def _install_source_is_desktop(self) -> bool:
+        # Check the autoinstall config first
+        source_config = self.app.autoinstall_config.get(self.autoinstall_key)
+        if (
+            source_config is not None
+            and (id := source_config.get("id")) is not None
```
dbungert, line 194: I think I'm vetoing this approach, as it relies on behavior in a string that is theoretically an arbitrary value.
What I want instead is for the `SourceController.start` logic to move earlier in the sequence - out of `start` entirely actually - so that we can answer this question before `SourceController.start()` has been called.  Then we should be able to just get the info needed.  What do you think?

dbungert (changes requested):

subiquity/server/controllers/filesystem.py
Chris-Peterson444: Ah yes this is much better, thanks!

subiquity/server/controllers/source.py
Chris-Peterson444: Yeah obsolete but I still think it's worth moving the magic var definition somewhere global

subiquity/server/controllers/source.py
Chris-Peterson444: When I originally looked at this bug I think I had reservations about moving the logic in `SourceController.start` to somewhere sooner, but looking at it again now I don't see why we can't. We can probably move that logic to `SourceController.load_autoinstall_data`. I'll give it a shot.

subiquity/server/controllers/filesystem.py
```diff
@@ -294,6 +294,18 @@ def load_autoinstall_data(self, data):
         # log.debug("self.ai_data = %s", data)
         self.ai_data = data
 
+    # The identity and user-data section are optional if we are only installing
+    # the reset partition, however the identity controller needs to know this
+    # before the filesystem controller naturally sets this. So this is a
+    # temporary function to inspect the outcome.
+    # See: https://github.com/canonical/subiquity/pull/1965
+    def _reset_parition_only(self):
+        # Cheating and checking the server's config makes this function
```
dbungert, line 303: I think this isn't cheating.  Filesystem is answering filesystem questions.

subiquity/server/controllers/filesystem.py
```diff
@@ -294,6 +294,18 @@ def load_autoinstall_data(self, data):
         # log.debug("self.ai_data = %s", data)
         self.ai_data = data
 
+    # The identity and user-data section are optional if we are only installing
+    # the reset partition, however the identity controller needs to know this
+    # before the filesystem controller naturally sets this. So this is a
+    # temporary function to inspect the outcome.
```
dbungert, line 300: Sometimes "temporary" functions aren't.  Do we have a plan to remove this?

dbungert (approved):
I have suggested a few tweaks but this seems fine.  What further work do you imagine for this draft PR?   What sort of VM testing has been done at this point?

subiquity/server/controllers/filesystem.py
Chris-Peterson444: Ah fair, thanks. It was "cheating" in the identity controller.

subiquity/server/controllers/filesystem.py
```diff
@@ -294,6 +294,18 @@ def load_autoinstall_data(self, data):
...
+    # See: https://github.com/canonical/subiquity/pull/1965
+    def _reset_parition_only(self):
```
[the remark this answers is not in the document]
Chris-Peterson444: Ah man I really can't type partition lately. Thanks. 

>  Probably should not have the leading underscore either, as the intention is that it can be used by other controllers.

Yeah good point. Renamed to `is_reset_partition_only` to not conflict with the instance variable `reset_partition_only`

Chris-Peterson444:
Thanks for the suggestions! Currently I am running the following VM tests before I mark this ready:

- Server autoinstall:
    - Includes only Identity section, successful autoinstall
    - Includes only user-data section, successful  autoinstall
    - Identity section is explicitly interactive, successful (interactive) autoinstall
    - Identity section is implicitly interactive (i.e. * ), successful (interactive) autoinstall
    - reset-partition-only is true, successful autoinstall
    - none of the above scenarios and missing both identity and user-data sections, installer halts and displays error
- Desktop autoinstall:
    - Provides neither identity or userdata, successful autoinstall and does user creation on first boot

Chris-Peterson444:
The above tests all pass

Chris-Peterson444:
The above tests all pass

PR #1966, opened 2024-04-10
filesystem: datasets for /root and /home

dbungert:
Partial implementation of LP:#2057837.  To create a specific dataset for /home/$user, this requires moving where and when users are created, which is not in plan for 24.04.

FR-7081

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -616,6 +616,11 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         rpool.create_zfs(f"ROOT/ubuntu_{uuid}/var/spool")
         rpool.create_zfs(f"ROOT/ubuntu_{uuid}/var/www")
 
+        userdata_uuid = gen_zsys_uuid()
+        rpool.create_zfs("USERDATA", canmount="off", mountpoint="none")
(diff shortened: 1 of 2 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -616,6 +616,11 @@ def guided_zfs(self, gap, choice: GuidedChoiceV2):
         rpool.create_zfs(f"ROOT/ubuntu_{uuid}/var/spool")
         rpool.create_zfs(f"ROOT/ubuntu_{uuid}/var/www")
 
+        userdata_uuid = gen_zsys_uuid()
+        rpool.create_zfs("USERDATA", canmount="off", mountpoint="/")
```
ogayot, line 620: Do we have to specify the mountpoint here? For "ROOT" we use the string "none" and `create_zfs` has `mountpoint=None` by default.

dbungert: Works for me.  I think it doesn't matter with `canmount="off"`  but consistency sounds good.  Fixed and vm tested!

ogayot: Thanks for checking!

ogayot (approved):

PR #1967, opened 2024-04-10
Doc makefile

dbungert:
* improve the doc makefile so `make install` multiple times DTRT
* fix deps of other targets, most so spelling -> install works but fix others while I'm here
* now the spelling check can be just `make -C doc spelling`

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -10,10 +10,12 @@ VENV          = .sphinx/venv/bin/activate
 
 
 # Put it first so that "make" without argument is like "make help".
-help:
-	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+help: install
+	@. $(VENV); $(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-install:
+install: .sphinx/venv
+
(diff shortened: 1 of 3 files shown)


doc/Makefile
```diff
@@ -30,16 +32,20 @@ install:
                 "* check links: make linkcheck \n" \
                 "* check inclusive language: make woke \n" \
 		"--------------------------------------------------------------- \n"
-run:
+
+.sphinx/requirements.txt:
+	:
```
Chris-Peterson444, line 37: What does this colon do?

Chris-Peterson444 (approved):
Just a curiosity question on one of the targets. Very nice though, thanks!

doc/Makefile
dbungert: There is a default target at the end that tries to run otherwise-unhandled targets with `sphinx-build`.  Merely declaring `.sphinx/requirements.txt` isn't enough and you get the following:

```
. .sphinx/venv/bin/activate; sphinx-build -M .sphinx/requirements.txt "." "_build"                   
Running Sphinx v7.2.6                                                                                
matplotlib is not installed, social cards will not be generated                                      
                                                                                                     
Sphinx error:                                                                                        
Builder name .sphinx/requirements.txt not registered or available through entry point                
```
By running anything we can avoid that, so I'm just using the no-op `:` command.

Chris-Peterson444: Ahh it's a no-op. Thanks!

PR #1968, opened 2024-04-11
mirror: if the mirror test fails, suggest an offline install

ogayot:
Client / TUI change
-----------
In the mirror screen, if the test fails and the user decides to ignore the failure, we used to continue the installation normally ; which in most scenarios resulted in an error at a later stage of the installation.

Instead, we now revert to an installation without network (i.e., only packages from the pool are considered for installation) if the user decides to ignore the failure.

Current design:
2024-04-11T11:07:57,433958092+02:00

Proposed design:
<details>
<summary>Previously proposed design</summary>

2024-04-11T11:07:27,317845628+02:00
</details>

2024-04-17T09:58:43,258322845+02:00

Note that the behavior hasn't changed if the user decides to skip the test _while it is running,_ 

Backend change
------------------

The /mirror GET and POST endpoints now include a boolean field named `use_during_installation`.

* if set to `True`, the mirror information will be used during installation to fetch packages online.
* if set to `False`, we will only fetch packages from the pool.

In either case, the mirror information will still be used to build the `etc/apt/` directory in the target system.

Currently, the way use_during_installation is implemented is coupled with the force_offline property of the network model. This means that Ubuntu Pro and other stuff will be disabled too if we're skipping the use of the archive.

NOTE: the default value for "use_during_installation" in the POST endpoint is `null` ; which means that we should not change the current value of force_offline.

LP:#2059898 and similar reports

### DIFF
--- a/subiquity/ui/views/mirror.py
+++ b/subiquity/ui/views/mirror.py
@@ -47,8 +47,10 @@ MIRROR_CHECK_CONFIRMATION_TEXTS = {
     MirrorCheckStatus.FAILED: (
         _("Mirror check failed"),
         _(
-            "The check of the mirror URL failed. You can continue, but it is very"
-            " likely that the installation will fail."
+            "The check of the mirror URL failed. If you decide to continue, only"
+            " packages present on the installation media will be considered for"
+            " installation. Remember to install security updates after booting"
+            " your newly installed system."
         ),
     ),
     None: (
@@ -217,7 +219,7 @@ class MirrorView(BaseView):
         self.last_status = check_state.status
 
     def done(self, result):
-        async def confirm_continue_anyway() -> None:
+        async def confirm_continue_anyway(continue_skip_archive: bool) -> None:
             title, question = MIRROR_CHECK_CONFIRMATION_TEXTS[self.last_status]
             confirmed = await self.ask_confirmation(
                 title=title,
@@ -227,7 +229,8 @@ class MirrorView(BaseView):
             )
 
             if confirmed:
-                self.controller.done(result.url.value)
+                skip_archive = True if continue_skip_archive else None
+                self.controller.done(result.url.value, skip_archive=skip_archive)
 
         log.debug("User input: {}".format(result.as_data()))
         if self.has_network and self.last_status in [
@@ -235,9 +238,10 @@ class MirrorView(BaseView):
             MirrorCheckStatus.FAILED,
             None,
         ]:
-            async_helpers.run_bg_task(confirm_continue_anyway())
+            status_is_failed = self.last_status == MirrorCheckStatus.FAILED
+            async_helpers.run_bg_task(confirm_continue_anyway(status_is_failed))
         else:
-            self.controller.done(result.url.value)
+            self.controller.done(result.url.value, skip_archive=None)
 
     def cancel(self, result=None):
         self.controller.cancel()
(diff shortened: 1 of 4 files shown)


dbungert (commented):
I propose that we take the needed parts from this PR and add in the policy change from #1816 but do so across the board - OFFLINE_INSTALL as the fallback across the board.  WDYT?

ogayot:
> I propose that we take the needed parts from this PR and add in the policy change from #1816 but do so across the board - OFFLINE_INSTALL as the fallback across the board. WDYT?

Honestly, I've never been a fan of `fallback: offline-install` by default ; at least without a clear indication that an offline installation is going on. Otherwise:

1. after installation, you might boot into an outdated (potentially vulnerable) system without realizing it
2. if you provide a wrong network configuration (e.g., missing proxy and stuff), I'd argue that you'd probably want the installation to stop and fix your config
3. it can't work properly (yet?) with other directives like `packages: [...]`. (unless everything is in the pool)

With that in mind, I'll still implement the change and we can discuss further.

Chris-Peterson444 (approved):

subiquity/ui/views/mirror.py
```diff
@@ -227,17 +228,19 @@ async def confirm_continue_anyway() -> None:
...
+                self.controller.done(result.url.value, skip_archive=skip_archive)
 
         log.debug("User input: {}".format(result.as_data()))
         if self.has_network and self.last_status in [
             MirrorCheckStatus.RUNNING,
             MirrorCheckStatus.FAILED,
             None,
         ]:
-            async_helpers.run_bg_task(confirm_continue_anyway())
+            status_is_failed = bool(self.last_status == MirrorCheckStatus.FAILED)
```
dbungert, line 240: nitpick: `bool()` is redundant

dbungert (approved):

subiquity/ui/views/mirror.py
```diff
@@ -227,17 +228,19 @@ async def confirm_continue_anyway() -> None:
             )
 
             if confirmed:
-                self.controller.done(result.url.value)
+                skip_archive = True if continue_skip_archive else None
```
mwhudson, line 231: The True/False/None tristate is a bit confusing but well. One thing at a time.

mwhudson (approved):
Just one (annoying) question but this looks good to me otherwise.

subiquity/ui/views/mirror.py
```diff
@@ -47,8 +47,9 @@
     MirrorCheckStatus.FAILED: (
         _("Mirror check failed"),
         _(
-            "The check of the mirror URL failed. You can continue, but it is very"
-            " likely that the installation will fail."
+            "The check of the mirror URL failed. If you decide to continue, only"
+            " packages present on the installation media will be considered for"
+            " installation."
```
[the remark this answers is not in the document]
ogayot: I was thinking of
```diff
-            "The check of the mirror URL failed. You can continue, but it is very"
-            " likely that the installation will fail."
+            "The check of the mirror URL failed. If you decide to continue, only"
+            " packages present on the installation media will be considered for"
+            " installation. Remember to install security updates after booting"
+            " your newly installed system."
```

subiquity/ui/views/mirror.py
ogayot: Sorry about that :upside_down_face:

subiquity/ui/views/mirror.py
ogayot: Well... black does not like the parentheses so I guess that's why I used bool() in the first place. Will do `status_is_failed = self.last_status == MirrorCheckStatus.FAILED` then

PR #1970, opened 2024-04-11
models/subiquity: disable cc_dpkg_grub

dbungert:
On first boot, there appear to be cases where the cc_dpkg_grub module in cloud-init is not doing the right thing, which are to be sorted out. But for first boot to happen we have a working boot config, so let's not perturb that.

LP:#2060695 / FR-7389

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -437,7 +437,9 @@ class SubiquityModel:
                 },
             }
         )
+        grub_dpkg = yaml.dump({"grub_dpkg": {"enabled": False}})
         files = [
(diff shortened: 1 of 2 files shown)


Chris-Peterson444 (approved):

mwhudson (approved):

ogayot (approved):

kukrimate (approved):
Looks good

ogayot:
Sorry, I might have clicked merged too quickly. @blackboxsw was still pending review.

blackboxsw (commented):
Thanks @ogayot @dbungert for this. I agree this will properly disable  cloud-init's grub_dpkg operations which probe and attempt to set primary boot device avoiding unwanted debconf set-selections of `grub-pc/install_devices` on subiquity-driven installations.

PR #1971, opened 2024-04-12
mirror: switch default policy to OFFLINE_INSTALL by default

ogayot:
Automated installations and desktop installations will now switch to an offline installation by default. In this mode, only packages from the pool will be fetched.

This should allow the installation to proceed when installing from a private network without external access to the internet.

This is similar in essence to #1816 but applies the policy to automated server installs as well.

If you are doing a desktop installation (or an automated server installation) in a private network with no Internet access, the mirror connectivity test will eventually timeout and the installer will switch to an "offline" installation. (let's call it semi-offline). 

Going forward, I think we should have a new directive in autoinstall configuration mimicking `use_during_installation` from #1968, so that users can do a semi-offline installation without going through the mirror connectivity test at all.

### DIFF
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -327,7 +327,7 @@ fallback
 ^^^^^^^^
 
 * **type:** string (enumeration)
-* **default:** abort
+* **default:** offline-install
 
 Controls what Subiquity does when no primary mirror is usable. Supported values are:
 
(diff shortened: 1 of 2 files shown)


dbungert (approved):

ogayot:
@d-loose FYI this will change the default behavior on desktop too. Subiquity will automatically switch to a semi-offline installation if the mirror connectivity test fails.

PR #1972, opened 2024-04-13
filesystem: reset-partition size bug

Chris-Peterson444:
In python, bool is a subclass of int so we can't use isinstance to check if the user specified a size for the reset partition. This causes autoinstall with "reset-partition: True" and "reset-partition-only: true" to crash the installer. (LP: #2061042).

I'm not too sure how useful the following autoinstall config is:
```yaml
autoinstall:
  version: 1
  storage:
    layout:
      name: direct
      reset-partition: true
      reset-partition-only: true
```
but this will cause subiquity to report the following:

`2024-04-12 22:16:10,913 INFO subiquity.server.controllers.filesystem:1463 autoinstall: will install reset partition of size 1`

and then crash during rsync.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -1457,7 +1457,8 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         rp_input = layout.get("reset-partition", None)
         if rp_input:
             reset_partition = True
-            if isinstance(rp_input, (str, int)):
+            # bool is a subclass of int -- check for int explicitly
+            if isinstance(rp_input, str) or type(rp_input) is int:
                 reset_partition_size = int(human2bytes(rp_input))
                 log.info(
                     "autoinstall: will install reset partition "
--- a/subiquity/server/controllers/tests/test_filesystem.py
+++ b/subiquity/server/controllers/tests/test_filesystem.py
@@ -546,6 +546,27 @@ class TestGuided(IsolatedAsyncioTestCase):
         self.assertEqual(None, d1p2.mount)
         self.assertEqual(DRY_RUN_RESET_SIZE, d1p2.size)
 
+    @parameterized.expand(
+        (
+            ({}, False, None),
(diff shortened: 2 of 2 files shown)


Chris-Peterson444:
Draft for now because I would like to add some tests next week to catch this in the future, but I wanted to open the PR sooner for visibility if it became a bigger issue.

Chris-Peterson444:
@medicalwei how disruptive is this bug for you? Currently you can get around this by specifying a size for the reset partition (e.g. `reset-patition: 4G` or something like this) so I've estimated this as low-priority issue.

medicalwei:
For now we only used fixed size reset partitions, so that's not a blocker issue for us.

Chris-Peterson444:
Good to hear, thanks!

subiquity/server/controllers/filesystem.py
```diff
@@ -1453,7 +1453,8 @@ async def run_autoinstall_guided(self, layout):
         rp_input = layout.get("reset-partition", None)
         if rp_input:
             reset_partition = True
-            if isinstance(rp_input, (str, int)):
+            # bool is a sublcass of int
+            if isinstance(rp_input, str) or type(rp_input).__name__ == "int":
```
ogayot, line 1457: Originally. I wanted to ask you to reverse the logic (i.e., starting with `if isinstance(rp_input, bool) ...  elif isinstance(rp_input, (str, int)`) but I'm not sure if other types besides bool were meant to be caught in the else block. So I'll let the choice up to you.

ogayot: s/sublcass/subclass/

ogayot: I supposed `type(rp_input) is int` should work as well, right?

Chris-Peterson444: yeah I think `type(...) is <type>` feels more pythonic

ogayot (approved):

PR #1974, opened 2024-04-15
curtin: desktop grub.replace_linux_default False

dbungert:
Grub provides "quiet splash" for GRUB_CMDLINE_LINUX_DEFAULT out of the box, but curtin rewrites that value and drops both "quiet" and "splash". On desktop, retain "quiet splash".

This should be configurable via autoinstall if this change is undesired.

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -517,6 +517,9 @@ class SubiquityModel:
             },
         }
 
+        if self.source.current.variant == "desktop":
+            config["grub"]["replace_linux_default"] = False
+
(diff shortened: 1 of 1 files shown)


Chris-Peterson444 (approved):
LGTM. How would you configure this via autoinstall? late commands?

dbungert:
> LGTM. How would you configure this via autoinstall? late commands?

`storage` may have a child key `grub`, and that may contain `replace_linux_default`.  I'm just setting a different default value for Desktop.

mwhudson (approved):

PR #1975, opened 2024-04-16
storage: log snapd response text when v2/systems/{system} or v2/snaps/{snap} fails

ogayot:
When the request to `v2/systems/enhanced-secureboot-desktop` (or equivalent) fails, Subiquity did not capture the reason. This makes it difficult to understand why this call sometimes fails. We now log the response text to understand better what's going on.

LP:#2061757

Also did the same change for the call to `v2/snaps/ubuntu-desktop-provision` intended to refresh the snap ; although we already know that sending two successive requests will return 409. LP:#2061756

### DIFF
--- a/apt-deps.txt
+++ b/apt-deps.txt
@@ -43,6 +43,7 @@ python3-pytest
 python3-pytest-xdist
 python3-pyudev
 python3-requests
+python3-requests-mock
 python3-requests-unixsocket
 python3-setuptools
 python3-systemd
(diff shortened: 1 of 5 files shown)


Chris-Peterson444 (approved):
Nice! TIL about `assertLogs`, although that double quote escaping weirdness is definitely frustrating. Thanks!

dbungert (approved):

mwhudson (approved):

PR #1982, opened 2024-04-25
quick-test: handle ubuntu-desktop-bootstrap

dbungert:


mwhudson (approved):
Given that I apparently was smart enough to have --inject-snap look at the snap's metadata and not the filename I think it might be less confusing all around to have this script _not_ call the new snap `subiquity_new.snap` but rather something more neutral like `new.snap`?

Chris-Peterson444 (approved):
Yes please to this

PR #1984, opened 2024-04-26
subiquity.common.serialize: add a way to have a "non-exhaustive" enum

mwhudson:
relates to #1980

### DIFF
--- a/subiquity/common/serialize.py
+++ b/subiquity/common/serialize.py
@@ -43,6 +43,13 @@ class SerializationError(Exception):
         return f"processing {self.obj}: at {p}, {self.message}"
 
 
+E = typing.TypeVar("E")
+
+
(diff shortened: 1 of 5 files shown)


dbungert (approved):
Thanks!

Looks fine, one open question, and a rebase to cleanup commits please.

mwhudson:
I rebased and cleaned up one bit of logic to hopefully more comprehensible. I don't see the open question though?

dbungert (approved):
Uh, I think I confused this with the other PR where I was suggesting to document one more possible state in the enum.  LGTM, thanks!

mwhudson:
I should tag @d-loose so he knows about the change in capabilities of the API here -- basically a dict with enum keys is serialized as an object, not a list of key, value pairs. there are no such types in the subiquity API today though.

PR #1985, opened 2024-04-26
cc-extract: don't unnecessarily access data

Chris-Peterson444:
We currently don't have plans to do anything with the _data_ of the keys that fail to validate by cloud-init; and trying to access this data has the potential to crash the installer if it doesn't exist, at least at the top level (LP:#2062988).

### DIFF
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -783,12 +783,18 @@ class SubiquityServer(Application):
                 log=log,
             )
 
-            # Raise AutoinstallError if we found any autoinstall as a cause
-            # of the schema validation error, otherwise continue
+            # Use filter_autoinstall on bad_keys to find potential autoinstall
+            # keys as the cause of the schema validation error. If so,
+            # raise AutoinstallError; else continue.
+            #
+            # Intentionally not attempting to extract bad key data since it is
+            # not guaranteed that the offending keys will be top-level (or
+            # even in?) in the combined config. Although still constructing
+            # a dict since filter_autoinstall expects a dict.
+            # LP: #2062988
 
-            # Filter only the bad keys
-            potential_autoinstall: dict[str, Any] = dict(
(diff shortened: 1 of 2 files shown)


subiquity/server/server.py
```diff
@@ -787,8 +787,13 @@ async def _extract_autoinstall_from_cloud_config(
             # of the schema validation error, otherwise continue
 
             # Filter only the bad keys
+            # Intentionally (key, None) ! Have a good plan (and reason) for
+            # getting the data of a key. It is not guaranteed that the
+            # offending keys to be top-level (or in?) in the combined config.
+            # LP: #2062988
+
             potential_autoinstall: dict[str, Any] = dict(
```
dbungert, line 795: I guess this should be a set instead

subiquity/server/tests/test_server.py
```diff
@@ -456,6 +456,22 @@ async def test_autoinstall_from_cloud_config(self, cloud_cfg, expected, throws):
...
 
+    async def test_cloud_config_extract_KeyError(self):
+        """Test autoinstall extract from cloud config resilient to missing data."""
+
+        self.server.base_schema = SubiquityServer.base_schema
+        self.pseudo_load_controllers()
+
+        with patch("subiquity.server.server.validate_cloud_init_schema") as val_mock:
+            val_mock.side_effect = CloudInitSchemaValidationError(
+                keys=["broadcat", "foobar"],
```
dbungert, line 467:
```suggestion
                keys=["broadcast", "foobar"],
```

subiquity/server/server.py
Chris-Peterson444: It needs to be a dict to pass to `filter_autoinstall`. We could re-work `filter_autoinstall` to take either a dict or a set but I feel like that starts to make it overly complex.

dbungert (approved):

dbungert (approved):

PR #1986, opened 2024-04-28
relicense code outside of subiquity to GPLv3-only

mwhudson:
Not sure of best practices for this wrt the LICENSE file and so on.


LICENSE.AGPL-3
dbungert, line 1: nitpick: https://www.gnu.org/licenses/agpl-3.0.txt is slightly different, with updated (https) urls.

dbungert (approved):

PR #1987, opened 2024-04-29
Add a way to forward user errors to the client when consuming /storage/v2 POST endpoints

ogayot:
Today, when a client sends a POST request to /storage/v2/ endpoints, if the data is inconsistent with the current state of the model, the Subiquity server process creates an error report and returns with HTTP status 500.

This is what makes the desktop installer display so many crash dialogs. We would want to a way to:
* display a user-friendly error in a dialog of the UI
* avoid creating a crash report every time

This PR allow to configure Subiquity in such a way that, in case of a user-error in /storage/v2:

* the error is sent back to the client with a 422 HTTP status (Unprocessable Content).
* no crash report is created

Here's the code we could use when consuming /storage/v2/add_boot_partition

```python
try:
    await self.endpoint.v2.add_boot_partition.POST(disk.id)
except RecoverableClientError as exc:
    self.ui.body.show_recoverable_error(str(exc)))  # Basically a function that shows an overlay
```

This is very similar that what I did for https://github.com/canonical/subiquity/pull/1227 at the time

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -16,6 +16,7 @@
 import asyncio
 import contextvars
 import inspect
+import json
 import logging
 import os
 import signal
@@ -57,6 +58,10 @@ log = logging.getLogger("subiquity.client.client")
 class Abort(Exception):
     def __init__(self, error_report_ref=None):
         self.error_report_ref = error_report_ref
+
+
+class RecoverableClientError(ValueError):
+    pass
 
 
 DEBUG_SHELL_INTRO = _(
@@ -206,7 +211,10 @@ class SubiquityClient(TuiApplication):
             raise Abort(ref)
         try:
             response.raise_for_status()
-        except aiohttp.ClientError:
+        except aiohttp.ClientError as exc:
+            if isinstance(exc, aiohttp.ClientResponseError) and exc.status == 422:
+                raise RecoverableClientError(json.loads(headers["x-error-msg"]))
+
             report = self.error_reporter.make_apport_report(
                 ErrorReportKind.SERVER_REQUEST_FAIL,
                 "request to {}".format(response.url.path),
--- a/subiquity/cmd/server.py
+++ b/subiquity/cmd/server.py
@@ -23,6 +23,7 @@ import sys
 
 import attr
 
+from subiquity.server.controllers.filesystem import set_user_error_reportable
 from subiquitycore.log import setup_logger
 
 from .common import LOGDIR, setup_environment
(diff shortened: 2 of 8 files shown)


Chris-Peterson444:
I like this idea a lot! A couple of thoughts:

- Instead of conditionally including the exception in the response dict (to avoid generating an apport report in the middleware) we should adjust the logic in the middleware to be `NonReportableException` aware IMO
- I feel like we should be making a hard distinction between non-reportable and recoverable errors. While non-reportable errors in the request handlers are potentially recoverable, non-reportable errors elsewhere are fatal to the installer. An explicit `RecoverableError` child type, or something like that, would at the very least make the situation less ambiguous on read I think. 
- I like the idea of conditionally changing the error type, although I assume the flag is just for testing? We could probably use this to start implementing the behavior on server before desktop is ready to consume it.

dbungert:
The core idea is solid - we absolutely need to turn these server-process exceptions into recoverable errors - but I want to confirm that the implementation is one u-d-b can use.

@d-loose - do you mind trying this branch out and confirming that you can consume these runtime errors?

dbungert:
> I like the idea of conditionally changing the error type, although I assume the flag is just for testing? 

We wouldn't want the flag around long-term, I see it as a feature flag so that we can do a main branch merge and let clients opt-in to try it out, and when we're convinced, make it the default.

d-loose:
Thanks, this looks great! Just played around with this branch a bit and can confirm that this works well for us. We'll need to figure out some general error handling and UX details, but in principle we can catch subiquity errors with a `4xx` status code and update the UI accordingly.
I'd assume subiquity would then simply discard the request causing the error, right?

ogayot:
> Thanks, this looks great! Just played around with this branch a bit and can confirm that this works well for us. We'll need to figure out some general error handling and UX details, but in principle we can catch subiquity errors with a `4xx` status code and update the UI accordingly. I'd assume subiquity would then simply discard the request causing the error, right?

Yes, that is the plan. I guess if a request fails but we can't revert it, we'll continue sending a HTTP 500 + crash report.

dbungert:
> Yes, that is the plan. I guess if a request fails but we can't revert it, we'll continue sending a HTTP 500 + crash report.

Non-blocking for this PR, but I guess we want a transaction-like system so that we don't half-update the model and then error out.

mwhudson (approved):
LGTM

ogayot:
> * Instead of conditionally including the exception in the response dict (to avoid generating an apport report in the middleware) we should adjust the logic in the middleware to be `NonReportableException` aware IMO
> 
> * I feel like we should be making a hard distinction between non-reportable and recoverable errors. While non-reportable errors in the request handlers are potentially recoverable, non-reportable errors elsewhere are fatal to the installer. An explicit `RecoverableError` child type, or something like that, would at the very least make the situation less ambiguous on read I think.

Thanks @Chris-Peterson444 ! Sorry for taking time to respond.

I've implemented the changes suggested. The middleware now knows about `NonReportableException` and will conditionally create the error report based on the type of exception encountered.

I also introduced RecoverableError, which will be reworked in a future PR. It felt wrong to me to add an inheritance relationship between `NonReportableException` and `RecoverableError` so I've made NonReportableSVE inherit from both.

Chris-Peterson444 (approved):
LGTM, thanks!

> I also introduced RecoverableError, which will be reworked in a future PR. It felt wrong to me to add an inheritance relationship between NonReportableException and RecoverableError so I've made NonReportableSVE inherit from both.

I think this makes sense. I didn't have a particular solution in mind previously but I just wanted to point out that it could potentially become confusing later trying to differentiate the different error classes and how they should behave. I think the separate parent classes is the right direction.

PR #1988, opened 2024-05-01
apt: fix mountpoint-install-tree test write dir

dbungert:
test_run_apt_config_check currently writes to curdir, update that to a tempdir instead.

`git diff -w` makes this easier to read, it's a two line change ignoring indent changes.


Chris-Peterson444 (approved):
`git diff -w` is nifty, thanks for the tip. LGTM.

PR #1993, opened 2024-05-06
allow system definition to be in live layer

mwhudson:
Rather than always assuming it has to be mounted in from the layer to be installed (snapd will now populate the seed in the target system when its install API is called if it is empty).

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -429,11 +429,11 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
 
     def _maybe_disable_encryption(self, info: VariationInfo) -> None:
         if self.model.bootloader != Bootloader.UEFI:
-            log.debug("Disabling core boot based install options on non-UEFI " "system")
+            log.debug("Disabling core boot based install options on non-UEFI system")
             info.capability_info.disallow_if(
                 lambda cap: cap.is_core_boot(),
                 GuidedDisallowedCapabilityReason.NOT_UEFI,
-                _("Enhanced secure boot options only available on UEFI " "systems."),
+                _("Enhanced secure boot options only available on UEFI systems."),
             )
         search_drivers = self.app.base_model.source.search_drivers
         if search_drivers is not SEARCH_DRIVERS_AUTOINSTALL_DEFAULT and search_drivers:
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -340,6 +340,13 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
     async def _get_system(
         self, variation_name, label
     ) -> Tuple[Optional[SystemDetails], bool]:
+        """Return system information for a given system label.
+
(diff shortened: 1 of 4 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -336,20 +337,29 @@ async def _unmount_systems_dir(self):
...
-        except requests.exceptions.HTTPError as http_err:
-            log.warning("v2/systems/%s returned %s", label, http_err.response.text)
-            raise
-        finally:
-            await self._unmount_systems_dir()
-        log.debug("got system %s", system)
-        return system
+    async def _get_system(
+        self, variation_name, label
+    ) -> Tuple[Optional[SystemDetails], bool]:
```
dbungert, line 342: I'd like a docstring or similar clarifying the intent of the return value, please

subiquity/server/controllers/filesystem.py
```diff
@@ -410,14 +420,37 @@ def disallowed_encryption(msg):
 
         return info
 
+    def _maybe_disable_encryption(self, info: VariationInfo) -> None:
+        if self.model.bootloader != Bootloader.UEFI:
+            log.debug("Disabling core boot based install options on non-UEFI " "system")
```
dbungert, line 425:
```suggestion
            log.debug("Disabling core boot based install options on non-UEFI system")
```

subiquity/server/controllers/filesystem.py
```diff
@@ -410,14 +420,37 @@ def disallowed_encryption(msg):
...
+            info.capability_info.disallow_if(
+                lambda cap: cap.is_core_boot(),
+                GuidedDisallowedCapabilityReason.NOT_UEFI,
+                _("Enhanced secure boot options only available on UEFI " "systems."),
```
dbungert, line 429: same

dbungert (approved):
What sort of things need to happen to test this?  ppa build of matching livecd-rootfs changes + a slimy check of existing ISOs?

subiquity/server/controllers/filesystem.py
mwhudson: Added a docstring. Would prefer to do it in a more type-y way but that just seems verbose.

subiquity/server/controllers/filesystem.py
mwhudson: Huh I thought our black config did that!

subiquity/server/controllers/filesystem.py
mwhudson: same

mwhudson:
> What sort of things need to happen to test this? ppa build of matching livecd-rootfs changes + a slimy check of existing ISOs?

Hmm yes, good question. One way involves me getting my ubuntu core installer stuff into shape for review but we can also do it by moving the system for the TPMFDE stuff into the live layer, which I think is pretty easy. I'll get back to you when I have that test-built!

dbungert (approved):

mwhudson:
OK tested this with a hacked up desktop ISO and the subiquity part works (there are some bits and pieces that need to change with the desktop ISO side but that's OK).

PR #1997, opened 2024-05-23
fix: cloud-init datasource_list must be one-line for ds-identify

blackboxsw:
Ensure datasource_list configuration from subiquity is a single line: datasource_list: [None] to support ds-identify limitation in parsing single-line datasource_list values.

cloud-init's ds-identify doesn't actually support multi-line YAML and never has.

We can see this behavior in /run/cloud-init/ds-identify.log which opts to use a default datsource_list and logs:

  /etcloud/cloud.cfg.d/99-installer.cfg set datasource_list:
  no datasource_list found, using default: ....

This issue doesn't affect subiquity in practice because it also places /etc/cloud/ds-identify.cfg containing `policy: enabled`. Which means cloud-init will be enabled regardless of whether ds-identify finds a valid datasource.

We see this in /run/cloud-init/ds-identify.log with the following logs:
```
  DSNAME=
  ...
  mode=enabled returning 0
```
Once cloud-init python code runs and merges system configuration in /etc/cloud/cloud.cfg.d with a proper YAML parser, it correctly parses multi-line YAML config and successfully limits the merged datasource_list: [ None ] provided in
/etc/cloud/cloud.cfg.d/99-installer.yaml.

If subiquity were to drop the /etc/cloud/ds-identify.cfg file in the future, cloud-init would actually be disabled on first boot because it never detected a viable datasource during systemd generator timeframe and ds-identify policy doesn't mandate being enabled by default.

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -420,10 +420,13 @@ class SubiquityModel:
         return (before, during)
 
     def _cloud_init_files(self):
-        # TODO, this should be moved to the in-target cloud-config seed so on
-        # first boot of the target, it reconfigures datasource_list to none
-        # for subsequent boots.
-        # (mwhudson does not entirely know what the above means!)
+        # cloud_init_files are placed in-target to be run on first boot due to
+        # /etc/cloud/cloud.cfg.d/99-installer.cfg which limits viable
+        # datasource_list to None. This only allow cloud-init to discover
+        # subiquity-defined configuration on first boot.
+        # During first boot, write_files creates /etc/cloud/cloud-init.disabled
+        # which will prevent cloud-init from running again on subsequent boots
+        # without first running: cloud-init clean.
         userdata = "#cloud-config\n" + yaml.dump(self._cloud_init_config())
         metadata = {"instance-id": str(uuid.uuid4())}
         config = yaml.dump(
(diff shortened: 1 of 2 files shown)


subiquity/models/subiquity.py
```diff
@@ -435,7 +435,10 @@ def _cloud_init_files(self):
                         "metadata": metadata,
                     },
                 },
-            }
+            },
```
dbungert, line 438: This all looks fine Chad, thanks!
While we're here, do you mind fact-checking the above TODO?

dbungert (approved):

subiquity/models/subiquity.py
blackboxsw: @dbungert nice. looking to drop this comment, I'll mention it in the commit message too. This historic comment was from a point in time 2017 36bcc539c when cloud-init was only seeding /var/lib/cloud/seed/nocloud-net/(meta-data|user-data) files via write_files in the target in POST_INSTALL_CONFIG.  This is no longer relevant as subiquity is providing DataSourceNone configuration now with merged user_data, also sets `datasource_list: [None]` for first boot and writes `/etc/cloud/cloud-init.disabled` to prevent cloud-init from ever running on subsequent boots without someone first running `cloud-init clean` which will remove subiquity's supplemental /etc/cloud/cloud-init.disabled flag file and the 99-installer.cfg which limits datasource_list to None.

dbungert: Excellent, I appreciate it.

dbungert (approved):

PR #1999, opened 2024-05-28
Replace calls to v1 guided and v1 reset with v2 equivalents

ogayot:
This is a first step in the switch from v1 storage API to the v2 alternative.

This also demonstrates the use of GET /storage (which is a v1 storage endpoint) as a way to introduce a compatibility layer during development. Going forward, we should drop calls to v1 GET /storage as well but going step by step should make the whole process simpler.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -1207,6 +1207,12 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         log.debug(data)
         self.locked_probe_data = True
         await self.guided(data)
+        if not data.capability.supports_manual_customization():
+            # Going forward, we probably want the client to call POST
+            # /storage/v2 when they are done ; rather than conditionally
+            # marking the model configured here. This requires a way to tell
+            # the client whether manual customization is possible though.
+            await self.configured()
         return await self.v2_guided_GET()
 
     async def v2_reformat_disk_POST(self, data: ReformatDisk) -> StorageResponseV2:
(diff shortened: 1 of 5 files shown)


dbungert (approved):
Looks good, very happy to see progress on this.

subiquity/server/controllers/filesystem.py
```diff
@@ -998,16 +998,10 @@ async def finish_install(self, context):
         )
 
     async def guided_POST(self, data: GuidedChoiceV2) -> StorageResponse:
-        log.debug(data)
-        await self.guided(data)
-        if not data.capability.supports_manual_customization():
-            await self.configured()
-        return self._done_response()
+        raise NotImplementedError
```
mwhudson, line 1001: Better to just remove these from the API than do this?

mwhudson (approved):
Love to see more progress in this direction! I'd encourage you to be more bold in just deleting API methods that are no longer called though :-)

(Side note, I notice we now have `GuidedChoiceV2` and `GuidedStorageResponseV2` types but removed the non-V2 types -- I guess we can rename them to remove the V2s at some point?)

subiquity/server/controllers/filesystem.py
ogayot: Yes, no problem!

ogayot:
> (Side note, I notice we now have GuidedChoiceV2 and GuidedStorageResponseV2 types but removed the non-V2 types -- I guess we can rename them to remove the V2s at some point?)

Good point. I'll do the renaming (maybe also with a `v2` namespace?) once I'm sure we don't have any V1 types hanging around.

PR #2004, opened 2024-06-04
List of matchers

dbungert:
Add unit tests for existing functionality (some overlap with `TestAutoInstallConfig` but some is new)
Permit match specs to be done as a list of match rules.
Refactor `disk_for_match` as well to make it simpler to follow.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1591,8 +1591,6 @@ class FilesystemModel:
         )
 
     def _make_matchers(self, match):
-        matchers = []
-
         def _udev_val(disk, key):
             return self._probe_data["blockdev"].get(disk.path, {}).get(key, "")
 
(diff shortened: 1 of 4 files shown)


dbungert:
@medicalwei FYI

subiquity/models/filesystem.py
```diff
@@ -1638,31 +1644,38 @@ def match_install_media(disk):
...
+            disks.sort(key=lambda d: d.size)
+        elif match.get("size") == "largest":
+            disks.sort(key=lambda d: d.size, reverse=True)
+        return disks
+
+    def _filtered_matches(self, disks: Sequence[_Device], match: dict):
+        matchers = self._make_matchers(match)
+        return [disk for disk in disks if all(match_fn(disk) for match_fn in matchers)]
+
+    def disk_for_match(
```
mwhudson, line 1663: Nothing really to do with your change but do any of the methods related to this actually refer to self? Maybe we can move them to a helper module at some point.

mwhudson (approved):
Looks good, one tangential comment

PR #2006, opened 2024-06-05
Cherry-pick fix for Netboot bug LP: #2062988 to ubuntu/noble

Chris-Peterson444:
I can confirm the fix works on the stable branch after local testing. For both Desktop and Server, and both BIOS and UEFI boot on each.

### DIFF
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -783,12 +783,18 @@ class SubiquityServer(Application):
                 log=log,
             )
 
-            # Raise AutoinstallError if we found any autoinstall as a cause
-            # of the schema validation error, otherwise continue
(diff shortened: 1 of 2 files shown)


Chris-Peterson444:
I was able to test this with QEMU's tftp server. Note that if you're trying to test the desktop image you'll need to provide the VM ~14G of ram.

mwhudson (approved):

ogayot (approved):
LGTM

I took a peek at `filter_autoinstall` and noticed it a potential problem so I filed https://bugs.launchpad.net/subiquity/+bug/2068603 but it is not a blocker for this PR in any way.

dbungert (approved):

PR #2008, opened 2024-06-06
ai-filter: be aware of autoinstall_key_alias

Chris-Peterson444:
Make sure we catch autoinstall key name aliases so we don't incorrectly warn (or eventually error) on actually correct key names (LP:#2068603).


ogayot (approved):
Thanks!

PR #2010, opened 2024-06-11
Move storage types to a different module

ogayot:
The types.py module is getting difficult to navigate and I would like to split it into multiple units. This PR moves all the storage related types to a different module.

I have already opened a PR in udi to support multiple files in the generator: https://github.com/canonical/ubuntu-desktop-provision/pull/765

which should allow to apply the following in https://github.com/canonical/ubuntu-desktop-provision/blob/main/packages/subiquity_client/Makefile:

```diff
-types_py = ../subiquity/subiquity/common/types.py
+types_py = ../subiquity/subiquity/common/types/__init__.py \
+           ../subiquity/subiquity/common/types/storage.py
```


Chris-Peterson444 (approved):
Nice work! I agree it was getting unwieldy. LGTM.

mwhudson (approved):

PR #2013, opened 2024-06-14
Late commands vs done

dbungert:


subiquity/common/types/__init__.py
```diff
@@ -91,6 +91,8 @@ class ApplicationState(enum.Enum):
     UU_RUNNING = enum.auto()
     UU_CANCELLING = enum.auto()
```
ogayot, line 92: I wonder if we want to drop this as well.

ogayot (commented):
I would like to separate the removal of the unattended-upgrades cancellation from the late-commands status change. I will split the commits into two accordingly.

subiquity/common/types/__init__.py
ogayot: @d-loose Do you see a problem with removing one this value from the enumeration?

d-loose: Not at all, we're not using it anywhere in the desktop installer

ogayot:
So I just tested this PR against the desktop installer. Although the meta/status reports `{"state": "LATE_COMMANDS"}`, the UI would still show the Installation complete screen (and the restart now button). So it requires more work.

subiquity/common/types/__init__.py
```diff
@@ -86,10 +86,11 @@ class ApplicationState(enum.Enum):
     WAITING = enum.auto()
     RUNNING = enum.auto()
 
-    # States reported while unattended-upgrades is running.
-    # TODO: check if these should be dropped in favor of RUNNING.
+    # State reported while unattended-upgrades is running.
+    # TODO: check if this should be dropped in favor of RUNNING.
```
mwhudson, line 90: They should only be distinct if a UI is going to care about the distinction and I think the way things are going the answer there is "no"?

subiquity/server/controllers/install.py
```diff
@@ -758,8 +751,7 @@ async def run_unattended_upgrades(self, context, policy):
         apt_conf_path = Path(aptdir) / "zzzz-temp-installer-unattended-upgrade"
         apt_conf_path.write_bytes(apt_conf_contents)
         try:
-            self.unattended_upgrades_ctx = context
-            self.unattended_upgrades_cmd = await start_curtin_command(
+            uu_cmd = await start_curtin_command(
```
mwhudson, line 754: This could be run_curtin_command I think?

ogayot: Good point!

subiquity/common/types/__init__.py
ogayot: Currently, we have a distinction in the text UI so I'm tempted to leave it as is for now, wdyt?

mwhudson: ok then!

Chris-Peterson444 (approved):

dbungert (commented):
LGTM.  I can't Approve since I opened the PR ;)

PR #2014, opened 2024-06-14
adopt curtin capabilities related to kernel in lowest layer

dbungert:


Chris-Peterson444 (approved):

subiquity/models/kernel.py
```diff
@@ -40,11 +40,9 @@ def needed_kernel(self) -> Optional[str]:
...
-
-        return {
-            "kernel": {
-                "package": self.needed_kernel,
-            },
-        }
+            result["kernel"]["install"] = False
+        else:
+            result["kernel"]["package"] = self.needed_kernel
+        return result
```
mwhudson, line 48: tangent: would be nice if we could use the rich curtin config types we've added here wouldn't it ...

subiquity/models/kernel.py
```diff
@@ -40,11 +40,9 @@ def needed_kernel(self) -> Optional[str]:
         return self.metapkg_name
 
     def render(self):
+        result = {"kernel": {"remove_existing": True}}
         if self.curthooks_no_install:
```
mwhudson, line 44: I don't think any core remains to set curthooks_no_install to True, so can this be tidied up?

mwhudson (commented):
nice simplification, I think you can go a touch further though

subiquity/models/kernel.py
dbungert: I feel tentative on if it's going to be useful or not, so I'm dropping it but isolating that change to a distinct commit.

PR #2015, opened 2024-06-15
network: ensure disconnected interfaces are disabled

Chris-Peterson444:
According to LP:#2063331, Subiquity policy should be to disable interfaces which are not automatically configured in the installation environment on the target machine. Subiquity already has this logic mostly in place, via a call to `BaseNetworkController.update_initial_configs` in the right places. The subiquitycore network controller calls this function on the first view of the network screen, but the non-core server network controller doesn't. Existing logic and comments in the non-core controller imply this is already the expected behavior in non-core installs as well (see comments in the existing autoinstall logic). This PR adds the `update_initial_configs` call in two places in the server network controller:

1. In the `GET` request handler. This is the endpoint clients should call to get information to populate the network screen, so we should update the device configurations before sending them to the client.
2. In the `configured` function. It's possible to complete the install by simply marking the network controller as configured without ever interacting with it. However, because the screen was not "shown" (i.e. server received a GET request), the target system will end up with the not updated network config. The Desktop installer specifically does this, likely to avoid changing control of the live session's network from NetworkManager to networkd. Calling `update_initial_config` here without calling `apply_config` will allow the target system to receive the correct config without modifying the live sessions network settings.

Lastly, I included a fix for `NetworkDev.ip_networks_for_version` that makes sure static configuration information is really removed from the config when setting ipv4 and ipv6 to either disabled or automatic. Existing behavior would allow static configuration options to be appended to the existing config when resubmitting the form and also persist when presented as "disabled" in the UI.


Chris-Peterson444:
There are some other improvements I think we could make to make the network screen more robust but I'll open a separate PR for those. These changes alone should resolve [LP:#2063331](https://launchpad.net/bugs/2063331).

subiquity/server/controllers/network.py
```diff
@@ -236,6 +237,15 @@ async def GET(self) -> NetworkStatus:
         )
 
     async def configured(self):
+        # There may be some instances in which the network controller
+        # is interactive but is only ever marked configured (through a
+        # POST to either /network or /meta/mark_configured) and is never
+        # interacted with otherwise (via GET) such that we don't disable
+        # interfaces which don't have a global IP.
+        if self.interactive() and not self.view_shown:
+            self.view_shown = True
```
ogayot, line 246: The name of the variable does not reflect what happens (anymore?) does it?

Chris-Peterson444: I should probably drop modifying that variable in that clause. It was to avoid re-invoking the `update_inital_config` logic if the client later did a `GET` but that should probably be fine.

ogayot (approved):

dbungert (approved):

PR #2016, opened 2024-06-18
Add instructions for netboot with QEMU

Chris-Peterson444:
Publicly documenting my workflow for leveraging QEMU's built in tftp server functionality to perform much faster iteration with netboot testing. This practically emulates the ISO tracker instructions but allows for much faster downloads (in the VM) and testing with modified ISOs.

I put this in a separate file from the other testing instructions so it wouldn't blow up the size. I wonder though if we should make some sort of `test_instructions` directory, sort of like the `test_data` directory, for manual instructions like this.

### DIFF
--- a/NETBOOT_TESTS.md
+++ b/NETBOOT_TESTS.md
@@ -1,0 +1,71 @@
+# Testing changes with netboot
+
+The ISO tracker already contains [test cases for performing netboot](https://iso.qa.ubuntu.com/qatracker/milestones/454/builds/304977/testcases/1734/results), but the setup can be cumbersome when iterating. This is a quick guide for using QEMU and Apache to perform VM-based netboot tests.
+
+## Serving your custom ISO
+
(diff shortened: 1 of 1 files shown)


NETBOOT_TESTS.md
```diff
@@ -0,0 +1,71 @@
...
+
+
+## Starting QEMU
+
+QEMU can start up a tftp server by passing the tftp directory to one of the NICs on the virtual machine.
+
+- When using UEFI bios, pass `-nic <other options>,tftp=/tmp/tftp/amd64,bootfile=bootx64.efi`.
+- When using non-UEFI bios, pass `-nic <other options>,tftp=/tmp/tftp/amd64,bootfile=pxelinux.0`.
+
+The `kvm-test.py` script currently doesn't support netbooting (it requires passing an ISO to mount to `/cdrom`), but you can use it to generate generate the command line arguments and then modify the `-nic` argument accordingly.
```
ogayot, line 59: s/generate generate/generate/

dbungert (approved):

ogayot (approved):

PR #2018, opened 2024-06-25
Merge late-commands status fix to ubuntu/noble

ogayot:
This is a cherry pick of patches from https://github.com/canonical/subiquity/pull/2013

Essentially, this introduces a new state "LATE_COMMANDS" before moving to "DONE" ; so that a reboot is not performed while late commands are still executing.

It also removes the ability to cancel the unattended-upgrades step. We can choose not to include this change if needed.

### DIFF
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -92,6 +92,8 @@ class ApplicationState(enum.Enum):
     UU_RUNNING = enum.auto()
     UU_CANCELLING = enum.auto()
 
+    LATE_COMMANDS = enum.auto()
+
     # Final state
     DONE = enum.auto()
     ERROR = enum.auto()
--- a/subiquity/common/types.py
+++ b/subiquity/common/types.py
@@ -87,10 +87,9 @@ class ApplicationState(enum.Enum):
     WAITING = enum.auto()
     RUNNING = enum.auto()
 
-    # States reported while unattended-upgrades is running.
-    # TODO: check if these should be dropped in favor of RUNNING.
+    # State reported while unattended-upgrades is running.
+    # TODO: check if this should be dropped in favor of RUNNING.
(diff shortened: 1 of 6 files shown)


ogayot:
I'm working on a fix for the CI since the change breaks system-setup (which we removed from main).

ogayot:
I added the following to system-setup to fix the CI:

```python
            self.app.update_state(ApplicationState.LATE_COMMANDS)
            await self.app.controllers.Late.run()
```
This was not needed on main because we dropped system-setup.  Before the patch, the LateCommands controller would automatically run after the application would enter the DONE state ; which obviously we don't want to do anymore.

dbungert (approved):
LGTM but I would like to hear that VM testing has been done with autoinstall with an interesting late-commands to show that this is working correctly.

ogayot:
> LGTM but I would like to hear that VM testing has been done with autoinstall with an interesting late-commands to show that this is working correctly.

Before opening the PR, I did some testing on the server installer ; and it worked as intended. That said, I didn't test autoinstall specifically so I might do more testing tomorrow. 

For my testing, I relied on a script placed under /etc/subiquity/postinst.d since Subiquity has a `run-parts`  command that is a builtin of the LateController:

```
$ cat /etc/subiquity/postinst.d/20-wait-fifo
#!/bin/bash
mkfifo /tmp/fifo
cat /tmp/fifo   # blocks until something writes to /tmp/fifo
```

The status was LATE_COMMANDS until I manually wrote to /tmp/fifo.

ogayot:
Okay, I did testing on the desktop installer, first using the etc/subiquity/postinst.d trick and then using autoinstall + late-commands. 

This worked as intended.

Chris-Peterson444 (approved):
LGTM. Thanks for the tests!

PR #2019, opened 2024-06-25
Use TypedDict-s for storage specs

ogayot:
Storage "specs" are deprecated but they exist today and it is difficult to know what variables are expected to be present. By using typing.TypedDict, we can make it more explicit what keys are expected and what types.

Running the type checker after the change revealed the following error so I added a FIXME at first (and finally removed the "size" key)

```
subiquity/server/controllers/filesystem.py:1259: error: Extra key "size" for TypedDict "FileSystemSpec"  [typeddict-unknown-key]
```

### DIFF
--- a/subiquity/common/filesystem/spec.py
+++ b/subiquity/common/filesystem/spec.py
@@ -1,0 +1,74 @@
+# Copyright 2024 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Types defined in this module are deprecated and will at some point be
+replaced by `attr.s` structures or similar. """
+
+
+import typing
+from typing import Any, TypedDict
+
+try:
+    from typing import Required
+except ImportError:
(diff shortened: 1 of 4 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -1252,12 +1256,15 @@ async def v2_add_partition_POST(self, data: AddPartitionV2) -> StorageResponseV2
             requested_size = data.gap.size
         # empty string is an unformatted partition
         fstype = data.partition.format or None
-        spec = {
-            "size": requested_size,
-            "fstype": fstype,
-            "mount": data.partition.mount,
+        spec: FileSystemSpec = {
+            "size": requested_size,  # FIXME this looks unused, watch out!
```
mwhudson, line 1260: Uh yeah this can probably go, right? Score +1 for static analysis

mwhudson (approved):

subiquity/server/controllers/filesystem.py
ogayot: Right, I finally dropped the key instead of adding the FIXME

subiquity/common/filesystem/spec.py
```diff
@@ -0,0 +1,74 @@
...
+        "on-remote-storage": bool,
+    },
+    total=False,
+)
+
+
+class LogicalVolumeSpec(FileSystemSpec, total=False):
+    name: str
+    size: int
+
```
Chris-Peterson444, line 70: Is it just a matter of style to use the alternative syntax for the base class, or am I missing something?

Chris-Peterson444 (approved):
Yes please to more typing. One non-blocking question. LGTM!

subiquity/common/filesystem/spec.py
ogayot: I think the class syntax is better but it doesn't work with attributes that have "-" in their name. So I've used the other syntax when I couldn't use the former 😬

Chris-Peterson444: Aah right. Ugh that is frustrating. Thanks!

PR #2020, opened 2024-06-28
Draft: Add server v2 storage endpoints

ogayot:
NOTE: This isn't ready for review yet.


dbungert:
So far this is reasonable.  A general bit of feedback is that it is interesting that we are consistently transforming the incoming data to the outbound spec with validation and a bit of consultation to the model - I think there is little dependence on the controller otherwise.

PR #2021, opened 2024-06-28
Add /storage/v2 deletion endpoints for Raid, VGs and LVs

ogayot:
This PR adds storage v2 endpoints to delete LVs, VGs and Raids

* DELETE /storage/v2/raid?id="raid-0"
* DELETE /storage/v2/volume_group?id="lvm-volgroup-0"
* DELETE /storage/v2/logical_volume?id="lvm-partition-0"

I've decided to use a bit more of a REST-like approach where the `volume_group`, `raid`, and `logical_volume` classes can be reused for POST, GET (and PATCH or PUT).

Expectation:

```python
class volume_group:
    def GET(id: str) -> VolumeGroup: ...
    def POST(data: Payload[AddVolumeGroupV2]) -> StorageResponseV2: ...
    def PATCH(data: Payload[ModifyVolumeGroupV2]) -> StorageResponseV2 ...
    def DELETE(id: str) -> StorageResponseV2: ...
```


Chris-Peterson444 (commented):
This makes sense to me I think. Couple of suggestions:

- `s/virtual_group/volume_group/` in the third commit message

- > this will cascade-delete LVs as well

  I think this is worth putting as a comment in the handler too

- Is it feasible to add some API tests for these?

ogayot:
>     * `s/virtual_group/volume_group/` in the third commit message

Thanks! I keep making this mistake :upside_down_face: 

>     * > this will cascade-delete LVs as well
>       
>       I think this is worth putting as a comment in the handler too

Good point, will add this!

>     * Is it feasible to add some API tests for these?

I think it will be easier to do API testing after I land the GET and POST endpoints. But I will try to add unit tests that use `*_DELETE()` methods in the meantime.

Chris-Peterson444 (approved):
> I think it will be easier to do API testing after I land the GET and POST endpoints. But I will try to add unit tests that use *_DELETE() methods in the meantime.

Ah yes haha the probably makes more sense! I'm fine with waiting on those then.

PR #2023, opened 2024-07-03
storage: use probert to determine if firmware supports NVMe/TCP booting

ogayot:
Using the new probert-firmware, we now check if the firmware happens to be EDK II built using the timberland-sig tree. Of course, this won't be enough going forward because not all firmware that support NVMe/TCP booting will be EDK II ; but it should be enough for the PoC with EDK2.

Alternatively, one can supply the --[no-]supports-nvme-tcp-booting to act as if the firmware did/did not support NVMe/TCP booting.

Depends on canonical/probert#149

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -212,7 +212,7 @@ parts:
 
     source: https://github.com/canonical/probert.git
     source-type: git
-    source-commit: f34ae3c3f942a9c479fe928911053a302e146251
+    source-commit: 83d25c873a4a7d93c158d76da86cb9c612c29572
 
     override-build: *pyinstall
 
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -1099,6 +1099,19 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
     async def generate_recovery_key_GET(self) -> str:
         return self.model.generate_recovery_key()
 
+    async def supports_nvme_tcp_booting_GET(self, wait: bool = False) -> Optional[bool]:
+        if self.app.opts.supports_nvme_tcp_booting is not None:
+            # No need to wait for the task to finish if the CLI arg is present.
+            return self.supports_nvme_tcp_booting
+
+        if wait:
+            await self._probe_firmware_task.wait()
+
+        if not self._probe_firmware_task.done():
+            return None
+
+        return self.supports_nvme_tcp_booting
+
     async def v2_GET(
         self,
         wait: bool = False,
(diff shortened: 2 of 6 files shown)


snapcraft.yaml
```diff
@@ -210,9 +210,9 @@ parts:
   probert:
     plugin: nil
 
-    source: https://github.com/canonical/probert.git
```
ogayot, line 213: TODO update with the right source and commit before merging.

subiquity/server/controllers/filesystem.py
```diff
@@ -1423,6 +1438,29 @@ async def _probe(self, *, context=None):
...
             break
 
+    async def _probe_firmware(self) -> None:
+        fw = await self.app.prober.get_firmware()
+
+        log.debug("detected firmware information: %s", fw)
+
+        edk2_timberland_sig = {
+            "bios-vendor": "EFI Development Kit II / OVMF",
+            "bios-version": "0.0.0",
```
dbungert, line 1448: should date be a relative comparison?   Is a version comparison helpful if the base value is 0.0.0 - what else could we see other than `None` or `""` or something higher?

ogayot: I am using strict comparisons because `["EFI Development Kit II / OVMF" - "0.0.0" - "02/06/2015"] ` are the only known good values (currently) for the firmware to have boot over NVMe/TCP support.

> should date be a relative comparison?

Probably not. If it is more recent, it probably indicates a build of EDK 2 from the official tree (without the NVMe/TCP patches). 

> Is a version comparison helpful if the base value is 0.0.0 - what else could we see other than None or "" or something higher?

I think bios-version is more or less a free-text field. For instance, on my machine it is `R10ET39W (1.24 )` and when using /usr/share/qemu/OVMF.fd, it is set to the `ovmf` package version e.g., `2024.02-2` [1].

NOTE: the bios-vendor, bios-version and bios-release-date can be customized when building edk2. For the PoC, we can add a step to configure a different bios-vendor/bios-version before building ; and update this code accordingly, e.g.,

```diff
+     poc_firmware = {"bios-version": "0.0.0 + NVMe/TCP patches"}
-     if set(edk2_timberland_sig.items()).issubset(set(fw.items())):
+     if set(edk2_timberland_sig.items()).issubset(set(fw.items())) or
+        set(poc_firmware.items()).issubset(set(fw.items())):
             log.debug("firmware seems to support booting with NVMe/TCP")
             assume_supported = True

[1] https://pastebin.ubuntu.com/p/DmTFJNcyKq/

ogayot:
I added the /storage/supports_nvme_tcp_booting endpoint since the client needs the information as well.

subiquity/server/controllers/filesystem.py
dbungert: OK, I'm fine with this strict-looking check to get started, we'll probably need to revisit this in the future as the solution matures but that's OK.

dbungert (approved):

PR #2025, opened 2024-07-08
NVMe/TCP - disable bootfs & ESP restrictions if NVMe/TCP booting is supported

ogayot:
Currently, when installing to a NVMe/TCP drive, the bootfs and ESP must be placed on local storage. When the firmware supports NVMe/TCP booting though, we should drop this restriction.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1480,11 +1480,22 @@ class FilesystemModel:
         else:
             return Bootloader.BIOS
 
-    def __init__(self, bootloader=None, *, root: str):
+    def __init__(
+        self,
+        bootloader=None,
+        *,
+        root: str,
+        opt_supports_nvme_tcp_booting: bool | None = None,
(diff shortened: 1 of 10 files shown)


subiquity/models/filesystem.py
```diff
@@ -2266,13 +2291,18 @@ def is_boot_mounted(self) -> bool:
     def is_bootfs_on_remote_storage(self) -> bool:
         return self._mount_for_path("/boot").device.volume.on_remote_storage()
 
+    def _can_install_remote(self) -> bool:
+        if self.supports_nvme_tcp_booting:
+            return True
+
```
dbungert, line 2297: Quick comment explaining this please.

dbungert (approved):

subiquity/models/filesystem.py
dbungert: Thanks!

Chris-Peterson444 (approved):

PR #2026, opened 2024-07-11
ubuntu-drivers debug wrapper 

Chris-Peterson444:
Introduces:
- A  debug mode for wrapping all of the ubuntu-drivers calls with a umockdev wrapper script (Closes #1959)
- A sub-module, `subiquity.system_scripts`, to store scripts which subiquity can write to the system and call externally.


Chris-Peterson444:
After out of band discussion, the system_scripts approach isn't what we want to go with. There's a different approach with just using binaries vendored in `/bin` that I'll investigate alongside some other work. Marking this PR as draft until I refactor based on that.

scripts/kvm-test.py
```diff
@@ -590,6 +596,12 @@ def install(ctx):
             if ctx.args.update:
                 appends.append('subiquity-channel=' + ctx.args.update)
 
+            if ctx.args.fake_pci_devices:
+                appends.append('fake-pci-devices')
```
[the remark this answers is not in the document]
Chris-Peterson444: Ah yes, I like that better. Done, thanks.

.flake8
```diff
@@ -3,3 +3,5 @@
 builtins = _
 max-line-length = 88
 extend-ignore = E203
+exclude =
+    subiquity/system_scripts
```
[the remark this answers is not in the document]
Chris-Peterson444: Ack. I guess this is all moot since I'm going to tear this out anyways, but I was concerned something like

```python
scripts = """\
some text
                                          really long line  # noqa
                                          another really long line  # noqa
"""
```
would be weird since it technically embeds the comment in the string, but you can just do
```python
scripts = """\
some text
                                          really long line
                                          another really long line
"""  # noqa
```
it turns out.

subiquity/server/ubuntu_drivers.py
```diff
@@ -168,6 +170,98 @@ async def list_oem(self, root_dir: str, context) -> List[str]:
...
+    a modified ISOs with the packages added to the pool.
+    """
+
+    def __init__(self, app, gpgpu: bool) -> None:
+        super().__init__(app, gpgpu)
+
+        self.wrapper_script_source: str = umockdev_wrapper.script
+        self.script_path = "/tmp/umockdev_wrapper.py"
+
+        self.dev_config = configs.broadcom
```
dbungert, line 188: This is a bit hardcoded which means that we can't actually use the other config blobs, such as dell_certified_with_nvidia, without source modification.  Thoughts?

Chris-Peterson444: I'm thinking with moving away from the system_scripts module, we have a default config like:
```python
        self.dev_config = {
            "devices": [
                {
                    "modalias": "pci:v000014E4d00004353sv00sd01bc02sc80i00",
                    "vendor": "0x14E4",
                    "device": "0x4353",
                }
            ]
        }
```
which gets written on boot to a known location (i.e. `/tmp/umockdev_config.yaml`) and then we just modify that with early-commands or similar before continuing with the install. This is also the inspiration for 1d5654510893cc5813da6d72113e359c09090cb4.  What do you think?

Chris-Peterson444:
Apologies for the noise. Somehow I didn't have pre-commit installed.

This is now ready including wrapping the install step, albeit not very elegantly for now.

subiquity/server/ubuntu_drivers.py
```diff
@@ -168,6 +172,129 @@ async def list_oem(self, root_dir: str, context) -> List[str]:
...
+                    "vendor": "0x14E4",
+                    "device": "0x4353",
+                }
+            ]
+        }
+
+        self.dev_config_path = "/tmp/umockdev_config.yaml"
+
+        # write config to live environment
+        write_file(self.dev_config_path, yaml.safe_dump(self.dev_config), mode=0o777)
```
dbungert, line 201: The vendor and device values can be read from the modalias - note the `v` and `d` values at the beginning of the modalias string matching the vendor and device - which means that we could do a kernel command line custom string of `subiquity-fake-pci-devices=modalias` and get driver fakery in return without slimy snap cycle.

snapcraft.yaml
```diff
@@ -140,6 +144,7 @@ parts:
       bin/subiquity-service: usr/bin/subiquity-service
       bin/subiquity-server: usr/bin/subiquity-server
       bin/subiquity-cmd: usr/bin/subiquity-cmd
+      bin/subiquity-umockdev-wrapper: system_scripts/subiquity-umockdev-wrapper
```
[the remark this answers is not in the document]
Chris-Peterson444: Unfortunately usr/bin/ has the snap vendored python3 version we're trying to avoid in the first place. We could get around this on a per script basis, but my thought was to have the single system_scripts_env for all of these types of scripts.

scripts/kvm-test.py
```diff
@@ -240,6 +240,12 @@ def load_config(self):
                     package)''')
 parser.add_argument('--profile', default="server",
                     help='load predefined memory, disk size and qemu options')
+parser.add_argument('--fake-pci-devices', action='store_true', default=False,
```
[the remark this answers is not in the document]
Chris-Peterson444: Yeah that sounds good to me.

snapcraft.yaml
```diff
@@ -40,8 +40,12 @@ apps:
     environment:
       PYTHONIOENCODING: utf-8
       SUBIQUITY_ROOT: $SNAP
+      PYTHONPATH_ORIG: $PYTHONPATH
+      PYTHON_ORIG: $PYTHON
       PYTHON: $SNAP/usr/bin/python3.10
+      PY3OR2_PYTHON_ORIG: $PY3OR2_PYTHON
```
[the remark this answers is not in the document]
Chris-Peterson444: I honestly didn't know what all used this but was going for consistency. Happy to drop.

subiquity/server/ubuntu_drivers.py
Chris-Peterson444: Ah I hadn't realized, nice. I do like that idea more than changing with early-commands.

snapcraft.yaml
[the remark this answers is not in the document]
dbungert: got it, skip this suggestion then

subiquity/server/ubuntu_drivers.py
```diff
@@ -168,6 +172,129 @@ async def list_oem(self, root_dir: str, context) -> List[str]:
...
+
+        prefix: list[str] = [
+            "subiquity-umockdev-wrapper",  # vendored in system_scripts
+            "--config",
+            f"{self.dev_config_path}",
```
ogayot, line 206:
```suggestion
            self.dev_config_path,
```

subiquity/server/ubuntu_drivers.py
```diff
@@ -168,6 +172,129 @@ async def list_oem(self, root_dir: str, context) -> List[str]:
...
+            await arun_command(["sh", "-c", "command -v ubuntu-drivers"], check=True)
+        except subprocess.CalledProcessError:
+            raise CommandNotFoundError(
+                f"Command ubuntu-drivers is not available in {root_dir}"
+            )
+        # Install wrapper script prerequisites on live system
+        try:
+            await arun_command(self.pre_req_cmd, check=True)
+        except subprocess.CalledProcessError as err:
+            log.debug(f"ensure_cmd retuned with exit code {err.returncode}")
```
ogayot, line 231: s/retuned/returned/

subiquity/server/ubuntu_drivers.py
```diff
@@ -262,4 +389,14 @@ def get_ubuntu_drivers_interface(app) -> UbuntuDriversInterface:
...
 
+    if "subiquity-fake-pci-devices" in app.opts.kernel_cmdline:
+        log.debug("Using umockdev wrapper")
+        cls = UbuntuDriversFakePCIDevicesInterface
+
+    # For quickly testing MOK enrollment we install on server and force no gpgpu
+    # The caveat to this is that it also has to be an online install
+    if "subiquity-server-force-no-gpgpu" in app.opts.kernel_cmdline:
+        log.debug("Forcing no gpgpu drivers. Requires online install on server.")
+        is_server = False
```
ogayot, line 400: would you rename the `is_server` variable? `gpgpu` makes more sense now I think.

subiquity/server/ubuntu_drivers.py
```diff
@@ -168,6 +172,129 @@ async def list_oem(self, root_dir: str, context) -> List[str]:
...
+        prefix: list[str] = [
+            "subiquity-umockdev-wrapper",  # vendored in system_scripts
+            "--config",
+            f"{self.dev_config_path}",
+            "--",  # Don't let wrapper consume ubuntu-drivers feature flags
+        ]
+
+        self.sys_env = system_scripts_env()
+
+        self.pre_req_cmd = ["apt", "install", "-y", "umockdev", "gir1.2-umockdev-1.0"]
```
ogayot, line 212: use `apt-get` instead of `apt` ?

subiquitycore/utils.py
```diff
@@ -35,6 +35,30 @@ def _clean_env(env, *, locale=True):
...
 
+def system_scripts_env() -> dict[str, str]:
+    """Generate an environment for running all programs outside of the snap,
+    but also include those vendored in $SNAP/bin.
+    """
+
+    env: dict[str, str] = orig_environ(os.environ)
+    snap_path: str | None = env.get("SNAP")
+    if snap_path is None:
+        raise AssertionError("Could not find path to SNAP")
```
ogayot, line 46: if you want an AssertionError, you should use assert instead.

```
assert snap_path is not None, "Could not find path to SNAP"
```

system_scripts/subiquity-umockdev-wrapper
ogayot, line 1: is this the same as scripts/umockdev-wrapper.py? Can we keep a single script?

Chris-Peterson444: Yeah we probably should do that. My vote is to just move it out of scripts and into system_scripts, modifying the dryrun code to expect the new location. Does that sounds alright?

Chris-Peterson444:
Thanks for all the feedback. I integrated all the suggested changes and rebased on top of main. Since the diff is a little big, the summary is:

- Added a generic `--kernel-cmdline` options to kvm-test and dropped the previously suggested flags for appending to the kernel command line
- the kernel command line options `subiquity-fake-pci-devices` now expects a comma separated list of modalias you can use to generate faked devices
- migrated `scripts/umockdev-wrapper.py` to `system_scripts/subiquity-umockdev-wrapper`

subiquity/server/ubuntu_drivers.py
```diff
@@ -168,6 +173,145 @@ async def list_oem(self, root_dir: str, context) -> List[str]:
...
+
+        self.sys_env = system_scripts_env()
+
+        self.pre_req_cmd = [
+            "apt-get",
+            "install",
+            "-y",
+            "umockdev",
+            "gir1.2-umockdev-1.0",
+        ]
```
dbungert, line 215: there seems to be an race condition / apt conflict here, I had to launch 3 times to see this command pass.

dbungert (approved):
This has been open for a bit, I'm inclined to land it and improve as we go, but I did run into a problem.

subiquitycore/utils.py
```diff
@@ -35,6 +35,29 @@ def _clean_env(env, *, locale=True):
     return env
 
 
+def system_scripts_env() -> dict[str, str]:
+    """Generate an environment for running all programs outside of the snap,
+    but also include those vendored in $SNAP/bin.
+    """
+
+    env: dict[str, str] = orig_environ(os.environ)
+    snap_path: str | None = env.get("SNAP")
```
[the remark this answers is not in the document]
Chris-Peterson444: It would but I was being nitpicky about how I wanted the error message to show up.

subiquity/server/ubuntu_drivers.py
Chris-Peterson444: Thanks for pointing this out. Pushed a new commit to fix this by waiting for the apt lock with `-oDPkg::Lock::Timeout=-1`.

dbungert (approved):

PR #2030, opened 2024-07-16
scripts: add script to inject ProbeData.txt into existing machine-config

ogayot:
The script can be used to inject a ProbeData.txt file into an existing machine-config file.

By default, it will:
 * read the storage probe data from a file named ProbeData.txt in the current directory
 * read the machine-config from examples/machines/simple.json
 * write the resulting machine-config to stdout

The path to ProbeData.txt can be specified using the --storage-probe-data option.
The patch to the machine-config can be specified using the --machine-config option.
The --overwrite option can be passed to overwrite the original machine-config.

examples:
```
 $ scripts/inject-probe-data.py \
    --storage-probe-data ProbeData.txt \
    --machine-config examples/machines/virtio.json
    --overwrite
```
```
 $ scripts/inject-probe-data.txt > simple-with-injected-storage.txt
```

### DIFF
--- a/scripts/inject-probe-data.py
+++ b/scripts/inject-probe-data.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+
+"""Bugs filed against Subiquity on Launchpad include a file named
+"ProbeData.txt" ; which contains the storage data extracted by probert.
+Some of the bugs can easily be reproduced in dry-run mode if we feed the
+storage data back to Subiquity. To do so, we can replace the "storage" section
+of the machine-config with the contents of ProbeData.txt.
+
+This script is meant to simplify this process.
+"""
+
+import argparse
+import contextlib
+import json
+import sys
+
+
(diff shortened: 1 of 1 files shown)


scripts/inject-probe-data.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+
+import argparse
+import contextlib
+import json
+import sys
+
+
+def parse_cmdline() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
dbungert, line 10: Would you add a short help string explaining the problem being solved?

dbungert (approved):

scripts/inject-probe-data.py
ogayot: Sure thing!

Chris-Peterson444 (approved):
Nice, thanks!

PR #2031, opened 2024-07-17
storage: don't suggest use-gap if we can't fit more partitions

ogayot:
If a disk has already reached its limit of (primary) partitions, suggesting a use-gap scenario can only lead to failures. In the same vein, if a disk only has room for one more partition but we also need to add an ESP, the scenario would fail to apply too.

We now try to prevent Subiquity from suggesting such, bad, use-gap scenarios ; by guessing the number of primary partitions needed ; and comparing it with the limit.

Although it isn't a perfect solution (see below), this change should hopefully address the most common problems.

What makes it difficult to properly suggest scenarios is that we do not have all the cards in hand with the current API. For instance, the user can choose to add a recovery partition, which of course influences the number of partitions required. Sadly, we don't have a way with the current API to allow a scenario without recovery partition ; while rejecting the same scenario with a recovery partition. Capabilities can also influence the number of required partitions, supposedly.

LP:#2073378

### DIFF
--- a/subiquity/common/filesystem/boot.py
+++ b/subiquity/common/filesystem/boot.py
@@ -63,6 +63,9 @@ class MakeBootDevicePlan(abc.ABC):
     this a boot device" in sync.
     """
 
+    def new_partition_count(self) -> int:
+        return 0
+
     @abc.abstractmethod
     def apply(self, manipulator):
         pass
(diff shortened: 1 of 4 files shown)


dbungert (approved):
Thanks for this Olivier.  I believe this is correct.
Two non-blocking comments:
* as you've already observed, this also applies to resize and reformat.  My sense is that the primary place people run into this is with resize.
* To make this better down the road, the number of primaries required varies by capability, so that could be factored in as well

Chris-Peterson444 (approved):

subiquity/common/filesystem/boot.py
```diff
@@ -63,6 +63,9 @@ class MakeBootDevicePlan(abc.ABC):
     this a boot device" in sync.
     """
 
+    def iterate_steps(self):
```
mwhudson, line 66: Seems like it would be a little cleaner to have a "new_partition_count()" method on MakeBootDevicePlan?

mwhudson (approved):
Nice. One suggestion about where to put some of the logic.

PR #2032, opened 2024-07-17
kernel: fix explicitly_requested for default ISOs

dbungert:
When attempting an OEM kernel install, the ensure_no_kernel_conflict() check fails. explicitly_requested has been set for stock ISOs, but explicitly_requested was only intended for autoinstall cases.


ogayot:
This is weird. According to https://github.com/canonical/subiquity/commit/7e8c1eedd543b7f5697376a9e5c648a70bce1229, I was trying to raise an error on purpose when running on certified hardware with a kernel specified. I honestly don't remember the details. Did we have the kernel specified in stock 24.04 ISOs?

ogayot (approved):

PR #2034, opened 2024-07-23
No double kernel

dbungert:
Fixes the problem of ending up with 2 kernels installed that can be seen on the OEM codepath if the kernel is preinstalled in the source image.  Works by telling curtin what to remove, since we can get the preinstalled kernel before we start installing the OEM one.

Requires ~~https://code.launchpad.net/~dbungert/curtin/+git/curtin/+merge/469886~~ (meged)

Fixes other unrelated issues
* fix specification of the oem kernel package name - was unused but it should still be the correct value
* fix a corner case where if OEM metapackage installation is requested and a kernel is requested in autoinstall and they match, don't fail that case
* Update the exception where there is an oem/kernel autoinstall conflict to AutoinstallError

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -31,6 +31,7 @@ from typing import Any, Dict, List, Optional
 
 import yaml
 from curtin.config import merge_config
+from curtin.distro import list_kernels
 
 from subiquity.common.errorreport import ErrorReportKind
 from subiquity.common.pkg import TargetPkg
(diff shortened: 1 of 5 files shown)


dbungert:
In draft only for the unmerged curtin changes

subiquity/server/controllers/install.py
```diff
@@ -298,6 +299,23 @@ async def install_oem_metapackages(ctx):
...
+        # to remove that pre-installed kernel.  This can't use the curtin
+        # ensure_one_kernel logic as by the time it knows what to do, there are
+        # already two kernels, so instead we tell curtin explicitly what to do.
+        target = self.tpath()
+        if self.app.opts.dry_run:
+            existing_kernels = []
+        else:
+            existing_kernels = list_kernels(target=target)
+        for metapkg in self.model.oem.metapkgs:
+            if metapkg.name in existing_kernels:
```
mwhudson, line 313: isn't metapkg.name going to be something like oem-somerville-weedle-meta, not something like linux-image-6.5.0-1026-oem ?

subiquity/server/controllers/oem.py
```diff
@@ -196,13 +193,16 @@ async def ensure_no_kernel_conflict(self) -> None:
...
-        if self.model.metapkgs and kernel_model.explicitly_requested:
-            # TODO
-            # This should be a dialog or something rather than the content of
-            # an exception, really. But this is a simple way to print out
-            # something in autoinstall.
-            msg = _(
-                """\
+        if self.model.metapkgs:
+            for metapkg in self.model.metapkgs:
+                if kernel_model.metapkg_name == metapkg.name:
```
mwhudson, line 198: Again here I think kernel_model.metapkg_name is something like linux-image-generic or linux-image-oem-22.04d, and metapkg.name is oem-*-meta

mwhudson (commented):
I have some worries about the many many kinds of metapackage that are flying around...

subiquity/server/controllers/install.py
dbungert: correct

subiquity/server/controllers/oem.py
dbungert: while true, the way to resolve that is to walk the dependency chain

dbungert: per chat elsewhere, I do think this part is wrong and will revisit it

subiquity/server/controllers/install.py
dbungert: per chat elsewhere, I do think this part is wrong and will revisit it

mwhudson (approved):
Would still like better names curtion options but actual correct behaviour comes first I guess!

PR #2039, opened 2024-07-27
Coredesktop models

dbungert:
* Update models for core
* Fix an issue with interactivity where full interactive state was calculated after confirmation, but needed to be known before confirmation

### DIFF
--- a/subiquity/server/controller.py
+++ b/subiquity/server/controller.py
@@ -112,6 +112,7 @@ class SubiquityController(BaseController):
         override interactive(), so refer to the controller of interest for full
         details.
         """
+        # FIXME is this correct for refresh?
         if self.interactive_for_variants is None:
             if self.model_name is None:
                 return True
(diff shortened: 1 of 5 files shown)


subiquity/server/controller.py
```diff
@@ -110,6 +101,13 @@ async def apply_autoinstall_config(self, context):
         """
         pass
 
+    @property
+    def _active(self):
```
mwhudson, line 105: Could this get a docstring please?

mwhudson (approved):

sergio-costas:
I tested it removing the `autoinstall.yaml` file, but still shows all the pages. The `install-sources.yaml` file is placed at `/cdrom/casper/install-sources.yaml' , and has these contents:

```
- default: true
  description:
    en: Ubuntu Core Desktop.
  id: ubuntu-core-desktop
  locale_support: locale-only
  name:
    en: Ubuntu Core Desktop
  path: pc.img.xz
  type: dd-xz:file
  size: 1747720248
  variant: core
```

Do I have to do anything else to remove the unwanted pages?

PR #2040, opened 2024-08-05
guided: set _netdev more consistently (including when doing guided)

ogayot:
When using the text-installer with NVMe/TCP, selecting a NVMe drive in the guided page results in the absence of _netdev option for the filesystems. This makes curtin unaware that we want a NVMe/TCP setup and therefore eventually prevents booting.

~~Ensure _netdev is properly set in this scenario.~~

~~Going forward, we should set this option in every scenarios where we create a filesystem on remote storage. But for now, let's keep things simple.~~

~~Marking as a draft because I want to experiment with encryption and LVM~~

### DIFF
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -56,9 +56,7 @@ class FilesystemManipulator:
     def create_mount(self, fs, spec: FileSystemSpec):
         if spec.get("mount") is None:
             return
-        mount = self.model.add_mount(
-            fs, spec["mount"], on_remote_storage=spec.get("on-remote-storage", False)
-        )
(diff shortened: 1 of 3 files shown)


mwhudson:
It's not a very helpful comment but it would be nice if the guided_* methods somehow didn't have to care about details like this

ogayot:
> It's not a very helpful comment but it would be nice if the guided_* methods somehow didn't have to care about details like this

Yes, I think that was a lazy implementation on my part. Your wish has been granted :+1:

Chris-Peterson444 (approved):

dbungert (approved):

mwhudson (approved):
Hah nice and it even makes for net-negative code!

PR #2041, opened 2024-08-05
Bump curtin revision for NVMe/TCP changes

ogayot:


mwhudson (approved):
We don't need this if #2034 gets merged but fine!

PR #2044, opened 2024-08-07
Fixes related to debug shell in non-async context

ogayot:
If the `redraw_screen` coroutine was called while a debug shell was opened (or more rarely during the Subiquity shutdown sequence) urwid would fail with:

```
AssertionError(self._started).
```

We now ensure that we skip the redraw in that scenario. The screen gets now redrawn after the debug shell is closed anyway.

Also, while making the `debug_shell` function a coroutine function, I forgot to update some of the code that invokes it ; resulting in various `RuntimeWarning: coroutine 'SubiquityClient.debug_shell' was never awaited` errors. That should be fixed now.

### DIFF
--- a/subiquity/client/client.py
+++ b/subiquity/client/client.py
@@ -571,7 +571,7 @@ class SubiquityClient(TuiApplication):
             if not self.ui.right_icon.current_help:
                 self.ui.right_icon.open_pop_up()
         elif key in ["ctrl z", "f2"]:
-            self.debug_shell()
+            self.request_debug_shell()
         elif self.opts.dry_run:
             self.unhandled_input_dry_run(key)
         else:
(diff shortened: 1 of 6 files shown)


dbungert (approved):
Sounds reasonable.
Is there an existing bug about this?  If not maybe a one-line entry in the changelog will be enough.  "Fix crash that could happen when opening the the debug terminal under these conditions"

ogayot:
> Sounds reasonable. Is there an existing bug about this? If not maybe a one-line entry in the changelog will be enough. "Fix crash that could happen when opening the the debug terminal under these conditions"

There is no bug report about this AFAICT. I didn't make it obvious but this should be a bug only affecting the development branch.

What changelog are we talking about here?

dbungert:
> There is no bug report about this AFAICT. I didn't make it obvious but this should be a bug only affecting the development branch.
> 
> What changelog are we talking about here?

Oh, devel only, we don't need a changelog.  The changelog here would be the one we would write for 24.10.1, but it sounds like we can skip it.  Thanks!

Chris-Peterson444 (approved):
Very nice!

PR #2047, opened 2024-08-08
Autoinstall snippets

mwhudson:
Add example snippets for locale, timezone and keyboard.

I'm not actually sure about all the details of the behaviour of the timezone key -- like what does `timezone: ""` do??

### DIFF
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -1040,7 +1040,19 @@ timezone
 * **default:** no timezone
 * **can be interactive:** no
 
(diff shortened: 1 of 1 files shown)


Chris-Peterson444 (approved):
> what does timezone: "" do??

On a quick read of the code it looks like `""` and `"geoip"` are equivalent? I'd have to do some testing to be sure though.

Suggestion: Add some text to the timezone section to mention valid timezone values can be generated with `timedatectl list-timezones`

dbungert (approved):

ogayot (approved):

dbungert (approved):

PR #2050, opened 2024-08-08
predictably arbitrary disk selection

dbungert:
filesystem: arbitrary disk select more consistent

If two disks of the same size went to the disk matchers with the default
"largest" match, the one we would choose would be based on whatever
order in python they ended up, and across installs this means that we
could select arbitrarily among disks of a matching size.

Add sort criteria based on various unique identifiers.  Make it work for
other disk-like devices, though the main target is disks and raids
because that's what can go to the matcher system.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -617,6 +617,16 @@ class _Device(_Formattable, ABC):
     def size(self):
         pass
 
+    @property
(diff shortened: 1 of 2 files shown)


mwhudson (approved):

dbungert:
@medicalwei - FYI

PR #2056, opened 2024-08-14
Various fixes for error reports

ogayot:
1. On the development branch, when showing a crash dialog early (e.g., after an early-command fails), the installer would often raise an exception because the report is not yet loaded on the client side ; although it has finished generating on the server side. See LP:#2076914 
-> Fixed by showing the "Loading report..." animation in this scenario

2.  When showing a crash dialog early, the md5check result is not shown because the Integrity controller:
   a. is not yet started
   b. does not respond to GET requests until all controllers are started
-> Fixed by starting the Integrity controller earlier ; and allowing it to respond to GET requests before all controllers have started. LP:#2076997

3. If the result of the MD5 check was not yet available by the time the crash report was shown, the result will not be refreshed after it becomes available.
-> fixed by adding the option to GET /integrity?wait=True and refresh the report when the call returns.

4. When showing a crash dialog early, if the user clicks on the [Help] menu, the installer will get stuck fetching the SSH information. This happens because the meta/ssh_info GET endpoint blocks until all controllers are started.
-> Fixed by allowing the controller to respond to GET requests before all controllers have started.

### DIFF
--- a/subiquity/ui/views/error.py
+++ b/subiquity/ui/views/error.py
@@ -257,8 +257,14 @@ class ErrorReportStretchy(Stretchy):
 
         self.spinner.stop()
 
-        if self.error_ref.state == ErrorReportState.DONE:
-            assert self.report
+        state = self.error_ref.state
+        if state == ErrorReportState.DONE and self.report is None:
+            # It is possible that the report has finished generating
+            # server-side (and therefore the API returns ErrorReportState.DONE)
+            # but hasn't been loaded locally.
+            state = ErrorReportState.LOADING
+
+        if state == ErrorReportState.DONE:
             widgets.append(btns["view"])
             widgets.append(Text(""))
             widgets.append(Text(rewrap(_(submit_text))))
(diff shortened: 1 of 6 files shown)


subiquity/ui/views/error.py
```diff
@@ -257,8 +257,14 @@ def _pile_elements(self):
 
         self.spinner.stop()
 
-        if self.error_ref.state == ErrorReportState.DONE:
-            assert self.report
+        state = self.error_ref.state
+        if state == ErrorReportState.DONE and self.report is None:
+            # It is possible that the report has finished loading server-side
+            # (and therefore the API returns ErrorReportState.DONE) but hasn't
+            # been loaded locally.
```
mwhudson, line 267: Mmm this feels like something that has been around since the client/server split. Does the server side every load the reports now? (i.e. can the server ever return LOADING here? it feels a bit inappropriate to have in the API, somehow).

mwhudson (approved):

subiquity/ui/views/error.py
ogayot: I think the server only responds with `ErrorReportState.INCOMPLETE`, `ErrorReportState.DONE` or `ErrorReportState.ERROR_GENERATING`. 

So I guess

```diff
-            # It is possible that the report has finished loading server-side
-            # It is possible that the report has finished generating server-side
            # (and therefore the API returns ErrorReportState.DONE) but hasn't
            # been loaded locally.
```

is more appropriate.

dbungert (approved):

Chris-Peterson444 (approved):

PR #2057, opened 2024-08-14
apport: set the APPORT_DATA_DIR variable in snap yaml definition

ogayot:
The APPORT_DATA_DIR variable must be set before the apport Python module is imported. This was no longer true potentially because of this import:

  from subiquity.server.controllers.filesystem import set_user_error_reportable

The result was:

```
ERROR: hook /usr/share/apport/general-hooks/ubuntu.py crashed
  File "/snap/subiquity/5988/usr/lib/python3/lib/python3/dist-packages/apport/report.py", line 228, in _run_hook
    symb['add_info'](report, ui)
  File "/usr/share/apport/general-hook/ubuntu.py", line 34, in add_info
    _add_release_info(report)
  File "/usr/share/apport/general-hooks/ubuntu.py", line 472, in _add_release_info
    report.add_tags([release_codename])
AttributeError: 'Report' object has no attribute 'add_tags'.
```

Instead of trying to call setup_environment very early; make sure that the variable is set by snapd before Subiquity actually starts.

LP:#2076233

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -24,6 +24,7 @@ apps:
       PYTHON: $SNAP/usr/bin/python3.10
       # APPORT_DATA_DIR must be set before the `apport` python module is
       # imported.
+      APPORT_DATA_DIR_ORIG: $APPORT_DATA_DIR
       APPORT_DATA_DIR: $SNAP/share/apport
   probert:
     command: usr/bin/subiquity-cmd $SNAP/usr/bin/python3.10 $SNAP/bin/probert
@@ -48,6 +49,7 @@ apps:
       PATH: $PATH:$SNAP/bin:$SNAP/sbin
       # APPORT_DATA_DIR must be set before the `apport` python module is
       # imported.
+      APPORT_DATA_DIR_ORIG: $APPORT_DATA_DIR
       APPORT_DATA_DIR: $SNAP/share/apport
   subiquity-service:
     command: usr/bin/subiquity-service
@@ -66,6 +68,7 @@ apps:
       PYTHON: $SNAP/usr/bin/python3.10
       # APPORT_DATA_DIR must be set before the `apport` python module is
       # imported.
+      APPORT_DATA_DIR_ORIG: $APPORT_DATA_DIR
       APPORT_DATA_DIR: $SNAP/share/apport
 
 parts:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -22,6 +22,9 @@ apps:
       SUBIQUITY_ROOT: $SNAP
       PYTHON_ORIG: $PYTHON
       PYTHON: $SNAP/usr/bin/python3.10
+      # APPORT_DATA_DIR must be set before the `apport` python module is
+      # imported.
+      APPORT_DATA_DIR: $SNAP/share/apport
   probert:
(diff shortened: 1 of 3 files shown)


snapcraft.yaml
```diff
@@ -43,6 +46,9 @@ apps:
       PYTHON: $SNAP/usr/bin/python3.10
       PY3OR2_PYTHON: $SNAP/usr/bin/python3.10
       PATH: $PATH:$SNAP/bin:$SNAP/sbin
+      # APPORT_DATA_DIR must be set before the `apport` python module is
+      # imported.
+      APPORT_DATA_DIR: $SNAP/share/apport
```
Chris-Peterson444, line 51:
```suggestion
      # APPORT_DATA_DIR must be set before the `apport` python module is
      # imported.
      APPORT_DATA_DIR_ORIG: $APPORT_DATA_DIR
      APPORT_DATA_DIR: $SNAP/share/apport
```

Chris-Peterson444: I think if we add an _ORIG here for APPORT_DATA_DIR this would fix running `ubuntu-bug subiquity` in the debug shell too.

Chris-Peterson444 (approved):
Excellent work! I am +1 on moving these sort of things to snapcraft.yaml.

snapcraft.yaml
dbungert: @Chris-Peterson444 in the debug shell do you want the out-of-snap behavior or the in-the-snap behavior?

Chris-Peterson444: out-of-snap for now. I don't really have any hard opinions on either way yet, but my understanding of the current intent of the debug shell behavior is that it should perform like an out-of-snap shell ( i.e., it should be pretty close to what you could get by sshing into the live machine) so we should be consistent. Currently APPORT_DATA_DIR pointing to the snap path makes this sorta not true. Although, I can see the merits for eventually pointing this back to the snap path and having some apport hook we could update with a snap refresh.

ogayot: I did this change. But now when I run `ubuntu-bug subiquity` from the debug shell, I'm asked to choose between:
```
*** You have two versions of this application installed, which one do you want to report a bug against?

Choices:
  1: subiquity 24.10-devel+git213.443b8f43 () snap
  2: subiquity (x1 of snap /snap/subiquity/x1 of version 24.10-devel+git213.443b8f43) deb package
  C: Cancel
```

(yes, I injected an updated snap). 

Not sure if I should push this change yet?

ogayot: Maybe but it does not happen with an updated snap without the APPORT_DATA_DIR_ORIG variable. So I'm not sure, @Chris-Peterson444 what do you think?

Chris-Peterson444: That looks like the right prompt to me. I can't remember the specifics of how apport gets the first option on server, but either choice will still run the package hook (choosing "snap" just also appends some SnapInfo fields). 

The current behavior is that the "deb package" is not found since `$SNAP/usr/share/apport/package-hooks/subiquity.py` doesn't exist. So I believe it's some default snap info collection that runs otherwise.

ogayot: ACK. Thanks for the feedback. Merging with the change then.

PR #2060, opened 2024-08-20
doc: update autoinstall reference's apt section

Chris-Peterson444:
General formatting updates, add example sections for each sub-key, and add links to advanced usage examples in curtin. 

Also, because I think the reference here should focus on the parts where Subiquity and Curtin's apt sections differ and not repeating Curtin doc of the section, I removed the "add a PPA" example. I do think it's a good example though, so I opened a PR [1] in Curtin to add it to the common snippet's section.

[1] https://code.launchpad.net/~cpete/curtin/+git/curtin/+merge/471552

### DIFF
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -423,7 +423,7 @@ mirror-selection
 If the ``primary`` section is contained within the ``mirror-selection`` section, the automatic mirror selection is enabled. This is the default in new installations.
 
 primary (when placed inside the ``mirror-selection`` section)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
 * **type:** custom, see below
 
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -392,7 +392,7 @@ apt
 
 APT configuration, used both during the installation and once booted into the target system.
 
-This section historically used the same format as curtin, which is documented in the `APT Source <https://curtin.readthedocs.io/en/latest/topics/apt_source.html>`_ section of the curtin documentation. Nonetheless, some key differences with the format supported by curtin have been introduced:
+This section has historically used the same format as curtin, which is documented in the `APT Source <https://curtin.readthedocs.io/en/latest/topics/apt_source.html>`_ section of the curtin documentation. Nonetheless, some key differences with the format supported by curtin have been introduced:
 
 - Subiquity supports an alternative format for the ``primary`` section, allowing configuration of a list of candidate primary mirrors. During installation, Subiquity automatically tests the specified mirrors and selects the first one that appears usable. This new behaviour is only activated when the ``primary`` section is wrapped in the ``mirror-selection`` section.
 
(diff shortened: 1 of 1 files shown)


doc/reference/autoinstall-reference.rst
```diff
@@ -432,52 +435,102 @@ In the new format, the ``primary`` section expects a list of mirrors, which can
...
+  * ``uri`` (Required): The URI of the mirror to use, e.g., ``http://fr.archive.ubuntu.com/ubuntu``.
+  * ``arches`` (Optional): A list of architectures supported by the mirror. By default, this list contains the current CPU architecture.
+
+The URI for the archive mirror does not have to be a country mirror, although it may be the most convenient, and can take the URL of any valid Ubuntu mirror. A list of all registered archive mirrors can be found on `Launchpad <https://launchpad.net/ubuntu/+archivemirrors>`_.
+
+Examples:
+
+.. code-block:: yaml
+
+  # Use custom mirrors for all architectures.
```
ogayot, line 447: The comment makes it sound like both CUSTOM_MIRROR_* will be used ; while in practice the latter will be used only if the former fails validation. Can we make that clearer somehow?

suggestion:

```yaml
# Use the first custom mirror that seems to work. Do not restrict to specific architectures.
```

doc/reference/autoinstall-reference.rst
```diff
@@ -432,52 +435,102 @@ In the new format, the ``primary`` section expects a list of mirrors, which can
...
+  autoinstall:
+    apt:
+      mirror-selection:
+        primary:
+          - uri: "$CUSTOM_MIRROR_1"
```
ogayot, line 452: I think using URLs (they don't have to exist) will be more useful in an example.

```diff
- uri: "$CUSTOM_MIRROR_1"
+ uri: http://mirror1.internal/ubuntu
```

Wdyt?

doc/reference/autoinstall-reference.rst
```diff
@@ -432,52 +435,102 @@ In the new format, the ``primary`` section expects a list of mirrors, which can
...
-            - uri: YOUR_MIRROR_GOES_HERE
             - country-mirror
-            - uri: http://archive.ubuntu.com/ubuntu
-
-To add a PPA:
+            - uri: http://dk.archive.ubuntu.com/ubuntu
+        geoip: true
 
-.. code-block:: yaml
+    # Don't use the country mirror and immediately fallback to next listed mirror.
```
ogayot, line 520: This example is a bit confusing IMO :upside_down_face: 

It would make sense if the user could specify `country: <cc>` but as-is specifying `country-mirror` and `geoip: false` feels wrong.

doc/reference/autoinstall-reference.rst
```diff
@@ -392,15 +392,18 @@ apt
...
 
 - Subiquity supports an alternative format for the ``primary`` section, allowing configuration of a list of candidate primary mirrors. During installation, Subiquity automatically tests the specified mirrors and selects the first one that appears usable. This new behaviour is only activated when the ``primary`` section is wrapped in the ``mirror-selection`` section.
 
 - The ``fallback`` key controls what Subiquity does when no primary mirror is usable.
 
 - The ``geoip`` key controls whether to perform IP-based geolocation to determine the correct country mirror.
 
-The default is:
+
+All other sections behave as defined in curtin. See the curtin `documentation <https://curtin.readthedocs.io/en/latest/topics/apt_source.html#common-snippets>`_ and its `example apt configurations <https://git.launchpad.net/curtin/tree/examples/apt-source.yaml>`_ for usage examples of these sections.
```
ogayot, line 404: maybe mention that it covers adding a PPA? I think it's a common use-case and people will be confused if they can't find the example anymore.

doc/reference/autoinstall-reference.rst
Chris-Peterson444: Yeah, I like your suggestion but with a minor tweak.

`  # Use the first custom mirror that works. Do not restrict to specific architectures.`

doc/reference/autoinstall-reference.rst
Chris-Peterson444: Yeah that's much better, thanks!

doc/reference/autoinstall-reference.rst
Chris-Peterson444: Yes I was conflicted about this. I think the ability to set `geoip: false` with the new mirror selection format feels wrong/pointless but I wanted to document that it technically is just a skip. Although coming back to it yeah it is maybe more confusing. I'll just remove the example.

ogayot (approved):
Thanks, LGTM

doc/reference/autoinstall-reference.rst
Chris-Peterson444: Sure, I think that makes sense. I updated the text a bit, wdyt?

PR #2061, opened 2024-08-21
do not offer to enable Pro for core installs

mwhudson:


mwhudson:
wait this doesn't work

mwhudson:
oh, i was testing the wrong branch. but i added some logging and a test change so we can be a bit more confident that it does in fact work.

dbungert (approved):

PR #2067, opened 2024-08-27
Fix some deprecation warnings on core24

ogayot:
On oracular, this PR makes `make unit` produce 59 warnings instead of 146. The remaining warnings (emitted multiple times) are:

*  DeprecationWarning: 'crypt' is deprecated and slated for removal in Python 3.13
*  /usr/lib/python3/dist-packages/bson/__init__.py:116: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
    -> should be addressed with https://launchpad.net/ubuntu/+source/pymongo/4.7.3-1 (won't be addressed for core24 probably though)
*   /usr/lib/python3/dist-packages/cloudinit/config/schema.py:554: DeprecationWarning: Passing a schema to Validator.iter_errors is deprecated and will be removed in a future release. Call validator.evolve(schema=new_schema).iter_errors(...) instead.
    -> to be investigated
*   /usr/lib/python3/dist-packages/urwid/widget/wimp.py:693: DeprecationWarning: Don't use user_arg argument, use user_args instead.
*  /usr/lib/python3/dist-packages/urwid/widget/wimp.py:264: DeprecationWarning: Don't use user_arg argument, use user_args instead.
    -> internal to urwid, when using on_press or on_state_change. Won't be fixed anytime soon.

### DIFF
--- a/console_conf/ui/views/chooser.py
+++ b/console_conf/ui/views/chooser.py
@@ -128,7 +128,7 @@ class ChooserView(ChooserBaseView):
                     Action(label=act.title.capitalize(), value=act, enabled=True)
                 )
             menu = ActionMenu(actions)
-            connect_signal(menu, "action", self._system_action, s)
+            connect_signal(menu, "action", self._system_action, user_args=[s])
             srow = make_action_menu_row(
                 [
                     Text(s.label),
@@ -162,7 +162,7 @@ class ChooserView(ChooserBaseView):
             ),
         )
 
-    def _system_action(self, sender, action, system):
+    def _system_action(self, system, sender, action):
         self.controller.select(system, action)
 
     def back(self, result):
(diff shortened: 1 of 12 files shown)


mwhudson (approved):
Looks good. The argument order change when you switch from user_arg to user_args is unfortunate!

Should we stop using on_press / on_state_change?

ogayot:
> Should we stop using on_press / on_state_change?

That sounds like a lot of work! We're using 

Maybe we subclass urwid.Button, and do something like:

```python
class Button(urwid.Button):
    def __init__(self, *args, on_press=None, user_args=(), **kwargs) -> None:
        if "user_data" in kwargs:
             raise ValueError("For this implementation, use user_args instead")
        super().__init__(*args, **kwargs)
        if on_press is not None:
            connect_signal(self, "click", on_press, user_args=list(user_args))
```
and then adjust the argument order.

Something for another day, I guess :)

mwhudson:
> Something for another day, I guess :)

Oh definitely, nothing for this branch!!

mwhudson:
> Maybe we subclass urwid.Button, and do something like:

Looking at most of our actual uses, having a wrapper that discarded the 'sender' argument to the callback would be useful ....

PR #2068, opened 2024-08-27
test-in-lxd: add usage when not enough args are supplied

ogayot:
I can never remember what arguments test-in-lxd takes. Make it display an usage when called without arguments.

### DIFF
--- a/scripts/test-in-lxd.sh
+++ b/scripts/test-in-lxd.sh
@@ -1,7 +1,22 @@
 #!/bin/bash
 set -eux
+
+if [ "$#" -lt 2 ]; then
+    cat >&2 <<EOF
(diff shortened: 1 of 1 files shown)


dbungert (approved):
Does this script behave it self otherwise?  I always thought of it as a CI only thing but I see it as a positive that someone could use it for local CI runs.

ogayot:
> Does this script behave it self otherwise?

I admit I mostly use it to reproduce CI failures locally ; but I haven't had many problems with it. Maybe it needs a `--checkout-head` option or something to run the command against a cleared tree and does not copy over untracked files to the container. (I have Ubuntu ISOs in my working copy so...). 

Overall, it's useful a useful tool to have, I think.

PR #2069, opened 2024-08-27
pro: replace deprecated cloud-init key ubuntu_advantage

ogayot:
On cloud-init versions >= 24.1, having the ubuntu_advantage key leads to the following deprecation notice:

  > cc_ubuntu_pro.py[WARNING]: Deprecated configuraiton key(s) provided:
  > ubuntu_advantage. Expected "ubuntu_pro": will attempt to continue.

Let's use ubuntu_pro instead ; which should be available on 20.04.5, 22.04.5 and all the 24.04+ ISOs.

Marking as a draft before I figure out what we want to do for older ISOs + refresh.


scripts/runtests.sh
```diff
@@ -169,7 +169,7 @@ python3 scripts/check-yaml-fields.py "$tmpdir"/var/log/installer/curtin-install/
 python3 scripts/check-yaml-fields.py <(python3 scripts/check-yaml-fields.py $tmpdir/etc/cloud/cloud.cfg.d/99-installer.cfg datasource.None.userdata_raw) \
         locale='"en_GB.UTF-8"' \
         timezone='"Pacific/Guam"' \
-        ubuntu_advantage.token='"C1NWcZTHLteJXGVMM6YhvHDpGrhyy7"' \
```
blackboxsw, line 172: Encoding a token publicly in a repo gave me pause, but Just confirmed that this token is, in fact, invalid/not real and cannot be used to attach.

subiquity/models/ubuntu_pro.py
```diff
@@ -38,11 +38,8 @@ def make_cloudconfig(self) -> dict:
         """
         if not self.token:
             return {}
-        # Both "ubuntu_advantage" and "ubuntu-advantage" keys are accepted, but
-        # "ubuntu-advantage" is deprecated despite not being mentioned in the
-        # documentation.
         return {
-            "ubuntu_advantage": {
+            "ubuntu_pro": {
```
blackboxsw, line 42: @ogayot looks good to me.   Moving to `ubuntu_pro` is supported on 20.04 and later for cloud-init version 24.1 which published to the archive in June.


There are two uploads under review that will make this valid as well for 16.04 and 18.04 if that becomes something you need to support, but that hasn't landed yet.

blackboxsw (commented):
The `ubuntu_pro` support is available in cloud-init and ubuntu pro in Focal and later as of cloud-init version 24.1. Looks good to me on the initial approach.

PR #2070, opened 2024-08-29
Make relevant filesystem controller errors "user errors" and categorize them

ogayot:
This PR comes with 3 main changes:

* changed the way to configure whether StorageRecoverableError should produce crash reports. The main drawback of the previous implementation (using type aliasing) is that inheriting from StorageRecoverableError is resolved at class creation time, not instance creation time.
* use subclasses of StorageRecoverableErrors to categorize storage errors
* added title and error code to RecoverableError. This is a step towards implementing RFC9457 (Problem Details for HTTP APIs) or something somehow similar to RFC9457.

### DIFF
--- a/subiquity/common/api/recoverable_error.py
+++ b/subiquity/common/api/recoverable_error.py
@@ -15,4 +15,16 @@
 
 
 class RecoverableError(Exception):
-    """An error type that should not be treated as fatal."""
+    """An API-only error type that denotes a non-fatal error in a HTTP request
+    handler."""
+
+    # This field tells Subiquity what to do if an instance of this exception is
+    # raised in a request handler.
+    # By default we produce a crash report but one can decide otherwise by
+    # changing this value to False. This can be done globally or by setting
+    # produce_crash_report=False in a subclass. See
+    # subiquity.server.controllers.filesystem.set_user_error_reportable for an
+    # example.
+    # Currently, this value also controls whether the error would lead to a 500
+    # or 422 HTTP status.
+    produce_crash_report = True
--- a/subiquity/common/api/recoverable_error.py
+++ b/subiquity/common/api/recoverable_error.py
@@ -16,7 +16,19 @@
 
 class RecoverableError(Exception):
     """An API-only error type that denotes a non-fatal error in a HTTP request
-    handler."""
+    handler.
+
+    The intention of this class is to provide enough information about the
+    error to make it serializable into a Problem Details Object defined in RFC
+    9457 (Problem Details for HTTP APIs).
+
+    * The "code" attribute would be used as part of the "type" member of the
+    Problem Details Object. Maybe using the tag URI scheme:
+      tag:subiquity,2024-08-28:<code>
+    * The "title" attribute would be used as the "title" member of the Problem
+    Details Object.
+    * As for the message of the exception, it could be used as the "detail"
+    member of the Problem Details object."""
 
     # This field tells Subiquity what to do if an instance of this exception is
     # raised in a request handler.
@@ -28,3 +40,17 @@ class RecoverableError(Exception):
     # Currently, this value also controls whether the error would lead to a 500
     # or 422 HTTP status.
     produce_crash_report = True
+
+    # One must override title and code in subclasses with meaningful values.
+    # Note that different instances of the same class will have the same title
+    # and code. This is by design.
+    code: str = ""
+    title: str = ""
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        if not self.title or not self.code:
+            raise NotImplementedError(
+                "Please do not instantiate directly. Use subclasses"
+                " having a title and a user code properly set."
+            )
--- a/subiquity/common/api/server.py
+++ b/subiquity/common/api/server.py
@@ -169,9 +169,13 @@ def _make_handler(
                 )
             except Exception as exc:
                 tb = traceback.TracebackException.from_exception(exc)
+                if not isinstance(exc, RecoverableError) or exc.produce_crash_report:
+                    status = 500
+                else:
+                    status = 422
                 resp = web.Response(
                     text="".join(tb.format()),
-                    status=422 if isinstance(exc, RecoverableError) else 500,
+                    status=status,
                     headers={
                         "x-status": "error",
                         "x-error-type": type(exc).__name__,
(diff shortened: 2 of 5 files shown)


Chris-Peterson444 (approved):
LGTM. A couple of non-blocking questions/comments. I like the direction!

subiquity/common/api/recoverable_error.py
```diff
@@ -15,4 +15,35 @@
...
+    Problem Details Object. Maybe using the tag URI scheme:
+      tag:subiquity,2024-08-28:<code>
+    * The "title" attribute would be used as the "title" member of the Problem
+    Details Object.
+    * As for the message of the exception, it could be used as the "detail"
+    member of the Problem Details object."""
+
+    # If this exception is uncaught in a request handler, what should we do.
+    # By default we produce a crash report, but one can change set it to False.
+    produce_crash_report = True
```
Chris-Peterson444, line 35: This also controls whether the client actually sees (or should treat) this as a recoverable error or not right? Also, having `RecoverableError` produce crash reports by default feels a little weird without the immediate context of the `--no-report-stroage-user-errors` flag changing this globally. Could we add some text to the comment about this behavior?

subiquity/server/server.py
```diff
@@ -551,7 +552,7 @@ async def middleware(self, request, handler):
                 resp.headers["x-error-msg"],
                 exc_info=exc,
             )
-            if not isinstance(exc, NonReportableException):
+            if not isinstance(exc, RecoverableError) or exc.produce_crash_report:
```
Chris-Peterson444, line 555: The way this is written, if we ever throw a `NonReportableException` in a request handler then we'll get an apport report but we won't halt the installer, which is probably fine because we want to catch those and convert those to `RecoverableError` anyways?

subiquity/common/api/recoverable_error.py
ogayot: > This also controls whether the client actually sees (or should treat) this as a recoverable error or not right? 

Correct! I'm not sure if this is the proper design going forward. But it is indeed what happens currently. Added a comment to mention it.

> Also, having RecoverableError produce crash reports by default feels a little weird without the immediate context of the --no-report-stroage-user-errors flag changing this globally. Could we add some text to the comment about this behavior?

Right. I've added documentation to reflect that. Note that the `--no-report-storage-user-errors` is meant to disappear in the future. And `produce_crash_report` will probably default to `True` (or be removed entirely).

subiquity/server/server.py
ogayot: To be honest, I'm not sure what use-cases we will have for `NonReportableException` besides `AutoinstallError` (which obviously should not be raised in a request handler).

And if we do run into an `AutoinstallError` raised in a request handler, I would say yes, we should produce a crash report because it shouldn't happen.

The only scenario that I think would be valid is if the exception is raised when doing  `await task` in a request handler, and that task is wrapping a `load_autoinstall_data()` or `apply_autoinstall_data()`  coroutine or something.

subiquity/common/api/server.py
```diff
@@ -169,17 +169,24 @@ async def handler(request):
...
                 tb = traceback.TracebackException.from_exception(exc)
+                headers = {
+                    "x-status": "error",
+                    "x-error-type": type(exc).__name__,
+                    # aiohttp will reject a header if its value contains a
+                    # "\r" or "\n" character. By using compact JSON, we
+                    # ensure those characters are escaped.
+                    "x-error-msg": json.dumps(str(exc), indent=None),
+                }
+                if not isinstance(exc, RecoverableError) or exc.produce_crash_report:
```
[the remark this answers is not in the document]
ogayot: agreed.

subiquity/server/server.py
ogayot: Nevermind, I guess we shouldn't create a crash report for those. They are non reportable after all. I'll update the implementation.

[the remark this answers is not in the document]
ogayot: Yeah, I've been giving this some more thoughts and I agree with you. While it feels wrong to potentially get `AutoinstallError`s to slip through a request handler, we still should honor the fact that such error is by definition "NonReportable".

subiquity/server/controllers/filesystem.py
```diff
@@ -127,23 +126,39 @@
...
+    """Exception to raise when a storage endpoint is not properly used."""
 
-    # TODO Inherit from RecoverableError going forward.
+    code = "storage-invalid-usage"
+    title = _("Invalid use of storage endpoint")
 
 
-# Depending on config, we will let the SVE fail on the server side
-StorageRecoverableError: Type[NonReportableSVE | ReportableSVE] = ReportableSVE
+class StorageConstraintViolationError(StorageRecoverableError):
```
Chris-Peterson444, line 140: praise: these subclass definitions read really nice. 

At first I had wondered why you didn't use the usual abstract base class machinery in `RecoverableError` and use properties to get around the weird init, but this way definitely reads nicer I think.

Chris-Peterson444 (approved):
This looks really nice, thanks Olivier!

subiquity/server/controllers/filesystem.py
```diff
@@ -743,7 +764,9 @@ def start_guided_resize(
         gap = gaps.after(disk, partition.offset)
         if gap is None:
             pgs = gaps.parts_and_gaps(disk)
-            raise Exception(f"gap not found after resize, pgs={pgs}")
+            raise StorageConstraintViolationError(
+                f"gap not found after resize, pgs={pgs}"
```
dbungert: I suggest moving this "pgs" bit to a debug log statement, the discussion about "gaps" is going to be kind of inscrutable to an end user and not at all clear what they should do about it.  The case here is we're doing guided resize and the gap we expected to find after resize just isn't there, perhaps a different resize would work better?  We could say some sort of "try again with a different value" type message, maybe.

This is hopefully a case that never is hit!

dbungert (approved):

subiquity/server/controllers/filesystem.py
ogayot: Good idea. I converted the content of the original exception to a debug log - and went for "failed to locate gap after resizing partition" to mirror the "failed to locate gap after adding boot" that we already have.

> This is hopefully a case that never is hit!

Actually we have had multiple occurrences of this error in the past. IIRC, one of the reason was the desktop installer offering to users the ability to do a guided resize, go back, do a guided resize again - and all that without a reset in between.

PR #2074, opened 2024-09-04
Kernel crash dumps autoinstall

Chris-Peterson444:
Introduce the `KernelCrashDumpsController` and `KernelCrashDumpsModel`, a non-interactive controller-model pair for setting the kernel-crash-dumps curthooks config. Practically, we introduce a new autoinstall section with the same format as the relevant curthooks config to selectively enable or disable kernel crash dumps on the target system. The default
configuration is `{ "enabled": null }` which will let curtin perform dynamic enablement.

Currently this is a no-op since the `kernel-crash-dumps` curthooks section hasn't been implemented in curtin yet.


autoinstall-schema.json
```diff
@@ -103,6 +103,18 @@
                 }
             ]
         },
+        "kernel-crash-dumps": {
+            "type": "object",
+            "properties": {
+                "enabled": {
+                    "type": "boolean"
```
mwhudson, line 110: Can't this be null too?

mwhudson (commented):
Schema possibly aside this all looks good. Probably doesn't make sense to land it until curtin changes are done though?

autoinstall-schema.json
Chris-Peterson444: Ah yes it can, thanks. This came from the original plan and me thinking something like:

```yaml
enabled: null
min-memory: 4G
reserved-memory: 512M
```

Would be kinda weird (custom crashkernel but only if the default says we should enable it I guess?) but really that's fine too.

Will fix shortly.

Chris-Peterson444:
Either way really. I suppose if we wait it'd save us a slightly annoying revert if it ends up not landing in curtin.

subiquity/server/server.py
```diff
@@ -258,6 +259,7 @@ class SubiquityServer(Application):
         "Locale",
         "Refresh",
         "Kernel",
+        "KernelCrashDumps",
```
ogayot, line 262: When I see this, I sort of regret not thinking about making `crash-dump` a subkey of `kernel`. But it's a design discussion and probably too late to consider anyway. :)

doc/reference/autoinstall-reference.rst
```diff
@@ -1190,6 +1190,40 @@ Example:
...
+* **can be interactive:** no
+
+Toggle kernel crash dumps enablement.
+
+The default value is no configuration, which will result in dynamic enablement as described in the following document (doesn't exist yet).
+
+enabled
+^^^^^^^
+
+**type:** boolean
```
ogayot, line 1209: Maybe we should document `null` as well. And say that the default for kernel-crash-dumps is:

```yaml
kernel-crash-dumps:
    enabled: null
```

subiquity/models/kernel_crash_dumps.py
```diff
@@ -0,0 +1,30 @@
...
+
+log = logging.getLogger("subiquity.models.kernel_crash_dumps")
+
+
+class KernelCrashDumpsModel:
+    # Set to True/False via autoinstall. Defaults to None to let curtin know
+    # to do dynamic enablement based on release, arch, etc.
+    enabled: bool | None = None
+
+    def render(self) -> dict[str, str]:
```
ogayot, line 25:
```python
dict[str, typing.Any]
```

subiquity/server/controllers/kernel_crash_dumps.py
```diff
@@ -0,0 +1,53 @@
...
+
+log = logging.getLogger("subiquity.server.controllers.kernel_crash_dumps")
+
+KernelCrashDumpsConfig = TypedDict(
+    "KernelCrashDumpsConfig",
+    {
+        "enabled": bool,
+    },
+    total=True,
+)
```
ogayot, line 28: Alternative notation, (which I like better but it's up to you)

```python
class KernelCrashDumpsConfig(TypedDict, total=True):
    enabled: bool
```

subiquity/server/controllers/kernel_crash_dumps.py
```diff
@@ -0,0 +1,53 @@
...
+        "required": ["enabled"],
+        "additionalProperties": False,
+    }
+
+    def load_autoinstall_data(self, data: KernelCrashDumpsConfig | None) -> None:
+        if data is None:
+            return
+        self.model.enabled = data["enabled"]
+
+    def make_autoinstall(self) -> dict[str, str]:
```
ogayot, line 48: dict[str, bool]

or

KernelCrashDumpsConfig, if the type checker does not freak out :)

ogayot (commented):
A few quibbles, but looks good overall.

subiquity/server/server.py
Chris-Peterson444: I had this thought too, but from the controller (and curtin hook) perspective it's pretty separate from the things `kernel` does. At least it made the implementation easy :grinning:

doc/reference/autoinstall-reference.rst
Chris-Peterson444: Sure. I went back and revised that section a little too.

subiquity/server/controllers/kernel_crash_dumps.py
Chris-Peterson444: Sure, I like this format better too. We'll have to change it back if we eventually expand the config to allow the originally planned "memory-*" keys but we should use the better format where we can!

subiquity/server/controllers/kernel_crash_dumps.py
Chris-Peterson444: Whoops yes the intent was to use `KernelCrashDumpsConfig` and mypy likes it so I'll take it.

subiquity/server/server.py
ogayot: That's fair!

subiquity/server/controllers/kernel_crash_dumps.py
ogayot: Either way is fine, specially if you anticipate keys with `-`.

ogayot (approved):

doc/reference/autoinstall-reference.rst
```diff
@@ -1190,6 +1190,50 @@ Example:
...
+Toggle kernel crash dumps enablement.
+
+The default configuration will result in dynamic enablement of kernel crash dumps as described in the following document (doesn't exist yet).
+
+Default configuration:
+
+.. code-block:: yaml
+
+   autoinstall:
+      # Allow kernel crash dumps to be enabled dynamically.
```
dbungert, line 1211: only 24.10 sees this default behavior, pre-24.10 is default off, let's spell that out.

Chris-Peterson444: Good point. Done!

dbungert (approved):
As we're documenting defaults, we should mention the architecture bit as well for default enabled / disabled state, otherwise LGTM.

Chris-Peterson444:
> As we're documenting defaults, we should mention the architecture bit as well for default enabled / disabled state, otherwise LGTM.

Done! If I'm missing anything else let me know please.

Here is the MP for the relevant curtin changes: https://code.launchpad.net/~cpete/curtin/+git/curtin/+merge/473169

dbungert (approved):

btravouillon:
> In pre-24.10, the default configuration will result in kernel crash dumps being disabled.

In pre-24.10, it looks like using a different value from the default (`enabled: null`) also results in kernel crash dumps being disabled.

```
autoinstall:
  kernel-crash-dumps:
    enabled: true
[...]
```

On a 24.04.2 system:
```
$ sudo grep -r crash-dumps /var/log/installer/
/var/log/installer/subiquity-server-debug.log.2676:2025-02-26 21:57:42,808 WARNING subiquity.server.server:623 Unrecognized top-level key 'kernel-crash-dumps'
/var/log/installer/subiquity-server-debug.log.2676:2025-02-26 21:57:42,808 WARNING root:38 warning: subiquity/core_validation/top_level_keys: Unrecognized top-level key 'kernel-crash-dumps'
/var/log/installer/installer-journal.txt:Feb 26 21:57:42 ubuntu-server subiquity_event.2676[2676]: subiquity/core_validation/top_level_keys: Unrecognized top-level key 'kernel-crash-dumps'
/var/log/installer/installer-journal.txt:Feb 26 21:57:54 ubuntu-server subiquity.subiquity-server[2676]: warning: subiquity/core_validation/top_level_keys: Unrecognized top-level key 'kernel-crash-dumps'
/var/log/installer/subiquity-server-info.log.2676:2025-02-26 21:57:42,808 WARNING subiquity.server.server:623 Unrecognized top-level key 'kernel-crash-dumps'
/var/log/installer/subiquity-server-info.log.2676:2025-02-26 21:57:42,808 WARNING root:38 warning: subiquity/core_validation/top_level_keys: Unrecognized top-level key 'kernel-crash-dumps'
```

Maybe the documentation should be updated to reflect this parameter is not supported at all in pre-24.10?

PR #2075, opened 2024-09-05
apt: fix invocation of setup_overlay in dry-run mode

ogayot:
`setup_overlay()` expects a list of lower directories. That said, we have one invocation in dry-run mode where we pass it a path (which is a string) directly instead of wrapping the path in a list with one element.

The dry-run implementation of `setup_overlay()` accesses the first element of the list using `lowers[0]`, and expects a path out of it. But when `lowers` is a string, `lowers[0]` returns the first character of string, which usually is `"/"` and therefore corresponds to a valid path.

It makes the code accidentally work, but use a path that is not the one we expected.

### DIFF
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -375,7 +375,7 @@ class DryRunAptConfigurer(AptConfigurer):
 
     @contextlib.asynccontextmanager
     async def overlay(self):
-        yield await self.mounter.setup_overlay(self.install_tree.mountpoint)
+        yield await self.mounter.setup_overlay([self.install_tree.mountpoint])
 
     async def deconfigure(self, context, target):
         await self.cleanup()


ogayot:
It makes mypy a little bit happier \o/

```diff
@@ -675 +674,0 @@
-subiquity/server/apt.py:378: error: Argument 1 to "setup_overlay" of "Mounter" has incompatible type "str | Any"; expected "list[Mountpoint | str | OverlayMountpoint]"  [arg-type]
@@ -2008 +2007 @@
-Found 1911 errors in 142 files (checked 334 source files)
+Found 1910 errors in 142 files (checked 334 source files)
```

Chris-Peterson444 (approved):

mwhudson:
Thanks. If you can think of a more elegant way of handling dry-run mode for all this stuff, I am all ears ...

PR #2077, opened 2024-09-09
Unconditionally run `zpool export -a` / revert systemd-shutdown implementation

ogayot:
storage: export zpools unconditionally

In the _pre_shutdown function, we used to invoke `zpool export -a` if we found the presence of a zpool in the storage model.

That being said, zpools can be implicitly declared by users using {"type": "format", "fstype": "zfsroot"} in the curtin actions. This is valid curtin configuration and unfortunately causes Subiquity to think that there is no zpool.

When that happened, Subiquity did not invoke `zpool export -a` and therefore the target system couldn't boot without a call to `zpool import -f`.

We now do the call to `zpool export -a` unconditionally. Running this command when there is no zpool is a noop so it should not be a problem.

In theory there is a risk that we could export a zpool that was not meant for exporting. However, that would involve somebody importing (or force importing) a zpool in the live installer environment and not wanting it exported at the end. This sounds like an very unlikely use case. Furthermore, this could already be a problem today since we invoke `zpool export` with the `-a` option (which exports all zpools imported on the system).
    
LP:#2073772

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -2053,13 +2053,6 @@ class FilesystemModel:
             config["grub"] = self.grub
         return config
 
-    def systemd_shutdown_commands(self) -> list[list[str]]:
-        """Return a list of commands meant to be executed by systemd-shutdown.
-        We entrust the execution of `zpool export` commands to systemd-shutdown
-        instead of subiquity's _pre_shutdown hook, in hope that the commands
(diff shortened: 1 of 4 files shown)


ogayot:
> So, `zfsutils-linux` is not installed by default, which means `zpool` may not be in PATH.

Oops, my bad! https://github.com/canonical/subiquity/pull/2080

PR #2080, opened 2024-09-11
Only execute `zpool export -a` if zpool is available

ogayot:
We recently made the call to `zpool export -a` unconditional to account for "advanced" autoinstall scenarios. However, the zpool command is not always present in the live installer environment ; and it results in a Subiquity crash if the command fails to execute. Today, zpool (provided by zfsutils-linux) is dynamically installed in the live environment by curtin when it handles a ZFS installation.

Let's make sure we don't crash if zpool is not available when shutting down. It just means we're not using ZFS.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -20,6 +20,7 @@ import json
 import logging
 import os
 import pathlib
+import shutil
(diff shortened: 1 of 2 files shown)


dbungert (approved):
Thanks!

PR #2082, opened 2024-09-12
kvm-test: set cloud-config disk after real disks

Chris-Peterson444:
Ensure that the disks used for installation are passed to kvm first and not the disk with cloud-config/autoinstall data. Sometimes curtin will put paths to disk in the fstab like:

    /dev/disk/by-path/pci-000n-xxxx/by-uuid/UUID

and when we reboot without the cloud-config disk the path to the install disk will change since the number of pci devices connected to the machine has technically changed. We can make this a little more stable by letting the cloud-config drive be the last drive attached.

### DIFF
--- a/scripts/kvm-test.py
+++ b/scripts/kvm-test.py
@@ -577,21 +577,6 @@ def install(ctx):
                 kvm.append('-nographic')
                 appends.append('console=ttyS0')
 
-            if ctx.args.cloud_config is not None or ctx.args.cloud_config_default:
-                if ctx.args.cloud_config is not None:
-                    ctx.cloudconfig = ctx.args.cloud_config.read()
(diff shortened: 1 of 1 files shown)


ogayot:
I've never run into this issue I think. Is it something new in Oracular?

dbungert (approved):

Chris-Peterson444:
Yeah I hadn't run into it until recently. Definitely feels like like something new in Oracular.

dbungert:
> I've never run into this issue I think. Is it something new in Oracular?

Not  completely at least.  I ran into a conflict here with grub and upgrade testing during mantic, where grub was keeping track of the relevant device but I had changed the device due to kvm-test drive ordering.  The LP bug escapes me at the moment.

ogayot:
This looks relevant: https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/2081256

PR #2083, opened 2024-09-13
Honor match directives when installing in a gap using layout.mode: use_gap

ogayot:
Autoinstall directives allow to specify match directives when using guided `use_gap` mode:

```yaml
  storage:
    layout:
      name: direct
      mode: use_gap
      match:
        install-media: true
```

However, match directives are currently ignored ; unlike when using the `reformat_disk` mode.
Make sure they are honored.

LP:#2080608

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -25,7 +25,18 @@ import platform
 import secrets
 import tempfile
 from abc import ABC, abstractmethod
-from typing import Callable, Dict, List, Optional, Sequence, Set, Tuple, Union
+from typing import (
+    Callable,
+    Dict,
+    List,
+    Literal,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    TypedDict,
+    Union,
+)
 
 import attr
 import more_itertools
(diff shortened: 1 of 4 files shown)


ogayot:
Testing done using:

```yaml
storage:
  layout:
    name: direct
    mode: use_gap
    match:
      devpath: "*vdc"
```
and two empty disks: /dev/vdb (12GiB) and /dev/vdc (8GiB)

* Without the change, the installer picks up /dev/vdb because it is the largest disk, effectively ignoring the match directive.
* With the change, the installer picks up /dev/vdc, honoring the match directive.

The behavior looks correct to me.

On the other hand, I believe, that it will change the behavior when no match directives are specified ; from
 * pick the largest gap in the largest disk; to
 * pick the largest gap, not necessary in the largest disk

is it acceptable?

dbungert (commented):
Looks good but I'd like some test automation to go with.  `subiquity/server/controllers/tests/test_filesystem.py` has some tests that stretch the definition of Unit test but I think are effective at testing this area, and it has some `use_gap` cases already.

dbungert:
> pick the largest gap, not necessary in the largest disk

That was arguably what I had in mind, so I'd call this a bug fix.  Would be worth release noting, to clarify that the `use_gap` behavior is more intentional and can now be better controlled.

ogayot:
> Looks good but I'd like some test automation to go with. `subiquity/server/controllers/tests/test_filesystem.py` has some tests that stretch the definition of Unit test but I think are effective at testing this area, and it has some `use_gap` cases already.

I spent a bit of time doing that. I would like to cover more scenarios but it's a start..

dbungert (approved):
I see you have more testing in mind, merge when you think it's ready.

ogayot:
> I see you have more testing in mind, merge when you think it's ready.

Thanks! I'm going to hit merge but will revisit this in a follow-up PR.

PR #2090, opened 2024-09-23
Don't suggest target-resize scenarios if we don't have room for partitions

ogayot:
In https://github.com/canonical/subiquity/pull/2031, I added code to ensure that we only suggest "use-gap" scenarios if we have enough room for primary partitions.

This is basically a follow-up where we do the same for "target-resize" scenarios.

Same as in the other PR, there are some limitations:
* recovery partition is not accounted for
* only the requirements for the DIRECT capability are accounted for.

LP:#2081724

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -1147,7 +1147,7 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
         return sorted(classic_capabilities)
 
     def available_use_gap_scenarios(
-        self, install_min
+        self, install_min: int
     ) -> list[tuple[int, GuidedStorageTargetUseGap]]:
         scenarios: list[tuple[int, GuidedStorageTargetUseGap]] = []
         for disk in self.potential_boot_disks(with_reformatting=False):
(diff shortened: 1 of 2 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -1200,6 +1200,60 @@ def available_use_gap_scenarios(
             scenarios.append((gap.size, use_gap))
         return scenarios
 
+    def resize_has_enough_room_for_partitions(self, disk, resized: Partition) -> bool:
+        """Check if we have enough room for all the primary partitions. This
+        isn't failproof but should limit the number of TargetResize scenarios
+        that are suggested but can't be applied because we don't have enough
+        room for partitions."""
+        new_primary_parts = 0
```
dbungert, line 1208: A common util function between this and use_gap would be nice.

dbungert (approved):

subiquity/server/controllers/filesystem.py
ogayot: Refactored! Thanks

dbungert (approved):

PR #2093, opened 2024-09-25
kvm-test: add the option to "overwrite" an existing installation without recreating the disk from scratch

ogayot:
When running `kvm-test` with `--install`, we currently have two options related to the target disk:

 * Abort if the target already exists, This is the default behavior.
 * If the `--overwrite` option (or `-o`) is passed, we instead recreate the target image from scratch.

This makes us unable to supply an already partitioned disk to kvm-test, which is a bit of a shame considering the use-cases we have currently.

This change adds 3 new options (well, only one is truly new):

* `--preserve-target` which is an explicit way to abort if the target image already exists (this is the default behavior)
* `--recreate-target` which mimics what `-o` / `--overwrite` does.
* `--reuse-target` which is the new one. When the option is supplied, we will reuse the disk without recreating it (if it already exists).

When we install to an existing disk with `--reuse-target`, we need to ask the firmware to boot from the installation media (or it will prefer booting from the disk if there is something installed to it). In BIOS mode, this means adding the `-boot order=d` option. However, in UEFI mode, it is more difficult to do. For now, I've added a note to stderr to tell how to proceed with UEFI (i.e., mashing ESC and then selecting `Boot Manager` -> `UEFI QEMU DVD-ROM`). Suggestions welcome :)

### DIFF
--- a/scripts/kvm-test.py
+++ b/scripts/kvm-test.py
@@ -2,7 +2,7 @@
 
 '''kvm-test - boot a kvm with a test iso, possibly building that test iso first
 
-kvm-test -q --install -o --boot
+kvm-test -q --install --recreate-target --boot
    slimy build, install, overwrite existing image if it exists,
    and boot the result after install
 
@@ -14,6 +14,7 @@ import contextlib
 import copy
 import crypt
 import dataclasses
+import enum
 import os
 import pathlib
 import shlex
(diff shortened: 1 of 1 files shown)


Chris-Peterson444 (approved):
Thanks for this, LGTM. I think the only way we can fix this for UEFI is to either specify the cdrom or firmware as a full device and use `bootindex` or something like that.

For now we could add `-boot menu=on` to make the ESC mashing easier?

ogayot:
> Thanks for this, LGTM. I think the only way we can fix this for UEFI is to either specify the cdrom or firmware as a full device and use `bootindex` or something like that.
> 
> For now we could add `-boot menu=on` to make the ESC mashing easier?

Good point, TIL `-boot menu=on`, thanks!

dbungert (approved):

PR #2094, opened 2024-10-01
storage: fix partial reformat not removing all candidate partitions

ogayot:
When reformatting a disk that has in-use (i.e., mounted) partitions, we do a "partial" reformat, which means we attempt to delete all the partitions that are *not* in use.

Unfortunately, the implementation was not correct. While iterating over the partitions, calling delete_partition affects the iterator. As a consequence, some partitions that should have been removed end up preserved.

Fixed by iterating over a copy of the container.

LP:#2083322


dbungert (approved):

mwhudson:
argh!

PR #2095, opened 2024-10-01
doc: security overview

dbungert:


rkratky (changes requested):
Review submitted as https://github.com/dbungert/subiquity/pull/2

PR #2096, opened 2024-10-01
doc: move custom words to custom_wordlist

dbungert:
The intention of .wordlist.txt is to match the general wordlist, so make future merges easier by moving custom words to the intended custom_wordlist location.


rkratky (approved):

Chris-Peterson444 (approved):
Thanks! I remember looking into doing this before when @rkratky mentioned the difference but I couldn't find the original `.wordlist` to compare to.

PR #2097, opened 2024-10-04
cloud-init: check for cloud-init-base package also

dbungert:
In the future, cloud-init will be available in a more minimal
`cloud-init-base` package, which is expected to provide the
functionality we expect.  Version check that one, if present, along with
the existing `cloud-init` package.

### DIFF
--- a/subiquity/cloudinit.py
+++ b/subiquity/cloudinit.py
@@ -72,11 +72,14 @@ def get_host_combined_cloud_config() -> dict:
 
 def cloud_init_version() -> str:
     # looks like 24.1~3gb729a4c4-0ubuntu1
-    cmd = ["dpkg-query", "-W", "-f=${Version}", "cloud-init"]
-    sp = run_command(cmd, check=False)
-    version = re.split("[-~]", sp.stdout)[0]
-    log.debug(f"cloud-init version: {version}")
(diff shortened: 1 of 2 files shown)


subiquity/cloudinit.py
```diff
@@ -72,11 +72,14 @@ def get_host_combined_cloud_config() -> dict:
...
-    cmd = ["dpkg-query", "-W", "-f=${Version}", "cloud-init"]
-    sp = run_command(cmd, check=False)
-    version = re.split("[-~]", sp.stdout)[0]
-    log.debug(f"cloud-init version: {version}")
-    return version
+    for pkg in "cloud-init", "cloud-init-base":
+        cmd = ["dpkg-query", "-W", "-f=${Version}", pkg]
+        sp = run_command(cmd, check=False)
+        if version := re.split("[-~]", sp.stdout)[0]:
+            log.debug(f"cloud-init version: {version}")
```
Chris-Peterson444, line 79: suggestion: It might be useful to log which variant we found too

```suggestion
            log.debug(f"{pkg} version: {version}")
```

Chris-Peterson444 (approved):

ogayot (approved):

PR #2099, opened 2024-10-08
Oracular / storage: log disk-id in POST /storage/v2/add_boot_partition

ogayot:
It seems that we have a strange behavior when
/storage/v2/add_boot_partition is called on a disk different from the disk partitioned for installation. However, we did not log the disk ID so post-mortem analysis is very hard.

(cherry picked from commit de5356b15247cebe142e4179aa1a41a941b8b6b2)

CC: @d-loose


dbungert (approved):
LGTM, but please confirm it has seen at least one VM test.

Chris-Peterson444 (approved):

PR #2102, opened 2024-10-14
doc: Add ubuntu-image security overview

upils:
Fixes: FR-9280


doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+ubuntu-image security overview
+==============================
+
+Overview of security aspects of ubuntu-image.
+
+
+Privileged execution
+--------------------
+ 
+``ubuntu-image`` requires **elevated permissions** to properly run. It is recommended to use a dedicated building machine, make sure ``ubuntu-image`` is installed from a trusted source and the provided configuration is trusted.
```
rkratky, line 12:
```suggestion
``ubuntu-image`` requires **elevated permissions** to properly run. It is recommended to use a dedicated building machine. Make sure ``ubuntu-image`` is installed from a trusted source, and the provided configuration is trusted. 
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+Privileged execution
+--------------------
+ 
+``ubuntu-image`` requires **elevated permissions** to properly run. It is recommended to use a dedicated building machine, make sure ``ubuntu-image`` is installed from a trusted source and the provided configuration is trusted. 
+
+
+Cryptography
+------------
+
+``ubuntu-image`` is a wrapper around several lower level tools to build a Ubuntu image. Use of cryptographic technologies can then be grouped in two categories: direct usage in the tool or in one of the wrapped tools.
```
rkratky, line 18:
```suggestion
``ubuntu-image`` is a wrapper around several lower level tools to build an Ubuntu image. Therefore, the use of cryptographic technologies can be divided into two categories:

* Direct use in the ``ubuntu-image`` tool
* Indirect use in one of the wrapped tools
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+
+
+ubuntu-image itself and Go libraries
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
rkratky, line 22:
```suggestion
Cryptography in ubuntu-image itself and Go libraries
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+
+* Tarball SHA256 checksum: When building a ``classic`` image, user can provide a tarball serving as the rootfs content of the image. The SHA256 checksum of the tarball can optionally be provided to verify it has not been altered. This verification `crypto/sha256`_ Go standard library package, implementing the SHA256 hash algorithm as defined in FIPS 180-4.
```
rkratky, line 24:
```suggestion
* Tarball SHA256 checksum: When building a ``classic`` image, user can provide a tarball serving as the rootfs content of the image. The SHA256 checksum of the tarball can optionally be provided to verify it has not been altered. This verification uses the `crypto/sha256`_ Go standard library package, which implements the SHA256 hash algorithm as defined in FIPS 180-4.
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+
+* Disk UUIDs: To generate unique disk IDs, ``ubuntu-image`` uses the ``Read()`` function from the `crypto/rand`_ Go standard library package, implementing a cryptographically secure random number generator.
```
rkratky, line 26:
```suggestion
* Disk UUIDs: To generate unique disk IDs, ``ubuntu-image`` uses the ``Read()`` function from the `crypto/rand`_ Go standard library package, which implements a cryptographically secure random number generator.
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+
+* Model assertion signature: When building a ``snap`` image, user must provide a signed model assertion. Verification of this signature is handled by the `snapd library`_, relying on SHA3-384 and SHA512 for hashing and OpenPGP v4 signatures with RSA 4096/8192 keys.
```
rkratky, line 28:
```suggestion
* Model assertion signature: When building a ``snap`` image, the user must provide a signed model assertion. Verification of this signature is handled by the `snapd library`_,  which relies on SHA3-384 and SHA512 for hashing and on OpenPGP v4 signatures with RSA 4096/8192 keys.
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+
+
+Underlying tools
+~~~~~~~~~~~~~~~~
```
rkratky, line 32:
```suggestion
Cryptography in wrapped tools
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+
+* PPA handling: During the build, PPAs can be added to the ``apt`` configuration of the image. The fingerprint of the PPA can be declared in the configuration provided to ``ubuntu-image``. This value is passed down to ``gpg`` to fetch the keys to validate the packages signature. The correct handling of these keys is delegated to ``gpg`` up to the point where ``ubuntu-image`` write them to the resulting ``apt`` configuration.
```
rkratky, line 34:
```suggestion
* PPA handling: During the build, PPAs (Personal Package Archives) can be added to the ``apt`` configuration of the image. The fingerprint of the PPA can be declared in the configuration provided to ``ubuntu-image``. This value is passed to ``gpg`` to fetch the keys to validate the package signatures. The correct handling of these keys is delegated to ``gpg`` up to the point where ``ubuntu-image`` writes them to the resulting ``apt`` configuration.
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+
+* Downloaded resources: During the build various resources are fetched from remote sources (seeds, packages, gadget trees). ``ubuntu-image`` is not enforcing the usage of encrypted communication channels but if they are used (HTTPS, git over SSH, etc.) their correct handling is delegated to the tool actually pulling the resource (germinate, git, etc.). In this case, if any security-related error is making a tool fail, ubuntu-image will fail too and will display the error so the user can be alerted and take any appropriate measure to remediate.
```
rkratky, line 36:
```suggestion
* Downloaded resources: During the build, various resources are fetched from remote sources (seeds, packages, gadget trees). ``ubuntu-image`` is not enforcing the use of encrypted communication channels, but if they are used (HTTPS, Git over SSH, etc.), their correct handling is delegated to the tool actually pulling the resource (Germinate, Git, etc.). In this case, if any security-related error is causing a tool to fail, ``ubuntu-image`` also fails and displays the error, so the user is alerted and can take appropriate measures to remedy the problem.
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+Underlying tools
+~~~~~~~~~~~~~~~~
+
+* PPA handling: During the build, PPAs can be added to the ``apt`` configuration of the image. The fingerprint of the PPA can be declared in the configuration provided to ``ubuntu-image``. This value is passed down to ``gpg`` to fetch the keys to validate the packages signature. The correct handling of these keys is delegated to ``gpg`` up to the point where ``ubuntu-image`` write them to the resulting ``apt`` configuration.
+
+* Downloaded resources: During the build various resources are fetched from remote sources (seeds, packages, gadget trees). ``ubuntu-image`` is not enforcing the usage of encrypted communication channels but if they are used (HTTPS, git over SSH, etc.) their correct handling is delegated to the tool actually pulling the resource (germinate, git, etc.). In this case, if any security-related error is making a tool fail, ubuntu-image will fail too and will display the error so the user can be alerted and take any appropriate measure to remediate. 
+
+
+Misc.
+~~~~~
```
rkratky, line 40:
```suggestion
Miscellaneous
~~~~~~~~~~~~~
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+
+* Secrets (passwords and hashes) can be present in the configuration files (image definition YAML) provided to ``ubuntu-image`` to build images. Specifically:
+
+  * In the "manual" customization section, user accounts can be defined with plain text or hashed passwords. These values are directly passed down to ``chpassword`` without any treatment.
```
rkratky, line 44:
```suggestion
  * In the ``manual`` customisation section, user accounts can be defined with plain text or hashed passwords. These values are directly passed to the ``chpassword`` utility without any treatment.
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+  * In the "cloud-init" customization section, given cloud-init configuration can contain hashed password. These values are written in `cloud-init`_ configuration files without any treatment.
```
rkratky, line 45:
```suggestion
  * In the ``cloud-init`` customisation section, the given cloud-init configuration can contain hashed passwords. These values are written into `cloud-init`_ configuration files without any treatment.
```

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,55 @@
...
+
+  These configuration files should then be securely store and if secrets are used they should ideally be injected at runtime.
```
rkratky, line 47:
```suggestion
  These configuration files should then be securely stored, and if secrets are used, they should ideally be injected at runtime.
```

doc/.wordlist.txt
```diff
@@ -1,38 +1,49 @@
...
 API
 APIs
-CLI
+balancer
 Charmhub
+CLI
+crypto
+cryptographic
+cryptographically
+customization
```
rkratky, line 10:
```suggestion
customisation
```
Could we use 'customisation', please, to keep up the "BritEng by default" pretense?

rkratky (changes requested):
Thanks, @upils! A handful of suggestions.

doc/.wordlist.txt
upils: At first I used 'customization' because the key in the configuration is written with a `z`. But it is a matter of choosing with what we want to be coherent and I have no strong opinion about it so 'customisation' is fine.

doc/explanation/ubuntu-image-security-overview.rst
upils: I am not sure about the suggestion to replace `packages signature` with `package signature`. I was talking about the signature of each package, not the multiple signatures of a single package.

rkratky (approved):
Thanks!

doc/.wordlist.txt
rkratky: >  the key in the configuration is written with a z

Which would be perfectly reasonable, but Canonical wants to insist on BritEng for docs, so we're essentially documenting our AmEng software in a different language 🤦‍

doc/explanation/ubuntu-image-security-overview.rst
```diff
@@ -0,0 +1,58 @@
...
+
+* Downloaded resources: During the build, various resources are fetched from remote sources (seeds, packages, gadget trees). ``ubuntu-image`` is not enforcing the use of encrypted communication channels, but if they are used (HTTPS, Git over SSH, etc.), their correct handling is delegated to the tool actually pulling the resource (Germinate, Git, etc.). In this case, if any security-related error is causing a tool to fail, ``ubuntu-image`` also fails and displays the error, so the user is alerted and can take appropriate measures to remedy the problem.
+
+
+Miscellaneous
+~~~~~~~~~~~~~
+
+* Secrets (passwords and hashes) can be present in the configuration files (image definition YAML) provided to ``ubuntu-image`` to build images. Specifically:
```
sil2100, line 45: Should we also include information about private PPA support? Since when a private PPA is added, the yaml needs to have the token included in as well (user:auth).

sil2100 (changes requested):
Looking all good, but suggested one addition.

sil2100 (approved):
Perfect.

PR #2105, opened 2024-10-15
Bridge kernel

dbungert:


subiquity/server/controllers/kernel.py
```diff
@@ -109,12 +109,12 @@ def _confirmed(self):
...
-            self._maybe_set_bridge_kernel(BridgeKernelReason.NVIDIA, False)
+            self._maybe_set_bridge_kernel(BridgeKernelReason.DRIVERS, False)
 
     def _drivers_decided(self):
         drivers_controller = self.app.controllers.Drivers
         self._maybe_set_bridge_kernel(
-            BridgeKernelReason.NVIDIA,
+            BridgeKernelReason.DRIVERS,
             drivers_controller.model.do_install
             and any("nvidia" in driver for driver in drivers_controller.drivers),
```
mwhudson, line 119: I guess this is still correct as if ubuntu-drivers suggests some random dkms driver then we don't need to fall back to the bridge kernel. reads a little strange though, maybe it needs a comment?

subiquity/server/controllers/source.py
```diff
@@ -114,8 +114,10 @@ def start(self):
         self.app.hub.subscribe(
             (InstallerChannels.CONFIGURED, "locale"), self._set_locale
         )
-        if self.model.catalog.version != 1:
-            raise Exception("unknown source catalog version")
+        if self.model.catalog.version not in range(1, 3):
```
mwhudson, line 117: I think I'd prefer (1, 2) here

subiquity/models/source.py
```diff
@@ -67,7 +67,7 @@ def __attrs_post_init__(self):
 
 
 class BridgeKernelReason(enum.Enum):
```
mwhudson, line 69: Maybe this enum should have some documentation?

mwhudson (approved):
A couple of take-it-or-leave-it comments but basically fine thanks!

PR #2107, opened 2024-10-30
Start building Subiquity as a core24 snap

ogayot:
Let's move from core22 to core24, this makes us benefit from python 3.12 and updated python modules. The new version of urwid is known to break things (screens not refreshing without external events e.g., keyboard inputs) but at this point most of the issues we ran into should be resolved.

Marking this as a draft until we get the first 25.04 daily builds working.

@d-loose, please have a peek at the changes in the snapcraft file, you will also need to replicate https://github.com/canonical/subiquity/pull/2106


dbungert (approved):

Chris-Peterson444 (approved):

ogayot:
I'm going to drop the jammy CI test (otherwise we need to keep compatibility with python 3.10) but it would be good to figure out a way to run the test on older LTSs using the python version from the snap.

ogayot:
+ @seb128. This is the change we want to land in Subiquity for core24. https://github.com/canonical/subiquity/pull/2106 is also needed.

ogayot:
Done some testing using the plucky ubuntu-server daily ISO and didn't run into any problem. Merging.

PR #2108, opened 2024-11-01
Replace crypt module with passlib

Chris-Peterson444:
~~In python3.13, the crypt submodule will be removed from the standard library. This won't affect snap builds until core26, but makes dryrun testing difficult. This PR introduces some logic to shell out to perl to compute the hash in dry-run mode only.~~

Replace `crypt` module usage with the `passlib` module, which is provided by the  `python3-passlib` package. The package is already in Main and doesn't appear to have any open CVEs. This will fix the issue when trying to import the `crypt` module in python3.13 or later, where it was finally removed.

### DIFF
--- a/subiquitycore/utils.py
+++ b/subiquitycore/utils.py
@@ -13,7 +13,6 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import asyncio
-import crypt
 import logging
 import os
 import random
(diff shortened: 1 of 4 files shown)


subiquitycore/utils.py
```diff
@@ -234,21 +235,41 @@ def log_process_streams(
     log.log(level, "--------------------------------------------------")
 
 
-# FIXME: replace with passlib and update package deps
+def _generate_salt() -> str:
```
ogayot, line 238: Could we do the following instead of manually generating a salt and slicing?

```python
handler.using(salt_chars="abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789./")
```

I believe we should be able to mock.patch https://passlib.readthedocs.io/en/stable/lib/passlib.utils.handlers.html?highlight=_generate_salt#passlib.utils.handlers.HasSalt._generate_salt

Chris-Peterson444: So double checking the docs (e.g. for [sha512](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.sha512_crypt.html)), if unspecified a salt will be generated automatically and the default salt character set is  `[./0-9A-Za-z]` for all of the algorithms. So yeah we can remove the manual salt generation entirely now, and `@patch("passlib.utils.handlers.HasSalt._generate_salt")` works great.

Chris-Peterson444:
I think reviewing by commit makes sense still, but we can probably just squash all three commits now once we're happy with the changes.

subiquitycore/utils.py
ogayot: Excellent

ogayot (approved):

dbungert (approved):
LGTM, please give it a VM test if you haven't yet.

Chris-Peterson444:
Just tested this on an Oracular install and I was able to login as expected. Based on some out-of-band discussion, I'm going to keep all three commits for historical accuracy but I'll update the commit messages so they're a little more interesting.

PR #2111, opened 2024-11-12
Fix switching back from console-conf to getty

valentindavid:
When a password for a user is set, then we should use getty rather than console-conf. Unit conditions did not work along conflicts, because conflicts are resolved during loading of systemd, and conditions are resolved right before executing the unit.

Instead we always use console-conf. If we find that we should have switched to getty, then we return a special value that is marked as not restarting (we always restart otherwise). An ExecStopPost also matches that exit status to start getty (the default job mode is "replace", which means the conflict will also kick console-conf out).

### DIFF
--- a/bin/console-conf-wrapper
+++ b/bin/console-conf-wrapper
@@ -17,9 +17,10 @@ if [ -e /run/snapd-recovery-chooser-triggered ]; then
         # when recovery chooser is invoked it takes over the terminal
         exec /usr/lib/snapd/snap-recovery-chooser
     fi
-    if [ -e /var/lib/console-conf/complete ]; then
-        exit 0
-    fi
+fi
+
+if [ -e /var/lib/console-conf/complete ]; then
+    exit 22
 fi
 
 # always prefer to use modeevn if it is available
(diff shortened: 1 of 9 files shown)


bin/console-conf-wrapper
```diff
@@ -80,6 +81,7 @@ if [ "$(snap managed)" = "true" ]; then
         done
     else
         touch /var/lib/console-conf/complete
```
dbungert, line 83: Does this file still matter?

dbungert (commented):
Looks fine.  Some questions:
* can we get this merged to main also?
* what is being done for core24?
* what sort of testing has this seen?

bin/console-conf-wrapper
valentindavid: Yes for 2 reasons.

The next time we start this service, we will check for this file without having to go through all the checks. And that keeps it consistent with previous behavior, that is if a password was there, but later removed, we should still show the login screen.

The second reason is that snapd does look at this file.

mwhudson:
> Console-conf is a snap now in uc24. And those systemd files are part of the core-base. So we can fix it on our side. There was already a partial fix there.

Could/should we delete these files from main then?

valentindavid:
> > Console-conf is a snap now in uc24. And those systemd files are part of the core-base. So we can fix it on our side. There was already a partial fix there.
> 
> Could/should we delete these files from main then?

I think so. I can have a look next week and test what we can remove.

PR #2112, opened 2024-11-15
console-conf@.service: Remove left over conditions

valentindavid:
Something got lost in patch #2111. This fixes starting login on VT when console-conf got disabled.


valentindavid:
I redid all the manual testing with #2111 merged. But found that there was something missing from #2111 that I had in the original patch I tested. Sorry about that.

dbungert (approved):

PR #2113, opened 2024-11-18
debian: remove systemd units for console-conf

valentindavid:
Those units are now in core-base and console-conf snap only care about files in /usr/share/subiquity

I have tested manually and console-conf works correctly on UC24 without these files.


dbungert (approved):

mwhudson:
Thank you!!

PR #2114, opened 2024-11-19
scripts: mirror-checker

Chris-Peterson444:
Introduce a small script to generate a schroot for testing country mirror health. I generally found this approach just as reliable to reproduce mirror issues and has been faster than firing up a whole VM to test.


dbungert (approved):
very nice!

PR #2115, opened 2024-11-20
scripts: mirror-checker check all mirrors

Chris-Peterson444:
An update to the mirror checker script to add a command for checking all of the country mirrors. The `check-all` command takes all known country codes and then calculates the actual mirrors to test based on DNS records. The script will print out the DNS information for each attempted mirror (i.e., `cc.archive.ubuntu.com`), the calculated real mirrors, and a table with the status of each mirror. It will also save the results of each mirror check in the single mirror check mode style (which is now available under the `check` command), which can be specified with the `-o` flag.

By default it will omit reporting information about mirrors which don't exist (i.e., fallback to `archive.ubuntu.com`) but can be instructed to provide it with the `--all` flag.  

Some examples of the output: https://people.canonical.com/~cpete/mirror-checker-example-output/
- Default output: https://people.canonical.com/~cpete/mirror-checker-example-output/check-all-noble.txt
- All country codes: https://people.canonical.com/~cpete/mirror-checker-example-output/check-all-noble-with-all.txt
- check-all results for us.archive.ubuntu.com: https://people.canonical.com/~cpete/mirror-checker/mirror_checker_output/us.txt

We could clean this up a little more at some point, filtering for real mirrors using DNS information got a little messy. I also expect some more iteration on this if we want to set it up for regular monitoring somewhere.

### DIFF
--- a/scripts/mirror-checker.py
+++ b/scripts/mirror-checker.py
@@ -414,12 +414,30 @@ def check_all_mirrors(
         for cc in aliases:
             mirror_status[cc] = mirror_status[mirror]
 
+    # Print the table so that bad mirrors are at the bottom and are easier to
+    # find.
+
     table = PrettyTable(header)
-    table.add_row(["primary", *mirror_status.pop("primary")])
+
+    primary_status = mirror_status.pop("primary")
+
+    if primary_status[0] == "ok":
+        table.add_row(["primary", *primary_status])
+
+    # "ok" or "NA" statuses first
     for cc, status in sorted(mirror_status.items()):
         if cc in filtered_mirrors["non_mirrors"] and not show_non_mirrors:
             continue
-        table.add_row([cc, *status])
+        if status[0] != "not ok":
+            table.add_row([cc, *status])
+
+    # Print "not ok" status second
+    if primary_status[0] == "not ok":
+        table.add_row(["primary", *primary_status])
+
+    for cc, status in sorted(mirror_status.items()):
+        if status[0] == "not ok":
+            table.add_row([cc, *status])
 
     print(table)
 
--- a/scripts/mirror-checker.py
+++ b/scripts/mirror-checker.py
@@ -8,9 +8,8 @@ import subprocess
 import sys
 import tempfile
 from datetime import datetime, timezone
-from typing import TextIO
 from pathlib import Path
-from typing import Any
+from typing import Any, TextIO
 
 import dns.resolver
 from prettytable import PrettyTable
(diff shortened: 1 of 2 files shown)


scripts/mirror-checker.py
```diff
@@ -1,77 +1,500 @@
...
+    output: TextIO = sys.stdout,
+) -> int:
+    mirror: str = "http://archive.ubuntu.com/ubuntu/"
+    return check_mirror(mirror, release, output=output)
+
+
+def get_cname(domain: str) -> str:
+    """Get CNAME records for domain."""
+    try:
+        answers = dns.resolver.Resolver().resolve(domain, "CNAME")
```
dbungert, line 92: suggestion: https://serverfault.com/a/574077 is claiming that it's technically possible but out of spec.  what about:
```suggestion
        [answers] = dns.resolver.Resolver().resolve(domain, "CNAME")
```
which means a return from `resolve()` with more than 1 element will `raise`.

scripts/mirror-checker.py
```diff
@@ -1,77 +1,500 @@
...
+    )
+
+    # Fill in results for aliases
+    for mirror, aliases in mirror_aliases.items():
+        for cc in aliases:
+            mirror_status[cc] = mirror_status[mirror]
+
+    table = PrettyTable(header)
+    table.add_row(["primary", *mirror_status.pop("primary")])
+    for cc, status in sorted(mirror_status.items()):
```
dbungert, line 389: suggestion: it's great news that all the mirrors look healthy right now, but we expect later that won't be the case.  What do you think about sorting on Status, with `!ok` at the bottom?  It would be nice to have failing mirrors on the bottom, which means we can just check the last lines of the report for failing mirrors and if that isn't the case, we don't need to read the log much further.  WDYT?

dbungert (approved):
Thank you for putting this together, this is a great tool to help identify problem archive mirrors.  I have 2 non-blocking suggestions.

scripts/mirror-checker.py
Chris-Peterson444: Yeah I like this better. Then we could also report information about poorly configured mirrors too.

scripts/mirror-checker.py
Chris-Peterson444: Yeah that sounds much better to me.

Chris-Peterson444:
Thanks for the feedback. I think both suggestions were great and went ahead and implemented them. I'm trying something new and added you as co-author for the commits that include your suggested changes.

PR #2116, opened 2024-11-20
scripts: add mount-layered-fs

dbungert:
Add a helper script to mount a layered squashfs, as is used on the Ubuntu ISOs.


Chris-Peterson444 (approved):
Very cool!

mwhudson:
Did you see https://github.com/mwhudson/livefs-editor/commit/dd74dee9466ee95a16a2432999c60b51d8105d51? not quite the same but not completely different too...

PR #2118, opened 2024-11-22
Fix logical partitions renumbering

ogayot:
When deleting a partition, we used to renumber the logical partitions. There was two problems associated with this:
1. The new partition numbers start at N+1 where N is the partition number of the partition deleted. However, there is no guarantee that N+1 corresponds to a valid logical partition number. For instance, if we remove partition 1 (on DOS partition table), then we would start renumbering logical partitions with number 2, 3, 4, ... Instead, we should start at 5 (which is the smallest partition number for a logical partition). To address this issue, we only renumber after removing a logical partition.
2. The renumbering also applied to partitions to preserve. This can break existing systems so let's make sure we stop renumbering if we run into a partition to preserve.


ogayot:
@ogayot to check if leaving a "gap" between logical partitions numbers is acceptable.

mwhudson:
> 2\. The renumbering also applied to partitions to preserve. This can break existing systems so let's make sure we stop renumbering if we run into a partition to preserve.

We can't though, this is just how logical partitions work -- logical partitions are basically a linked list and are numbered as you walk along the list so if you delete one, the partitions after that in the list get renumbered. At least that's my understanding -- I might be wrong! But that's why things are the way they are, for better or worse.

ogayot:
> > 2. The renumbering also applied to partitions to preserve. This can break existing systems so let's make sure we stop renumbering if we run into a partition to preserve.
> 
> We can't though, this is just how logical partitions work -- logical partitions are basically a linked list and are numbered as you walk along the list so if you delete one, the partitions after that in the list get renumbered. At least that's my understanding -- I might be wrong! But that's why things are the way they are, for better or worse.

Unfortunately, that is my understanding as well after reading more about the topic. Thanks for pointing this out. I'll drop my second patch.

ogayot:
> @ogayot to check if leaving a "gap" between logical partitions numbers is acceptable.

An actual gap (e.g., having partition 6 with no partition 5) is not a thing *for logical partitions* because of how numbering works. And while it seems theoretically possible to have partition N+1 before N in terms of physical location ; many partitioning tools expect the opposite.

dbungert (approved):
Thank you!

mwhudson (approved):

PR #2120, opened 2024-12-02
doc: fix use of "flavour" (GB spelling) in examples instead of "flavor"

ogayot:
This bug was reported at:

https://bugs.launchpad.net/subiquity/+bug/2088366


Chris-Peterson444 (approved):
Thanks for this! We should probably do a pass over the other keys and check for other cases where we may have alternated between American English vs British English spelling.

PR #2123, opened 2024-12-26
install: remove previous uuid= and rp-partuuid= in the factory reset grub config

medicalwei:
We received a report that, after creating a USB reset media from the system and reinstalled, the system cannot boot into factory reset partition again.  We found out that there are two uuid= entries in `etc/grub/grub.cfg` in the reinstalled factory reset partition.

This is to remove the additional uuid= and rp-partuuid= entries if they are found in the kernel cmdline we are going to manipulate.

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -506,6 +506,13 @@ class InstallController(SubiquityController):
                 words = shlex.split(line)
                 if words and words[0] == "linux" and "---" in words:
                     index = words.index("---")
+                    words = [
+                        word
+                        for word in words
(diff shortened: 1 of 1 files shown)


dbungert (commented):
Thanks, that sounds correct.  Looks like we don't have a unit test on `adjust_rp` yet. 
 Would you add one for this case to `subiquity/server/controllers/tests/test_install.py`?

medicalwei:
Test is added for checking the grub config modification and casper uuid generation, and ensure they match each other.

The above CI failure seems to be infra issue, please confirm.

PR #2124, opened 2025-01-07
storage: add debugging TUI view to debug storage v2 guided scenarios

ogayot:
The TUI gained a new CLI option (i.e., --debug-sv2-guided) causing the normal guided storage view to be replaced by a debugging view.

The debugging view is blunt but offers at least some support for all v2 guided scenarios (target reformat, use_gap, resize, erase & install, ...) which the normal view lacks support for.

The list of scenarios is taken straight from what the API returns. There is no filtering (capabilities are basically ignored) and currently no customization is possible (e.g., resizing always takes the recommended size).

Nevertheless, this new view could be very helpful to reproduce storage bugs that occurred on the desktop installer.

Also, the makefile now has a `dryrun-debug-sv2` target that runs subiquity in dry-run mode with the relevant options to make the new debugging view usable.

Example with threebuntu-on-msdos.json:

Screenshot from 2025-01-07 13-36-03

### DIFF
--- a/subiquity/ui/views/filesystem/guided.py
+++ b/subiquity/ui/views/filesystem/guided.py
@@ -21,17 +21,23 @@ import attr
 from urwid import Text, connect_signal
 
 from subiquity.common.types.storage import (
+    Disk,
     Gap,
     GuidedCapability,
     GuidedChoiceV2,
     GuidedDisallowedCapabilityReason,
+    GuidedStorageTarget,
+    GuidedStorageTargetEraseInstall,
     GuidedStorageTargetManual,
     GuidedStorageTargetReformat,
+    GuidedStorageTargetResize,
+    GuidedStorageTargetUseGap,
     Partition,
     RecoveryKey,
 )
 from subiquity.models.filesystem import humanize_size
-from subiquitycore.ui.buttons import other_btn
+from subiquitycore.ui.buttons import forward_btn, other_btn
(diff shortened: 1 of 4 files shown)


subiquity/ui/views/filesystem/guided.py
```diff
@@ -319,6 +325,117 @@ def _toggle_guided(self, sender, new_value):
...
+                label = "Manual partitioning"
+            elif isinstance(target, GuidedStorageTargetReformat):
+                label = "Install to {} after wiping it".format(target.disk_id)
+            elif isinstance(target, GuidedStorageTargetUseGap):
+                label = "Install to {} in {} gap at offset {}".format(
+                    target.disk_id,
+                    humanize_size(target.gap.size),
+                    target.gap.offset,
+                )
+            elif isinstance(target, GuidedStorageTargetEraseInstall):
```
dbungert, line 365: suggest unifying {EraseInstall,Resize} and just fixing the label

dbungert (approved):
Nice step 1 work toward an implementation using v2 APIs.  One suggestion to consider.

Chris-Peterson444 (approved):

subiquity/ui/views/filesystem/guided.py
ogayot: done, thanks

PR #2127, opened 2025-01-13
Support "core" variant

Chris-Peterson444:
Introduce support for a "core" variant in Subiquity Server and the TUI. The changes here can be broadly summarized into two parts:

1. Refactor and update some server logic around variant handling.
2. Give the TUI a list of controllers to use for the "core" variant and add it as a supported variant in the server.

With these changes, we should be able to construct an `ubuntu-core-installer` ISO and have it display only the desired screens:
- Language
- Keyboard
- (Refresh)
- Network
- Disk

Instead of merging the changes here, I would like to open smaller PRs for each of the above changes and keep this PR for any meta-discussion around the necessary changes.

See #2128 for the first part, updating the variant handling. 
See #2129 for the second, adding support for core in the TUI.

I have performed ISO testing with these changes on Plucky dailies for Server and Desktop, as well as with the available core-installer ISO, with success.

### DIFF
--- a/examples/answers/core-desktop.yaml
+++ b/examples/answers/core-desktop.yaml
@@ -7,8 +7,6 @@ Refresh:
   update: no
 Keyboard:
   layout: us
-Zdev:
-  accept-default: yes
 Network:
   accept-default: yes
 Proxy:
@@ -16,9 +14,5 @@ Proxy:
 Filesystem:
   guided: yes
   guided-index: 0
-UbuntuPro:
-  token: ""
 InstallProgress:
   reboot: yes
-Drivers:
-  install: yes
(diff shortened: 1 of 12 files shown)


Chris-Peterson444:
CC: @sergio-costas

Chris-Peterson444:
> the type hint in the interactive method and the "return False" at the end maybe should go in a different patch

Yeah you're probably right. It would also be a good opportunity to add some tests to there for the logic.

Chris-Peterson444:
I ended up adding the changes to `interactive` in #2128 anyways, but I did add some tests. Both PRs have landed now, so I'm closing this one.

PR #2128, opened 2025-01-13
Update variant handling in SubiquityServer

Chris-Peterson444:
Part 1 of 2 for adding support for a "core" variant from #2127.

This PR introduces 1 behavior change and 1 bug fix surrounding variant handling, and 1 function signature fix for a semi-related function in the `SubiquityController` class.

- Behavior change: On startup, the server's variant will be set to the variant of the selected source in the catalog. Previously, the source controller would pick the first entry in the source catalog but not update the variant for the server. 
- Bug fix: Fixes a bug where configuring the source controller only updates the SubiquitModel (`app.base_model`) and not the server's notion of the variant (`app.variant`). Now selecting a new source will update both.
- Minor fix: Update `SubiquityController.interactive` to return `False` instead of `None` for non-interactive controllers, since this is always used in truthy contexts.

There is one temporary commit here that adds "core" as a supported variant in `subiquity/server/server.py`, just so that the `examples/sources/core-desktop.yaml` answers test will pass. (The TUI will get "core" as the client and then try to post it to the server, which would otherwise fail without this commit.) We can drop this commit before merging, as the logic to add "core" as a supported variant is handled in the other PR.

### DIFF
--- a/subiquity/server/controller.py
+++ b/subiquity/server/controller.py
@@ -111,7 +111,35 @@ class SubiquityController(BaseController):
         """
         pass
 
-    def interactive(self):
+    def interactive(self) -> bool:
+        """Return whether this controller is "interactive".
+
+        An "interactive" controller is one the installer client is expected to
+        ask the user questions about.  Why would a controller not be
+        interactive? There are a few reasons:
+
+         1. Some controllers do not have an associated UI and can only be
+            configured via autoinstall (e.g. the kernel controller)
+         2. When autoinstalling, the default is that _no_ controllers are
+            interactive, although this can be controlled via
+            `interactive-sections` in the autoinstall config.
+         3. Some controllers are not relevant when installing a particular
+            variant, e.g., we do not ask the user about timezone when
+            installing a server system.
+
+        Interactive controllers are marked configured when the user moves on
+        from the screen for that controller. Non-interactive controllers are
+        configured at different times:
+
+         1. Controllers with no UI are always marked configured by the
+            SubiquityServer during startup, after running the controller's
+            apply_autoinstall_config method.
+         2. During an autoinstall, controllers not listed under
+            'interactive-sections' become non-interactive and are marked
+            configured at the same time as (1).
+         3. Controllers not applicable for a variant are marked configured
+            when the install is confirmed.
+        """
         if not self.app.autoinstall_config:
             return self._active
         i_sections = self.app.autoinstall_config.get("interactive-sections", [])
@@ -127,6 +155,8 @@ class SubiquityController(BaseController):
             and self.autoinstall_key_alias in i_sections
         ):
             return self._active
+
+        return False
 
     async def configured(self):
         """Let the world know that this controller's model is now configured."""
--- a/subiquity/server/server.py
+++ b/subiquity/server/server.py
@@ -119,7 +119,6 @@ class MetaController:
     async def client_variant_POST(self, variant: str) -> None:
         if variant not in self.app.supported_variants:
             raise ValueError(f"unrecognized client variant {variant}")
-        self.app.base_model.set_source_variant(variant)
         self.app.set_source_variant(variant)
 
     async def client_variant_GET(self) -> str:
@@ -291,6 +290,10 @@ class SubiquityServer(Application):
         root = "/"
         if self.opts.dry_run:
             root = os.path.abspath(self.opts.output_base)
+        # TODO: Set the model source variant before returning it?
+        #       This _will_ eventually get set by the source controller,
+        #       but before then it's in a state that only requires the
+        #       "default" models i.e., the base set all variants require.
         return SubiquityModel(
             root,
             self.hub,
@@ -302,7 +305,7 @@ class SubiquityServer(Application):
     def __init__(self, opts, block_log_dir):
         super().__init__(opts)
         self.dr_cfg: Optional[DRConfig] = None
-        self.set_source_variant(self.supported_variants[0])
+        self._set_source_variant(self.supported_variants[0])
         self.block_log_dir = block_log_dir
         self.cloud_init_ok = None
         self.state_event = asyncio.Event()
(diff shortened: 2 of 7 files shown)


subiquitycore/tests/mocks.py
```diff
@@ -23,6 +24,7 @@ class MockedApplication:
     autoinstall_config = {}
     answers = {}
     opts = None
+    set_source_variant = SubiquityServer.set_source_variant
```
ogayot, line 27: What's going on here? Even though it's in tests code, I'd rather avoid the dependency from `subiquitycore` to `subiquity`.

subiquity/server/controller.py
```diff
@@ -128,6 +128,8 @@ def interactive(self):
         ):
             return self._active
 
+        return False
```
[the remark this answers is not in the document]
Chris-Peterson444: Yeah I think adding a comment would be good for explaining the uses of this function, if not an eventual refactor of how `interactive` is determined.  

Generally, we will hit the `return False` line if we are performing an autoinstall and that controller wasn't specified as an interactive controller. In this view, I believe we could replace each `return self._active` call with a `return True` and it would functionally be the same.

However, the `interactive_for_variants` concept changes this and may turn a controller into a non-interactive one after the install confirmation step, regardless of it being specified in the `interactive-sections` section of the autoinstall config.

subiquity/server/server.py
```diff
@@ -354,6 +353,11 @@ def __init__(self, opts, block_log_dir):
     def set_source_variant(self, variant):
         self.variant = variant
 
+        # Update the model variant too.
+        # Guard for init. The "base_model" attribute isn't set until .run() is called.
+        if getattr(self, "base_model", None) is not None:
```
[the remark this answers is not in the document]
Chris-Peterson444: Handy, thanks! I forgot about that one.

subiquitycore/tests/mocks.py
Chris-Peterson444: That's a good point. Originally I had this in place because the identity controller tests needed access to this function through a dependency on the source controller and I thought it better to abstract that logic in mock application than have the identity controller tests need to know what to mock. But that's resolved with #2131 so I'll update this to just patch it in the source controller.

Chris-Peterson444: I rebased on the updated tests and moved the logic to patch that function in just the source controller tests. Wdyt?

subiquity/server/controller.py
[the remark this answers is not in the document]
Chris-Peterson444: I decided to make it a docstring instead, which I think describes the usage well. Wdyt?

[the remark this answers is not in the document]
ogayot: Works for me, thanks!

subiquitycore/tests/mocks.py
ogayot: Yes, that's much better IMO.

ogayot (approved):

dbungert (approved):

subiquity/server/controller.py
```diff
@@ -111,7 +111,17 @@ async def apply_autoinstall_config(self, context):
...
+        """Get the interactivity status of this controller.
+
+        A controller's interactive state is True by default, unless performing
+        an autoinstall and the controller was not implicitly nor explicitly
+        included in the `interactive-sections` section of the autoinstall config.
+
+        This exception to this is for controllers that opt-in to the
+        `interactive_for_variants` system, which will automatically configure
+        the controller and mark it non-active after the install confirmation step.
+        """
```
mwhudson, line 124: Ugh the fact that this is such a headscratcher is possibly an indication that Michael made some mistakes a few years ago. I don't think we have a good explanation in the code about what an "interactive" controller means so maybe let's write a bit of an essay here (seems as good a place as any other) as to what it means. How about this:

```

"""
Return whether this controller is "interactive".

An "interactive" controller is one the installer client is expected to
ask the user questions about.  Why would a controller not be
interactive? There are a few reasons:

 1. some controllers do not have a UI and can only be configured via autoinstall
    (e.g. the kernel controller)
 2. when autoinstalling, the default is that _no_ controllers are interactive,
    although this can be controlled via `interactive-sections` in the
    autoinstall config.
 3. some controllers are not relevant when installing a particular variant,
    e.g. we do not ask the user about timezone when installing a server system.

Interactive controllers are marked configured when the user moves on
from the screen for that controller. Non-interactive controllers are
configured at different times:

 1. controllers with no UI are configured XXX i don't know when??
 2. during an autoinstall, controllers are marked configured after
    their apply_autoinstall_config method has run
 3. controllers not applicable for a variant are marked configured
    when the install is confirmed.
"""
```

It is slightly ridiculous that I can't remember / find when controllers with no UI don't get configured!

subiquity/server/server.py
```diff
@@ -354,6 +353,11 @@ def __init__(self, opts, block_log_dir):
     def set_source_variant(self, variant):
         self.variant = variant
 
+        # Update the model variant too.
+        # Guard for init. The "base_model" attribute isn't set until .run() is called.
```
mwhudson, line 357: Ugh that's a bit nasty.

mwhudson (approved):
Looks ok to me. I think we can beef up that docstring a bit though.

subiquity/server/controller.py
Chris-Peterson444: I like this much better, thanks!

I misspoke in my original docstring a bit, sorry. A `SubiquityController` is generally always interactive unless performing autoinstall, but (1) implies a `NonInteractiveController` which is always False. So they always do (2) where the server applies its autoinstall config and then configures it.

subiquity/server/server.py
Chris-Peterson444: I agree and now I'm looking at it again, I think we could do this a bit smarter. The problem is I wanted to set the variant through this interface on `__init__` but the model hasn't been attached yet. This way also doesn't solve the problem of the informing the model about our variant when we attach it (though the way the code exists now it's not strictly necessary).

subiquity/server/controller.py
Chris-Peterson444: Alright, I've taken this with some slight modifications. I also went ahead and added some unit tests just to show it does what is says. What do you think?

subiquity/server/server.py
Chris-Peterson444: I refactored this slightly and added some comments to `make_model` to hopefully make this a little less gross.

subiquity/server/controller.py
mwhudson: LGTM, thanks for this.

mwhudson (approved):

PR #2130, opened 2025-01-14
Kvm test swtpm

dbungert:
Several fixes to swtpm integration allowing TPMFDE test automation, including having TPM state persist across VM boots, allowing multiple VMs with TPM to run in parallel, and an apparmor integration workaround.

Sample command line
```
./scripts/kvm-test.py -r24.04.1-desktop --install -o -b --boot --profile desktop --with-tpm2
```

### DIFF
--- a/scripts/kvm-test.py
+++ b/scripts/kvm-test.py
@@ -107,13 +107,14 @@ class Context:
         except KeyError:
             pass
         self.curdir = os.getcwd()
-        self.iso = f'/tmp/kvm-test/{self.release}-test.iso'
         self.hostname = f'{self.release}-test'
-        self.target = f'/tmp/kvm-test/{self.hostname}.img'
-        parent = pathlib.Path(self.target).parent
+        self.rundir = pathlib.Path('/tmp/kvm-test')
+        self.iso = self.rundir / f'{self.release}-test.iso'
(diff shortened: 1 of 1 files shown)


scripts/kvm-test.py
```diff
@@ -571,7 +572,7 @@ def install(ctx):
     if os.path.exists(ctx.target):
```
ogayot, line 572: Maybe we want ctx.vmstate here as well. If we pass `--no-disk` (e.g., nvme/tcp), this condition will always be false.

scripts/kvm-test.py
```diff
@@ -722,7 +718,7 @@ def main() -> None:
     if ctx.args.base and ctx.args.build:
         raise Exception('cannot use base iso and build')
 
-    os.makedirs('/tmp/kvm-test', exist_ok=True)
+    os.makedirs(ctx.rundir, exist_ok=True)
```
ogayot, line 722: `parents=True`?

ogayot (commented):
Thanks! Two suggestions. Otherwise, LGTM

scripts/kvm-test.py
dbungert: Fixed

ogayot (approved):

PR #2134, opened 2025-01-16
Erase install - disable renumbering instead of excluding scenarios

ogayot:
This is another option to address the issue with partition renumbering. Instead of completely hiding a erase-and-install scenario just because renumbering is a problem, we disable it for erase-and-install scenarios only.

The (possibly weak) assumption is that because we are erasing one partition and installing a new system **at the same place**, we will never end up with less logical partitions that we had before (and therefore we can avoid renumbering).

This makes it possible to do erase-and-install on a logical partition that is not at the end (e.g., p5 on threebuntu-on-msdos.json).

### DIFF
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -120,7 +120,7 @@ class FilesystemManipulator:
         self.create_filesystem(part, spec)
         return part
 
-    def delete_partition(self, part, override_preserve=False):
+    def delete_partition(self, part, override_preserve=False, allow_renumbering=True):
         if (
             not override_preserve
             and part.device.preserve
@@ -128,7 +128,7 @@ class FilesystemManipulator:
         ):
             raise Exception("cannot delete partitions from preserved disks")
         self.clear(part)
-        self.model.remove_partition(part)
+        self.model.remove_partition(part, allow_renumbering=allow_renumbering)
 
     def create_raid(self, spec: RaidSpec):
         for d in spec["devices"] | spec["spare_devices"]:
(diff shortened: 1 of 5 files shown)


dbungert (commented):
One suggestion.

Do you envision this as mostly a workaround?

ogayot:
> Do you envision this as mostly a workaround?

Depends ...

Today, it is functionally an improvement because it makes it possible to run erase-and-install on a logical partition that is not at the end. Excluding the scenario because of renumbering (which is what the other PR does) feels more like a workaround to me.

Now, the assumption that erase-and-install will remove **one** partition and add a (at least) one new at the same place feels a bit weak. What if we decide to support erase-and-install with an existing installation that uses multiple partitions? Suddenly, there will be no guarantee that renumbering can be avoided.

And from the code standpoint:

Calling `disk.renumber_logical_partitions()` unconditionally as part of `remove_partition` felt wrong. Changing partition numbers has implications and IMO we should only do it when we have no other option. So the `allow_renumbering` option is definitely an improvement to me.
I also think that there are more valid use-cases for `allow_renumbering=False` than there are for `allow_renumbering=True` but I'll refrain from changing the default (for now?) because I'm afraid of regressions.

Typically, I'd prefer a single, explicit call to `disk.renumber_logical_partitions()` before proceeding with the install if we have a "hole" in partition numbers. For manual partitioning, this is another story because we do not want the UI to "lie" about partition numbers that might change after the user validates their config.

dbungert (approved):

Chris-Peterson444 (approved):

PR #2137, opened 2025-01-16
extract some functionality from FilesystemController

dbungert:
That class is way too big.


subiquity/server/snapd/system_getter.py
```diff
@@ -0,0 +1,103 @@
+# Copyright 2024 Canonical, Ltd.
```
ogayot, line 1: year++

ogayot (approved):
Thanks!

PR #2141, opened 2025-01-20
console-conf@.service: do not exit as error when preventing restart

valentindavid:
Upstreaming of https://github.com/canonical/core-base/pull/296

This does not affect behavior apart from systemctl not showing the service in red.

### DIFF
--- a/bin/console-conf-check-error
+++ b/bin/console-conf-check-error
@@ -2,6 +2,6 @@
 
 set -eu
 
-if [ "${SERVICE_RESULT}" == "exit-code" ] && [ "${EXIT_STATUS}" -eq 22 ]; then
+if [ "${SERVICE_RESULT}" == "success" ] && [ "${EXIT_STATUS}" -eq 22 ]; then
   systemctl start --no-block "${1}"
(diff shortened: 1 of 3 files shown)


dbungert:
Thanks Valentin.

I have a question - I'm looking at this combination and it seems a little unexpected
```
RestartPreventExitStatus=22
SuccessExitStatus=22
Restart=always
```

We're deliberately using `exit 22` as success (fine) and not restarting with that (fine), but that means that an `exit 0` would be a restart while `exit 22` would not restart, is this intended?

I wonder if we instead want this, which means an exit 0 doesn't restart and we don't need `RestartPreventExitStatus`.  Thoughts?
```
SuccessExitStatus=22
Restart=on-failure
```

valentindavid:
> We're deliberately using exit 22 as success (fine) and not restarting with that (fine), but that means that an exit 0 would be a restart while exit 22 would not restart, is this intended?

Yes, this is exactly what we want. We always want to restart, unless the script tells use not to.

dbungert (approved):

PR #2144, opened 2025-01-22
Fix the pc-kernel issue

ogayot:
**Initial problem** 

The `SystemGetter.get` method essentially does 3 important things:
 * it mounts stuff in `/var/lib/snapd/seed`
 * it queries snapd (and expect it to read contents from `/var/lib/snapd/seed`)
 * it then tries to cleanup by unmounting what it previously mounted
    
Unfortunately, if we cancel this task in the middle, especially while it is querying snapd, the cleanup code will fail to unmount things because snapd is still using `/var/lib/snapd/seed` (cancelling the query does not mean that snapd will immediately release the resources - or even stop processing the query). Typically this results in a failure to unmount pc-kernel.snap with EBUSY.
```
umount: /var/lib/snapd/seed/snaps/pc-kernel_2010.snap: target is busy.
```
    
Effectively, if `SystemGetter.get` gets cancelled, we would need to wait until the query to snapd finishes before we run the cleanup code.

But the approach that this PR takes is to make sure `SystemGetter.get` is not cancelled in the first place.

`SystemGetter.get` was subject to cancellation because it runs as part of  `_examine_systems` - which is cancelled and restarted when the source changes.

We now make sure that we shield `SystemGetter.get` from cancellation when we execute it from `_examine_systems`. If `_examine_system` gets cancelled, `SystemGetter.get` will now continue running in the background.

**Which leads to problem 2**

If we cancel `_examine_system` and rerun it, we end up with two concurrent calls to `SystemGetter.get` (the new one + the one we want to finish). Obviously, things go sideways if this happens. 

To avoid the problem, we now acquire a lock at the beginning of `SystemGetter.get` - so that only one task runs the critical section at a given time.

**Which leads to problem 3**

Internally, `SystemGetter.get(variation)` expects that the variation exists for the currently configured source.

However, because of the lock we added to `SystemGetter.get`, it is now more likely that the configured source has changed by the time we start mounting stuff and querying snapd.
    
When that happens, `SystemGetter.get(variation)` fails with `KeyError` ; because the variation is no longer present in the currently configured source.

Fixed by making sure that we use the source handler that was configured when `SystemGetter.get(variation)` was first invoked.

This bug was already present but way more difficult to reproduce timing-wise.

**Which leads to problem 4**
    
With the lock added to `SystemGetter.get`, when we change the source multiple times quickly, we create a "queue" of `SystemGetter`.get tasks. This can take a long time to consume, and really we only care about the result of the last query.

When `_examine_systems` gets cancelled, we now check if the `SystemGetter.get` task has successfully acquired the lock. If it hasn't, then we cancel it.

LP:#2084032

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -434,7 +434,20 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             system = None
             label = variation.snapd_system_label
             if label is not None:
-                system, in_live_layer = await self._system_getter.get(name, label)
+                # We do not want to propagate cancellation to
+                # _system_getter.get. If it gets cancelled during its critical
+                # section, it won't be able to properly clean up after itself
+                # (see LP: #2084032).
+                # Therefore we use an asyncio.Task (coupled with
+                # asyncio.shield) so we can prevent propagation.
+                task = asyncio.create_task(self._system_getter.get(name, label))
+
+                system, in_live_layer = await asyncio.shield(task)
+                # _system_getter.get is marked async_helpers.exclusive
+                # so it should be safe to let it finish "unsupervised" if we
+                # get cancelled even though it might be called again
+                # concurrently.
+
             log.debug("got system %s for variation %s", system, name)
             if system is not None and len(system.volumes) > 0:
                 if not self.app.opts.enhanced_secureboot:
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -434,21 +434,32 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
             system = None
             label = variation.snapd_system_label
             if label is not None:
-                # We do not want to propagate cancellation to
+                # We do not want to unconditionally propagate cancellation to
                 # _system_getter.get. If it gets cancelled during its critical
                 # section, it won't be able to properly clean up after itself
                 # (see LP: #2084032).
                 # Therefore we use an asyncio.Task (coupled with
                 # asyncio.shield) so we can prevent propagation.
+                in_critical_section = asyncio.Event()
                 task = asyncio.create_task(
-                    self._system_getter.get(name, label, source_id=catalog_entry.id)
+                    self._system_getter.get(
+                        name,
+                        label,
+                        source_id=catalog_entry.id,
+                        started_event=in_critical_section,
+                    )
                 )
-
-                system, in_live_layer = await asyncio.shield(task)
-                # _system_getter.get is marked async_helpers.exclusive
-                # so it should be safe to let it finish "unsupervised" if we
-                # get cancelled even though it might be called again
-                # concurrently.
+                try:
+                    system, in_live_layer = await asyncio.shield(task)
+                except asyncio.CancelledError:
+                    if not in_critical_section.is_set():
+                        # Just to make sure we don't end up with a large queue of
+                        # _system_getter.get() tasks.
+                        task.cancel()
+                    # _system_getter.get is marked async_helpers.exclusive
+                    # so it should be safe to let it finish "unsupervised" even
+                    # though it might be called again concurrently.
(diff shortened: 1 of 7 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -434,7 +434,28 @@ async def _examine_systems(self):
...
             label = variation.snapd_system_label
             if label is not None:
-                system, in_live_layer = await self._system_getter.get(name, label)
+                # We do not want _system_getter.get to be cancelled if
+                # _examine_systems is cancelled if it is running its critical
+                # section - because it needs to cleanup after itself. However,
+                # it should be safe to call _system_getter.get multiple times
+                # concurrently. There is a lock to prevent mount/unmounts from
+                # being done concurrently.
+                in_critical_section = asyncio.Event()
```
dbungert, line 443: So we have the SingleInstanceTask concept, do you think it would help in this context?  It seems like it's close in concept but not a direct match, which suggests to me that we want to abstract this out so we can offer a variation of this shielded SingleInstanceTask - given how treacherous cancellation is, I bet we have other places where we should be using a concept like this.

dbungert (commented):
see note

Chris-Peterson444 (commented):
Wow, very nice! 

I'm +1 on the `SingleInstanceTask`-like abstraction. On the one hand I feel like it's not necessarily blocking for this PR, but on the other it might help with unit testing some of this which otherwise sounds nightmarish.

subiquity/server/controllers/filesystem.py
ogayot: I've considered something like SingleInstanceTask, yes. It's true that in essence it can prevent concurrent execution of a task. But it is inconvenient to use.

1. we must make sure that all callers of `async def unsafe()` share the same `SingleInstanceTask(unsafe)` instance. if we call `unsafe()` from anywhere else, there could still be concurrent executions.
2. `SingleInstanceTask.start()` will either raise an exception or cancel the previous task if there was one.
3. I don't trust the `except BaseException` from its implementation...

I've just implemented the `@aync_helpers.exclusive` decorator (maybe the name is inappropriate) and used it on `SystemsGetter.get`. This does not solve the cancellation problem. But it does prevent concurrent execution of a function/method. So it is a start (and makes the code a bit easier to read).

Chris-Peterson444 (approved):

subiquity/server/controllers/filesystem.py
dbungert: Thanks.  I don't have a better suggestion than `exclusive` so unless a new suggestion arrives shortly, please use that.

dbungert (approved):

subiquitycore/async_helpers.py
```diff
@@ -117,3 +118,22 @@ def done(self):
...
+def exclusive(coroutine_function):
+    """Can be used to decorate a coroutine function that we do not want to run
+    multiple times concurrently. It uses a lock internally.
+    If the caller needs to know when the decorated coroutine starts executing
+    (i.e., when it has acquired the exclusive lock), they can pass an
+    asyncio.Event as the "started_event" keyword-only argument.
+    """
+    lock = asyncio.Lock()
+
+    async def wrapped(*args, started_event: Optional[asyncio.Event] = None, **kwargs):
```
mwhudson, line 132: Does it make sense to have started_event be kwarg-only?

mwhudson (approved):
It's kind of distressing that all this machinery is needed but I don't see any real scope for simplifications.

I have one very minor suggestion otherwise this LGTM! (and well done for fighting through it all!)

subiquitycore/async_helpers.py
ogayot: I think so? I mean, if I change it to not kwarg-only, it would have to be the first positional argument, right? I'm slightly confused.

mwhudson: Oh ignore me I misread that it already is :-)

PR #2145, opened 2025-01-23
Pin pass stub

dbungert:
Start work on pin/passphrase with TPMFDE with a stub entropy calculation endpoint and some basic API validation.

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -70,6 +70,7 @@ from subiquity.common.types import (
 from subiquity.common.types.storage import (
     AddPartitionV2,
     Disk,
+    EntropyResponse,
     GuidedChoiceV2,
     GuidedStorageResponseV2,
     ModifyPartitionV2,
@@ -428,6 +429,17 @@ class API:
                 def DELETE(id: str) -> StorageResponseV2:
                     """Delete the Raid specified by its ID. Any associated
                     partition will be deleted as well."""
+
+            class calculate_entropy:
+                def POST(
+                    passphrase: Optional[str] = None,
+                    pin: Optional[str] = None,
+                ) -> EntropyResponse:
+                    """Calculate the entropy associated with the supplied
+                    passphrase or pin.  Clients must use this endpoint to
+                    confirm that the pin or passphrase is suitable prior to
+                    configuring CORE_BOOT_ENCRYPTED, and may use it in other
+                    scenarios."""
 
     class codecs:
         def GET() -> CodecsData:
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -42,6 +42,7 @@ from subiquity.common.types.storage import (
     AddPartitionV2,
     Bootloader,
     Disk,
+    EntropyResponse,
     GuidedCapability,
     GuidedChoiceV2,
     GuidedDisallowedCapability,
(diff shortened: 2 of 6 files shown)


subiquity/server/controllers/filesystem.py
```diff
@@ -262,6 +263,23 @@ def dd(cls, name: str, min_size: int):
...
+    if passphrase is not None and pin is not None:
+        raise StorageRecoverableError("must supply at most one of pin and passphrase")
+    if not pin_allowed and pin is not None:
+        raise StorageRecoverableError("unexepected pin supplied")
+    if not passphrase_allowed and passphrase is not None:
+        raise StorageRecoverableError("unexepected passphrase supplied")
+
+    if pin is not None:
+        try:
+            int(pin)
```
ogayot, line 278: `int()` is quite permissive. For instance `int(+4)` and `int(-3)` are valid. Instead we could use `str.isdecimal`.
```python
if not pin.isdecimal():
    raise
```

subiquity/common/apidef.py
```diff
@@ -429,6 +430,17 @@ def DELETE(id: str) -> StorageResponseV2:
                     """Delete the Raid specified by its ID. Any associated
                     partition will be deleted as well."""
 
+            class calculate_entropy:
```
ogayot, line 433: I would have moved this outside the `v2` namespace but maybe my understanding of this namespace is incorrect.

subiquity/server/controllers/tests/test_filesystem.py
```diff
@@ -2343,3 +2345,68 @@ def test_is_reset_partition_only_utility(self, config, expected):
...
+            (GuidedCapability.LVM_LUKS, False, True),
+            (GuidedCapability.CORE_BOOT_ENCRYPTED, True, True),
+        )
+    )
+    def test_capability_pin_pass_validation(self, capability, pin_ok, pass_ok):
+        def maybe_assert_raises(ok: bool):
+            if ok:
+                return contextlib.nullcontext()
+            else:
+                return self.assertRaises(StorageRecoverableError)
```
Chris-Peterson444, line 2372: praise: I really like this construction. I've written enough tests like:
```python
if ok:
    # do thing
else:
    with self.assertRaises(...):
        # do bad thing
```
and it makes me want to go fix them now.

subiquity/common/apidef.py
dbungert: Added to meeting agenda to discuss

subiquity/server/controllers/filesystem.py
```diff
@@ -262,6 +263,23 @@ def dd(cls, name: str, min_size: int):
...
 
+def validate_pin_pass(
+    passphrase_allowed: bool, pin_allowed: bool, passphrase: str, pin: str
+) -> None:
+    if passphrase is not None and pin is not None:
+        raise StorageRecoverableError("must supply at most one of pin and passphrase")
+    if not pin_allowed and pin is not None:
+        raise StorageRecoverableError("unexepected pin supplied")
+    if not passphrase_allowed and passphrase is not None:
+        raise StorageRecoverableError("unexepected passphrase supplied")
```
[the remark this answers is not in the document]
dbungert: Fixed

subiquity/server/controllers/filesystem.py
dbungert: Much more sensible

ogayot (approved):

PR #2148, opened 2025-01-27
When changing format of a new parittion, don't require wipe

ogayot:


subiquity/server/controllers/filesystem.py
```diff
@@ -1398,10 +1398,15 @@ async def v2_edit_partition_POST(
...
         spec: PartitionSpec = {"mount": data.partition.mount or partition.mount}
+        # NOTE from ogayot: it is my understanding that in this context, having
+        # format=None (which is the default value) means to keep the current
+        # format. To change from ext4 to "unformatted", one must specify
+        # format="".
         if data.partition.format is not None:
             if data.partition.format != partition.original_fstype():
-                if data.partition.wipe is None:
-                    raise ValueError("changing partition format requires a wipe value")
+                if data.partition.wipe is None and partition.preserve:
```
ogayot, line 1407: Should we check the value of `preserve` at the FS level (which may or may not exist) ? This is what we do to label "to be reformatted as ...." and "to be formatted as ....".

https://github.com/canonical/subiquity/blob/ad2dd0aa8330ec909db648c844e8254ee93e3e05/subiquity/common/filesystem/labels.py#L228

ogayot: @dbungert do you know? i.e., the difference between partition.preserve and partition.fs.preserve?

dbungert: I will defer to the curtin documentation on this one. https://curtin.readthedocs.io/en/latest/topics/storage.html 

From a label standpoint this is a very subtle difference.

subiquity/server/controllers/filesystem.py
```diff
@@ -1398,10 +1398,16 @@ async def v2_edit_partition_POST(
...
```
Chris-Peterson444, line 1407: Just to be clear, this works because `preserve` is already set for existing objects by default?

ogayot: Yes, we set `preserve=True` for everything coming from probert.

https://github.com/canonical/subiquity/blob/b3df1cf64d74a32c6656b5d23d4075d94ccd5087/subiquity/models/filesystem.py#L1653-L1661

https://github.com/canonical/subiquity/blob/b3df1cf64d74a32c6656b5d23d4075d94ccd5087/subiquity/models/filesystem.py#L2022-L2023

(sorry for taking a year to respond ...)

subiquity/server/controllers/filesystem.py
ogayot: The v2 storage API only exposes the FS as a property of a partition. So the presence of `preserve` should be checked at the partition level too. So a partition with `preserve=True` prevents the FS from being modified and `preserve=False` allowed the FS to be modified.

Let's check `preserve` at the partition level like we're doing at the moment.

dbungert (approved):
Thanks for digging this up, would be worth a desktop VM test with manual partitioning.

subiquity/server/controllers/filesystem.py
```diff
@@ -1723,11 +1723,15 @@ async def v2_edit_partition_POST(
                     "edit_partition does not support changing partition name"
                 )
         spec: PartitionSpec = {"mount": data.partition.mount or partition.mount}
+        # NOTE from ogayot: it is my understanding that in this context, having
+        # format=None (which is the default value) means to keep the current
+        # format. To change from ext4 to "unformatted", one must specify
+        # format="".
```
ogayot: Unfortunately, this was not be true. Using `format=""` to make the partition unformatted might work accidentally but we should not rely on that.

When doing so, the logs will show `"adding  to Partition..."` where `""` is the filesystem name.

Chris-Peterson444 (approved):

PR #2149, opened 2025-01-27
storage: always accept to resize a partition that is being wiped

ogayot:
If we are wiping a partition, don't bother checking if the filesystem that was on it supports resizing. The filesystem is going to be erased anyway. The new filesystem (if any) will be created using mkfs.* so there is no need for resize support.

LP:#2096799

### DIFF
--- a/subiquity/common/filesystem/tests/test_manipulator.py
+++ b/subiquity/common/filesystem/tests/test_manipulator.py
@@ -754,3 +754,9 @@ class TestCanResize(unittest.TestCase):
         part = make_partition(self.manipulator.model, disk, preserve=True)
         make_filesystem(self.manipulator.model, partition=part, fstype="asdf")
         self.assertFalse(self.manipulator.can_resize_partition(part))
+
+    def test_resize_invalid_but_wipe(self):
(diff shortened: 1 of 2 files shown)


subiquity/common/filesystem/tests/test_manipulator.py
```diff
@@ -743,6 +746,22 @@ def test_resize_unpreserved(self):
         part = make_partition(self.manipulator.model, disk, preserve=False)
         self.assertTrue(self.manipulator.can_resize_partition(part))
 
+    def test_resize_non_resizable(self):
+        disk = make_disk(self.manipulator.model, ptable=None)
+        part = make_partition(self.manipulator.model, disk, preserve=True)
+        # If XFS becomes resizable, replace with something else.
```
dbungert, line 752: maybe use a non-real fs name instead?

dbungert (approved):

subiquity/common/filesystem/tests/test_manipulator.py
ogayot: Done!

PR #2150, opened 2025-01-27
kernel: fix install hang with autoinstall + kernel

dbungert:
Fix an install hang that is reproducible today on main branch (oracular and earlier not affected) when supplying a kernel to install via autoinstall.


Chris-Peterson444 (approved):
Nice catch!

PR #2155, opened 2025-01-30
storage: fix use of ptable parameter when calling reformat

ogayot:
When reformatting a disk using the following call:

```python
manipulator.reformat(disk, ptable="something")
```

The ptable parameter would only appear to be honored if the disk did not contain any partition. This happened because the `reformat()` function sets the ptable of the disk first and then starts removing partitions using `delete_partition()` in a loop.

However, `delete_partition()` has a mechanism to reset the partition table to `None` when it is invoked on the last partition of a disk.

Therefore, the disk ended up with `None` as the partition table (ignoring the ptable value) if the disk contained any partition.

Fixed by moving the assignment of `disk.ptable` after removing partitions. Arguably, it is a mistake to change the partition table value before removing the partitions anyway.

### DIFF
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import logging
+from typing import Optional
 
 from curtin.block import get_resize_fstypes
 
@@ -268,12 +269,13 @@ class FilesystemManipulator:
         for subobj in obj.fs(), obj.constructed_device():
             self.delete(subobj)
 
-    def reformat(self, disk, ptable=None, wipe=None):
+    def reformat(self, disk, ptable: Optional[str], wipe=None):
+        """Reformat the specified disk. If ptable is None, use the default
+        partition table type."""
         disk.grub_device = False
         for p in list(disk.partitions()):
             self.delete_partition(p, True)
-        if ptable is not None:
-            disk.ptable = ptable
+        disk.ptable = ptable
         self.clear(disk, wipe)
 
     def can_resize_partition(self, partition, *, wipe=None):
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -270,10 +270,10 @@ class FilesystemManipulator:
 
     def reformat(self, disk, ptable=None, wipe=None):
         disk.grub_device = False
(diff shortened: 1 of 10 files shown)


subiquity/common/filesystem/manipulator.py
```diff
@@ -270,10 +270,10 @@ def clear(self, obj, wipe=None):
 
     def reformat(self, disk, ptable=None, wipe=None):
```
ogayot, line 271: Note that using the default value for `ptable` will still have a behavior on disks that have partitions vs on disks that have no partition.

For disks with partitions, `disk.ptable` will be reset to `None`
For disks without partition, `disk.ptable` will be left untouched.

This feels inconsistent.

ogayot:
Updated to change the meaning of `reformat(disk, ptable=None)`/

Previously this meant "keep the current type of partition table" but it was only honored when the disk had zero partition. The new meaning is "use the default partition type", which is what we have always, accidentally, done for disks with partitions.

The reason for introducing this potentially breaking change is that formatting a disk that had partitions is a more common use-case than formatting a disk that has zero partition. At this point, we sort of expect `reformat(disk, ptable=None)`  to create a default/GPT partition table so "fixing" the original issue will be more disruptive than changing the meaning of `reformat(ptable=None)`.

I introduced the ability in autoinstall to specify the type of partition table when doing reformat, so that if anyone wanted to keep the current partition table format, they can still do it (provided they know what the previous type was).

subiquity/common/types/storage.py
```diff
@@ -236,6 +236,8 @@ class GuidedResizeValues:
 @attr.s(auto_attribs=True)
 class GuidedStorageTargetReformat:
     disk_id: str
+    # ptable=None means to use the default (GPT in most scenarios)
+    ptable: Optional[str] = None
```
dbungert, line 240: You're absolutely correct that this exposes the capability to autoinstall users of layout.  The following gave me a msdos ptable as expected:

```
autoinstall:
  storage:
    layout:
      ptable: msdos
      name: direct
```
While I'll admit to mixed feelings about that, this will be interesting to some users so we should document that.  Would you add a note about that in the autoinstall references?

dbungert (approved):

Chris-Peterson444 (approved):

subiquity/common/types/storage.py
ogayot: updated

PR #2157, opened 2025-02-05
Don't crash and allow reformat of a disk with unsupported partition table

ogayot:
This PR allows Subiquity to work with partition tables that are marked "unsupported". In the presence of such disk, Subiquity used to crash, either in the UI, or when generating guided scenarios.

### Machine config

**TODO**

Locally, I have been using probe-data from LP:#2065516 but it requires curtin patches to mark the ptable from the dynamic disk as "unsupported". Otherwise, it still crashes. See https://code.launchpad.net/~ogayot/curtin/+git/curtin/+merge/480694

### API change

Disks that have an unsupported partition table will now be marked `"requires_reformat": true` in the /storage/v2 endpoint:

```json
"disks": [
    {
      "id": "disk-sda",
      "label": "TOSHIBA_HDWD110_999E6AZFS",
      "type": "local disk",
      "size": 1000204886016,
      "usage_labels": [
        "unsupported partition table"
      ],
      "partitions": [
[...]
      ]
      "ptable": "unsupported",
      "requires_reformat": true
```

Technically, looking at `"requires_reformat": true` is the same as looking at `"ptable": "unsupported"` but I assume that,  in the future we might not use the value `"unsupported"` for all the different unsupported ptables.
    
Clients are expected to prevent modifications of a disk which is marked `"requires_reformat": true`. Modifications include:
    
 * creating a new partition
 * removing a partition
 * modifying a partition
 * marking the disk as the boot device
    
Essentially, the only action that the client should support for such disk is to reformat it.

### TUI change

Screenshot from 2025-02-05 14-40-21
Screenshot from 2025-02-05 14-40-34

### Testing

TODO an actual install that includes reformat

### DIFF
--- a/subiquity/common/filesystem/gaps.py
+++ b/subiquity/common/filesystem/gaps.py
@@ -197,6 +197,8 @@ def find_disk_gaps_v2(device, info=None):
 def parts_and_gaps_disk(device, ignore_disk_fs=False):
     if device._fs is not None and not ignore_disk_fs:
         return []
+    if device.ptable == "unsupported":
+        return device.partitions()
     if device._m.storage_version == 1:
         return find_disk_gaps_v1(device)
     else:
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -121,7 +121,7 @@ class FilesystemManipulator:
         self.create_filesystem(part, spec)
         return part
 
-    def delete_partition(self, part, override_preserve=False, allow_renumbering=True):
+    def delete_partition(self, part, *, override_preserve=False, allow_renumbering=True):
         if (
             not override_preserve
             and part.device.preserve
@@ -146,7 +146,7 @@ class FilesystemManipulator:
         for v in raid._subvolumes:
             self.delete_raid(v)
         for p in list(raid.partitions()):
-            self.delete_partition(p, True)
+            self.delete_partition(p, override_preserve=True)
         for d in set(raid.devices) | set(raid.spare_devices):
             d.wipe = "superblock"
         self.model.remove_raid(raid)
@@ -274,7 +274,7 @@ class FilesystemManipulator:
         partition table type."""
         disk.grub_device = False
         for p in list(disk.partitions()):
-            self.delete_partition(p, True)
+            self.delete_partition(p, override_preserve=True)
         disk.ptable = ptable
         self.clear(disk, wipe)
 
--- a/subiquity/common/filesystem/manipulator.py
+++ b/subiquity/common/filesystem/manipulator.py
@@ -121,7 +121,14 @@ class FilesystemManipulator:
         self.create_filesystem(part, spec)
         return part
 
-    def delete_partition(self, part, *, override_preserve=False, allow_renumbering=True):
+    def delete_partition(
+        self,
+        part,
+        *,
+        override_preserve=False,
+        allow_renumbering=True,
+        allow_moving=True,
+    ):
         if (
             not override_preserve
             and part.device.preserve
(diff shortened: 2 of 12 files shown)


ogayot:
@d-loose would you have a look at the `API change` section and tell me if that seems manageable from the UI standpoint?

subiquity/common/filesystem/gaps.py
```diff
@@ -197,6 +197,8 @@ def maybe_add_gap(start, end, in_extended):
 def parts_and_gaps_disk(device, ignore_disk_fs=False):
     if device._fs is not None and not ignore_disk_fs:
         return []
+    if device.ptable == "unsupported":
+        return device.partitions()
```
dbungert, line 201: this or empty list?  I guess this is safe enough if the other manipulations actions are guarding against doing so on an unsupported ptable.

subiquity/common/filesystem/manipulator.py
```diff
@@ -121,15 +121,24 @@ def create_partition(self, device, gap, spec: FileSystemSpec, **kw):
         self.create_filesystem(part, spec)
         return part
 
-    def delete_partition(self, part, override_preserve=False, allow_renumbering=True):
+    def delete_partition(
+        self,
+        part,
+        override_preserve=False,
```
dbungert, line 127: suggestion: maybe that one should be keyword-only as well.

subiquity/common/types/storage.py
```diff
@@ -114,6 +114,7 @@ class Disk:
     model: Optional[str] = None
     vendor: Optional[str] = None
     has_in_use_partition: bool = False
+    requires_reformat: Optional[bool] = None
```
dbungert, line 117: as I understand it, `resuires_reformat` is meant for this case where there is a disk and we could work with it, but only if we reformat it and start from scratch.  This differs from other cases where we might be able to work with it, such as more regular dos and gpt ptable scenarios where we can act more selectively like deleting a single partition.

I ask because this field also sounds like we might set this to false to indicate that we want a disk to be formatted, which is a bit different.  So I think this merits a comment at least, and maybe give a think on the name (not that I have a better suggestion).

subiquity/common/filesystem/manipulator.py
ogayot: +1

subiquity/common/filesystem/gaps.py
ogayot: Maybe it depends on context, parts_and_gaps has many call sites nowadays. But in general I would say `device.partitions()` is superior. It shows we acknowledge that there might be content on the disk potentially worth keeping. But if we run into problems with this approach, I'm inclined to try with an empty list.

subiquity/common/types/storage.py
ogayot: I see your point and I'm not particularly happy with "requires_reformat" either.

In a way, this field wouldn't be needed if we added that list of allowed operations that we've been talking about for a long time. e.g.:

```json
{"disks": [
    {"id": "disk-sda", "allowed-actions": ["reformat", "add-partition"], 
     "partitions": [
       {"id": "part-sda1", "allowed-actions": ["delete", "shrink", "expand", "overwrite"]},
       {"id": "part-sda2", "allowed-actions": ["delete", "shrink", "expand", "overwrite"]},
    ]},
    {"id": "disk-sdb", "allowed-actions": ["reformat"], 
     "partitions": [
       {"id": "part-sda1", "allowed-actions": []},
       {"id": "part-sda2", "allowed-actions": []},
    ]},
]}
```
Maybe we hold-off on adding the `requires_reformat` field for now? We can ask clients to inspect `ptable` and disable actions if it's `"unsupported"`.

dbungert: Yes, I think you're correct, that this is a step towards a list of allowed operations.  How about we move forward with the field with a comment similar to this discussion?

subiquity/common/filesystem/gaps.py
dbungert: Works for me, please proceed.

subiquity/common/types/storage.py
ogayot: Works me me, comment added :)

Chris-Peterson444 (approved):
Looks good to me, thanks!

PR #2158, opened 2025-02-06
examples: add machine configs for bitlocker dual-boot scenarios

d-loose:
I've been using these machine configs to test dual-boot scenarios with bitlocker in the desktop installer. They've been generated by setting up a Windows 11 VM and injecting the block probe data from `/var/log/installer/block/probe-data.json` into the existing `simple.json` machine config using `scripts/inject-probe-data.py` from this repo.
I'd prefer adding those to the examples here, rather than than adding them to the [ubuntu-desktop-provision](https://github.com/canonical/ubuntu-desktop-provision/) repo. They might be useful for subiquity tests as well.


ogayot (approved):
Looking at the "os" section, this looks interesting for erase-and-install. There's no "windows11" detected but the EFI partition is labeled as "windows boot manager"

We might want to filter out "windows boot manager" from the possible "erase-and-install" scenarios.

PR #2159, opened 2025-02-12
Pickup curtin fix and bring machine-config for MS Windows dynamic disks

ogayot:
I was able to install Ubuntu over a disk that was formatted as a dynamic disk.


dbungert (approved):
Double-check the wording on the snapcraft rev commit then it LGTM.

ogayot:
> Double-check the wording on the snapcraft rev commit then it LGTM.

Good catch, thanks!

PR #2163, opened 2025-02-20
ci: migrate actions runners off ubuntu 20.04

Chris-Peterson444:
During some testing today I got a notification that GitHub actions will be deprecating the 20.04 image on 2025-04-01[1] so let's migrate away. Unless we have any reason to use the latest supported version, let's start using `ubuntu-latest`. It looks like we were on 22.04 but moved back to 20.04 because of setup issues[2], which I believe I was able to replicate and fix here. 

The problem: The GitHub Actions runner images have Docker pre-installed and connectivity issues with LXD and Docker both installed on a system is a known issue[3]. Though I don't know why the previously used `ubuntu-20.04` images didn't have this issue even though they also have Docker installed. I'll also note that I did try the "simpler" solution in [3] but for whatever reason I couldn't get that work.

[1] https://github.com/actions/runner-images/issues/11101
[2] https://github.com/canonical/subiquity/pull/1852
[3] https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker


.github/workflows/ci.yaml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
dbungert, line 13: I want an explicit reference to `ubuntu-24.04` rather than the floating reference.  We may well find that `ubuntu-26.04` needs some migration and I'd rather control that timing.

dbungert (commented):
Thanks for investigating this.

.github/workflows/ci.yaml
Chris-Peterson444: Fair enough, done!

dbungert (approved):

scripts/test-in-lxd.sh
```diff
@@ -21,6 +21,20 @@ TESTER=subiquity-${IMAGE##*:}
...
+if [ "active" == "$(systemctl is-active docker)" ]
+then
+    iptables  -I DOCKER-USER -i lxdbr0 -j ACCEPT
+    iptables  -I DOCKER-USER -o lxdbr0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
+    # GitHub doesn't use IPV6 (yet?)
+    # ip6tables -I DOCKER-USER -i lxdbr0 -j ACCEPT
+    # ip6tables -I DOCKER-USER -o lxdbr0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
+fi
+
+
```
ogayot, line 37: Can we move this to a separate script that runs only in the CI please? `test-in-lxd.sh` can be used locally and this would:
1. force us to run the script as root if we have docker enabled
2. insert firewall rules that are not deleted when the script exits

ogayot (changes requested):
Glad to see we're moving to more recent runners, but the workaround for the firewall  thing is misplaced IMO

scripts/test-in-lxd.sh
ogayot: Yes, that's fine!

.github/workflows/ci.yaml
```diff
@@ -54,7 +58,7 @@ jobs:
   static-typing:
     # In this job, we compare the output of mypy before and after the PR.
     if: github.event_name == 'pull_request'
-    runs-on: ubuntu-22.04  # minimimum 22.04 for python3-typeshed
+    runs-on: ubuntu-24.04 # minimimum 22.04 for python3-typeshed
```
ogayot, line 61: Not your bug if you would like to drop the comment or s/minimimum/minimum/, feel free :laughing:

ogayot (approved):
Thanks, LGTM!

PR #2166, opened 2025-02-21
Snap components

dbungert:


subiquity/server/controllers/install.py
```diff
@@ -668,6 +668,24 @@ async def platform_postinstall(self):
...
+        kernel_components = info.available_kernel_components
+        for driver in self.app.controllers.Drivers.drivers:
+            m = re.match("nvidia-driver-([0-9]+)", driver)
+            if not m:
+                continue
+            v = m.group(1)
+            ko = f"nvidia-{v}-ko"
+            user = f"nvidia-{v}-user"
+            if ko in kernel_components and user in kernel_components:
+                return [ko, user]
```
Chris-Peterson444, line 686: thought: I think the driver list is generally ascending so that means we'll get the oldest compatible driver first here. I don't know how many kernel component versions we'll have so maybe it doesn't actually matter in practice.

dbungert: I suspect that only one component version is ever going to be offered, I'm checking on that.
I could do a reverse sort here and cause us to prefer the larger number over the smaller one, which sounds like good defensive programming anyhow.  Will implement.

dbungert:
FFE: LP:#2099950

subiquity/server/snapd/types.py
```diff
@@ -205,16 +205,67 @@ class StorageEncryption:
...
+@snapdtype
+class Model:
+    architecture: str
+    snaps: List[ModelSnap]
+
+    def snap_of_type(self, typ: ModelSnapType) -> Optional[ModelSnap]:
+        for snap in self.snaps:
+            if snap.type == typ:
+                return snap
+        return None
```
Chris-Peterson444, line 249: nit: this only gets the first of the type in the list. From the name and usage it seems expected that there's only one of this type, which is true for the kernel (could we ever have more?), but it wouldn't work for say `typ=ModelSnapType.APP` where we already have multiple.

I think instead we want something like:
```suggestion
    def snaps_of_type(self, typ: ModelSnapType) -> List[ModelSnap]:
        return [snap for snap in self.snaps if snap.type == typ]
```
and have the caller make the assertion about the number of them?

I wouldn't say it's blocking and think it's fine to fix this in a follow up though.

Chris-Peterson444 (approved):
One inline comment but everything else looks great to me, thanks!

subiquity/server/controllers/tests/test_install.py
```diff
@@ -379,3 +380,44 @@ async def test_adjust_rp(self):
...
+                ["nvidia-2-ko", "nvidia-2-user"],
+            ),
+            (
+                ["nvidia-1-ko", "nvidia-1-user", "nvidia-2-ko", "nvidia-2-user"],
+                ["nvidia-driver-2", "nvidia-driver-1"],
+                ["nvidia-2-ko", "nvidia-2-user"],
+            ),
+        )
+    )
+    def test_bad(self, comps, drivers, expected):
```
ogayot, line 420: s/test_bad/test_kernel_components/ ?

subiquity/server/controllers/tests/test_install.py
```diff
@@ -379,3 +380,44 @@ async def test_adjust_rp(self):
                 (d / ".disk/casper-uuid-generic").read_text().strip()
             )
             self.assertEqual(new_casper_uuid, casper_uuid_from_file)
+
+
+class TestInstallControllerDriverMatch(unittest.IsolatedAsyncioTestCase):
```
ogayot, line 385: Optionally, you can make it a `unittest.TestCase`.

subiquity/server/snapd/types.py
dbungert: https://github.com/canonical/subiquity/pull/2168/commits/42ca09c59ea3d5c0b3aa620f76d966ff6e8e0139

PR #2167, opened 2025-02-24
Zdevs autoinstall

ogayot:
This is to support doing basic actions (i.e., enable/disable) on zdevs with autoinstall. The work is a joint effort with @Chris-Peterson444.

```yaml
zdevs:
  - id: 0.0.1506
    state: enabled
  - id: 0.0.1507
    state: enabled
  - id: 0.0.1508
    state: disabled
  # or using an alternative YAML syntax
  - {id: 0.0.1509, state: enabled}
```

In the first implementation we did, we used separate sections for `disable` and `enable`, but the order is important for DASDs.


subiquity/server/controllers/zdev.py
```diff
@@ -607,20 +643,60 @@ def __init__(self, app):
...
+            await self.zdev_handling_task
+
+    def start(self) -> None:
+        if self.ai_data:
+            self.zdev_handling_task = schedule_task(self.handle_zdevs())
+
+    def make_autoinstall(self) -> ZdevAi:
+        # Small "optimization" to avoid producing a config that enables or
+        # disables a given device multiple times in a row.
+        return [x for x, _ in itertools.groupby(self.done_actions)]
```
Chris-Peterson444, line 661: praise: TIL about `itertools.groupby` (very nifty) and this is much better than our original plan of putting some of the logic in `handle_zdevs`. Thanks!

Chris-Peterson444 (approved):
This LGTM. Thanks for the extra tests too. 

I typically like the multi-commit merge but I think this one could be squashed into a single commit. What do you think?

autoinstall-schema.json
```diff
@@ -139,6 +139,23 @@
             ],
             "additionalProperties": false
         },
+        "zdevs": {
```
dbungert, line 142: Every commit that modifies the schema should be matched with an autoinstall-reference update, please.

subiquity/server/controllers/zdev.py
```diff
@@ -593,10 +596,43 @@
 id="0.0.c0fe" type="generic-ccw" on="no" exists="yes" pers="no" auto="no" failed="yes" names=""'''  # noqa: E501
 
 
+ZdevAiItem = TypedDict(
```
dbungert, line 599: For most things like this we been doing attrs, what is the motivation here for TypedDict instead?

subiquity/server/controllers/zdev.py
```diff
@@ -607,20 +643,60 @@ def __init__(self, app):
...
+        return [x for x, _ in itertools.groupby(self.done_actions)]
+
+    async def handle_zdevs(self) -> None:
+        for item in self.ai_data:
+            if item["state"] == "enabled":
+                await self.chzdev("enable", self.zdevinfos[item["id"]])
+            elif item["state"] == "disabled":
+                await self.chzdev("disable", self.zdevinfos[item["id"]])
+            else:
+                raise ValueError('state must be "enabled" or "disabled"')
```
dbungert, line 670:
```suggestion
                raise AutoinstallError('state must be "enabled" or "disabled"')
```

subiquity/server/controllers/zdev.py
```diff
@@ -607,20 +643,60 @@ def __init__(self, app):
...
+        self, action: Literal["enable", "disable"], zdev: ZdevInfo
+    ) -> None:
+        if action == "enable":
+            on = True
+            state = "enabled"
+        elif action == "disable":
+            on = False
+            state = "disabled"
+        else:
+            raise ValueError("action must be 'enable' or 'disable'")
```
dbungert, line 687:
```suggestion
            raise AutoinstallError("action must be 'enable' or 'disable'")
```

subiquity/server/controllers/zdev.py
```diff
@@ -593,10 +596,43 @@
 id="0.0.c0fe" type="generic-ccw" on="no" exists="yes" pers="no" auto="no" failed="yes" names=""'''  # noqa: E501
 
 
+ZdevAiItem = TypedDict(
+    "ZdevAiItem",
+    {
+        "id": str,
+        "state": Literal["enabled", "disabled"],
```
dbungert, line 603: suggest using enable/disable here to match the apidef and to indicate that we are taking this action, not asserting initial state matches expectations.

subiquity/server/controllers/zdev.py
```diff
@@ -607,20 +643,60 @@ def __init__(self, app):
...
+
         if self.opts.dry_run:
-            await asyncio.sleep(random.random() * 0.4)
-            on = action == "enable"
             self.zdevinfos[zdev.id].on = on
             self.zdevinfos[zdev.id].pers = on
-        else:
-            chzdev_cmd = ["chzdev", "--%s" % action, zdev.id]
-            await arun_command(chzdev_cmd)
+            await asyncio.sleep(random.random() * 0.4)
```
dbungert, line 694:
```suggestion
            await asyncio.sleep(random.random() * 0.4 / self.app.scale_factor)
```

dbungert (commented):
Thanks for working on this. I have a few suggestions to discuss.

subiquity/server/controllers/zdev.py
ogayot: Agreed, it makes the code cleaner in the `chzdev` function, which is more important :)

subiquity/server/controllers/zdev.py
ogayot: `chzdev` is also called by non-autoinstall code. Maybe we catch the ValueError and wrap it in an AutoinstallError in autoinstall code.

ogayot:
> I typically like the multi-commit merge but I think this one could be squashed into a single commit. What do you think?

Yeah, I'll rework the history a bit. I want to make the move `arun_command` to `CommandRunner` a separate change, for instance.

subiquity/server/controllers/zdev.py
dbungert: got it.  I still lean towards the attrs implementation to stay consistent with other places in Subiquity.  Can we try it in a new commit and see if it reads differently?

ogayot: Chris and I updated the code to use attrs as well (in the last commit)

doc/reference/autoinstall-reference.rst
```diff
@@ -1465,3 +1465,44 @@ The following example provides user-data which is approximately equivalent to wh
...
+
+state
+^^^^^
+
+**type:** string (enumeration)
+
+Controls whether the device should be enabled or disabled. Supported values are:
+
+* ``enabled``
+* ``disabled``
```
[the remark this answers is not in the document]
ogayot: I had a chat with @frank-heimes and it sounds like `ignore` / `cio_ignore` is better used as a kernel CL argument. I'm leaning forward `enabled: true/false` now.

dbungert (approved):
Thoughts on one of the tests, then please proceed.  Thanks!

subiquity/server/controllers/tests/test_zdev.py
```diff
@@ -0,0 +1,141 @@
...
+        ]
+
+        expected = [
+            {"id": "0.0.1507", "enabled": True},
+            {"id": "0.0.1508", "enabled": False},
+            {"id": "0.0.1509", "enabled": True},
+        ]
+        self.assertEqual(expected, self.ctrler.make_autoinstall())
+
+    def test_make_autoinstall_with_dupes(self):
```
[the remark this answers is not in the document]
dbungert: I don't think we need logic changes, I want to see another test with a disable and enable of a single device, and to confirm that this looks correct to you in terms of the generated ai_actions.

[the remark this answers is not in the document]
ogayot: Test added, thanks!

dbungert (approved):
Thanks!

Chris-Peterson444 (approved):

PR #2168, opened 2025-02-24
components: fix version matching

dbungert:
~~Prerequisite PR: https://github.com/canonical/subiquity/pull/2166~~

install: handle nvidia driver version != component version

Often (always?) the `ubuntu-drivers list --recommended` version is going
to be newer than what's actually available as a pc-kernel component.
Handle that version mismatch.

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -214,10 +214,10 @@ class VariationInfo:
     def available_kernel_components(self) -> List[str]:
         if not self.system.available_optional:
             return []
-        kernel = self.system.model.snap_of_type(snapdtypes.ModelSnapType.KERNEL)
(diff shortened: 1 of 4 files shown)


Chris-Peterson444 (commented):
I think this makes sense to me, but at this point do we even care what version `ubuntu-drivers` reports for the apt drivers then? Or should we just say if that list of drivers reported is non-empty then just use the latest component driver available?

dbungert:
> Or should we just say if that list of drivers reported is non-empty then just use the latest component driver available?

It can theoretically list non-nvidia things, so I think the check here is reasonable if mildly overengineered.

Chris-Peterson444 (approved):

Chris-Peterson444 (approved):
LGTM, thank you!

PR #2174, opened 2025-03-11
pin/pass: volume auth, plus capability related fixes

dbungert:
* Pass pin / passphrase info along, if supplied in GuidedChoiceV2.
* Fix pin / passphrase validation, which should consider both CORE_BOOT_{,PREFER_}ENCRYPTED.
* Fix use_tpm check to also consider both CORE_BOOT_{,PREFER_}ENCRYPTED.


ogayot (approved):

Chris-Peterson444 (approved):
very nice. LGTM, thanks!

PR #2175, opened 2025-03-12
kernel: fix explicitly_requested for default ISOs

ogayot:
When attempting an OEM kernel install, the ensure_no_kernel_conflict() check fails. explicitly_requested has been set for stock ISOs, but explicitly_requested was only intended for autoinstall cases.

(cherry picked from commit 39feac24e3041b0150a1f7d74b1ea58862b39d4c)

LP:#2100961


ogayot:
Tested on a Dell XPS 9320.

Chris-Peterson444 (approved):

PR #2176, opened 2025-03-13
filesystem: install all optional snaps

Chris-Peterson444:
The snapd API expects that by specifying a value for `optional_install` on a request to `/v2/systems/` you only want those items installed. Existing logic to support only installing particular components meant that no optional snaps were installed. Update the logic in `finish_install` to always install all of the optional snaps available for the model.

I found the behavior surrounding `optional_install` on the snapd API a little confusing so I added some extra comments surrounding the type definitions in subiquity to document the expected behavior.

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -35,7 +35,7 @@ validate () {
         autoinstall-reset-only)
             python3 scripts/validate-yaml.py --no-root-mount "${cfgs[@]}"
             ;;
-        answers-core-desktop|answers-uc24)
+        answers-core-desktop|autoinstall-ubuntu-core-installer|answers-uc24)
             ;;
         *)
             python3 scripts/validate-yaml.py "${cfgs[@]}"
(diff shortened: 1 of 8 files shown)


dbungert (approved):
This looks correct, thanks!

scripts/runtests.sh
```diff
@@ -234,6 +234,23 @@ validate
...
+    --dry-run \
+    --output-base "$tmpdir" \
+    --machine-config examples/machines/simple.json \
+    --autoinstall examples/autoinstall/core-installer.yaml \
+    --bootloader uefi \
+    --snaps-from-examples \
+    --kernel-cmdline autoinstall \
+    --source-catalog examples/sources/core-installer.yaml
+validate
+grep -q "finish_install: optional_install=OptionalInstall(snaps=\['console-conf'\], components={'pc-kernel': \[\]}, all=False)" \
```
ogayot, line 251: suggestion: use `grep -F` and drop the escape characters.


>       -F, --fixed-strings
>              Interpret PATTERNS as fixed strings, not regular expressions.

subiquity/server/controllers/filesystem.py
```diff
@@ -1041,12 +1041,21 @@ async def finish_install(self, context, kernel_components):
...
+
             optional_install = snapdtypes.OptionalInstall(
-                components={kernels[0].name: kernel_components}
+                components={kernels[0].name: kernel_components},
+                snaps=optional_snaps,
             )
         else:
             log.error(f"unexpected number of kernel snaps {len(kernels)}")
-            optional_install = None
+            # multi-kernel model case unkown, let snapd try to install all
```
ogayot, line 1054: minor: s/unkown/unknown/

ogayot (approved):

PR #2177, opened 2025-03-14
Log snapd responses

dbungert:
If `subiquity-log-snapd` is found on the kernel command line, log snapd responses to /tmp.  Off by default.


Chris-Peterson444 (approved):
This has been super handy already, thanks!

ogayot (approved):

PR #2179, opened 2025-03-17
Revert `69bb830`: "network: do not accept route metric > 20000"

MggMuggins:
Fixes [LP#2079222](https://bugs.launchpad.net/subiquity/+bug/2079222).

This reverts commit 69bb8307eb03255ad60fa20ba5d8a148bd726e91.

This behavior was first introduced in #1569. `has_network` is currently used to determine if a mirror check should happen, to decide whether to install drivers, and to display a sensible default Ubuntu Pro setting in the UI. It is possible to use the full functionality provided by all those controllers without access to https://connectivity-check.ubuntu.com/ (via repo mirrors & the airgapped Ubuntu Pro contracts server).

This raises the broader question of whether `has_network` is a sufficient abstraction to capture the nuance of "airgapped environment with a repo mirror and local contracts server", "airgapped environment with no mirror or contracts server", "offline install", or any permutation of those. That conversation is likely out of scope for this bugfix though.

### DIFF
--- a/subiquitycore/controllers/network.py
+++ b/subiquitycore/controllers/network.py
@@ -84,17 +84,7 @@ class SubiquityNetworkEventReceiver(NetworkEventReceiver):
         self.controller.update_has_default_route(self.has_default_route)
 
     def _default_route_exists(self, routes):
-        for route in routes:
-            if int(route["table"]) != 254:
-                continue
-            if route["dst"]:
-                continue
-            if int(route["priority"]) >= 20000:
-                # network manager probes routes by creating one at 20000 +
-                # the real metric, but those aren't necessarily valid.
-                continue
-            return True
-        return False
+        return any(route["table"] == 254 and not route["dst"] for route in routes)
 
     def probe_default_routes(self):
         with pyroute2.NDB() as ndb:
--- a/subiquitycore/controllers/tests/test_network.py
+++ b/subiquitycore/controllers/tests/test_network.py
@@ -73,16 +73,6 @@ class TestRoutes(unittest.IsolatedAsyncioTestCase):
                 "priority": 100,
                 "gateway": None,
             },
-            {
-                "target": "localhost",
-                "tflags": 0,
-                "table": 254,
-                "ifname": "ens3",
-                "dst": "10.0.2.0",
-                "dst_len": 24,
-                "priority": 20100,
-                "gateway": None,
-            },
         ]
 
         self.assertTrue(self.er._default_route_exists(routes))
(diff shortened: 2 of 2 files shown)


MggMuggins:
Thanks for the clarity, I'll go back to the drawing board.

dbungert:
Thanks for your understanding @MggMuggins .  You're absolutely correct about the general has_network problem, but I don't want to make it worse by taking cases that are treated as offline today, and attempting to reach out to the archive when that won't work.

The workaround you posted to LP:#2096660 (`printf "[connectivity]\nenabled=false\n" > /etc/NetworkManager/conf.d/disable-connectivity.conf`) looks very reasonable.  

Specifically for apt, what's interesting is if we can reach the mirror, and there is a concept of mirror checking already.  Note that this is closer in scope to feature development than a bug fix, so if you plan to spend some extended time on the topic we should probably discuss direction.

MggMuggins:
That's totally reasonable, I was mostly hoping to gauge your appetite for a revert. :slightly_smiling_face: 

Strongly agree on discussing direction. The workaround I provided has been more or less accepted by the customer so this isn't urgent, but I totally forsee other customers stumbling into this; it would be ideal to see a fix. The sum of [`autoinstall.ssh.install-server`](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#install-server) and [configurable apt sources](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#apt) implies to me that both features work in a network-restricted environment. Would you prefer that I raise a product feedback request so that this can be handled by your team rather than SEG?

I'm wondering if there's any reasonable alternative to reaching out to the configured apt sources though. It seems to me that skipping the mirror check entirely based only on `has_network` as it currently exists is inherently flawed; there's no easy way to know a network's topology without probing for specific resources.

Looking at the uses of `has_network` and adopting a probing strategy:
- `has_network`, based only on the presence of a default route (not considering the NM conncheck); without a default route there's no reason to probe anything
- `has_reachable_mirror`, based on the result of the mirror check

Difficulty with probing is how to avoid scope creep; if the mirror selections are probed, then why shouldn't subiquity probe the snapstore or all the other documented [pro requirements](https://canonical-ubuntu-pro-client.readthedocs-hosted.com/en/latest/references/network_requirements/)?

Would be curious to hear your thoughts on this. Thanks!

dbungert:
> Would you prefer that I raise a product feedback request so that this can be handled by your team rather than SEG?

You're certainly welcome to do the feedback request, but that doesn't imply that we have the capacity to pick up the work near term, especially when workarounds are known and not too unreasonable.    Further capacity discussions can happen on our internal channels.

If SEG has the interest in pursuing this, I would be curious to see a solution that permits apt operations when we know we can reach the mirror building on Subiquity mirror checking capabilities and to separate that from the big `has_network` switch - even if `has_network` appears false but we can reach the mirror, then apt operations should be OK.

PR #2180, opened 2025-03-18
Add CL arg to disable the WLAN listener

ogayot:
Since the desktop installer does not rely on the WLAN listener, we expect the ubuntu-desktop-bootstrap snap to run subiquity with `--no-wlan-listener` (CC: @d-loose).

This should be a good way to avoid running into Python crashes, such as LP:#2071620 until we get them addressed.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -232,7 +232,7 @@ parts:
 
     source: https://github.com/canonical/probert.git
     source-type: git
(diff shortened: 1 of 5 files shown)


ogayot:
I tested this on real hardware (XPS 9320) and it didn't seem to affect the behavior at all.

dbungert (approved):
LGTM, please merge.

Chris-Peterson444 (approved):

PR #2181, opened 2025-03-18
API implementation for entropy check

dbungert:


dbungert:
Opened for review purposes, only intended for merge outside of the future `ubuntu/plucky` branch absent new information.

PR #2183, opened 2025-03-26
Fix crash for automated installs with zdevs

ogayot:
During real installs on s390x, passing a `zdevs` autoinstall section with content leads to the following error:

`ZdevController' object has no attribute 'zdevinfos'`

This happens because the `zdevinfos` variable is only defined when running in dry-run mode, but the tests did not catch that.

The variable is now renamed `dr_zdevinfos` (dr stands for "dry-run") to make it more obvious.

When the list of `zdevinfos` is needed, we now call `lszdev` to populate it. On some hardware, calling `lszdev` takes seconds so we do not invoke it in the initializer of the controller. Instead we delay its use until we know for sure we need it.

We also use an `async` version of `lszdev` so that we don't block the whole event loop when it is running.

LP:#2104267

### DIFF
--- a/subiquity/server/controllers/tests/test_zdev.py
+++ b/subiquity/server/controllers/tests/test_zdev.py
@@ -117,7 +117,7 @@ class TestZdevController(unittest.IsolatedAsyncioTestCase):
     async def test_chzdev_wrong_action(self):
         self.ctrler.done_ai_actions = []
         with self.assertRaises(ValueError):
-            await self.ctrler.chzdev("enAble", self.ctrler.zdevinfos["0.0.1507"])
+            await self.ctrler.chzdev("enAble", self.ctrler.dr_zdevinfos["0.0.1507"])
         self.assertFalse(self.ctrler.done_ai_actions)
 
     @patch("asyncio.sleep", AsyncMock())
@@ -126,7 +126,7 @@ class TestZdevController(unittest.IsolatedAsyncioTestCase):
 
         self.ctrler.app.command_runner = Mock()
         with patch.object(self.ctrler.app.command_runner, "run", AsyncMock()) as m_run:
-            await self.ctrler.chzdev("enable", self.ctrler.zdevinfos["0.0.1507"])
+            await self.ctrler.chzdev("enable", self.ctrler.dr_zdevinfos["0.0.1507"])
 
         m_run.assert_called_once_with(["chzdev", "--enable", "0.0.1507"])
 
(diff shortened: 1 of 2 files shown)


subiquity/server/controllers/zdev.py
```diff
@@ -703,12 +708,25 @@ async def chzdev_POST(self, action: str, zdev: ZdevInfo) -> List[ZdevInfo]:
...
+
+    def _parse_lszdev(self, output: str) -> list[ZdevInfo]:
+        devices = output.splitlines()
         devices.sort()
         return [ZdevInfo.from_row(row) for row in devices]
+
+    def lszdev_sync(self) -> list[ZdevInfo]:
+        """Synchronous version of lszdev - which we can drop once we move the
+        call to lszdev outside the initializer."""
+        return self._parse_lszdev(self._raw_lszdev_sync())
```
Chris-Peterson444, line 729: thought: We could potentially refactor how we do the assignment - something along the lines of scheduling another task in `start` - to avoid this sync vs async business? Especially if we only like having the sync version around to support the dry-run scenario. Which is probably what you're getting at with the comment.

Then again, that seems like even more mechanics just to support dry-run and this seems like a perfectly reasonable approach to me. Does this the sync copy have value outside of dry-run? We could just rename to make it clear this is a synchronous helper function for dry-run.

Chris-Peterson444 (approved):
Oh dang, somehow we both missed this! Thanks for the fix! 

Two small comments but otherwise LGTM. Draft because we're waiting for real-hardware tests?

subiquity/server/controllers/tests/test_zdev.py
```diff
@@ -70,28 +76,58 @@ async def test_handle_zdevs__none(self):
...
+                ),
+            ]
 
         self.assertEqual(expected_calls, m_chzdev.mock_calls)
 
+        if not dry_run:
+            m_raw_lszdev.assert_called_once()
+        else:
+            m_raw_lszdev.assert_not_called()
+
```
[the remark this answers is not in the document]
ogayot: Oops, good catch! Removing

ogayot:
> Oh dang, somehow we both missed this! Thanks for the fix!
> 
> Two small comments but otherwise LGTM. Draft because we're waiting for real-hardware tests?

Yes, but now Frank confirmed that it worked for him, so I'll mark it ready.

subiquity/server/controllers/zdev.py
Chris-Peterson444: Oh yes absolutely a follow-up PR thing, sorry for not making that clearer!

PR #2184, opened 2025-03-31
Docs: Upgrade the Docs Starter Pack to v1.0.0

rkratky:
A big update of the docs publishing setup:

* Starter Pack updated to v1.0
  - Switch to extension-based Starter Pack
  - Many unnecessary files removed
  - Includes an update mechanism
  - New Makefile
  - Some file reshuffle
* The 'vale' linter used instead of the 'woke' tool
* Sphinx config merged into one file
* Manpage link rendering changed to standard Sphinx way
* Minor improvements

~~Note: There's a known regression in the Starter Pack in that the "Give feedback" button currently doesn't work with Launchpad (the link in the footer for opening LP issues is there though). As soon as this is fixed upstream, I'll update the config. I believe this is acceptable in the short term.~~ (Fixed.)


dbungert:
Hey @rkratky - conceptually this is no problem, but the doc check seems a bit unhealthy.  The first one was cancelled for reasons that aren't clear to me, I re-ran the task, it ran for over an hour without completing.  Thoughts?

rkratky:
It was me who cancelled the check - for the same reason that it stalled. I'm looking into it.

rkratky:
> Hey @rkratky - conceptually this is no problem, but the doc check seems a bit unhealthy. The first one was cancelled for reasons that aren't clear to me, I re-ran the task, it ran for over an hour without completing. Thoughts?

@dbungert, I have now fixed the failing docs CI, so I believe it's good to be merged.

doc/.sphinx/metrics/source_metrics.sh
```diff
@@ -0,0 +1,66 @@
...
+        meanval=$readabilityWords
+    fi
+
+    readabilityAverage=$(echo "scale=2; 0.39 * ($readabilityWords / $readabilitySentences) + (11.8 * ($readabilitySyllables / $readabilityWords)) - 15.59" | bc)
+
+    # cast average to int for comparison
+    readabilityAverageInt=$(echo "$readabilityAverage / 1" | bc)
+
+    # value below 8 is considered readable
+    if [ "$readabilityAverageInt" -lt 8 ]; then
```
dbungert, line 52: This is not enforce by the workflows yet, I think?  If so, if we update the docs and trigger this, what sort of things (broadly speaking) should we do to correct?

doc/.sphinx/spellingcheck.yaml
```diff
@@ -1,28 +1,30 @@
...
-      - link
-      - title
-      - div.relatedlinks
-      - div.visually-hidden
-      - img
-      - a.p-navigation__link
+  - name: rST files
+    aspell:
+      lang: en
+      d: en_GB
```
dbungert, line 5: en_US now I think?

rkratky: Docs updated

dbungert (approved):
LGTM, thank you!

PR #2186, opened 2025-03-31
Provision issue 1001

dbungert:
Provide information about the effective mount / format / encrypted state for complex devices constructed by guided installs.

Field names may not be final.

### DIFF
--- a/subiquity/common/filesystem/labels.py
+++ b/subiquity/common/filesystem/labels.py
@@ -20,6 +20,7 @@ from subiquity.common.types import storage as types
 from subiquity.models.filesystem import (
     ZFS,
     Disk,
+    DM_Crypt,
     LVM_LogicalVolume,
     LVM_VolGroup,
     Partition,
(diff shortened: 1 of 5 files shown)


subiquity/tests/api/test_api.py
```diff
@@ -2379,3 +2396,204 @@ async def test_post_source_update_server_variant(self):
...
+                {
+                    "target": reformat,
+                    "capability": capability,
+                    "password": password,
+                },
+            )
+            resp = await inst.get("/storage/v2")
+            [d] = resp["disks"]
+            for expected, actual in zip(partitions, d["partitions"]):
+                self.assertDictSubset(expected, actual, f"partnum {actual['number']}")
```
Chris-Peterson444, line 2599: suggestion: We could use `itertools.zip_longest` to guarantee we don't truncate one of the lists. I think the scenario in which we do that is unlikely but I generally favor zip_longest when comparing objects. 

```suggestion
            for expected, actual in itertools.zip_longest(partitions, d["partitions"], fillvalue={}):
                self.assertDictSubset(expected, actual, f"partnum {actual['number']}")
```

Chris-Peterson444 (approved):
One suggestion but otherwise LGTM. Thanks!

subiquity/tests/api/test_api.py
dbungert: ah, nice.  That's better, thanks.  I'm going with fillvalue=None as I want to see that case fail - we shouldn't randomly be getting a different number of partitions here and I want that to make itself evident in the failed test.

Chris-Peterson444: Oh yeah even better. I can see the empty dict allowing some annoying cases!

ogayot (approved):

subiquity/common/filesystem/labels.py
```diff
@@ -338,8 +339,37 @@ def _for_client_disk(disk, *, min_size=0):
     )
 
 
+def _resolve_constructed_object(partition: Partition) -> (str, str, bool):
```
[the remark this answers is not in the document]
dbungert: oops

d-loose (approved):
Thank you - that works for the UI!

PR #2188, opened 2025-04-02
models: we don't need curtin's grub replace_linux_default=False

dbungert:
Curtin knows to carry forth kernel command line after ---.  We set replace_linux_default=False in subiquity with the goal of having `quiet splash` be in the default grub config, but `--- quiet splash` is already specified in the ISO grub config.

This matters because on arm64 we need more parameters, so we want those to flow through to the default grub config on the target system.

LP:#2104310

### DIFF
--- a/subiquity/models/subiquity.py
+++ b/subiquity/models/subiquity.py
@@ -486,9 +486,6 @@ class SubiquityModel:
             },
         }
 
-        if self.source.current.variant == "desktop":
-            config["grub"]["replace_linux_default"] = False
-
         if os.path.exists("/run/casper-md5check.json"):
             with open("/run/casper-md5check.json") as fp:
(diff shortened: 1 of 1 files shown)


Chris-Peterson444 (approved):
This looks correct, thanks!

ogayot (approved):
LGTM.

When you say "the ISO", are you talking about the plucky desktop ISO specifically?

dbungert:
> LGTM.
> 
> When you say "the ISO", are you talking about the plucky desktop ISO specifically?

Correct.  Basically, with this change, we are expecting that grub command line is built correctly (for desktop, containing `--- quiet splash`).  Also kvm-test should be updated to match, as since it often supplies a kernel command line it can obscure this behavior.

PR #2190, opened 2025-04-02
kvm-test: support carryover parameters

dbungert:
When supplying kernel cmdline to the VM, look at grub.cfg and, if the triple dash is found, take the parameters from that and add that to our kernel command line.

This is a more accurate form of testing, because under the current implementation, the target system grub defaults have a different value depending on if kvm-test was used with or without a kernel command line.

(implied by https://github.com/canonical/subiquity/pull/2188)


Chris-Peterson444 (approved):
Thanks for this!

scripts/kvm-test.py
```diff
@@ -582,6 +582,19 @@ def create_disk(path: Path, size: str):
...
 
 
+def get_grub_appends(ctx, mntdir: str) -> list[str]:
+    with (Path(mntdir) / "boot/grub/grub.cfg").open() as fp:
+        for line in fp:
+            # something like
+            # linux /casper/vmlinuz --- quiet splash
+            expected_sep = " --- "
+            before, actual_sep, after = line.partition(expected_sep)
+            if actual_sep != expected_sep:
```
ogayot, line 592: I'd suggest looking for `linux` and `/casper/vmlinuz` tokens as well. If ISOs have multiple boot entries we want the one that matches `linux /casper/vmlinuz` since it is the kernel we're booting.

Maybe `re` will help?

```python
m = re.search(r"linux\s+/casper/vmlinuz\s+---\s+(.*)", line)
print(m.groups(1))
> quiet splash
```

scripts/kvm-test.py
```diff
@@ -582,6 +582,19 @@ def create_disk(path: Path, size: str):
     run(['qemu-img', 'create', '-f', 'qcow2', str(path), size])
 
 
+def get_grub_appends(ctx, mntdir: str) -> list[str]:
+    with (Path(mntdir) / "boot/grub/grub.cfg").open() as fp:
```
ogayot, line 586: nit: if you want to get rid of one indentation level:

```python
for line in Path(mntdir).read_text().splitlines():
```

scripts/kvm-test.py
dbungert: implemented, thanks!

ogayot (approved):
LGTM!

PR #2191, opened 2025-04-08
storage: fix display of potential devices for MD or LVM

ogayot:
When creating or editing an LVM or MD container, the list of potential devices is shown. Unfortunately, when one of the devices has an overly long label (> 50 chars), all of the devices became invisible, making it challenging to create a RAID / LVM properly.

Ensure, the list of devices is now visible no matter the length of labels.

The actual fix is the `can_shrink=True` property. But we can also clip the name of the device using `wrap="ellipsis"` if we want a cleaner (but less detailed) output.

LP:#2045688

**Current behavior when a device has a long label:**

Screenshot from 2025-04-08 12-38-58

**New behavior with the `wrap="ellipsis"`**

Screenshot from 2025-04-08 12-37-44

**Alternative new behavior without the `wrap="ellipsis"`**

Screenshot from 2025-04-08 12-37-06

### DIFF
--- a/subiquity/ui/views/filesystem/compound.py
+++ b/subiquity/ui/views/filesystem/compound.py
@@ -123,7 +123,7 @@ class MultiDeviceChooser(WidgetWrap, WantsToKnowFormField):
                 self.all_rows.append(
                     TableRow(
                         [
-                            Text("    " + labels.label(device), wrap="ellipsis"),
+                            Text("    " + labels.label(device)),
                             Text(humanize_size(device.size), align="right"),
                         ]
                     )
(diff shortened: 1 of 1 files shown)


ogayot:
[probe-data.json](https://github.com/user-attachments/files/19647812/probe-data.json)

^ This is the probe data file i used for testing.

Chris-Peterson444 (approved):
Very nice, thank you! 

In terms of with vs without `wrap="ellipsis"`, I would say I prefer the non-ellipsis version. The long name extending the view doesn't bother me and I am more concerned about truncating important information if the view gets too small.

dbungert (approved):
The ellipsis version looks quite nice.   Somebody out there probably has devices where the truncated version is insufficient to tell the difference.  I think the non-ellipsis form is safer, and if we want to improve it I think that should take the form of more ways to reach the info dialog.  I'm electing to merge the non-ellipsis version.

PR #2193, opened 2025-04-11
apt: remove /var/lib/apt/lists from apt runs

dbungert:
In testing against a country mirror, apt-get update may fail with:
  File has unexpected size (x != y). Mirror sync in progress?

On inspection, we found that the copy of /var/lib/apt/lists was in a bad state and that further apt-get update invocation would not correct it.

By removing this cached information, we have to download more from the mirror (not great), but the result should be more robust.

LP:#2105480

### DIFF
--- a/subiquity/server/apt.py
+++ b/subiquity/server/apt.py
@@ -263,6 +263,11 @@ class AptConfigurer:
             mode=0o644,
         )
 
+        # workaround LP: #2105480 and many many many like it
+        apt_lists = self.install_tree.pp("var/lib/apt/lists")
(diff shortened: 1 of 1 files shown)


Chris-Peterson444 (approved):
I tested this extensively with Dan and it LGTM. This will probably (hopefully) solve a lot of our mirror syncing bugs!

dbungert:
Utkarsh Gupta also approved this merge on behalf of Ubuntu Release, so merging here and plucky and rebuilding the 25.04 RC snap.

PR #2194, opened 2025-04-15
storage: prevent using the install media in the guided TUI

ogayot:
Ever since we introduced support for booting from a recovery partition, storage v2 has been advertising "partial" guided reformat scenarios (without making a clear distinction from a normal reformat scenario).

However, only the /storage/v2 API supports partial reformats.

For the TUI, which uses storage v1, a partial reformat is not possible, so we should not offer the installation media as a potential target to reformat, even though the storage v2 API advertises it.

**Before**
Screenshot from 2025-04-15 10-24-08

**After**
Screenshot from 2025-04-15 10-25-18

LP:#2106499

### DIFF
--- a/subiquity/ui/views/filesystem/guided.py
+++ b/subiquity/ui/views/filesystem/guided.py
@@ -99,7 +99,9 @@ def summarize_device(disk):
     label = disk.label
     rows = [
         (
-            disk,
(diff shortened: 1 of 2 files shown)


Chris-Peterson444 (approved):
Nice. LGTM, thank you!

dbungert (approved):

PR #2197, opened 2025-04-21
network: write netplan config directly

dbungert:
As an extension of
https://code.launchpad.net/~dbungert/livecd-rootfs/+git/livecd-rootfs/+merge/484436 write netplan cfg directly instead of having cloud-init in the middle.


holmanb (commented):
This matches my expectations, but I have not tested it. 

+1 from me

ogayot (approved):
This makes the code cleaner as well, thanks!

Chris-Peterson444 (approved):

subiquity/models/network.py
```diff
@@ -38,42 +37,24 @@ def render_config(self):
...
-                        "content": self.stringify_config(netplan),
-                        "permissions": "0600",
-                    },
+        r = {
+            "write_files": {
+                # Disable cloud-init networking
+                "no_cloudinit_net": {
+                    "path": (
+                        "etc/cloud/cloud.cfg.d/"
+                        "subiquity-disable-cloudinit-networking.cfg"
```
blackboxsw, line 46: Can we name this prefix `00-subiquity-disable-cloudinit-networking.cfg` so that this config can be overridden easily if desired inf lexicographic order? Typically clouds or custom envs end up providing ##-something.cfg and a leading integer prefix gives them more namespaces to order overrides without having to lay down a zzzz-my-override.cfg

dbungert: Done, thanks!

PR #2199, opened 2025-05-06
Noble - Backport the ability to deal with unsupported partition tables (including Windows Dynamic disks)

ogayot:
This is a backport to noble of the "support" for unsupported partition tables.

The originals PRs are #2157, #2159 and #2155 (not directly related to unsupported ptables, see below)

I initially didn't want to to include the fixes for LP:#2097091 but the behavior of `POST /storage/v2/reformat_disk` with `ptable=None` (which is what the desktop installer does when clicking on "New partition table") is not suitable for unsupported partition tables (e.g. trying to reformat an unsupported partition table would attempt to create another unsupported partition table.

Also requires https://code.launchpad.net/~ogayot/curtin/+git/curtin/+merge/485463


ogayot:
https://github.com/canonical/subiquity/pull/2200 is needed to fix the CI

ogayot:
> Just for reference this is #2155 and #2157, right?

Correct, and #2159 as well - which I've just added to get the machine config. I've added references in the PR description, thanks!

> This looks good to me already. I think the slimy VM test is sufficient, but since this is still in Draft do you anticipate more testing?

> I'm happy with the testing I've done so far. If desktop implements the relevant change on their end, I'll do more testing though.

> Also, CI just looks unhappy because help.ubuntu.com is unhappy right now so we shouldn't block on that.

Yep, and it passed on retry so I wouldn't worry too much about it.

snapcraft.yaml
```diff
@@ -63,10 +63,11 @@ parts:
   curtin:
     plugin: nil
 
-    source: https://git.launchpad.net/curtin
+    # TODO update
```
ogayot, line 66: Let's fix this before merging

dbungert (approved):

PR #2201, opened 2025-05-07
slimy: build and update probert as well

ogayot:


Chris-Peterson444 (approved):

scripts/slimy-update-snap.sh
```diff
@@ -71,8 +72,14 @@ rm -rf "${subiquity_dest}/subiquity"
 rm -rf "${subiquity_dest}/subiquitycore"
 
 (cd "${src}" && ./scripts/update-part.py curtin)
+(cd "${src}" && ./scripts/update-part.py probert)
+
+# Build probert for the C extensions
+probert_tmpdir=$(mktemp -d)
+(cd "${src}"/probert && python3 setup.py build --build-lib "${probert_tmpdir}")
```
dbungert, line 79: This differs a bit from what we do in the `Makefile`, would you explain?

scripts/slimy-update-snap.sh
```diff
@@ -71,8 +72,14 @@ rm -rf "${subiquity_dest}/subiquity"
 rm -rf "${subiquity_dest}/subiquitycore"
 
 (cd "${src}" && ./scripts/update-part.py curtin)
+(cd "${src}" && ./scripts/update-part.py probert)
+
+# Build probert for the C extensions
+probert_tmpdir=$(mktemp -d)
```
dbungert, line 78: I would like to see this as a subdir of `tmpdir` which is covered by the handler registered with `trap`.

ogayot: I guess this comment doesn't apply anymore since I'm building the C extensions in place now.

dbungert (approved):
thanks for doing this.  I run so many quick-tests, it's helpful to have coverage for probert.

PR #2214, opened 2025-06-03
slimy: fix for where default python != py3.12

dbungert:
slimy ISOs built from questing fail on _rtnetlink due to using python 3.13 instead of 3.12.  fix that.


Chris-Peterson444 (approved):
Thanks!

PR #2215, opened 2025-06-04
Fetch core boot recovery key from snapd

ogayot:
Instead of generating a fake recovery key for core boot installations, we now lean on a new feature from snapd to fetch an actual recovery key.

The main problem I ran into is that the Subiquity defines the return type of `/v2/systems/{system-label}` as a `ChangeID`. But this is only correct if the endpoint returns an async response. Depending on the content of the request, the response is sync or async.

I have tried to add support for `Union[str, SystemActionResponse]` as a return type but with no success for now. In the meantime, I introduced a workaround to use an alternative snapd API definition.

~~Marking as a draft since this depends on canonical/snapd#15503~~

### DIFF
--- a/subiquity/server/snapd/types.py
+++ b/subiquity/server/snapd/types.py
@@ -273,6 +273,7 @@ class SystemAction(enum.Enum):
 
 class SystemActionStep(enum.Enum):
     SETUP_STORAGE_ENCRYPTION = "setup-storage-encryption"
+    GENERATE_RECOVERY_KEY = "generate-recovery-key"
     FINISH = "finish"
 
 
@@ -319,3 +320,8 @@ class SystemActionRequest:
 @snapdtype
 class SystemActionResponseSetupEncryption:
     encrypted_devices: Dict[NonExhaustive[Role], str] = attr.Factory(dict)
+
+
+@snapdtype
+class SystemActionResponseGenerateRecoveryKey:
+    recovery_key: str
--- a/subiquity/server/snapd/types.py
+++ b/subiquity/server/snapd/types.py
@@ -317,5 +317,5 @@ class SystemActionRequest:
 
 
 @snapdtype
-class SystemActionResponse:
+class SystemActionResponseSetupEncryption:
     encrypted_devices: Dict[NonExhaustive[Role], str] = attr.Factory(dict)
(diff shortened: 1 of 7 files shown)


subiquity/server/snapd/types.py
```diff
@@ -317,5 +318,10 @@ class SystemActionRequest:
...
 
 @snapdtype
-class SystemActionResponse:
+class SystemActionResponseSetupEncryption:
     encrypted_devices: Dict[NonExhaustive[Role], str] = attr.Factory(dict)
+
+
+@snapdtype
+class SystemActionResponseGenerateRecoveryKey:
+    recovery_key: str
```
ogayot, line 327: we might want to keep a single type and define `encrypted_devices` and `recovery_key` as `Optional` fields.

dbungert: That would work as well but I do like the clear contract here on what fields are relevant for this case.  Up to you.

dbungert (commented):
Looks reasonable so far.  I would prefer the serializer fix, of course, but I appreciate the technical challenges of fixing that layer.  I think we're due a conversation about a move to cattrs.

ogayot:
I tested this with a manually built snapd and was able to fetch the recovery key via /storage/v2/core_boot_recovery_key :heavy_check_mark:

ogayot:
@d-loose FYI this seems to be working (provided we have a recent enough snapd build)

dbungert (approved):

subiquity/server/snapd/types.py
ogayot: Keeping the types separate for now then. Thanks.

mwhudson:
> I have tried to add support for `Union[str, SystemActionResponse]` as a return type but with no success for now. In the meantime, I introduced a workaround to use an alternative snapd API definition.

Yeah that's not surprising because the way Union is implemented in my serialization thingy is to add an extra field describing the actual type, and so that's extremely unlikely to work with any other way of serializing things. I suppose it would be possible to do some magic to make unions work by looking at the field names of the union-ed types to be able to tell the options apart but eh. I also think an API fundamentally changing type/behaviour depending on what the arguments are is not great design!

PR #2216, opened 2025-06-05
Tics deps

dbungert:
* The request is to trigger on a cron basis, rather than more frequently
* Run the tests so we can show the coverage report
* Install the dependencies, both for tests and so we can reduce warnings of unrecognized stuff

### DIFF
--- a/.github/workflows/tics.yaml
+++ b/.github/workflows/tics.yaml
@@ -1,8 +1,9 @@
 name: TICS
+
 on:
-  push:
-    branches:
-      - main
+  workflow_dispatch:
+  schedule:
+    - cron: '17 5 * * 6'  # Run at 5:17a (arbitrary) on Saturday
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/tics.yaml
```diff
@@ -1,22 +1,32 @@
 name: TICS
+
 on:
-  push:
-    branches:
-      - main
+  workflow_dispatch:
+  schedule:
+    - cron: '17 5 * * 6'  # Run at 5:17a (arbitrary) on Saturday
```
Chris-Peterson444, line 6: praise: thanks for the comment on the cron time. I always have to look up the syntax and it's nice to know if the exact timing is important.

Makefile
```diff
@@ -64,9 +64,11 @@ flake8:
 
 .PHONY: unit
 unit: gitdeps
+	mkdir -p .coverage || true
```
Chris-Peterson444: question: What's the motivation behind `|| true` here? I'm trying to imagine scenarios where `mkdir -p` would fail that we'd like to ignore.

Chris-Peterson444 (approved):
One question that's kind of nitpicky, but LGTM. Thanks!

Makefile
dbungert: I saw a failure at one point while debugging and I can't reproduce it now so I'm removing the `|| true`, thanks for flagging it.

dbungert: oh, I see, I had a file there by that same name so the directory creation rightfully fails.  Yes, I prefer the form without the `|| true`

Chris-Peterson444: Ahh that makes sense. Sounds good to me, thanks!

PR #2217, opened 2025-06-06
identity: fix warnings when running unit tests

ogayot:
When running the unit tests, the following warnings are raised:

```
  subiquity/ui/views/tests/test_installprogress.py::IdentityViewTests::test_event_other_formatting
    subiquity/ui/views/installprogress.py:69: PileWarning: <MagicMock name='Text()' id='136247309836912'> is not a Widget
      self.event_pile = Pile(event_body)

  subiquity/ui/views/tests/test_installprogress.py::IdentityViewTests::test_event_other_robust_splitting
    subiquity/ui/views/installprogress.py:69: PileWarning: <MagicMock name='Text()' id='136247310463200'> is not a Widget
      self.event_pile = Pile(event_body)

  subiquity/ui/views/tests/test_installprogress.py::IdentityViewTests::test_event_other_robust_splitting
    subiquitycore/ui/container.py:368: ColumnsWarning: <MagicMock name='Text()' id='136247310463200'> is not a Widget
      super().__init__(*args, **kw)
```

We patch urwid.Text so we can use assert_has_calls() and similar functions. But as a result, the objects returned by urwid.Text(...) are not urwid widgets - which makes urwid complain.

Fixed by ensuring that the mock actually returns urwid Text widgets using the wraps construct.

### DIFF
--- a/subiquity/ui/views/tests/test_installprogress.py
+++ b/subiquity/ui/views/tests/test_installprogress.py
@@ -2,6 +2,8 @@ import unittest
 from unittest import mock
 from unittest.mock import patch
 
+import urwid
+
 from subiquity.client.controllers.progress import ProgressController
 from subiquity.common.types import ApplicationState
 from subiquity.ui.views.installprogress import ProgressView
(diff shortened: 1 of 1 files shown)


Chris-Peterson444 (approved):
oops, thank you!

PR #2220, opened 2025-06-12
Entropy calculation against newer snapd (2.71)

ogayot:
This PR is now based on #2221 and adapts the `/storage/v2/calculate_entropy` endpoint to the current entropy rework in snapd (> 2.70) (see https://github.com/canonical/snapd/pull/15607)

 * The entropy values are expressed as integers, not floats
 * We provide access to the `optimal-entropy-bits` field
 * Successful entropy-checks return a non-null response with entropy information.
 * We provide access to the failure-reasons and a success boolean.

<details>
<summary>Original description (before split):</summary>
Up to 925caabec4fe61562353adbbc5fe5ebc2b02fd5c, this is basically a rework of https://github.com/canonical/subiquity/pull/2181 where the implementation is based on what snapd 2.68 offers.

In subsequent commits though, we lean on future snapd work or work that isn't yet part of "stable" snapd.

If we want to, I can split this PR into two parts:
* the implementation against snapd 2.68.
* the implementation against future snapd work.

#### Main differences from original branch:
* When consuming `/v2/systems/{label} action="check-{passphrase|pin}"`, we expect a sync response from snapd (we use a workaround with an alternative `@api` definition for it).
* When consuming `/v2/systems/{label} action="check-{passphrase|pin}"`, we accept non-success HTTP status codes. Snapd returns HTTP 400 if the entropy is too low.
   * The implementation is definitely not ideal but can be improved later.
   * It currently relies on a new optional boolean kwarg `raise_for_status` that can be passed to `*.POST()` and `*.GET()` methods. When true, this boolean disables the call to `requests.Response.raise_for_status`, therefore parsing the response as if it was a success. The main drawback is that it disable exceptions for all HTTP statuses, not just 400.
 * It supplies the passphrase (or the pin) to snapd.
 * We use a new implementation of `check-passphrase` / `check-pin` (in the last few commits):
    * The entropy values are expressed as integers, not floats
    * We provide access to the `optimal-entropy-bits` field
    * Successful entropy-checks return a non-null response with entropy information.

#### TODO
[ ] Rework how we handle the 400 response. The `raise_for_status=True` implementation feels wrong.
[ ] Do some more testing
[ ] Maybe write tests that use a flask-based implementation of a mini-snapd that returns fake values.
</details>

### DIFF
--- a/subiquity/common/types/storage.py
+++ b/subiquity/common/types/storage.py
@@ -412,5 +412,5 @@ class CalculateEntropyRequest:
 
 @attr.s(auto_attribs=True)
 class EntropyResponse:
-    entropy: float
-    minimum_required: float
+    entropy_bits: int
+    min_entropy_bits: int
--- a/subiquity/common/types/storage.py
+++ b/subiquity/common/types/storage.py
@@ -412,6 +412,8 @@ class CalculateEntropyRequest:
 
 @attr.s(auto_attribs=True)
 class EntropyResponse:
+    success: bool
+
     entropy_bits: int
     min_entropy_bits: int
     optimal_entropy_bits: int
--- a/subiquity/common/types/storage.py
+++ b/subiquity/common/types/storage.py
@@ -417,3 +417,6 @@ class EntropyResponse:
     entropy_bits: int
     min_entropy_bits: int
     optimal_entropy_bits: int
+
+    # Set to None if success is True
+    failure_reasons: Optional[List[str]] = None
--- a/subiquity/common/types/storage.py
+++ b/subiquity/common/types/storage.py
@@ -414,3 +414,4 @@ class CalculateEntropyRequest:
 class EntropyResponse:
     entropy_bits: int
     min_entropy_bits: int
+    optimal_entropy_bits: int
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -390,15 +390,12 @@ class API:
                     partition will be deleted as well."""
 
             class calculate_entropy:
-                def POST(
-                    data: Payload[CalculateEntropyRequest],
-                ) -> Optional[EntropyResponse]:
+                def POST(data: Payload[CalculateEntropyRequest]) -> EntropyResponse:
                     """Calculate the entropy associated with the supplied
                     passphrase or pin.  Clients must use this endpoint to
                     confirm that the pin or passphrase is suitable prior to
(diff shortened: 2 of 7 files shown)


subiquity/common/types/storage.py
```diff
@@ -406,5 +406,6 @@ class ReformatDisk:
 
 @attr.s(auto_attribs=True)
 class EntropyResponse:
-    entropy: float
-    minimum_required: float
+    entropy_bits: int
+    min_entropy_bits: int
+    optimal_entropy_bits: int
```
ogayot, line 419: @d-loose I figured we haven't discussed in details what the API should return. 

Would you like to see:
* a boolean that tells if the check is a success or not?
* the list of "reasons" for a check to be a failure? (currently, this is only "low-entropy" I think)
* the message from snapd (e.g., "passphrase did not pass quality checks")

Anything else?

d-loose: Thanks, yes I think it's a good idea to include a boolean as well as the list of "reasons", to be prepared for future quality check failures beyond `entropy_bits < min_entropy_bits`. I don't think we need the raw message from snapd.

dbungert:
> If we want to, I can split this PR into two parts: the implementation against snapd 2.68, the implementation against future snapd work.

That doesn't feel needed unless you want to land the first part now.

subiquity/common/types/storage.py
```diff
@@ -406,5 +406,11 @@ class ReformatDisk:
...
-    entropy: float
-    minimum_required: float
+    success: bool
+
+    entropy_bits: int
+    min_entropy_bits: int
+    optimal_entropy_bits: int
+
+    # Set to None if success is True
+    failure_reasons: Optional[list[str]] = None
```
d-loose: Nitpick, noticed while testing this in the installer:
```suggestion
    failure_reasons: Optional[List[str]] = None
```

ogayot: Thanks!

ogayot:
> > If we want to, I can split this PR into two parts: the implementation against snapd 2.68, the implementation against future snapd work.
> 
> That doesn't feel needed unless you want to land the first part now.

Yep, I went forward and split the PR. I think merging the first part could help us detect potential bugs in the v2/calculate_entropy a bit earlier.

ogayot:
Merging to the tpmfde branch to unlock the current situation. Will open a separate PR to merge that back to main.

PR #2221, opened 2025-06-16
Entropy backend with snapd 2.68

ogayot:
This is a split of #2220 that only includes changes that are compatible with what snapd 2.68 offers.

#### Main differences from original branch from Dan:
* When consuming `/v2/systems/{label} action="check-{passphrase|pin}"`, we expect a sync response from snapd (we use a workaround with an alternative `@api` definition for it).
* When consuming `/v2/systems/{label} action="check-{passphrase|pin}"`, we accept non-success HTTP status codes. Snapd returns HTTP 400 if the entropy is too low.
   * The implementation is definitely not ideal but can be improved later.
   * It currently relies on a new optional boolean kwarg `raise_for_status` that can be passed to `*.POST()` and `*.GET()` methods. When true, this boolean disables the call to `requests.Response.raise_for_status`, therefore parsing the response as if it was a success. The main drawback is that it disable exceptions for all HTTP statuses, not just 400.
 * It supplies the passphrase (or the pin) to snapd
 * Pin/Passphrase is passed in the body of /v2/storage/calculate_entropy, not the query string

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -69,6 +69,7 @@ from subiquity.common.types import (
 )
 from subiquity.common.types.storage import (
     AddPartitionV2,
+    CalculateEntropyRequest,
     Disk,
     EntropyResponse,
     GuidedChoiceV2,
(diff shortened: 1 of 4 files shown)


d-loose:
Thanks a lot!
I can prepare a PR for the desktop installer based on those changes, so that we have a UI for testing.

dbungert (approved):

PR #2223, opened 2025-06-16
storage: generate the recovery later, so it doesn't expire

ogayot:
When calling `/v2/systems/{label}` with `action="generate-recovery-key"`, a recovery key is created with an expire duration of 5 minutes (defined in snapd).

Snapd expects that the `/v2/systems/{label}` with `action="finish-install"` gets called before the key expires.

In case the installer doesn't, we end up with:

```
    File "subiquity/server/snapd/api.py", line 142, in post_and_wait
      raise aiohttp.ClientError(result.err)
  aiohttp.client_exceptions.ClientError: cannot perform the following tasks:
  - Finish setup of run system for "enhanced-secureboot-desktop" (recovery key has expired)
```

We now move the "generate-recovery-key" step just before the "finish-install" step. This should drastically lower the window for key expiration.

The drawback is that subiquity's `/v2/storage/core_boot_recovery_key` endpoint can now only be used late during the installation, but UDB consumes it at the very end so that is ok.

LP:#2114738

### DIFF
--- a/subiquity/server/controllers/install.py
+++ b/subiquity/server/controllers/install.py
@@ -384,7 +384,6 @@ class InstallController(SubiquityController):
             )
             if fs_controller.use_tpm:
                 await fs_controller.setup_encryption(context=context)
-                await fs_controller.fetch_core_boot_recovery_key()
             await run_curtin_step(
                 name="formatting",
                 stages=["partitioning"],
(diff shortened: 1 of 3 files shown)


subiquity/server/controllers/install.py
```diff
@@ -732,6 +732,13 @@ async def postinstall(self, *, context):
 
         fs_controller = self.app.controllers.Filesystem
         if fs_controller.use_snapd_install_api():
+            if fs_controller.use_tpm:
+                # This will generate a recovery key which will initially expire
+                # after a very short duration (e.g., 5 minutes).
+                # We need to reach the finish_install step within that timeframe,
+                # otherwise the key will be considered expired. So let's keep
+                # the two calls next to one another.
+                await fs_controller.fetch_core_boot_recovery_key()
```
dbungert, line 740: This all looks fine but I was expecting to see the existing `fs_controller.fetch_core_boot_recovery_key()` call removed, I assume we want to remove the other one?

ogayot: Oops, of course ! Thanks for noticing. I'll redo the testing!

PR #2227, opened 2025-06-19
Entropy fix when response is {"type": "error"} (and fix entropy expected type)

ogayot:
The `/v2/systems/{label}` action="check-passphrase|check-pin" endpoint doesn't return `{"type": "sync"}` on error, it returns `{"type": "error"}` instead. This code path in Subiquity was untested and there were several issues with it:

* The `make_request` context manager yields a `_FakeError` instance, but then could yield `_FakeResponse` subsequently. This caused a:
```python
RuntimeError: "generator didn't stop"
```
* The caller or `_FakeError` didn't provide the data argument that the initializer expected.
* If `raise_for_status` is False, then `_FakeError`.json gets called, but the method didn't exist.

I also adapted the dry-run implementation so that it returns `{"type": "error"}`

Last, it turned out during testing that snapd 2.68 returns `min-entropy-bits` as an integer and `entropy-bits` as a floating point number. Adapt the code to match that. In the future, both values should be int though.

LP:#2114918

### DIFF
--- a/subiquity/server/snapd/types.py
+++ b/subiquity/server/snapd/types.py
@@ -358,4 +358,6 @@ class InsufficientEntropyDetails:
 class EntropyCheckResponse:
     kind: EntropyCheckResponseKind
     message: str
-    value: InsufficientEntropyDetails
+
+    # Set to None if kind="unsupported"
+    value: Optional[InsufficientEntropyDetails] = None
--- a/subiquity/server/snapd/types.py
+++ b/subiquity/server/snapd/types.py
@@ -351,7 +351,7 @@ class InsufficientEntropyReasons(enum.Enum):
 class InsufficientEntropyDetails:
     reasons: List[InsufficientEntropyReasons]
     entropy_bits: float
-    min_entropy_bits: float
+    min_entropy_bits: int
 
 
 @snapdtype
--- a/subiquity/common/api/client.py
+++ b/subiquity/common/api/client.py
@@ -51,7 +51,8 @@ def _wrap(make_request, path, meth, serializer, serialize_query_args):
             params=query_args,
             raise_for_status=raise_for_status,
         ) as resp:
-            resp.raise_for_status()
+            if raise_for_status:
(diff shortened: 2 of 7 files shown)


subiquity/server/snapd/types.py
```diff
@@ -351,11 +351,13 @@ class InsufficientEntropyReasons(enum.Enum):
 class InsufficientEntropyDetails:
     reasons: List[InsufficientEntropyReasons]
     entropy_bits: float
-    min_entropy_bits: float
+    min_entropy_bits: int
```
Chris-Peterson444, line 354: question(non-blocking): Ideally the mixed type situation will resolve early enough that we'll only have to support the int/int case but if not, can we do:
```python
entropy_bits: int | float
min_entropy_bits: int
```

and have the serializer just do the right thing?

Definitely not something we have to deal with right now, but I was trying to anticipate what the future changes might look like and I haven't worked with the serializer code enough to know how much of a pain it's going to be, if at all.

Chris-Peterson444 (approved):
LGTM, thank you!

subiquity/server/snapd/types.py
ogayot: Yes I agree this would be an improvement. I've been willing to make similar changes to the (de)serializer to support more union types but have run into obstacles when I tried. Maybe it would be easier if we start only with the deserialization implementation (this is what we'd need here).

dbungert (approved):

subiquity/server/snapd/types.py
mwhudson: It feels like union support for scalars out to be reasonably simple? But I also feel like unions in your API are a bad smell so...

mwhudson (approved):
LGTM

subiquity/server/snapd/types.py
ogayot: I'll try to get this implemented in a separate PR!

PR #2228, opened 2025-06-19
storage: when using the @api workaround, inherit the value of log_responses

ogayot:
Currently, when building an alternative snapd API client (i.e., because the types annotations are wrong), the client is created with disabled logging of the snapd responses. But if the standard snapd API client has logging enabled, we should also enable it for the alternative clients.


Chris-Peterson444 (approved):
nice, thanks!

PR #2229, opened 2025-06-19
snapd: set Content-Type header appropriately when POSTing to snapd

ogayot:
When POSTing to Snapd, we send it JSON data. Previously, we did not specify a Content-Type header, which is fine with snapd. We now rely on requests to set the Content-Type header appropriately (i.e., to application/json).

When debugging, it makes it easier to work with captured data. It also makes it easier to implement a mock version of snapd with flask.


Chris-Peterson444 (approved):
Makes sense to me, thanks!

(also yay for the requests package making this easy)

PR #2231, opened 2025-06-24
tests(api): increase timeouts for slow tests

Chris-Peterson444:
Increase timeouts for consistently slow and flaky tests.

I can consistently get these tests to fail during the test cleanup phase by reducing the default timeout to 5 seconds.

For the two autoinstall related tests, these are flaky for the following reason:  POSTs to `/shutdown` don't really work when the server is both non-interactive and in an error state (the POST handler waits [here](https://github.com/canonical/subiquity/blob/073b19e7edb8f791657ea39db72d2de3353ab7ef/subiquity/server/controllers/shutdown.py#L62) forever). If the test runner is slow or the global timeout is sufficiently short, we end up waiting for the server to shutdown when the global timeout happens. The shutdown call in the test cleanup has a 5 second timeout (that doesn't fail the test of course), so as long as there is at least that much time left in the global timeout then the test will pass. The shutdown behavior is a bit surprising but not necessarily a huge bug IMO considering the niche case. Instead of messing with the shutdown behavior, the cleanest solution for now is to increase the test timeouts.

The probert related test here is also failing while waiting on a POST to `/shutdown` but is seemingly unrelated to the other two tests. Let's also increase the timeout here for now.


subiquity/tests/api/test_api.py
```diff
@@ -2070,7 +2070,7 @@ async def test_interactive(self):
                 )
                 self.assertTrue(expected.issubset(resp))
 
-    @timeout()
+    @timeout(multipler=2)
```
ogayot: s/multipler/multiplier/

Chris-Peterson444: Doh, wow. How did I even get that to pass? Fixed, thank you!

PR #2233, opened 2025-06-25
storage: use locking when handling multiple v2/calculate_entropy requests

ogayot:
Similarly to the systems-getter object, our implementation of `/storage/v2/calculate_entropy` relies on bind-mounting stuff to `/var/lib/snapd/seed/{systems,snaps}`, calling snapd, and then unmounting.

And just like the systems-getter object, the same limitations apply. One cannot expect things to go well if multiple query handlers for /storage/v2/calculate_entropy run at the same time without locking.


ogayot:
@d-loose this patch makes things much more reliable for the entropy check.

PR #2236, opened 2025-07-08
Serializer: add support for unions of basic types

ogayot:
This makes is possible to use `typing.Union[int, str]`, `typing.Union[bool, float, int, None]`, etc when serializing / deserializing data exchanged with APIs.

### DIFF
--- a/subiquity/common/tests/test_serialization.py
+++ b/subiquity/common/tests/test_serialization.py
@@ -207,6 +207,31 @@ class TestSerializer(CommonSerializerTests, unittest.TestCase):
         }
         self.assertSerialization(typing.Union[Data, Container], data, expected)
 
+    def test_serialization_union_none_attrs(self):
+        @attr.s(auto_attribs=True)
+        class A:
+            x: int
+
+        @attr.s(auto_attribs=True)
+        class B:
+            y: int
+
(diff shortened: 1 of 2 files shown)


subiquity/common/tests/test_serialization.py
```diff
@@ -207,6 +219,61 @@ def test_serialize_union(self):
...
+
+        self.assertSerialization(typing.Union[type(None), A, B], None, None)
+        self.assertSerialization(
+            typing.Union[type(None), A, B], A(10), {"$type": "A", "x": 10}
+        )
+        self.assertSerialization(
+            typing.Union[type(None), A, B], B(10), {"$type": "B", "y": 10}
+        )
+
+        self.assertSerialization(typing.Union[type(None), A, B], None, None)
```
Chris-Peterson444, line 269: This working is surprising to me. I need to revisit how this pans out. Does this work with non-attrs types?

dbungert (approved):

subiquity/common/tests/test_serialization.py
Chris-Peterson444: Oh nevermind, I see it now. Sorry!

Chris-Peterson444 (approved):

PR #2237, opened 2025-07-08
security: add SECURITY.md file

ogayot:


SECURITY.md
```diff
@@ -0,0 +1,17 @@
+# Security policy
+
+## Reporting a vulnerability
+
+If you discover a security vulnerability, follow the steps outlined below to report it:
+
+1. Do not publicly disclose the vulnerability before discussing it with us.
+2. Report a bug at https://bugs.launchpad.net/subiquity
+
+    **Important**: Remember to set the information type to *Private Security*. You will see a field with the text *This bug contains information that is:*
```
Chris-Peterson444: suggestions: I think if we quote the field text, instead of italicizing it, it would make the hint a little more clear. 

>   **Important**: Remember to set the information type to *Private Security*. You will see a field with the text "This bug contains information that is:"

Wdyt? 

I might also add some extra text to the end of the hint:

>   **Important**: Remember to set the information type to *Private Security*. You will see a field with the text "This bug contains information that is:" that will allow you to change this.

Chris-Peterson444 (approved):
That was fast, thanks!

SECURITY.md
dbungert: +1 to a wording tweak here.  Here's my take on it

Important: Remember to set the information type to *Private Security*.  This is set with the field below the Bug Description.  Click the edit icon under "This bug contains information that is:", and choose *Private Security*.

ogayot: Implemented, thanks for the suggestions!

PR #2243, opened 2025-07-30
storage: add endpoint to check if the TPM key can be encrypted

ogayot:
This introduces a new endpoint:
```
GET /storage/v2/core_boot_encryption_features
```

e.g.:

```sh
curl --unix-socket /run/subiquity/socket http://a/storage/v2/core_boot_encryption_features
["PASSPHRASE_AUTH"]
curl --unix-socket /run/subiquity/socket http://a/storage/v2/core_boot_encryption_features
["PASSPHRASE_AUTH", "PIN_AUTH"]
```

Which returns a list of encryption "features" supported by a TPM/FDE variation.
When booting a TPM/FDE system, different authentication methods are supported.

NOTE: This change was originally part of a larger change that included a rework of how we handle variations info.

### DIFF
--- a/subiquity/common/apidef.py
+++ b/subiquity/common/apidef.py
@@ -70,6 +70,7 @@ from subiquity.common.types import (
 from subiquity.common.types.storage import (
     AddPartitionV2,
     CalculateEntropyRequest,
+    CoreBootEncryptionFeatures,
     Disk,
     EntropyResponse,
     GuidedChoiceV2,
@@ -399,6 +400,9 @@ class API:
 
             class core_boot_recovery_key:
                 def GET() -> str: ...
+
+            class core_boot_encryption_features:
+                def GET() -> List[CoreBootEncryptionFeatures]: ...
 
     class codecs:
         def GET() -> CodecsData: ...
--- a/subiquity/common/types/storage.py
+++ b/subiquity/common/types/storage.py
@@ -420,3 +420,8 @@ class EntropyResponse:
 
     # Set to None if success is True
     failure_reasons: Optional[List[str]] = None
+
+
+class CoreBootEncryptionFeatures(enum.Enum):
+    PASSPHRASE_AUTH = "passphrase-auth"
+    PIN_AUTH = "pin-auth"
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -44,6 +44,7 @@ from subiquity.common.types.storage import (
     AddPartitionV2,
     Bootloader,
     CalculateEntropyRequest,
+    CoreBootEncryptionFeatures,
     Disk,
     EntropyResponse,
     GuidedCapability,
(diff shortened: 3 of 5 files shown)


ogayot:
I'm not sure if the terminology if "Core-Boot-TPM-Key-Encryption" is a sensible name (both for the endpoint and for the types). I'm open for suggestions.

ogayot:
@d-loose fyi

d-loose:
Great stuff, thanks!

subiquity/common/apidef.py
```diff
@@ -400,6 +401,9 @@ def POST(data: Payload[CalculateEntropyRequest]) -> EntropyResponse:
             class core_boot_recovery_key:
                 def GET() -> str: ...
 
+            class core_boot_tpm_key_encryption:
+                def GET() -> List[CoreBootTPMKeyEncryption]: ...
+
```
Chris-Peterson444: suggestion: what about something like `core_boot_tpm_support`? 

I think we can remove the word "encryption" since "tpm" already implies encryption (right?). 

I think we should also not put "key" in the name here since we're not really returning a key. I'm comparing this with the `core_boot_recovery_key` endpoint that really does return a key.

subiquity/common/types/storage.py
```diff
@@ -420,3 +420,8 @@ class EntropyResponse:
 
     # Set to None if success is True
     failure_reasons: Optional[List[str]] = None
+
+
+class CoreBootTPMKeyEncryption(enum.Enum):
```
Chris-Peterson444: suggestion: `CoreBootTPMMethod` or `CoreBootTPMEncryptionMethod` if the former is too vague.

If you were to take the above suggestion then we could talk about encryption "methods" that we "support". So the GET request would give us a list of supported methods:
```python
class core_boot_tpm_support:
                def GET() -> List[CoreBootTPMMethod]: ...
```

If you wanted to mirror the snapd type we could use "features" instead of "methods" too.

dbungert: +1 to mirroring the type names.  Not something I've specifically done yet, IIRC, but I think it's a good idea.  So something along the lines of `StorageEncryptionFeature`.

subiquity/common/apidef.py
dbungert: `core_boot_encryption_features` ?

subiquity/server/controllers/filesystem.py
```diff
@@ -1781,6 +1782,28 @@ async def v2_core_boot_recovery_key_GET(self) -> str:
...
+                    variation.system.storage_encryption.features
+                )
+            except AttributeError:
+                continue
+
+            if features is None:
+                # Snapd is too old
+                return []
+
+            return [CoreBootTPMKeyEncryption(feature.value) for feature in features]
```
dbungert: the return is inside the for loop, is that expected?
would this produce the correct result if multiple variations supported tpmfde?

I wonder if we should accumulate the features into a set, cast that to a list, and then return.

ogayot: >  would this produce the correct result if multiple variations supported tpmfde?

Unfortunately, I don't have a very satisfying answer.

I wish the endpoint would be more reliable if we have multiple variations that support tpm/fde with different "features". Currently it returns the "features" supported by the first variation. This sounds like an arbitrary choice.

OTOH, it seems consistent with Subiquity's behavior when the user later sends a POST request to /storage/v2/guided with Capability=CORE_BOOT_ENCRYPTED, which is to pick the first variation that supports it.

https://github.com/canonical/subiquity/blob/7ab87892f6e9543f28345920493db31889732f4a/subiquity/server/controllers/filesystem.py#L789-L792
https://github.com/canonical/subiquity/blob/7ab87892f6e9543f28345920493db31889732f4a/subiquity/server/controllers/filesystem.py#L806

So I would say that returning the "features" supported by the first tpm/fde variation is a little more useful to the client (at least for now) than giving them an accumulation that includes features that are in inaccessible variations.

Happy to hear your thoughts though.

subiquity/common/apidef.py
ogayot: I went for `core_boot_encryption_features`. Thanks!

subiquity/common/types/storage.py
ogayot: Ok, I went forward and renamed to `CoreBootEncrptionFeatures`. I think `StorageEncryptionFeature` is too vague.

Chris-Peterson444 (approved):

dbungert (approved):
Thanks!

PR #2248, opened 2025-08-13
.github: drop cla-check in favor of company-wide workflow

ogayot:
There is now a company-wide CLA check. We don't need two cla-checks in the workflows.


dbungert (approved):

ogayot:
We might need to disable this for main but to keep it for other (future) stable branches.

Chris-Peterson444 (approved):
> We might need to disable this for main but to keep it for other (future) stable branches.

I like this plan. LGTM, thanks!

PR #2251, opened 2025-09-02
snapcraft: build with new python-only probert-network

ogayot:
Now that canonical/probert#160 has landed, we can use the new probert-network implemented only in python.

FFe was approved https://bugs.launchpad.net/subiquity/+bug/2121429

### DIFF
--- a/apt-deps.txt
+++ b/apt-deps.txt
@@ -1,4 +1,3 @@
-build-essential
 cloud-init
 curl
 dctrl-tools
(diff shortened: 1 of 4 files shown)


apt-deps.txt
Chris-Peterson444, line 1: praise: thanks for cleaning up this file while you're here too!

Chris-Peterson444 (approved):
This is some great work, thanks Olivier!

dbungert (approved):
Nice work!

PR #2255, opened 2025-09-05
network: fix network loss if Wi-Fi IP configured before AP settings

ogayot:
network: fix network loss if Wi-Fi IP configured before AP settings

From a netplan standpoint, it is invalid to declare a Wi-Fi configuration without defining the access points settings.

And if a user of Subquity decides to configure the Wi-Fi IP settings before the "Wi-Fi" (i.e., AP) settings, the *intermediate* netplan configuration produced becomes invalid. As soon as the user validates the IP settings, the network is lost immediately (for all interfaces).

The logs show that "netplan apply" fails with:

```
Error in network definition: {iface}: No access points defined
```
Furthermore, if using console-conf instead of Subiquity, the exception is uncaught and terminates the console-conf process.

Fixed by not producing the netplan configuration when a Wi-Fi device is incompletely configured.

LP:#2121854

### DIFF
--- a/subiquitycore/models/network.py
+++ b/subiquitycore/models/network.py
@@ -17,7 +17,7 @@ import ipaddress
 import logging
 from gettext import pgettext
 from socket import AF_INET, AF_INET6
-from typing import Dict, List, Optional
+from typing import Any, Dict, List, Optional
 
 import attr
 import yaml
@@ -193,7 +193,7 @@ class NetworkDev:
         self._model = model
         self._name = name
         self.type = typ
-        self.config = {}
+        self.config: Optional[dict[str, Any]] = {}
 
         # import done here to break a chain where anybody importing
         # subiquity.common.types has to have probert
@@ -461,7 +461,8 @@ class NetworkModel(object):
     """ """
 
     def __init__(self, project):
-        self.devices_by_name = {}  # Maps interface names to NetworkDev
+        # Maps interface names to NetworkDev
+        self.devices_by_name: dict[str, NetworkDev] = {}
         self._has_network = False
         self.project = project
         self.force_offline = False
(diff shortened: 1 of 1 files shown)


subiquitycore/models/network.py
```diff
@@ -449,12 +449,20 @@ def remove_routes(self, version):
         else:
             self.config.pop("routes", None)
 
+    def has_incomplete_config(self) -> bool:
+        """Netplan will be upset if devices have incomplete configuration, such
+        as IP addressing but no SSID configured for Wi-Fi interfaces."""
+        if self.type == "wlan" and self.configured_ssid == (None, None):
```
dbungert, line 455: the configured_ssid interface feels weird, if we were building it in this PR I think it might have a different interface like returning the value for `unconfigured` if `type != wlan`.

Nothing actionable here - your `has_incompleate_config()` is correct for the interface, just making a side comment.

subiquitycore/models/network.py
```diff
@@ -580,8 +588,20 @@ def render_config(self):
...
+            if not dev.config and not dev.is_used:
+                continue
+
+            if dev.has_incomplete_config():
+                if not dev.is_used:
+                    continue
+
+                log.warning(
+                    "netdev %s is marked as used but has an incomplete"
+                    " configuration. Netplan will probably be upset.",
```
dbungert, line 607: Is that actually true given that it's been suppressed?

ogayot: Yes, the interface is *not* suppressed in that scenario (assuming such a scenario actually exists). The `configs[dev.name] = dev.config` instruction is executed since we didn't hit the "continue".

dbungert: oh, right.  Maybe we should suppress such a config?  What is the value of writing a partial config, especially when we know that netplan will "probably" be unhappy with it?

ogayot: If that's okay, I'll add a TODO. If we suppress this configuration, we also need to suppress the configuration of the device(s) that depend on it (i.e., bond, vlans, <other>). Otherwise, netplan will fail too.

Chris-Peterson444: +1 to capturing this in a comment for future excursions here

ogayot: TODO added!

dbungert: OK

PR #2300, opened 2026-01-27
vmtests: create

dbungert:
Add a vmtests directory with a framework based on pytest, virt-install, and autoinstall to automate test runs in a VM.

Sample invocation:
```
pytest-3 --iso=/srv/iso/resolute/resolute-live-server-amd64.iso
```

### DIFF
--- a/vmtests/tests/conftest.py
+++ b/vmtests/tests/conftest.py
@@ -1,0 +1,463 @@
+import contextlib
+import json
+import re
+import subprocess
+import time
+from datetime import datetime
+from pathlib import Path
+from subprocess import DEVNULL, PIPE, CompletedProcess
+from tempfile import TemporaryDirectory
+from typing import Optional, Sequence
+
+import pytest
+import yaml
+
+from subiquity.models.source import SourceModel
+
+from . import Firmware
+
+
+class VM:
+    def __init__(
+        self,
+        vmm,
+        firmware=Firmware.UEFI,
+        memory_GiB=4,
+        disk_sizes_GiB=[10],
+        cloud_config=None,
+    ):
+        self.vmm = vmm
+        self.firmware = firmware
+        self.memory_GiB = memory_GiB
+        self.disk_sizes_GiB = disk_sizes_GiB
+        self.cloud_config = cloud_config
+        self.parent_tempdir = vmm.tempdir
+        self.live_boot = True
+        self.ip = None
+
+    def __enter__(self):
+        self.exit_stack = es = contextlib.ExitStack()
+        dir = self.parent_tempdir
+        self.tempdir = Path(es.enter_context(TemporaryDirectory(dir=dir)))
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.exit_stack.close()
+
+    def ssh(self, cmd: Sequence[str], check=True) -> CompletedProcess:
+        """
+        SSH to system and run the command, either the installation environment
+        if before first reboot, or to the target system.
+        """
+        if self.ip is None:
+            raise RuntimeError("attempted to ssh prior to SUT having an IP")
+        cmd = ["ssh", *self._ssh_common(), f"ubuntu@{self.ip}", "--", *cmd]
+        return _runcmd(cmd, stdout=PIPE, stderr=DEVNULL, check=check)
+
+    def in_target(self, cmd: Sequence[str]) -> CompletedProcess:
+        """
+        Run the supplied command underneath `curtin in-target`.  Only
+        meaningful during the live install boot, use ssh() instead if testing
+        the first boot of the target system.
+        """
+        assert self.live_boot
+
+        in_target = [
+            "sudo",
+            *self._curtin_cmd(),
+            "in-target",
+            "-i",
+            "-t",
+            "/target",
+            "--",
+        ]
+        return self.ssh(in_target + cmd)
+
+    def lsblk(self) -> dict:
+        """
+        The dictionary representation of `lsblk -Jb`.
+        """
+        sp = self.ssh(["lsblk", "-Jb", "/dev/vda"])
+        return json.loads(sp.stdout.decode())
+
+    def reboot(self) -> None:
+        """
+        Reboot the VM.  If this is the live install boot, signal to the
+        installer that the install should finish cleanly.  Wait for poweroff,
+        start the VM again, and wait for SSH to be available.
+        """
+        self._poweroff()
+        _virsh(["start", self.domain], stdout=DEVNULL)
+        self._wait_for_poweron()
+        self._wait_for_ip()
+        self._wait_for_ssh()
+
+    def _merge_cloud_config(self, cloud_config, pubkey) -> Path:
+        """
+        merge test_cloud_config with the base config, write the result to a
+        temp file, add a few additional test requirements, and return that file
+        """
+        with open("vmtests/base-cloud-config.yaml", "r") as fp:
+            base_cc_data = yaml.safe_load(fp)
+
+        # default to the minimal source.  As the id names vary by ISO, we have
+        # to guess that out, but they should be the one that ends in -minimal.
+        # this default to minimal behavior is done before the merge of test
+        # cloud-config data, as it's intended to be part of the
+        # base-cloud-config.
+        source = SourceModel()
+        with _mounter(self.vmm.option_iso) as iso:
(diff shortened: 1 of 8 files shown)


vmtests/tests/conftest.py
```diff
@@ -0,0 +1,360 @@
...
+
+from . import Firmware
+
+
+# FIXME timeout
+
+def runcmd(cmd, **kwargs):
+    kwargs.setdefault("check", True)
+    print(f'running command "{cmd}"')
+    return subprocess.run(cmd, **kwargs)
```
ogayot: Non blocking, preference based alternative:
```suggestion
def runcmd(cmd, *, check=True, **kwargs):
    print(f'running command "{cmd}"')
    return subprocess.run(cmd, check=check, **kwargs)
```

vmtests/tests/conftest.py
```diff
@@ -0,0 +1,360 @@
...
+        cloud_config = merge_data(test_config, base_config)
+
+        # add test ssh key for install system
+        cloud_config["users"][1]["ssh_authorized_keys"] = pubkey
+
+        # add test ssh key for target system
+        cloud_config["autoinstall"]["ssh"]["authorized-keys"] = [pubkey]
+
+        result = self.tempdir / "cloud-config.yaml"
+        with open(result, "w") as fp:
```
ogayot: I like `result.write_text(...)` but it's up to you!

https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text

vmtests/tests/conftest.py
```diff
@@ -0,0 +1,360 @@
...
+    def _merge_cloud_config(self, test_config_str, pubkey) -> Path:
+        """
+        merge test_config_str with the base config, write the result to a
+        temp file, and return that file
+        """
+        with open("base-cloud-config.yaml", "r") as fp:
+            base_config = yaml.safe_load(fp)
+
+        test_config = yaml.safe_load(test_config_str)
+        cloud_config = merge_data(test_config, base_config)
```
ogayot: Seems to me that if we have late-commands in test_config, they will be added after the `while [ ! -f /run/finish ] ; do sleep 1 ; done` construct.

Should we manually push that late command after calling merge_data?

```python
def add_final_late_command(config):
    config.setdefault("autoinstall", {})
    config["autoinstall"].setdefault("late-commands", [])
    config["autoinstall"]["late-commands"].append("while [ ! -f /run/finish ] ; do sleep 1 ; done")
```

vmtests/tests/conftest.py
```diff
@@ -0,0 +1,360 @@
...
+            self._wait_for_socket()
+            self._wait_for_install_end()
+            yield
+
+            if self.live_boot:
+                # we haven't done first boot yet, so do that now as a crude
+                # check that first boot works.
+                self.reboot()
+
+        except:
```
ogayot: Would `except Exception:` work here?

vmtests/tests/conftest.py
```diff
@@ -0,0 +1,360 @@
...
+            fp.write(yaml.dump(cloud_config))
+
+        return result
+
+    def _assert_install_success(self, data: dict) -> None:
+        match data["state"]:
+            case "LATE_COMMANDS":
+                # the good scenario - we have done autoinstall almost to the
+                # end, and are only being held up by the late-command check
+                # waiting for the /run/finish file to exist
```
ogayot: If there are user-supplied late-commands, this won't necessarily be true. Maybe this is acceptable for now? I suggest adding a note though.

ogayot (approved):
Thanks for making vmtest a thing! I have some suggestions but nothing blocks merging in my opinion.

I would like to have a way to specify the libvirt instance to use (i.e., `qemu:///session` vs `qemu:///system`). 

It can be passed to `virsh` and `virt-install` using `virsh --connect qemu:///session` and `virt-install --connect qemu:///session`.

Maybe, define a `virsh` function?

```python
def virsh(args: list[str], **kwargs):
    prefix = ["virsh"]
    if libvirt_instance is not None:
        prefix.extend(["--connect", libvirt_instance])
    return runcmd(prefix + args, **kwargs)
```

dbungert:
Thanks @ogayot for the first round feedback.  Updates since that round:
* more type info on functions
* a few docstring updates
* rethink the keyword argument name on `VM()` for the test cloud config partial snippet, going with `cloud_config`
* fix handling `VM(cloud_config=None)`
* dynamically determine the minimal source ID, thus allowing this to run desktop-style ISOs
* some function names that should be clearer, for instance s/_assert_install_success/_assert_state_late_commands
* log directory also uses seconds in the directory name
* some FIXMEs closed, some added
* default all waits with a timeout to 300s
* handle tests that have late-commands

vmtests/tests/conftest.py
dbungert: Done

vmtests/tests/conftest.py
dbungert: Done

vmtests/tests/conftest.py
dbungert: I want to collect logs in ctrl-C cases.  We re-raise here - do you have further concerns?

vmtests/tests/conftest.py
dbungert: Comment clarified + I distinguish this now, if late-commands are part of the test then the test assertions start after (most) late-commands now.

dbungert:
I've added a few straightforward fixes.  I have more queued and will pursue after this PR.  Fixes added as distinct commits to separate from a simple rebase on top of latest main.
I've added a virsh refactor to test `session` vms, session vms don't work yet, suggest we investigate post merge.

dbungert:
Moving out of draft, ready for merge after a rebase, commits only split for review reasons.

vmtests/tests/conftest.py
```diff
@@ -0,0 +1,458 @@
...
+
+        except (Exception, KeyboardInterrupt):
+            now = datetime.now()
+            ts = f"{now:%Y-%m-%d-%H-%M-%S}"
+            label = f"subiquity-vmtest-{self.vmm.test_name}-{ts}"
+            dest = Path(f"/tmp/{label}")
+            dest.mkdir(exist_ok=False)
+            self._collect_logs(label, dest)
+            print(f"logs written to {dest}")
+            raise
```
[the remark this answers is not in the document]
dbungert: future pr

vmtests/tests/conftest.py
```diff
@@ -0,0 +1,458 @@
...
+        print("wait for end state.", end="", flush=True)
+        while True:
+            sp = self.ssh(
+                ["curl", "--unix-socket", "/run/subiquity/socket", "a/meta/status"],
+            )
+            data = json.loads(sp.stdout)
+            if data["state"] in ("LATE_COMMANDS", "DONE", "ERROR", "EXITED"):
+                break
+            print(".", end="", flush=True)
+            time.sleep(1)
```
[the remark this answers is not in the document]
dbungert: future pr

vmtests/tests/conftest.py
```diff
@@ -0,0 +1,458 @@
...
+            "--disk",
+            f"size={self.disk_sizes_GiB[0]},target.dev=vda",
+            "--location",
+            f"{str(self.vmm.option_iso)},kernel=/casper/vmlinuz,initrd=/casper/initrd",
+            "--cloud-init",
+            f"user-data={cc_file},meta-data={self.empty}",
+            "--extra-args",
+            "autoinstall",
+            "--noautoconsole",
+        ]
```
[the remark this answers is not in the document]
dbungert: oops.  future pr

vmtests/tests/conftest.py
```diff
@@ -0,0 +1,458 @@
...
+            with Path(iso / "casper/install-sources.yaml").open() as fp:
+                source.load_from_file(fp)
+        for entry in source.catalog.sources:
+            if entry.id.endswith("-minimal"):
+                minimal_id = entry.id
+                break
+        else:
+            ids = [entry.id for entry in source.catalog.sources]
+            raise RuntimeError(f"minimal catalog entry not found in {ids}")
+        merged_cc["autoinstall"].setdefault("source", {})["id"] = minimal_id
```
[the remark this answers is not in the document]
dbungert: fixed by moving before _merge_data() call.  Comments added clarifying intent.

ogayot (approved):
Thanks!

Maybe squashing the patches would clean things up.

Great work overall.

PR #2343, opened 2026-03-18
workflows: implement coverage reporting

dbungert:
Generated with assistance from claude-sonnet-4.6.
Assumes #2342

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,0 +1,40 @@
+name: Coverage
+
+# FIXME compare entire workflow to other workflows we use today
+
+on:
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
+
+jobs:
+  coverage:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y $(cat apt-deps.txt)
+
+      - name: Fetch git dependencies (curtin, probert)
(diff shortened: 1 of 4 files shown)


.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
+
+jobs:
+  coverage:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```
sergio-costas, line 14: Why not use?

```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Run unit tests with coverage
+        run: make coverage | tee pytest-coverage.txt
+        continue-on-error: true
+
+      - name: Save PR number for coverage comment workflow
+        run: echo "${{ github.event.pull_request.number }}" > pr-number.txt
+
+      - name: Upload coverage artifact
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
```
sergio-costas, line 32: Idem;
```suggestion
        uses: actions/upload-artifact@v4
```

.github/workflows/coverage.yaml
dbungert: tags can be recreated, and this facility is being used in supply chain attacks when someone's actions repo is taken over.  the hash is much safer.

this is minorly paranoid for things under actions/* but easier to just apply the concept universally than pick and choose which it make sense for.

PR #2348, opened 2026-03-27
refresh: don't restart server if /refresh/progress change ID is not ours

ogayot:
Currently, the GET handler for /refresh/progress can trigger a server restart if the change ID specified corresponds to a DONE change.
    
This is generally a bad practice. First, because the server should be able to restart without the client querying /refresh/progress.  But more importantly, because the GET method is considered a safe method, and clients are allowed to retry safe requests on failure.

Therefore, if a client sends a GET /refresh/progress and this causes the server to restart, the client can decide to retry the request when the server becomes available again.

And this can cause the server to restart in a loop, until the client gives up. This causes various issues like clients receiving `ServerDisconnectedError` exceptions.

Ideally, we would never trigger a server restart in the GET handler. But doing so properly requires major code changes (i.e, currently the server does not monitor the refresh progress, without the client's intervention).

For now, we'll make sure that we only trigger a server restart if the change ID requested actually corresponds to a refresh operation **that was initiated by the current server process**.

Otherwise, if the change ID does not belong to us, we'll simply return the snapd change with status DONE.

LP:#2146422

### DIFF
--- a/subiquity/server/controllers/refresh.py
+++ b/subiquity/server/controllers/refresh.py
@@ -64,6 +64,13 @@ class RefreshController(SubiquityController):
         self.app.hub.subscribe(
             InstallerChannels.SNAPD_NETWORK_CHANGE, self.snapd_network_changed
         )
+        # List of snapd change IDs for refresh operations that we initiated.
+        # We use a list because refresh operations can fail and be restarted.
+        # This was introduced to avoid accidentally restarting the server when
+        # a client queries progress using a change_id that that it not ours (or
+        # in practice, belongs to a server process that was running before the
+        # refresh.  See LP: #2146422.
+        self.initiated_changes: list[str] = []
 
     def load_autoinstall_data(self, data):
         if data is not None:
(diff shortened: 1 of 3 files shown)


subiquity/server/server.py
```diff
@@ -1078,6 +1078,8 @@ async def _proxy_set(self):
         self.hub.broadcast(InstallerChannels.SNAPD_NETWORK_CHANGE)
 
     def restart(self):
+        # FIXME this method can only work if we're running tle subiquity snap.
```
dbungert:
```suggestion
        # FIXME this method can only work if we're running the subiquity snap.
```

subiquity/server/controllers/tests/test_refresh.py
```diff
@@ -142,3 +144,45 @@ async def test_start_update_api_error_logged(self):
...
+            id=cid,
+            kind="refresh-snap",
+            summary='Refresh "Subiquity" snap',
+            status=change_status,
+            tasks=[],
+            ready=False,
+        )
+
+        if is_our_change_id:
+            self.rc.initiated_changes.append(cid)
```
dbungert: I suppose it'd be more correct for the test to have a different change ID in initiated_changes, to go along with a test with an empty list of changes, but I believe your underlying implementation handles both cases.

dbungert (approved):
Thanks for your work on this bug

subiquity/server/controllers/tests/test_refresh.py
ogayot: You're right,! I'll add an extra tests to cover that scenario, thanks!

subiquity/server/server.py
ogayot: Thanks!

dbungert (approved):

subiquity/server/controllers/refresh.py
```diff
@@ -219,15 +226,21 @@ async def start_update(self, context):
...
 
     async def get_progress(self, change_id: str) -> Change:
         change = await self.app.snapdapi.v2.changes[change_id].GET()
         if change.status == TaskStatus.DONE:
-            # Clearly if we got here we didn't get restarted by
-            # snapd/systemctl (dry-run mode)
-            self.app.restart()
+            # TODO instead of relying on self.initiated_changes, we should
+            # move the call to self.app.restart away from the GET handler.
+            # Maybe we shouls start a monitoring task when we initiate the
```
Chris-Peterson444:
```suggestion
            # Maybe we should start a monitoring task when we initiate the
```

Chris-Peterson444 (approved):
Thank you!

subiquity/server/controllers/refresh.py
ogayot: Thanks!

PR #2351, opened 2026-03-31
fix assign_omitted_offsets exiting early on multi-disk systems

altendky:
## Summary

`assign_omitted_offsets()` uses `return` instead of `continue` when a disk has
no logical partitions (line 1876 of `subiquity/models/filesystem.py`). Since the
function iterates over all disks in a `for disk in self._all(type="disk"):` loop,
this causes it to exit entirely after processing the first disk, leaving
partitions on subsequent disks with `offset=None`.

Later, `partitions_by_offset()` calls `sorted(..., key=lambda p: p.offset)`,
which crashes with:

```
TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'
```

Any storage v2 autoinstall config with two or more GPT disks triggers this
because GPT disks never have MBR logical partitions.

LP:#2146858

## Changes

- **`return` -> `continue`** so the loop processes all disks
- Replace bare `next(filter(...))` for extended partition lookup with a guarded
  `list(filter(...))` to avoid `StopIteration` if the data is inconsistent
- Add `TestAssignOmittedOffsets` test class (this function was previously
  untested)

## Related

- LP:#2093314 crashes at the same site (`partitions_by_offset`) but the root
  cause identified there is partitions with a literal offset of 0 on USB media.
  This is a separate code path where offsets are never assigned at all.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1873,11 +1873,18 @@ class FilesystemModel:
                 prev_end = part.offset + part.size
 
             if not logical_parts:
-                return
+                continue
 
-            extended_part = next(
+            extended_parts = list(
                 filter(lambda x: x.flag == "extended", disk.partitions())
             )
+            if not extended_parts:
+                log.warning(
+                    "Disk %s has logical partitions but no extended partition",
+                    disk.path,
+                )
+                continue
+            extended_part = extended_parts[0]
 
             prev_end = extended_part.offset
             for part in logical_parts:
--- a/subiquity/models/tests/test_filesystem.py
+++ b/subiquity/models/tests/test_filesystem.py
@@ -2269,3 +2269,46 @@ class TestDiskForMatch(SubiTestCase):
         ]
         self.assertEqual(vdb, m.disk_for_match([vda, vdb], match))
         self.assertEqual([vdb], m.disks_for_match([vda, vdb], match))
+
+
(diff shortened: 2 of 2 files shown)


ogayot (commented):
Thank you!

The patch looks right to me, and the new tests are very much appreciated.

The only request I have is to include the `LP: #2146858` reference in the commit message and run black using `make format`, please.

We'd also need you to sign the [Canonical CLA](https://canonical.com/legal/contributors) so we can merge your changes.

subiquity/models/filesystem.py
```diff
@@ -1873,11 +1873,19 @@ def ad(v):  # ad == "align down"
...
+                continue
 
-            extended_part = next(
+            extended_parts = list(
                 filter(lambda x: x.flag == "extended", disk.partitions())
             )
+            if not extended_parts:
+                log.warning(
+                    "Disk %s has logical partitions but no extended"
+                    " partition",
```
ogayot: To make the CI happy:
```diff
-                    "Disk %s has logical partitions but no extended"
-                    " partition",
+                    "Disk %s has logical partitions but no extended partition",
```

altendky:
@ogayot thanks for looking this over!  i believe i have complete the requested changes.

ogayot (approved):
LGTM! Thank again!

dbungert (approved):
The bug and PR are both solid.   Thank you Kyle, great contribution!

PR #2355, opened 2026-04-05
filesystem: add rd.auto=1 to grub for storage with md RAID

ahachete:
Ubuntu 26.04 switched from initramfs-tools to dracut. Dracut does not auto-assemble md arrays without the rd.auto=1 kernel parameter, causing boot failure when the root filesystem is on md RAID due to MD device UUID not being found (as it is not assembled).

Write a grub.d config snippet via curtin write_files so rd.auto=1 is included in GRUB_CMDLINE_LINUX whenever md RAID arrays are present.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -2155,6 +2155,12 @@ class FilesystemModel:
             config["swap"] = {"size": 0}
         if self.grub is not None:
             config["grub"] = self.grub
+        if self.all_raids():
+            config.setdefault("write_files", {})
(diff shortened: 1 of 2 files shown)


ahachete:
Sure, makes sense. Just for completeness: do you have a reference (link, PR, etc) of how this will be handled by `dracut`? Always try to auto-assemble, even without `rd.auto=1`? (I think that is a better user experience anyway, but I'm no expert here, just curious)

dbungert:
Closing, was fixed in dracut, so we don't need this here.

PR #2360, opened 2026-04-15
snapcraft: default user in group lxd

dbungert:
Add lxd to the list of groups that the default user should be in. Note that the local users-and-groups file is just a copy convenient for testing, the package user-setup is the source of truth.


dbungert:
vm testing done with `kvm-test --build`

Chris-Peterson444 (approved):
Thanks!

PR #2362, opened 2026-04-29
security: add scope and supported releases in SECURITY.md

ogayot:


SECURITY.md
```diff
@@ -15,3 +48,5 @@ If you discover a security vulnerability, follow the steps outlined below to rep
    - Suggested mitigation, if possible
 
 The [Ubuntu Security disclosure and embargo policy](https://ubuntu.com/security/disclosure-policy) contains more information about what you can expect when you contact us and what we expect from you.
+
+The Subiquity team will be notified of the issue and will work with you to determine whether the issue qualifies as a security issue. We will then handle figuring out a fix, getting a CVE assigned and coordinating the release of the fix.
```
Chris-Peterson444: I think we could tighten this up slightly. Being a little more pedantic about "vulnerability" vs "issue" in the first sentence and a some rephrasing in the second. Something like:
```suggestion
The Subiquity team will be notified of the vulnerability and will work with you to determine whether it qualifies as a security issue. From there, we will take charge of developing a fix, securing a CVE assignment if necessary, and coordinating the release.
```

Wdyt?

SECURITY.md
```diff
@@ -1,5 +1,38 @@
...
+The goal of Subiquity is to deploy a new Ubuntu system.
+
+In general, any vulnerability that allows a third-party, without authorized
+access to the machine, to gain partial or full control over what gets deployed
+will be considered a security issue.
+
+Subiquity runs in a live installer environment. We consider that the user of
+the live environment has root permissions. Therefore any vulnerability that
+only allows the live environment user to escalate privileges further will not
+be considered a security issue.
```
Chris-Peterson444: Slight style and grammar nitpicks: Two of these sentences could be combined. 
```suggestion
Subiquity runs in a live installer environment and we consider that the user of
the live environment has root permissions. Therefore, any vulnerability that
only allows the live environment user to escalate privileges further will not
be considered a security issue.
```

SECURITY.md
```diff
@@ -1,5 +1,38 @@
 # Security policy
 
+## Supported versions
+
+There are two main types of Subiquity releases.
+
+* Subiquity releases corresponding to an Ubuntu release, which are included in
+  corresponding installer ISOs, e.g.:
+    * 24.04, 24.04.1, 24.04.2, 25.10, 26.04, ...
+* Ad-hoc releases that are only offered as installer refreshes, e.g.,:
```
Chris-Peterson444: I don't have a great alternative, but "ad-hoc" sounds like we only do these releases reactively. While sometimes true (special bug fixes, security fixes, etc), we also have regularly scheduled releases in the same category.

I would suggest "interim" releases but that might get confusing for interim ubuntu releases. What do you think about calling these:
- ISO releases
- Refresh Releases

?

SECURITY.md
```diff
@@ -1,5 +1,38 @@
...
+    * 24.10.1, 24.04.4.1, ...
+
+We typically support all Subiquity versions that correspond to an Ubuntu
+release, until that Ubuntu release becomes EOL.
+
+For Ubuntu LTSes, we typically stop delivering bug fixes after the .5 ISO is
+out. But we can still cut out ad-hoc releases for high or critical security
+issues when relevant.
```
Chris-Peterson444, line 18: Grammar nitpick:


```suggestion
For Ubuntu LTSes, we typically stop delivering bug fixes after the .5 ISO is
out. However, we can still cut out ad-hoc releases for high or critical security
issues when relevant.
```

SECURITY.md
dbungert: I like where you're going with this.  I suggest calling them Patch releases.  They don't map perfectly to major.minor.patch with LTS point releases, but the concept should be clear otherwise.

SECURITY.md
```diff
@@ -1,5 +1,38 @@
...
+## What qualifies as a security issue
+
+The goal of Subiquity is to deploy a new Ubuntu system.
+
+In general, any vulnerability that allows a third-party, without authorized
+access to the machine, to gain partial or full control over what gets deployed
+will be considered a security issue.
+
+Subiquity runs in a live installer environment. We consider that the user of
+the live environment has root permissions. Therefore any vulnerability that
```
dbungert:
```suggestion
Subiquity runs in a live installer environment. Since `sudo` is configured in the live environment to allow root access without a password, any user process has a trivial path to full root privileges.  Therefore any vulnerability that
```

dbungert (approved):
Thanks for putting this together.  Generally LGTM, feel free to merge and get reviewed by the SSDLC team.

SECURITY.md
ogayot: Applied, thanks!

SECURITY.md
ogayot: Thanks!

SECURITY.md
ogayot: Thanks!

SECURITY.md
ogayot: I've applied your suggestion with slight tweaks, to preserve a subtle distinction between our expectations (i.e., installer user can elevate privileges) from how it's configured in practice in ISOs (i.e, sudo without password).

SECURITY.md
ogayot: Thank you!

PR #2363, opened 2026-04-30
Group partition actions together when rendering storage config

ogayot:
If curtin is given a partitioning configuration in the following order:

 1. create a partition p1 on disk d1
 2. create a filesystem on p1 and mount it
 3. create a partition p2 on disk d1

Then the underlying partitioning tool (e.g., `sfdisk(8)` , `fdasd(8)`) will be unable to re-read the partition table using `ioctl(BLKRRPART)`, after creating p2, because the modification to the disk has been made while the disk is busy (i.e., it had a mounted partition).

For `sfdisk`, it is okay because curtin passes the `--no-reread` option and immediately follow its invocation by a call to `partprobe(8)`.

However, there is no such option for `fdasd`. If the call to `ioctl(BLKRRPART)` fails, it is reported as an `fdasd` failure:
```
fdasd error: IOCTL error
Error while rereading partition table.
Please reboot!
```
While adding a `--no-reread` option to `fdasd` would be good, the simplest way to avoid the issue is to rearrange the actions in a slightly different order.

The new constraint ensures that after we create a first partition on a disk, we immediately create any other partition on the same disk.

LP:#2146873

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -26,8 +26,10 @@ import secrets
 import tempfile
 from abc import ABC, abstractmethod
 from typing import (
+    Any,
     Callable,
     Dict,
+    Iterator,
     List,
     Literal,
     Optional,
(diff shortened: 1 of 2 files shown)


subiquity/models/filesystem.py
```diff
@@ -2090,6 +2097,27 @@ def can_emit(obj):
...
+                if action["type"] == "partition":
+                    device_to_partitions[action["device"]].append(action)
+
+            for action in actions:
+                if action["type"] == "partition":
+                    if device_to_partitions[action["device"]][0] is action:
+                        yield from device_to_partitions[action["device"]]
+                else:
+                    yield action
+
```
Chris-Peterson444, line 2120: as much as we were wracking our brains over this yesterday, this came out easy to read. thanks!

subiquity/models/filesystem.py
```diff
@@ -2128,7 +2156,7 @@ def can_emit(obj):
                     act["volume"] = device["id"]
             r = devices + r
 
-        return r
+        return list(grouped_partition_actions(r))
```
Chris-Peterson444, line 2159: nit: future ordering modifications would have a smaller diff if we updated `r` in a separate line from the return.


```suggestion
        r =  list(grouped_partition_actions(r))
        
        return r
```

Chris-Peterson444 (approved):
One non-blocking comment, but otherwise LGTM. This came together nicely, thanks!

dbungert (approved):
I can accept the PR as a bugfix and a net improvment, but the function that this is a part of was already hard to follow and is now even more complex.

Without feeling required to implement right now, do you have thoughts on what we might do to make _render_actions() easier to understand?

ogayot:
> Without feeling required to implement right now, do you have thoughts on what we might do to make _render_actions() easier to understand?

I think the main factor that makes this function difficult to follow (and to update) is the side effects in `can_emit`.

The name makes it sound like it is just a function returning true/false. And the basic usage shows that the caller will push the work item back to the end of the queue if `can_emit` returns false:

https://github.com/canonical/subiquity/blob/da408a45b1f608ecd956ba266c5f0cde22f2fbbf/subiquity/models/filesystem.py#L2103-L2106

But actually, `can_emit` will by itself push extra work to the queue in some scenarios:

* If it finds implicit dependencies, it will push them at the end of the work queue, i.e., `next_work.append(dep)`
* It not all sibling partitions of disk are explicitly enumerated, it will push the missing ones at the end of the work queue via `ensure_partitions`

Getting rid of these side effects would improve the situation and make this function easier to work with.

Also, I think a `collections.deque` could replace the `work + next_work` constructs. This would be another improvement.

PR #2367, opened 2026-05-07
storage: don't specify alignment data when formatting a raw device

ogayot:
When formatting a disk on the TUI, we use the same form as we use for creating a partition but with removed fields (e.g., size). For partitions, a maximum size and alignement info are needed, so they are mandatory in PartitionForm's initializer.

When formatting a disk though, the alignment info is unused. We still have to pass a value because it is mandatory in PartitionForm but there is no need to give a sound value.

Previously, we would still pass the alignment info that corresponds to a GPT. This makes the code confusing.

Let's specify the alignment info as a zero value instead.


Chris-Peterson444 (approved):
I wonder if we should eventually create a new form for this, but the comment and explicit value is an improvement for sure. Thanks!

Agreed, that would be https://github.com/canonical/subiquity/pull/2369

PR #2373, opened 2026-05-28
tests: fix invalid use of assertRaises(..., msg=...)

ogayot:
I've been incorrectly using `assertRaises(..., msg=<something>)` thinking it would match the text of the exception. But msg has a different purpose. It is shown when the assertion fails.

Let's use assertRaisesRegex to actually check the content of the exception instead.


dbungert (approved):
Nice fix, thanks.

PR #2375, opened 2026-06-01
Fix typo in comment

ogayot:


dbungert (approved):
repeated words are so invisible apparently

thanks!

ogayot:
> repeated words are so invisible apparently

I swear only a bot can spot those :laughing:

PR #2377, opened 2026-06-01
refresh: store change IDs in a set

ogayot:
These two fixes were suggested by copilot as part of #2348  but we needed to act fast before the 26.04 release. This is essentially a follow on PR that addresses the suggestions.


ogayot:
Closing because we already have #2349

PR #2378, opened 2026-06-02
storage: make questionable classmethod a staticmethod instead

ogayot:
In commit 51cf1afc91f1720785e1966744003b2adba6d5b0, we introduced FilesystemModel.is_mounted_filesystem to replace the old FilesystemModel.get_fs_by_name method.

That said, we preserved the `@classmethod` decorator even though:
 * the first parameter is called `self`, not `cls`
 * we do not use `self` (or `cls`).

I think we originally wanted to drop the `@classmethod` decorator but forgot to do so. Either way, the method itself does not use any of the class' attributes. It essentially belongs to the class for name-spacing so let's make it a `@staticmethod` instead.

### DIFF
--- a/subiquity/models/filesystem.py
+++ b/subiquity/models/filesystem.py
@@ -1567,8 +1567,8 @@ class FilesystemModel:
         ),
     }
 
-    @classmethod
-    def is_mounted_filesystem(self, fstype):
+    @staticmethod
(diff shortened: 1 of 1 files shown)


dbungert (approved):
Yes, looking at the commit, this makes more sense.  The old form did use `cls`, but this one doesn't need it.

Thanks!

PR #2382, opened 2026-06-11
Introduce storage requirements system, including ext4 /boot requirement for SecureBoot

ogayot:
This PR moves the storage requirements / constraints to a file named `subiquity/common/storage/requirements.py`. The goal is to have all requirements there in the end.

Each requirement has:
* a severity - telling whether having this requirement violated is a warning or a blocker for installing.
* a method to check if the requirement should be applied (for instance, some requirements only apply to a given set of architectures)
* a method to check if the requirement is met / satisfied.
* a method to check if the requirement is violated.

For now, we added 3 existing requirements + one new one:

* [existing] a filesystem must be mounted at /
* [existing] a local /boot filesystem must exist for nvme-tcp if the firmware has no support for it
* [existing] a bootloader partition must be present (i.e., either an ESP or a grub spacer) - unless we're doing BIOS + MSDOS or other rare configs
* [new] the filesystem for /boot must be ext4, for platforms that use signed GRUB (i.e., SecureBoot)

Also part of this PR are:

* More complete autoinstall validation for storage configuration (i.e., previously the NVMe/TCP requirement was not checked). Now all requirements defined in requirements.Requirements are evaluated.
* `AutoinstallError` exceptions are now raised when requirements are not met as part of autoinstall.

LP:#2165727

### DIFF
--- a/subiquity/common/storage/requirements.py
+++ b/subiquity/common/storage/requirements.py
@@ -18,6 +18,8 @@ from typing import TYPE_CHECKING
 
 import attrs
 
+from subiquitycore.lsb_release import lsb_release
+
 if TYPE_CHECKING:
     # Avoid circular import: models/storage.py imports Requirements
     # from this module at the top level.
@@ -95,6 +97,14 @@ def _is_boot_ext4(model) -> bool:
     return mount.fstype == "ext4"
 
 
+def _needs_ext4_boot(model) -> bool:
+    """UEFI systems with signed GRUB require ext4 for /boot on 26.10+."""
+    if not model.is_root_mounted() or not model.uses_signed_grub():
+        return False
+    release = lsb_release(dry_run=model.dry_run)["release"]
+    return tuple(int(p) for p in release.split(".")) >= (26, 10)
+
+
 class Requirements:
     """Well-known install-readiness checks for storage setup.
 
@@ -124,7 +134,7 @@ class Requirements:
         guidance_message_kind=GuidanceMessageKind.USE_EXT4_BOOT,
         severity=RequirementSeverity.BLOCKING,
         check=_is_boot_ext4,
-        applies_to=lambda m: m.is_root_mounted() and m.uses_signed_grub(),
+        applies_to=_needs_ext4_boot,
     )
 
     @staticmethod
--- a/subiquity/common/storage/requirements.py
+++ b/subiquity/common/storage/requirements.py
@@ -1,0 +1,102 @@
+# Copyright 2026 Canonical, Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import enum
+from collections.abc import Callable
+from typing import TYPE_CHECKING
+
+import attrs
+
+if TYPE_CHECKING:
+    # Avoid circular import: models/storage.py imports Requirements
+    # from this module at the top level.
+    from subiquity.models.storage import StorageModel
+
+
+class RequirementSeverity(enum.Enum):
+    """How seriously a violated requirement should be treated."""
+
+    BLOCKING = "blocking"
+    WARNING = "warning"
+
+
+@attrs.define
+class StorageRequirement:
+    """A single install-readiness check with a user-facing guidance message.
+
+    Attributes
+    ----------
+    guidance_message:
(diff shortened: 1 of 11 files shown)


subiquity/common/filesystem/requirements.py
```diff
@@ -0,0 +1,120 @@
...
+    WARNING = "warning"
+
+
+@attrs.define
+class StorageRequirement:
+    """A single install-readiness check with a user-facing guidance message.
+
+    Attributes
+    ----------
+    guidance_message:
```
dbungert: following snapd error handling, I suppose we want an enum here also that is more stable for translation purposes for Desktop

subiquity/common/storage/requirements.py
```diff
@@ -0,0 +1,120 @@
...
+    # Avoid circular import: models/filesystem.py imports Requirements
+    # from this module at the top level.
+    from subiquity.models.filesystem import FilesystemModel
+
+
+class RequirementSeverity(enum.Enum):
+    """How seriously a violated requirement should be treated."""
+
+    BLOCKING = "blocking"
+    WARNING = "warning"
```
dbungert, line 33: do you have an example of what might be a warning?

subiquity/models/tests/test_filesystem.py
```diff
@@ -391,109 +391,96 @@ def test_supports_nvme_tcp_booting(
 
     @parameterized.expand(
         (
-            (True, True, True, True),
-            (True, True, False, True),
-            (True, False, False, True),
-            (False, True, True, False),
-            (False, True, False, True),
-            (False, False, False, False),
+            (False, False, False, False, False, False, False),
```
dbungert: this is getting rough to read - looking at the individual lines it's pretty tough to tell the intent of the individual cases.  Give this some thought please.  Also some of these look odd (no test case using bootfs_remote == True for instance means we could hard-code it to False, or do we want a test case with it True?)

subiquity/models/storage.py
```diff
@@ -2442,11 +2446,44 @@ def needs_bootloader_partition(self):
...
-    def _mount_for_path(self, path):
+    def uses_grub(self) -> bool:
+        """Return True when the system's bootloader is GRUB-based.
+
+        Only s390x systems (``Bootloader.NONE``) do not use GRUB; all
+        other boot methods (BIOS, UEFI, PREP) rely on GRUB.
+
+        Curtin also has a ``uses_grub(machine)`` function in
+        curtin/commands/curthooks.py, which determines GRUB usage from
+        the machine architecture instead.  Consider consolidating.
```
dbungert, line 2473: we will definitely want to consolidate, as the logic on these two differ.

subiquity/models/storage.py
```diff
@@ -2442,11 +2446,44 @@ def needs_bootloader_partition(self):
...
+    def uses_signed_grub(self) -> bool:
+        """Return True when the system uses signed GRUB (UEFI SecureBoot).
+
+        Signed GRUB is only required on UEFI systems.  BIOS and PREP use
+        unsigned GRUB.  In theory, not all UEFI platforms that use GRUB
+        have signed GRUB (riscv64 is an example).  In practice, we apply
+        the same requirement so that it's future-proof if those platforms
+        gain Secure Boot support.
+        """
+        return self.uses_grub() and self.bootloader == Bootloader.UEFI
```
dbungert, line 2492: do things get weird here if we use a system with secureboot turned off in the firmware?

dbungert (commented):
Thanks Olivier.  This looks close.  A few things to discuss.

subiquity/common/storage/requirements.py
ogayot: What I initially had in mind was NVMe/TCP where it's hard for the installer to really tell whether the firmware supports booting with no local disk at all. But since the current situation is to deny the configuration completely (i.e., the equivalent of `blocking` here), I kept it as is.

subiquity/models/storage.py
ogayot: AFAICT, signed GRUB is for systems that support SecureBoot, not for systems that have SecureBoot necessarily enabled.
If the user decides they don't want to use SecureBoot, it should be theoretically possible to get them an unsigned GRUB. But I don't think we want to support the use case (i.e. curtin does not support that today).

subiquity/common/filesystem/requirements.py
ogayot: Good point. While I don't have a full answer yet on how requirements fit in the storage API, I think introducing an enum now gets us one step closer.

subiquity/common/filesystem/requirements.py
```diff
@@ -0,0 +1,148 @@
...
+    return mount.fstype == "ext4"
+
+
+def _needs_ext4_boot(model) -> bool:
+    """UEFI systems with signed GRUB require ext4 for /boot on 26.10+."""
+    if not model.is_root_mounted() or not model.uses_signed_grub():
+        return False
+    release = lsb_release(dry_run=model.dry_run)["release"]
+    return tuple(int(p) for p in release.split(".")) >= (26, 10)
+
```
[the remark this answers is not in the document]
ogayot: it's fine.

dbungert (approved):
Thanks, I think this is ready.

PR #2383, opened 2026-06-19
Rename "filesystem" to "storage"

ogayot:
The "filesystem" model, controllers and view deal with everything related to storage. Not just filesystem.

While the API and some recent files have "storage" in the name, most of the codebase still used "filesystem". I've been willing to make this consistent for a long time.

~~To avoid invalidating existing answers files, I introduced an optional "legacy_name" attribute and set it to "Filesystem" in the newly renamed `StorageController`.~~

~~If an answers file has a `Storage` section, this is where we load answers from. Otherwise, if it has no `Storage` section but has a `Filesystem` section (i.e., legacy name), this is where we load answers from.~~


mwhudson (commented):
Seems OK on the face of it but I haven't looked closely enough to vote. Do you want to move `subiquity.models` to `subiquity.server.models` next? :)

ogayot:
> Do you want to move `subiquity.models` to `subiquity.server.models` next? :)

Let's do it :) Too bad the client uses the Storage / FilesystemModel

subiquitycore/tuicontroller.py
```diff
@@ -31,7 +31,12 @@ class TuiController(BaseController):
     def __init__(self, app):
         super().__init__(app)
         self.ui = app.ui
-        self.answers = app.answers.get(self.name, {})
+        if self.name in app.answers:
+            self.answers = app.answers[self.name]
+        elif hasattr(self, "legacy_name") and self.legacy_name in app.answers:
```
dbungert: So I think we can get away with just not supporting this.  Answers is not meant to be a supported automation path for subiquity, autoinstall is the intended answer there.  And realistically I'd like to replace answers with virt-install vmtests + something that properly pretends to provide TUI inputs instead of this system which requires special code paths in the TUI.

UTAH runs answers tests and it carries no answers files, relying on subiquity to provide them instead, so we can do the name change and just be done with it.

I acknowledge this isn't much of a code savings.

subiquity/client/controller.py
```diff
@@ -30,6 +30,5 @@ class SubiquityTuiController(TuiController):
 
     def __init__(self, app):
         super().__init__(app)
-        self.answers = app.answers.get(self.name, {})
```
dbungert, line 33: This is dropped, can you talk about this?  Is it unused?

subiquity/client/controllers/storage.py
dbungert:
```suggestion
        log.info("Resetting Storage model")
```

subiquity/client/controllers/storage.py
```diff
@@ -147,7 +148,7 @@ async def run_answers(self):
             self.ui.body.form.guided_choice.value = value
             self.ui.body.done(None)
             self.app.answers["filesystem-confirmed"] = True
```
dbungert: this one as well?  looks to need matching change in client/client.py

subiquity/client/client.py
dbungert: here

dbungert (commented):
Thanks for this, this name inconsistency is indeed worth resolving.  Local LLM review pointed out, correctly, that ./scripts/raid-size-tests.py is broken by this refactor - I haven't tried to run it on main so it might have been broken before, but at least address the renames there please.

subiquity/client/controller.py
ogayot: I've reverted the change since it was part of the backward compat with answers - but I originally dropped it because TuiController already does that.

ogayot:
> Thanks for this, this name inconsistency is indeed worth resolving. Local LLM review pointed out, correctly, that ./scripts/raid-size-tests.py is broken by this refactor - I haven't tried to run it on main so it might have been broken before, but at least address the renames there please.

Good catch! Updated thanks.

dbungert (approved):
Nice work!

PR #2386, opened 2026-06-19
storage: fix on-remote-storage not working with ZFS and zpools

ogayot:
`_mount_for_path` can sometimes return something that isn't a `Mount` object. When that happens, accessing `.device.volume` fails.

Typically, this happens with zpools and zfs datasets.

Instead of enumerating all the different scenarios, I'm adding a `on_remote_storage` method to all the potential fsobj.

### DIFF
--- a/subiquity/models/tests/test_filesystem.py
+++ b/subiquity/models/tests/test_filesystem.py
@@ -2076,6 +2076,106 @@ class TestOnRemoteStorage(SubiTestCase):
         with mock.patch.object(vg, "on_remote_storage", return_value=True):
             self.assertTrue(lv.on_remote_storage())
 
+    def test_filesystem(self):
+        m, d = make_model_and_disk()
+        p = make_partition(model=m, device=d)
+        fs = make_filesystem(model=m, partition=p)
+
+        with mock.patch.object(p, "on_remote_storage", return_value=False):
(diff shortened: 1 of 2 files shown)


subiquity/models/tests/test_filesystem.py
```diff
@@ -2076,6 +2076,106 @@ def test_lvm_logical_volume(self):
         with mock.patch.object(vg, "on_remote_storage", return_value=True):
             self.assertTrue(lv.on_remote_storage())
 
+    def test_filesystem(self):
+        m, d = make_model_and_disk(name="sda", serial="sata0")
+        p = make_partition(model=m, device=d)
+        fs = make_filesystem(model=m, partition=p, fstype="ext4")
```
dbungert: Many of these tests have kwargs to make_*() for things not needed by the tests. `name`, `serial`, and `fstype` can routinely be deleted in this diff with no consequence.

If you don't need them I suggest dropping them, or incorporating those in some meaningful way.

dbungert (approved):
Thanks, this is an improvement.

subiquity/models/tests/test_filesystem.py
ogayot: Good point, I dropped most of the arguments that were unnecessary and changed `model.add_zpool to make_zpool`. Thanks!

PR #2387, opened 2026-06-22
Guided ZFS with separate /boot partition when using signed grub + Ubuntu 26.10+

ogayot:
Signed GRUB on Ubuntu 26.10+ cannot read ZFS, so `/boot` cannot live on a ZFS bpool. When the requirement holds, guided ZFS now creates an ext4 filesystem for /boot on its dedicated partition instead of a bpool. Systems that don't need ext4 keep the existing bpool behaviour.

LP:#2165727


dbungert (commented):
Looks correct so far

ogayot:
Did some testing with ZFS guided on amd64:
* 26.04 -> bpool created
* 26.10 -> no bpool, /boot on separate partition

LGTM

dbungert (approved):
Thank you!

PR #2391, opened 2026-06-29
snapd: pass keyboard layout to snapd

ogayot:
Since snapd 2.76, setup_encryption expects the keyboard configuration. The purpose is to setup the keyboard correctly on first boot before asking for the passphrase.

In snapd before version 2.76, the keyboard configuration is unsupported and would be ignored.

### DIFF
--- a/subiquity/server/snapd/test/test_types.py
+++ b/subiquity/server/snapd/test/test_types.py
@@ -14,11 +14,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from unittest import TestCase
+from unittest.mock import Mock
 
 import attr
 import attrs
 
-from subiquity.server.snapd.types import snapdtype
+from subiquity.server.snapd.types import KeyboardConfig, snapdtype
 from subiquitycore.tests.parameterized import parameterized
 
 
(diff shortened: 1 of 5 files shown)


ogayot:
@d-loose fyi

subiquity/server/snapd/test/test_types.py
```diff
@@ -38,3 +39,60 @@ class MetadataMerge:
 
         [field] = attrs.fields(MetadataMerge)
         self.assertEqual(expected, field.metadata)
+
+
+class TestKeyboardConfig(TestCase):
```
dbungert, line 44: seems like a good candidate for a parameterized test, easily could be a place where we add more test cases to cover some keyboard quirk in the future.  Would that be net shorter than this version?

dbungert (approved):
Thanks - one question on tests.

subiquity/server/snapd/test/test_types.py
ogayot: Should be definitely shorter. Will implement. Thanks!

ogayot:
Documentation check is failing because of a dead link. I've addressed it as part of https://github.com/canonical/subiquity/pull/2395

Chris-Peterson444 (approved):
Nice, thanks!

dbungert (approved):
Thanks!

PR #2394, opened 2026-07-01
Add core boot encryption requirements

ogayot:
This introduces a new endpoint that mimics snapd's `encryption.requirements`. The only difference is that while snapd's `requirements` field will not be present if empty, Subiquity will return an empty list instead.

I've used the opportunity to rename `CoreBootEncryptionFeatures` to `CoreBootEncryptionFeature` since it's an enumeration, not a flag class, list or anything that calls for plural.

@d-loose fyi

### DIFF
--- a/subiquity/server/snapd/types.py
+++ b/subiquity/server/snapd/types.py
@@ -198,11 +198,9 @@ class EncryptionType(enum.Enum):
     DEVICE_SETUP_HOOK = "device-setup-hook"
 
 
-class EncryptionFeature(enum.Enum):
-    PASSPHRASE_AUTH = "passphrase-auth"
-    PIN_AUTH = "pin-auth"
-
-
+# Alias the snapd-side enums to the Subiquity API storage types so the two layers
+# cannot drift apart instead of duplicating the definitions here.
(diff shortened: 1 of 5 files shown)


subiquity/server/snapd/types.py
```diff
@@ -225,6 +225,8 @@ class StorageEncryption:
     # Introduced in snapd 2.68, but can be None if snapd does not want to offer
     # pin/passphrase.
     features: Optional[List[EncryptionFeature]] = None
+    # Introduced in snapd 2.76, but can be None if snapd has no requirement.
+    requirements: Optional[List[EncryptionRequirement]] = None
 
     # Since snapd 2.71 <-- to be confirmed once released.
```
dbungert, line 231: Is this confirmed?

dbungert (approved):
Thanks!

subiquity/server/snapd/types.py
ogayot: Yes. Or rather, it's confirmed that the value is not present in the response if snapd has no requirements (i.e. the field is marked `json:"omitempty"` on snapd side) - which we turn into `None` internally.

PR #2396, opened 2026-07-03
Add auto-proceed option for errors in TPM/FDE autoinstalls

ogayot:
A limitation of TPM/FDE unattended installations was the inability for the user to "interact" with pre-install check errors. Such errors block the installation even in scenarios where a trivial `Action.PROCEED` would normally fix the situation. Notably, this is an issue in VMs because the pre-install check returns a `"running-in-vm"` error. Therefore, we were unable to run automated TPM/FDE installations on VMs.

This patch brings the ability to automatically execute an `Action.PROCEED` for selective error kinds.

To do this, we introduce an `accepted-errors` autoinstall directive (available for the hybrid layout with encryption). The directive can be used to enumerate the different error kinds that should automatically be addressed with an `Action.PROCEED`, should snapd offer it as a solution.

One can specify `accepted-errors` as a list of error kinds, or as a single error kind.

The following configuration allows to perform a TPM/FDE unattended installation on a VM:
```yaml
storage:
  layout:
    name: hybrid
    encryption: yes
    accepted-errors: [running-in-vm]
    ## If only one is specified, no need for a list
    ## So the following is equivalent.
    # accepted-errors: running-in-vm
```

### DIFF
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -623,6 +623,33 @@ Additionally, TPM-backed encryption can be enabled by using the ``hybrid`` layou
           name: hybrid
           encrypted: yes
 
+TPM-backed encryption supports the additional ``auto-proceed`` property, specifying the list of pre-install errors that Subiquity should automatically attempt to proceed past.
+
+.. code-block:: yaml
+
+    autoinstall:
+      storage:
+        layout:
+          name: hybrid
+          encrypted: yes
+          auto-proceed: [running-in-vm]
+
+One can also specify ``auto-proceed`` as a single pre-install error (i.e., ``auto-proceed: running-in-vm``).
+
+The list of currently supported pre-install errors is:
+
+ * ``running-in-vm``
+ * ``insufficient-dma-protection``
+ * ``no-kernel-iommu``
+ * ``addon-drivers-present``
+ * ``sys-prep-applications-present``
+ * ``absolute-present``
+ * ``weak-secure-boot-algorithms-detected``
+ * ``pre-os-secure-boot-auth-by-enrolled-digests``
+ * ``no-hardware-root-of-trust``
+
+See the `upstream definition in Secboot <https://github.com/canonical/secboot/blob/457b03a16d19dadd703c5a33bf8565f16113023c/efi/preinstall/checks_context.go#L162-L172>`__
+
 Sizing-policy
 ^^^^^^^^^^^^^
 
--- a/doc/reference/autoinstall-reference.rst
+++ b/doc/reference/autoinstall-reference.rst
@@ -623,6 +623,45 @@ Additionally, TPM-backed encryption can be enabled by using the ``hybrid`` layou
           name: hybrid
           encrypted: yes
 
+TPM-backed encryption supports the additional ``accepted-errors`` property, which lists the pre-install errors that the user accepts. For each accepted error, Subiquity automatically executes the corresponding ``proceed`` action that snapd offers for it, allowing the unattended installation to continue.
+
+.. note::
+
+   The ``accepted-errors`` property is the autoinstall equivalent of selecting **Solution: Ignore** for each listed error in the desktop installer UI.
+
+.. code-block:: yaml
+
+    autoinstall:
+      storage:
+        layout:
+          name: hybrid
+          encrypted: yes
+          accepted-errors: [running-in-vm]
+
+One can also specify ``accepted-errors`` as a single pre-install error (i.e., ``accepted-errors: running-in-vm``).
+
+The list of currently supported pre-install errors is:
+
+ * ``running-in-vm``
+ * ``insufficient-dma-protection``
+ * ``no-kernel-iommu``
+ * ``addon-drivers-present``
+ * ``sys-prep-applications-present``
+ * ``absolute-present``
+ * ``weak-secure-boot-algorithms-detected``
+ * ``pre-os-secure-boot-auth-by-enrolled-digests``
+ * ``no-hardware-root-of-trust``
+
+See the `upstream definition in Secboot <https://github.com/canonical/secboot/blob/457b03a16d19dadd703c5a33bf8565f16113023c/efi/preinstall/checks_context.go#L162-L172>`__
+
+.. note::
+
+   Listing an error kind means you accept the associated risk. Several errors kinds such as ``weak-secure-boot-algorithms-detected`` indicate a reduced security posture; only list them when you understand the implications for the installed system.
+
+.. note::
+
+   Only errors for which snapd offers a ``proceed`` action can be honored. Other errors (those that require a different action such as ``reboot``, ``enable-tpm-via-firmware`` or ``clear-tpm``) are **not** handled and will still cause the installation to fail, even if listed under ``accepted-errors``.
+
 Sizing-policy
 ^^^^^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


doc/reference/autoinstall-reference.rst
```diff
@@ -623,6 +623,33 @@ Additionally, TPM-backed encryption can be enabled by using the ``hybrid`` layou
...
+TPM-backed encryption supports the additional ``auto-proceed`` property, specifying the list of pre-install errors that Subiquity should automatically attempt to proceed past.
+
+.. code-block:: yaml
+
+    autoinstall:
+      storage:
+        layout:
+          name: hybrid
+          encrypted: yes
+          auto-proceed: [running-in-vm]
```
dbungert: should we support `auto-proceed: [*]` ?

doc/reference/autoinstall-reference.rst
```diff
@@ -623,6 +623,33 @@ Additionally, TPM-backed encryption can be enabled by using the ``hybrid`` layou
           name: hybrid
           encrypted: yes
 
+TPM-backed encryption supports the additional ``auto-proceed`` property, specifying the list of pre-install errors that Subiquity should automatically attempt to proceed past.
```
dbungert: The name `auto-proceed` feels awkward.  Tell me about your thoughts on this name.
I see the relationship to the underlying snapd action behavior.
What do you think about `acceptable-errors` instead?

Is proceed the only FixAction we imagine pre-authorizing?  For instance, could CLEAR_TPM be viable?

subiquity/server/controllers/storage.py
```diff
@@ -2199,25 +2213,41 @@ def ensure_core_boot_autoinstall_capability_compatible(self, capability) -> None
...
+                    )
+                    await self.execute_fix_action(action_proceed)
+                    await self.ensure_core_boot_autoinstall_capability_compatible(
+                        capability, autoproceed=autoproceed
+                    )
+                    return
                 raise AutoinstallError(
                     _(
-                        f"hybrid layout ({encryption_text}) is not available: {first_error}"
+                        f"hybrid layout ({encryption_text}) is not available: {error.kind.value}"
```
dbungert, line 2216: This error is potentially a common one for hybrid install failures.  For instance, if an error arrives but it doesn't have PROCEED as a possible fix action, I think that would be unclear if this error was raised since it was on `auto-proceed` presumably.    Is the contents of error.kind.value clear enough here or do we need to decorate it with additional information on why things didn't work?

Further, if someone isn't familiar with auto-proceed, they may need to populate it with multiple error kinds.  If we only show one at a time here the user potentially needs to do multiple rounds to get all the approved errors that they want to ignore.

dbungert (changes requested):
Thanks Olivier.  Fixing autoinstall here for TPMFDE is absolutely something we should do.  I want to talk about naming and some of the behaviors here before we continue.

doc/reference/autoinstall-reference.rst
ogayot: I'd rather stay on the safe side and force people to enumerate the values for now. I'm not completely opposed to have something similar in the future though.

doc/reference/autoinstall-reference.rst
ogayot: > The name auto-proceed feels awkward. Tell me about your thoughts on this name.
> I see the relationship to the underlying snapd action behavior.
> What do you think about acceptable-errors instead?

I agree that `auto-proceed` feels a bit awkward. The reason why I didn't use the term "error" in the name is because I don't consider "running-in-vm" an actual error, although it's defined by secboot as an "error kind". But yeah, maybe `acceptable-errors`, `acceptable-preinstall-errors`, or something else reads better.

> Is proceed the only FixAction we imagine pre-authorizing? For instance, could CLEAR_TPM be viable?

Honestly I don't have a very good answer besides "maybe". At least not now? Proceed is by far the easiest action to deal with, because we can close our eyes and run it without too much risk.

subiquity/server/controllers/storage.py
ogayot: > This error is potentially a common one for hybrid install failures. For instance, if an error arrives but it doesn't have PROCEED as a possible fix action, I think that would be unclear if this error was raised since it was on auto-proceed presumably. Is the contents of error.kind.value clear enough here or do we need to decorate it with additional information on why things didn't work?

Right, maybe we check if there's an associated Action.PROCEED and add either:
* "cannot automatically proceed" if there's no Action.PROCEED
* "consider adding {error.kind.value} to storage: layout: auto-proceed" (or similar) otherwise?

> Further, if someone isn't familiar with auto-proceed, they may need to populate it with multiple error kinds. If we only show one at a time here the user potentially needs to do multiple rounds to get all the approved errors that they want to ignore.

I honestly feel like it's acceptable.

doc/reference/autoinstall-reference.rst
dbungert: > error

What term do we use in the Desktop UI for this concept?  We should probably match that.

subiquity/server/controllers/storage.py
dbungert: > Right, maybe we check if there's an associated Action.PROCEED and add either:
> 
> * "cannot automatically proceed" if there's no Action.PROCEED
> * "consider adding {error.kind.value} to storage: layout: auto-proceed" (or similar) otherwise?

That sounds reasonable.
 
> > Further, if someone isn't familiar with auto-proceed, they may need to populate it with multiple error kinds. If we only show one at a time here the user potentially needs to do multiple rounds to get all the approved errors that they want to ignore.
> 
> I honestly feel like it's acceptable.

I remain concerned about this part - we can iterate quickly with VMs but real users iterating on autoinstall on real hardware find the process to be slow.

How common is it expected to be that we should have multiple PROCEEDs to acknowledge?

doc/reference/autoinstall-reference.rst
ogayot: There isn't really a word like "error", "problem", ... in the UI.

They offer "solutions" and "ignore" is what applies to Action.PROCEED 

image

doc/reference/autoinstall-reference.rst
dbungert: OK, let's skip for now but consider for a future PR

subiquity/server/controllers/storage.py
dbungert: So I'm OK with not implementing this right now, but I think we should consider this for a follow-on add-on.

doc/reference/autoinstall-reference.rst
dbungert: I'm still stuck on this one.  What do we think about the term `ignore-errors` ?

ogayot: I've spent more time thinking about this. I think `"accepted-errors": ["running-in-vm"]` (or "acceptable-errors") is the strongest option.

I prefer "accepted" over "ignored" because it fits better when there is a security trade-off in my opinion.

For instance, if there is no hardware root of trust (i.e., `no-hardware-root-of-trust`), the user can accept the risk but they will need to enable PIN or passphrase. If they ignored the problem, they wouldn't expect to be forced to use a PIN/passphrase.

dbungert: I leaned toward "ignored" to draw similarities to the live actions taken through the UI.  I can work with "acceptable-errors", I think it's the stronger of the two as it sounds like a better fit for things that might happen (or maybe not).

Can we document this to make it clearer that the Ignore option in the Desktop UX refers to the same concept?

ogayot: Updated the doc, thanks for the suggestion!

dbungert (approved):
Thanks!

PR #2398, opened 2026-07-03
apidef: redact ubuntu_pro GET responses that return secret tokens

ogayot:
The `/ubuntu_pro` GET endpoints return secret material in their response bodies but were not decorated with @redacted_response.


dbungert (approved):
Thanks!

PR #2400, opened 2026-07-07
workflow: use workshop as method of running CI.

dbungert:
* Move to using workshop as the primary method of running containers for CI.  This simplifies the story of reproducing the CI environment locally.
* Unify lint + tests there.
* Move aside existing direct LXD tests for now, to be removed in following PR.
* sudo is still needed to run integration tests due to not having a tty, so we can fake it with the read of /dev/console.  (Anticipated integration tests improvements using pyte should resolve that.)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,31 +18,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image:
-          - ubuntu-daily:noble
-          - ubuntu-daily:resolute
+        series:
+          - noble     # match core snap
+          - resolute  # latest
+        make-target:
+          - lint
+          - check
     steps:
     - uses: actions/checkout@v4
+    - uses: canonical/launch-workshop@v1
+      with:
+        workshop: subiquity-${{ matrix.series }}
     - name: run
-      run: |
-        sudo ./scripts/github-ci-lxd-setup.sh
-        sudo ./scripts/test-in-lxd.sh ${{ matrix.image }} "make check"
-
-  lint:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -18,31 +18,16 @@ jobs:
...
-          - ubuntu-daily:noble     # match the core snap we're running against
-          - ubuntu-daily:resolute  # latest
-    steps:
-    - uses: actions/checkout@v4
-    - name: lint
-      run: |
-        sudo ./scripts/github-ci-lxd-setup.sh
-        sudo ./scripts/test-in-lxd.sh ${{ matrix.image }} "make lint"
-
+      run: workshop exec ${{ matrix.workshop }} -- sudo make lint check
```
ogayot: I think it would help to separate lint from check, so that check runs even if the lint failed. Wdth?

I don't know if `make --keep-going` could help or if we need to split `make lint` from `make check`

ogayot (commented):
Thanks!

dbungert:
OK, lint and check are now matrixed so we can split them on different runners and allow lint to fail without blocking check from running.  https://github.com/dbungert/subiquity/actions/runs/29455307317

Unbreaking lint for clean CI now.

.github/workflows/ci.yaml
dbungert: Using the matrix functionality to specify the target for make.

dbungert:
One last tweak - I've updated the matrix variable name and values, which results in titles of the checks being nicer to read.

ogayot (approved):

PR #2403, opened 2026-07-13
storage: fix use of type instead of instance of Reformat in tests

ogayot:


dbungert (approved):
ah, yes, that makes sense.  Thanks!

PR #2405, opened 2026-07-15
Add experimental support for TPM/FDE in a gap

ogayot:
Add experimental support for TPM/FDE installs in a gap.

When the `--experimental-use-gap-tpm-fde` option is passed to Subiquity server, Subiquity can start advertising TPM/FDE scenarios in a gap.

This only works if installing in the gap would still result in an install compliant with the requirements from snapd/the gadget.

@d-loose fyi

### DIFF
--- a/subiquity/common/types/storage.py
+++ b/subiquity/common/types/storage.py
@@ -311,6 +311,7 @@ class GuidedDisallowedCapabilityReason(enum.Enum):
     CORE_BOOT_ENCRYPTION_UNAVAILABLE = enum.auto()
     NOT_UEFI = enum.auto()
     THIRD_PARTY_DRIVERS = enum.auto()
+    INCOMPATIBLE_LOCATION = enum.auto()
 
 
 @attr.s(auto_attribs=True)
--- a/subiquity/common/types/storage.py
+++ b/subiquity/common/types/storage.py
@@ -30,11 +30,35 @@ class ProbeStatus(enum.Enum):
     DONE = enum.auto()
 
 
+class Bootloader(enum.Enum):
+    GRUB = "GRUB"
+
+    @classmethod
+    def from_snapd(cls, value: str):
+        # Typically, snapd uses "grub" as the bootloader.
+        # But it also supports:
+        # * "android-boot"
+        # * "lk"
+        # * "u-boot"
+        # * "piboot"
+        # https://github.com/canonical/snapd/blob/b247de9fd0682c56d463d2ab2e906a6dfa234296/gadget/gadget.go#L111-L115
+        return cls(value.upper())
+
+
 class FirmwareType(enum.Enum):
     NONE = "NONE"  # a system where the bootloader is external, e.g. s390x
     BIOS = "BIOS"  # BIOS, where the bootloader gets dd-ed to the start of a device
     UEFI = "UEFI"  # UEFI, ESPs and /boot/efi and all that (amd64 and arm64)
     PREP = "PREP"  # ppc64el, which puts grub on a PReP partition
+
+    def _bootloader(self) -> Bootloader | None:
+        """Returns the bootloader typically used on a platform with this
+        firmware type.
+        In the future, this function might be moved outside of FirmwareType,
+        thus the leading underscore in the name."""
+        if self == self.NONE:
+            return None
+        return Bootloader.GRUB
 
 
 @attr.s(auto_attribs=True)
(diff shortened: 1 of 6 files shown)


subiquity/common/types/storage.py
```diff
@@ -287,6 +287,7 @@ class GuidedDisallowedCapabilityReason(enum.Enum):
     CORE_BOOT_ENCRYPTION_UNAVAILABLE = enum.auto()
     NOT_UEFI = enum.auto()
     THIRD_PARTY_DRIVERS = enum.auto()
+    INCOMPATIBLE_GAP = enum.auto()
```
dbungert: TOO_SMALL is arguably also an INCOMPATIBLE_GAP.  INCOMPATIBLE_GAP would work but I wonder if we can be more specific.  The root cause here is that the install type has physical location requirements, and this gap doesn't satisfy those requirements.  What do you think about the name INVALID_LOCATION?

subiquity/server/controllers/storage.py
```diff
@@ -261,6 +265,28 @@ def is_core_boot_classic(self) -> bool:
...
+
+        if not isinstance(gap.device, ModelDisk):
+            return False
+
+        disk = gap.device
+
+        if disk.ptable is not None and disk.ptable != volume.schema:
+            return False
+
+        # TODO should we check the bootloader too? Typically it's "grub" in the
```
dbungert: I don't think it has to be this PR but I do think we should handle that case before calling this feature Done Done.  Or is it an easy check?

subiquity/server/controllers/storage.py
```diff
@@ -271,10 +297,19 @@ def capability_info_for_gap(
         else:
             gap_size = gap.size
         r = self.capability_info.copy()
-        if gap_size < install_min:
-            r.disallow_all(
-                reason=GuidedDisallowedCapabilityReason.TOO_SMALL,
-            )
+        if self.is_core_boot_classic():
+            # I guess we can ignore install_min.
```
dbungert: This should be correct but let's tighten up the comment with some certainty.

subiquity/server/controllers/tests/test_storage.py
```diff
@@ -2996,6 +3098,102 @@ async def test_resize_has_enough_room_for_partitions__logical(self, p_boot_plan)
...
+
+        target = GuidedStorageTargetUseGap(
+            disk_id=self.disk.id,
+            gap=labels.for_client(gaps.largest_gap(self.disk)),
+        )
+        choice = GuidedChoiceV2(
+            target=target, capability=GuidedCapability.CORE_BOOT_ENCRYPTED
+        )
+
+        self.app.opts.experimental_use_gap_tpm_fde = False
```
dbungert, line 3156: this test and the next look identical up to this point, would be nice to de-dupe the test setup

dbungert (commented):
Thanks Olivier, a couple comments.

subiquity/server/controllers/storage.py
ogayot: I agree we should do it. According to https://documentation.ubuntu.com/core/how-to-guides/image-creation/build-a-gadget-snap/, bootloader can take the values "grub" or "u-boot". 

What makes it awkward is that "bootloader" in Subiquity refers to UEFI, BIOS, PREP.

ogayot: https://github.com/canonical/subiquity/pull/2406 could help make the ambiguity less apparent.

subiquity/server/controllers/tests/test_storage.py
ogayot: Thanks, reworked!

subiquity/common/types/storage.py
ogayot: I think that makes sense, thanks for the suggestion! Does "INCOMPATIBLE_LOCATION" read better than "INVALID_LOCATION"?

Otherwise, I'll go for the INVALID_LOCATION.

dbungert: I like INCOMPATIBLE_LOCATION better, INVALID sounds like a more permanent problem where INCOMPATIBLE sounds more like something that might improve later (which I do believe will be the case)

subiquity/server/controllers/storage.py
ogayot: I've added an extra check for it.

subiquity/server/controllers/storage.py
ogayot: This seems not to be the case in practice so I'm making the `gap_size < install_min` test effective also for tpm/fde.

dbungert (approved):
Thanks!

PR #2406, opened 2026-07-16
storage: rename Bootloader -> FirmwareType

ogayot:
UEFI, BIOS and PREP use to be referred to bootloaders in Subiquity. That said, examples of what typically is referred to as a bootloader are Grub and U-Boot.

The integration with snapd makes this wording inconsistency worse.

This patch now changes the wording from "bootloader" to "firmware-type" when it comes to UEFI, BIOS or PREP.

I am not 100% sold on the name yet, feedback would be appreciated.

For now, the `--bootloader` cmdline argument has been kept for backward compatibility purpose bug I'm planning to get rid of it.

As for the API, `bootloader` is now an alias of `firmware_type` too. We can keep this one around longer if needed.

### DIFF
--- a/subiquity/common/storage/tests/test_actions.py
+++ b/subiquity/common/storage/tests/test_actions.py
@@ -17,7 +17,7 @@ import unittest
 
 from subiquity.common.storage import gaps
 from subiquity.common.storage.actions import DeviceAction
-from subiquity.models.storage import Bootloader
+from subiquity.models.storage import FirmwareType
 from subiquity.models.tests.test_storage import (
     make_disk,
     make_lv,
@@ -129,11 +129,11 @@ class TestActions(unittest.TestCase):
         self.assertActionNotSupported(disk, DeviceAction.DELETE)
 
     def test_disk_action_TOGGLE_BOOT_NONE(self):
-        model, disk = make_model_and_disk(Bootloader.NONE)
+        model, disk = make_model_and_disk(FirmwareType.NONE)
         self.assertActionNotSupported(disk, DeviceAction.TOGGLE_BOOT)
 
     def test_disk_action_TOGGLE_BOOT_BIOS(self):
-        model = make_model(Bootloader.BIOS)
+        model = make_model(FirmwareType.BIOS)
         # Disks with msdos partition tables can always be the BIOS boot disk.
         dos_disk = make_disk(model, ptable="msdos", preserve=True)
         self.assertActionPossible(dos_disk, DeviceAction.TOGGLE_BOOT)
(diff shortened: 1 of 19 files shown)


subiquity/common/types/storage.py
```diff
@@ -362,7 +362,9 @@ class GuidedDisallowedCapability:
 class StorageResponse:
     status: ProbeStatus
     error_report: Optional[ErrorReportRef] = None
-    bootloader: Optional[Bootloader] = None
+    firmware_type: Optional[FirmwareType] = None
+    # Deprecated, for backward compatiblity
```
dbungert:
```suggestion
    # Deprecated, for backward compatibility
```

subiquity/server/controllers/storage.py
```diff
@@ -1003,7 +1003,9 @@ def full_probe_error(self):
     def _done_response(self):
         return StorageResponse(
             status=ProbeStatus.DONE,
-            bootloader=self.model.bootloader,
+            firmware_type=self.model.firmware_type,
+            # Temporary, for backward compatiblity
```
dbungert:
```suggestion
            # Temporary, for backward compatibility
```

subiquity/common/storage/tests/test_actions.py
```diff
@@ -187,11 +187,11 @@ def test_disk_action_TOGGLE_BOOT_BIOS(self):
         )
         self.assertActionNotPossible(gpt_disk4, DeviceAction.TOGGLE_BOOT)
 
-    def _test_TOGGLE_BOOT_boot_partition(self, bl, flag):
+    def _test_TOGGLE_BOOT_boot_partition(self, fwt, flag):
         # The logic for when TOGGLE_BOOT is enabled for both UEFI and PREP
         # bootloaders turns out to be the same, modulo the special flag that
```
dbungert, line 192: This PR has a challenge of that bootloader is doing double duty to both mean firmware and actual bootloader, but I think bootloader referenced here is actually firmware.

subiquity/server/controllers/tests/test_identity.py
```diff
@@ -116,7 +116,7 @@ class TestControllerUserCreationFlows(SubiTestCase):
     # See subiquity/models/tests/test_subiquity.py for details.
     def setUp(self):
         self.app = make_app()
-        self.app.opts.bootloader = False
+        self.app.opts.firmware_type = False
```
dbungert: (preexisting problem) should give it a different value than False.

dbungert (approved):
Thank you for doing this.  This was another concept name that was off, I appreciate the rename.

PR #2407, opened 2026-07-23
kvm-test: several fixes

dbungert:
* fallback to a baseiso value from the --iso command line, if not present in the config
* fixes for Path() usage
* LIVEFS_EDITOR env variable to pass to sudo - letting quick-test handle this usually doesn't work since sudo clears the environment
* switch to sudo mount / umount as fuseiso is current incompatible with our usage and with apparmor
* stop adding core20, we don't need that anymore
* use -cpu host for amd64v3 support


Chris-Peterson444 (approved):
Nice, thanks!

PR #2409, opened 2026-07-28
storage: Allow Mount to accept filesystem types not backed by a device

ghadi-rahme:
This fixes tmpfs mounts causing autoinstall crashes.

LP:#2041725

### DIFF
--- a/subiquity/models/storage.py
+++ b/subiquity/models/storage.py
@@ -562,24 +562,25 @@ def asdict(inst, *, for_api: bool):
     r = collections.OrderedDict()
     for field in attr.fields(type(inst)):
         metadata = field.metadata
-        if not for_api or not metadata.get("for_api", False):
-            if field.name.startswith("_"):
-                continue
         name = field.name.lstrip("_")
         m = getattr(inst, "serialize_" + name, None)
         if m:
+            # Serialize_* should run for public and private fields.
             r.update(m())
-        else:
-            v = getattr(inst, field.name)
-            if v is not None:
-                if metadata.get("ref", False):
-                    r[name] = v.id
-                elif metadata.get("reflist", False):
-                    r[name] = [elem.id for elem in v]
-                elif isinstance(v, StorageInfo):
-                    r[name] = {v.name: v.raw}
-                else:
-                    r[name] = v
+            continue
+        if not for_api or not metadata.get("for_api", False):
+            if field.name.startswith("_"):
+                continue
+        v = getattr(inst, field.name)
+        if v is not None:
+            if metadata.get("ref", False):
+                r[name] = v.id
+            elif metadata.get("reflist", False):
+                r[name] = [elem.id for elem in v]
+            elif isinstance(v, StorageInfo):
+                r[name] = {v.name: v.raw}
+            else:
+                r[name] = v
     return r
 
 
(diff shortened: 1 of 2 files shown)


dbungert:
Hi @ghadi-rahme, thanks for the PR.
Would you add unit test case(s) covering this bugfix?  It's the best bet against future regressions.

ghadi-rahme:
Hello @dbungert, I have added the tests, let me know how it looks.

subiquity/models/storage.py
```diff
@@ -562,24 +562,25 @@ def asdict(inst, *, for_api: bool):
     r = collections.OrderedDict()
     for field in attr.fields(type(inst)):
         metadata = field.metadata
-        if not for_api or not metadata.get("for_api", False):
```
dbungert, line 565: serialize_ now runs on both public and private - this part is fine - but does so regardless of for_api status, and I'm not sure that's intended.

subiquity/models/storage.py
dbungert, line 1407: You're correct in pointing out that self.device can be none, but note that there are other cases (not exercised in the autoinstall case, but still) that deref self.device

dbungert, line 1410: and here

dbungert, line 1418: and here

dbungert (commented):
Hi Ghadi - thanks for the PR, I think it's broadly correct.  A few comments I'd like you to consider.

subiquity/models/storage.py
ghadi-rahme: This is currently not a problem since none of the fields that have `for_api` metadata start with `serialize_`. 

However I did notice that all the fields tagged with `for_api` are all private fields. I will rewrite this function with this in mind to simplify it and make it skip any field that is tagged `for_api` when the argument `for_api` is false.

so something like this towards the top of the function:
```
is_for_api = metadata.get("for_api", False)
if is_for_api and not for_api:
    continue
```

ghadi-rahme: Ignore my above comment, after double checking the function I realized that the solution I proposed is wrong.

ghadi-rahme:
Thank you for the review @dbungert !

I have added None checks in `on_remote_storage` and `can_delete` inside the `Mount` class.
However I did not get the problem about `serialize_*` running for both public and private regardless of the `for_api` field.

The previous method blocked all private fields that were not labeled with the `for_api` metadata or when not rendering for the API. This current implementation will allow these fields to be added to the dict as long as they have their own `serialize_*` method. If I were to keep the check where it remains, then fstype will get ignored since its private.

I guess the main point of confusion is understanding which fields are expected to be filtered out when not using the API and why `serialize_*` fields shouldn't be included when not rendering for the API.

I looked at the original commit that added this part of the method and could not find any clarification on this.

ogayot (changes requested):
Thank you Ghadi, and apologies for the delayed review!

I find it a bit sad that `Mount` objects behave differently based on where the filesystem being mounted is located. But that is an existing problem today.

I think in the future, I'd prefer `Mount` to have a `filesystem`/`fs` attribute that is always set. And the associated `Filesystem` object can have an optional `device`/`volume` associated to it. It would make things cleaner IMO.

But that isn't how curtin works today - and of course I'm not asking you to do it now.

I think your changes are sound, given the current situation. Good work!

A couple of things to address before merging:

* The TUI would probably choke if it ran into a configuration with a `Mount` that has no `.device`. Conveniently, I don't think we offer any way to customize a storage configuration submitted via autoinstall. Would you mind adding a comment (perhaps coupled with a FIXME) in `MountInfo` (i.e., subiquity/ui/views/storage/storage.py) explaining that it assumes `Mount` instances have a `.device` - which isn't true for tmpfs?
* In your commit message, please change `LP#2041725` to `LP: #2041725`.

Everything else looks good to me. Thank you!

ghadi-rahme:
Thank you @ogayot for the review! 

I have applied the requested changes.

subiquity/ui/views/storage/storage.py
```diff
@@ -56,6 +55,7 @@
 
 @attr.s
 class MountInfo:
+    # FIXME: `mount` is assumed to have a .device property here, which is the case for tmpfs mounts.
```
ogayot:
```diff
-# FIXME: `mount` is assumed to have a .device property here, which is the case for tmpfs mounts.
+# FIXME: `mount` is assumed to have a .device property here, which is not the case for tmpfs mounts.
```

ogayot (changes requested):
Thank you! Just a minor comment tweak and we can merge.

subiquity/ui/views/storage/storage.py
ghadi-rahme: fixed.
When I first wrote the line I missed the `not` and added it back, however when I came back to my editor now to see why it wasn't part of the commit I realized that I never saved the file before committing the change.

ogayot (approved):
Thank you! Merging

PR #2410, opened 2026-08-10
Read version from /etc/os-release, with fallback to /etc/lsb-release

ogayot:
The `/etc/lsb-release` file has been dropped from `stonking`.

And while the `/usr/bin/lsb_release` script has been adapted to read from `/etc/os-release`, Subiquity does not rely on this script (because it can't be used in a partial chroot).

I added support for reading `/etc/os-release` and replaced all occurrences of `lsb_release`. by `read_ubuntu_version`, 

`read_ubuntu_version` reads from `/etc/os-release` if present, and falls back to `/etc/lsb-release`.

Unlike `lsb_release()` which returns a dictionary of all properties, `read_ubuntu_version` returns an `attrs` object that contains the following attributes:

* `release` (i.e., the version number stored as a string)
* `codename` (e.g., stonking, resolute)
* `pretty_name` (e.g., `"Ubuntu Stonking Stingray (development branch)"`)

Note: initially I was storing the release as a tuple of integers, e.g., `(26, 4)` to make comparisons cleaner.
Unfortunately, it made it awkward to print the version number as text.

Tested successfully with a refreshed snap on a server 26.10 ISO 

LP:#2162082


subiquitycore/os.py
```diff
@@ -0,0 +1,121 @@
...
+    if dry_run and path is not None:
+        raise ValueError("Both dry_run and path are specified.")
+
+    if path is None:
+        path = LEGACY_LSB_RELEASE_EXAMPLE if dry_run else LEGACY_LSB_RELEASE
+
+    return lsb_release_from_path(path)
+
+
+def read_ubuntu_info(path: Path | None = None, dry_run=False) -> UbuntuInfo:
```
ogayot: Maybe we get rid of the `path` parameter here. It's awkward for this abstraction.

subiquity/client/controllers/ubuntu_pro.py
```diff
@@ -63,14 +63,14 @@ async def make_ui(self) -> UbuntuProView:
         dry_run: bool = self.app.opts.dry_run
         pre_release = False
 
-        lsb = lsb_release(dry_run=dry_run)
+        info = read_ubuntu_info(dry_run=dry_run)
 
-        if "LTS" not in lsb["description"]:
-            major, minor = lsb["release"].split(".")
+        if "LTS" not in info.pretty_name:
+            major, minor = info.release.split(".")
```
dbungert, line 70: I think it would be nice to have this functionality (is LTS? and getting the tuple of major, minor, ...) as direct functionality of UbuntuInfo, it feels like it's likely to be reused

subiquity/client/controllers/ubuntu_pro.py
```diff
@@ -63,14 +63,14 @@ async def make_ui(self) -> UbuntuProView:
...
 
-            # If running a pre-LTS (e.g., 24.04, 26.04, ...), show the SSH UI
-            # for testing, but with a warning.
             if int(major) >= 20 and int(major) % 2 == 0 and minor == "04":
```
dbungert: same here, a nice property or so for this functionality would help and obviate the need for the first part of the comment

subiquity/models/subiquity.py
```diff
@@ -353,7 +353,7 @@ def _cloud_init_config(self):
             merge_config(config, {"packages": list(self.cloud_init_packages)})
         if self.userdata is not None:
             merge_cloud_init_config(config, self.userdata)
-        if lsb_release()["release"] not in ("20.04", "22.04"):
+        if read_ubuntu_info().release not in ("20.04", "22.04"):
```
dbungert: is this wrong on point releases?  UbuntuInfo.release does mention that release versions can be in form x.y.z.

subiquity/server/apt.py
```diff
@@ -272,7 +272,7 @@ async def configure_for_install(self, context):
                     continue
                 self.install_tree.pp(relpath).unlink()
 
-        codename = lsb_release(dry_run=self.app.opts.dry_run)["codename"]
+        codename = read_ubuntu_info(dry_run=self.app.opts.dry_run).codename
```
dbungert, line 275: (non-blocking) existing code uses codename in some spots and series in others, do you want to unify?  We are doing some renames (see below)

subiquity/common/os.py
```diff
@@ -0,0 +1,121 @@
...
+            path.read_text(), key_predicate=lambda x: x.startswith("DISTRIB_")
+        )
+        return UbuntuInfo(
+            codename=props["DISTRIB_CODENAME"],
+            release=props["DISTRIB_RELEASE"],
+            pretty_name=props["DISTRIB_DESCRIPTION"],
+        )
+
+    @classmethod
+    def from_os_release(cls, path: Path) -> "UbuntuInfo":
```
dbungert, line 75: The os release logic we recently added to curtin from cpython has handling for quotes and special characters.  I'd like to have a common approach between curtin and subiquity.  What do you suggest?

examples/os-release-resolute
dbungert, line 1: does anything reference this file?  I see an os-release-focal path reference but the file is not in this PR.

subiquitycore/tests/test_os.py
```diff
@@ -0,0 +1,144 @@
...
+
+        with patch("subiquitycore.os.Path.read_text", return_value=content):
+            info = UbuntuInfo.from_lsb_release(path=Path("/dev/null"))
+
+        self.assertEqual("resolute", info.codename)
+        self.assertEqual("26.04", info.release)
+        self.assertEqual("Ubuntu 26.04 LTS", info.pretty_name)
+
+    def test_from_os_release(self):
+        content = """\
```
dbungert: do you want to load this from the example file?

dbungert (commented):
Thanks Olivier.  I have a larger question comparing to what we do in Curtin which I would like to unify.  After that the feedback is pretty minor.

subiquity/client/controllers/tests/test_ubuntu_pro.py
Dh2Cool, line 1: suggestion: these 4 tests (`lts`/`not_lts`/`noble_devel`/`26_04_future`) are the same shape, just different `UbuntuInfo` inputs and expected `pre_release`/`Skip` outcome. Could this follow the pattern you used in `test_oem.py`'s `test_install_on` where you can use `@parameterized.expand` over the values.

subiquity/models/subiquity.py
Dh2Cool: Olivier's PR description mentions release can look like (24, 4, 3), so "22.04.3" not in ("20.04", "22.04") would be True and disable cloud-init wrongly. Also there's two of these checks (line 356 and 417), so both would need the fix.

subiquitycore/os.py
```diff
@@ -0,0 +1,121 @@
...
+    if dry_run and path is not None:
+        raise ValueError("Both dry_run and path are specified.")
+
+    if path is not None:
+        # Let's assume it's an os-release file.
+        return UbuntuInfo.from_os_release(path=path)
+    else:
+        with contextlib.suppress(FileNotFoundError):
+            return UbuntuInfo.from_os_release(path=os_release)
+        return UbuntuInfo.from_lsb_release(path=legacy_lsb_release)
```
Dh2Cool: This silently falls back to lsb-release with no log line. Could we add a log.debug here so we know which path was used?

Dh2Cool (commented):
This is my first review here!
Left a few small comments and suggestions mostly around edge cases and test cleanup
Thanks for the PR, Olivier!

subiquity/client/controllers/ubuntu_pro.py
ogayot: Agreed. I've introduced `version_number()` and `is_marked_lts()`. 

I didn't go for `is_lts()` because during the development cycle of an LTS, the function returns `False`.

subiquitycore/os.py
ogayot: Done

subiquity/models/subiquity.py
ogayot: Apologies, I'm the one who got confused.

`DISTRIB_RELEASE` (from `/etc/lsb-release`) and `VERSION_ID` (from `/etc/os-release`) do **not** contain the patch number (e.g. 24.04 **.3**). I'll update the PR description.

If we ever want the patch number (we might at some point), we have to read from another field, maybe `DISTRIB_DESCRIPTION` or `VERSION` but the format of these fields is less strict. Unfortunately, there is no field that contains just the version number + the patch number.

So, to answer the original question, this code should work for point releases.

examples/os-release-resolute
ogayot: Not yet. We're still pretending that dry-run subiquity runs on focal:
```python
LSB_RELEASE_EXAMPLE = "examples/lsb-release-focal"
```
I can bump to `resolute` but I think other updates would be required. I'll do that in a separate PR.

I still think it's useful to include the example file though.

subiquitycore/tests/test_os.py
ogayot: Makes sense, done!

examples/os-release-resolute
ogayot: I've used the file as test-data now.

subiquity/common/os.py
ogayot: You're right. It sounds silly to introduce yet another implementation that resembles `freedesktop_os_release`. 

I guess importing `curtin.util.load_os_release` from `curtin` is an option.
That implies moving `subiquitycore.os` somewhere else like `subiquity.common.os` because `subiquitycore` should not depend on `curtin`.

I don't think copying the implementation from `curtin` to `subiquity` makes sense. That would be yet another place to update the code if `freedesktop_os_release` changes.

Maybe we could try adding support for alternative paths to cpython?

subiquity/server/apt.py
ogayot: I think unifying makes sense. That said, I'm hesitant on the term to use.

`codename` reflects `/etc/os-release` and `/etc/lsb-release` better but in documentation, `codename` typically includes the animal name e.g., "Resolute Raccoon"

https://ubuntu.com/project/docs/release-team/list-of-releases/#list-of-current-releases

IMO `series` is a bit more mysterious, but could be a decent alternative. Thoughts?

subiquitycore/os.py
ogayot: Good idea! Implemented.

subiquity/client/controllers/tests/test_ubuntu_pro.py
ogayot: Seems possible but I think that's outside the scope of this PR. I'll keep the idea in mind for a follow-on PR.

subiquity/server/apt.py
dbungert: In the immediate code sense, `suite` would also be a candidate name as it's the term documented in `man sources.list`.  I lean toward `codename` or `codename_adjective` if we want to be very clear about it, mirroring the term used in os-release.

dbungert (approved):
Thanks!

subiquity/server/apt.py
ogayot: Follow on PR: https://github.com/canonical/subiquity/pull/2412

PR #2415, opened 2026-08-19
Move core-boot install bits outside of storage.py

ogayot:
This PR introduces a new core_boot_installer.py file, containing the implementation of the different "steps" that Subiquity needs to run in a TPM/FDE installation.

For now, I've left the GET /v2/systems/<label> and GET /v2/systems operations in storage.py but we can move them in the future.

This makes storage.py smaller, which is always nice, and makes the overall structure more consistent.

### DIFF
--- a/subiquity/server/snapd/core_boot_installer.py
+++ b/subiquity/server/snapd/core_boot_installer.py
@@ -115,7 +115,7 @@ class CoreBootInstaller:
 
         return result.recovery_key
 
-    async def snapd_target_preseed(self, target: pathlib.Path) -> None:
+    async def target_preseed(self, target: pathlib.Path) -> None:
         async with AsyncExitStack() as es:
             # Bind-mount required filesystems
             # Some of these might already be mounted by curtin, but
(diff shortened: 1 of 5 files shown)


subiquity/server/snapd/core_boot_installer.py
```diff
@@ -0,0 +1,184 @@
...
+from subiquitycore.context import with_context
+
+if TYPE_CHECKING:
+    from subiquity.server.controllers.storage import VariationInfo
+
+log = logging.getLogger("subiquity.server.snapd.core_boot_installer")
+
+
+@attrs.define
+class CoreBootInstallPlan:
```
dbungert: Tell me about the Plan term here.  At first I was thinking it would be similar to the Boot Plan, but it's a bit different, and I think the hint on that is the near immediate use of the term "state" instead in the docstring.  When we follow OnVolume and VolumesAuth this is just data saved to tell future us what to do, whereas the plan has chained together the logic to actually do the changes needed for the boot plan.

ogayot: You're right. I didn't think about the existing boot plan. I agree it makes things slightly confusing. I think `CoreBootInstallData` will feel clearer.

dbungert (approved):
Thanks!

PR #2416, opened 2026-08-19
examples: make ubuntu-and-free-space resizable

ogayot:
Many `examples/machines` files were introduced before we added support for resizing filesystems. Therefore, they don't include ESTIMATED_MIN_SIZE properties on filesystems.

Adding ESTIMATED_MIN_SIZE on the last ext4 filesystem of the disk in `ubuntu-and-free-space.json`, so that we can exercise resize scenarios in dry-run.


dbungert (approved):
Thanks!

PR #2418, opened 2026-08-24
subiquity: bump dry-run version to stonking to add 26.10+ grub requirements

ogayot:
26.10+ boot requirements for signed grub are currently not exercised in dry-run mode because the default os-release / lsb-release files are from focal.

The main drawback of moving to stonking is that it's not a LTS - therefore the Ubuntu Pro screen will vanish. 

Marking as a draft until we figure out the options.

LP:#2165727

### DIFF
--- a/examples/answers/bridge.yaml
+++ b/examples/answers/bridge.yaml
@@ -1,4 +1,5 @@
 #source-catalog: examples/sources/bridge.yaml
+#dr-config: examples/dry-run-configs/bridge.yaml
 Source:
   source: ubuntu-server
   search_drivers: true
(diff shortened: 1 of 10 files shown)


dbungert (commented):
I suppose you could create an obviously fake release at version 28.04.  Heavy emphasis on obviously fake.

ogayot:
The problem with CI is that since 26.04, we search for OEM packages on server, and `Ubuntu-Oem-Kernel-Flavour` has priority over the bridge kernel we request.

ogayot:
> I suppose you could create an obviously fake release at version 28.04. Heavy emphasis on obviously fake.

Right, I think I'm fine dropping ubuntu-pro from the default dry-run screens. But on the long run, it would be good if dry runs could specify the release.

dbungert (approved):

ogayot:
I didn't realize #2326 was related :(

PR #2419, opened 2026-08-27
subiquity: fix AttributeError when building the About message

ogayot:
In the following commit, we introduced read_ubuntu_info(), returning an UbuntuInfo object instead of an lsb_release dictionary:

  515ddf1a subiquity: replace use of lsb-release by read_ubuntu_info

Unfortunately, replacing the dict by an UbuntuInfo object broke the ability to display help because _default_about_msg() calls info.update() and template.format(**info) on the object returned by read_ubuntu_info(). The UbuntuInfo attrs class has no update() method and is not a mapping, causing an AttributeError when opening the About dialog.

Fixed by converting the UbuntuInfo instance to a plain dict with attrs.asdict() before merging in the snap version/revision fields and formatting the template.

LP:#2165329

### DIFF
--- a/subiquity/ui/views/help.py
+++ b/subiquity/ui/views/help.py
@@ -16,6 +16,7 @@
 import logging
 import os
 
+import attrs
 from urwid import Divider, Filler, PopUpLauncher, Text, connect_signal
 
 from subiquity.common.os import read_ubuntu_info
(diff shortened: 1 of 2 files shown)


Dh2Cool:
there's no existing test file for this view, so unsure if it's necessary, but a small test calling _default_about_msg() would have caught this and any future template/field drift. Thanks!

ogayot:
> there's no existing test file for this view, so unsure if it's necessary, but a small test calling _default_about_msg() would have caught this and any future template/field drift. Thanks!

Good point, test added, thanks!

dbungert (approved):

PR #2422, opened 2026-09-02
workflow: +reposcan

dbungert:
* update format version for black
* pin actions to hash
* run pre-commit in CI, allowing some dedupe between CI and pre-commit


dbungert:
Moving PR to Draft - I think some of the reposcan encouraged fixes merit running in other PRs, such as updating our psf/black version, which we can merge first.

.github/workflows/ci.yaml
```diff
@@ -25,28 +28,27 @@ jobs:
...
         with:
-          version: "~= 25.1.0"
-          src: "console_conf subiquity subiquitycore"
-  format-isort:
-    runs-on: ubuntu-24.04
-    steps:
-      - uses: actions/checkout@v4
-      - uses: isort/isort-action@v1
+          persist-credentials: false
+          fetch-depth: 0
```
ogayot, line 47: does `fetch-depth: 0` speed things up? If so, should we use it for all the jobs?

.github/workflows/automatic-doc-checks.yml
```diff
@@ -17,6 +17,6 @@ concurrency:
 
 jobs:
   documentation-checks:
-    uses: canonical/documentation-workflows/.github/workflows/documentation-checks.yaml@main
+    uses: canonical/documentation-workflows/.github/workflows/documentation-checks.yaml@309dd382b6a07cbd23d84e3912fb05e6bc1f4ffb  # main
```
ogayot, line 22: This one is really bugging me but I don't have good suggestions to make 🤷‍♂️

ogayot (approved):
Thanks for moving this forward!

.github/workflows/automatic-doc-checks.yml
ogayot: Maybe add a comment mentioning https://redirect.github.com/canonical/documentation-workflows/issues/6

.github/workflows/ci.yaml
dbungert: Probably makes things slower actually - `fetch-depth: 0` fetches all commits.  I wonder if `fetch-depth: 2` would also work.

It was a requirement to get the doc checks to work, which isn't necessarily the objective of this PR - doc checks didn't like the default `fetch-depth: 1`.  https://github.com/canonical/subiquity/actions/runs/33656784293/job/100337199506#step:4:443

ogayot: Sounds good! If we're keeping `fetch-depth: 0`, would you mention that it's to make doc checks happy (because they don't like shallow repos)? It's not immediately obvious.

dbungert: Good call - added a note

ogayot (approved):
LGTM, thanks!

PR #2425, opened 2026-09-03
Fix issue when resizing LVM LVs

ogayot:
In e8dac16ea51ad6f2770b27e3a23d35c1c90563a1, we replaced 
    
```python
if gaps.largest_gap_size(vg) < 0:
```
with
```python    
if gaps.largest_gap_size(vg) < lv.size:
```
after noticing that the if body was unreachable (i.e., `largest_gap_size` always returns positive values).

However, we wrongly assumed why the branch had become unreachable. What made the branch actually unreachable was commit 83cfdbdf, which replaced:

```python
if vg.free_for_partitions < 0:    # free_for_partitions can be negative
```
by
```python
if gaps.largest_gap_size(vg) < 0:   # gaps.largest_gap_size(vg) can't return negative values
```

The intent of `vg.free_for_partitions < 0` was to check **after performing the resize** that we're not occupying more space than there is on the VG. But that does not work anymore when we switch to `largest_gap_size`.

The proper way to test if the LV would become too large is to do the comparison **before** resizing.

Also, LVM LV inconsistently aligns sizes to 1 MiB or 4 MiB. Changed to 4 MiB (i.e., LVM_CHUNK_SIZE) for consistency. 

LP:#2166333

### DIFF
--- a/subiquity/common/storage/manipulator.py
+++ b/subiquity/common/storage/manipulator.py
@@ -374,9 +374,11 @@ class StorageManipulator:
             if "name" in spec:
                 lv.name = spec["name"]
             if "size" in spec:
-                lv.size = align_up(spec["size"], LVM_CHUNK_SIZE)
-                if gaps.largest_gap_size(vg) < 0:
+                new_size = align_up(spec["size"], LVM_CHUNK_SIZE)
+                size_change = new_size - lv.size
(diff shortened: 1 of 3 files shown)


dbungert (approved):
Thank you!

PR #2426, opened 2026-09-03
ubuntu-advantage: remove empty try ... finally block

ogayot:
Not sure what exactly what I was trying to achieve but this construct looks clearly useless.


dbungert (approved):
heh, maybe the finally was a placeholder and it never was filled in.  Are you picturing some sort of cleanup here not handled by NamedTemporaryFile()?

ogayot:
> heh, maybe the finally was a placeholder and it never was filled in. Are you picturing some sort of cleanup here not handled by NamedTemporaryFile()?

I'm sure I had an idea in mind back then :laughing: 

I have actually been looking at our `contextmanager` methods, trying to see if things like https://redirect.github.com/canonical/curtin/pull/55 are more common than expected. But this is the only weirdness I found so far.

PR #2427, opened 2026-09-03
feat(keyboard): add localectl support for X11 keyboard layout persistence

eeickmeyer:
As ubuntu-desktop-bootstrap was built to be used by all Ubuntu flavors, we need to make sure we're not using too many GNOME-specific idioms. Otherwise desktops, such as KDE Plasma, miss out on certain features like setting the keyboard layout. Currently, this is completely missed when one uses this installer on Ubuntu Studio.

This pull request makes the keyboard layout setting more DE-agnostic and, since subiquity runs as root, this can be applied directly with --no-ask-password unlike the (unprivileged) installer frontend, which would otherwise have to either fail or show a polkit prompt asking for credentials that do not yet exist.

### DIFF
--- a/subiquity/server/controllers/keyboard.py
+++ b/subiquity/server/controllers/keyboard.py
@@ -16,6 +16,7 @@
 import logging
 import os
 import pwd
+import subprocess
 from typing import Dict, Optional, Sequence, Tuple
 
 import attr
@@ -256,6 +257,38 @@ class KeyboardController(SubiquityController):
             f"[('xkb','{xkb_value}')]",
         ]
         await self._run_gui_command(gsettings, **kwargs)
+        await self._set_localectl_x11_keyboard(layout, variant)
+
+    async def _set_localectl_x11_keyboard(self, layout: str, variant: str) -> None:
+        """Persist the layout via logind's org.freedesktop.locale1
+        SetX11Keyboard (through localectl), so desktops that honour it (e.g.
+        KDE Plasma with FollowLocale1=true) pick it up without a
+        desktop-specific call.
+        """
+        new = latinizable(layout, variant)
+        if new is not None:
+            layout, variant = new
+        cmd = [
+            "localectl",
+            "--no-ask-password",
+            "--no-convert",
+            "--",
+            "set-x11-keymap",
+            layout,
+            "",  # model
+            variant,
+        ]
+        if self.opts.dry_run:
+            scale = os.environ.get("SUBIQUITY_REPLAY_TIMESCALE", "1")
+            cmd = ["sleep", str(1 / float(scale))]
+        try:
+            # Run directly rather than via _run_gui_command: that helper
+            # de-privileges to the target user via systemd-run, whereas this
+            # call relies on subiquity's own root privileges to set the
+            # keymap without a polkit prompt.
+            await arun_command(cmd, check=True)
+        except subprocess.CalledProcessError as exc:
+            log.warning("Could not localectl set-x11-keymap: %r", exc)
 
     @with_context(description="configuring keyboard")
     async def setup_target(self, context):
--- a/subiquity/server/controllers/tests/test_keyboard.py
+++ b/subiquity/server/controllers/tests/test_keyboard.py
@@ -15,7 +15,7 @@
 
 import os
 import unittest
-from unittest.mock import Mock, patch
+from unittest.mock import Mock, call, patch
 
 import jsonschema
 from jsonschema.validators import validator_for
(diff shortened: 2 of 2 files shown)


subiquity/server/controllers/keyboard.py
```diff
@@ -256,6 +257,36 @@ async def set_input_source(self, layout: str, variant: str, **kwargs):
             f"[('xkb','{xkb_value}')]",
         ]
         await self._run_gui_command(gsettings, **kwargs)
+        await self._set_localectl_x11_keyboard(layout, variant)
+
+    async def _set_localectl_x11_keyboard(self, layout: str, variant: str) -> None:
+        """Persist the layout via logind's org.freedesktop.locale1
+        SetX11Keyboard (through localectl), so desktops that honour it (e.g.
+        KDE Plasma with FollowLocale1=true) pick it up without a
+        desktop-specific call. Subiquity runs as root, so this can be applied
```
dbungert: This last sentence of the docstring is odd - it's a couple of concepts that are baseline assumptions of the subiquity/server/* logic - that things are running as root and that the frontends are likely unprivileged.  I'd prefer to just remove this sentence entirely.

subiquity/server/controllers/keyboard.py
```diff
@@ -256,6 +257,36 @@ async def set_input_source(self, layout: str, variant: str, **kwargs):
...
+            "--no-ask-password",
+            "--no-convert",
+            "set-x11-keymap",
+            layout,
+            "",
+            variant,
+        ]
+        if self.opts.dry_run:
+            scale = os.environ.get("SUBIQUITY_REPLAY_TIMESCALE", "1")
+            cmd = ["sleep", str(1 / float(scale))]
```
dbungert, line 283: we probably don't need the sleep here in dry-run but this is matching the nearby code, so I'm reluctant to call it wrong.  For this PR I suggest we leave this style of dry run sleep but outside this PR we should address this.

subiquity/server/controllers/tests/test_keyboard.py
```diff
@@ -97,4 +102,16 @@ async def test_input_source(self, layout, variant, expected_xkb):
...
+                "--no-ask-password",
+                "--no-convert",
+                "set-x11-keymap",
+                expected_locale_layout,
+                "",
+                expected_variant,
+            ]
+            self.assertEqual(
+                mock_arun_command.call_args_list,
+                [call(cmd), call(localectl_cmd, check=True)],
```
dbungert: nit: since it's two commands now it'd be nice to give the `cmd` variable a diferent name, `gsettings_cmd` maybe.

dbungert (commented):
Hi Erich, thanks for the PR.   I have a few minor suggestions.

Please give this a try and confirm that it does what you expect, if you haven't already.  the script at `scripts/quick-test-this-branch.sh` can allow you to, relatively quickly, inject modified subiquity code into an existing ISO.

eeickmeyer:
Hi Dan,

I enacted those suggestions you made, and I'll check in a .iso run tomorrow. So, to summarize:

 - Removed the second sentence from the docstring
 - Renamed `cmd` to `gsettings_cmd`.
 - Left the sleep intact, though you're probably right about removing it and the other one eventually. I was simply following suit.

And I completely agree with both changes as 1) I guess I was overdocumenting, and 2) I'd normally be more specific with variable names, so I don't know why I didn't do that. :sweat_smile:

subiquity/server/controllers/keyboard.py
```diff
@@ -256,6 +257,33 @@ async def set_input_source(self, layout: str, variant: str, **kwargs):
...
+        try:
+            await arun_command(cmd, check=True)
+        except subprocess.CalledProcessError as exc:
+            log.warning("Could not localectl set-x11-keymap: %r", exc)
```
ogayot, line 291: Would it work to use 

```python
await self._run_gui_command(cmd)   # maybe with user="root" if needed
```

or do we need something that avoids systemd-run?

subiquity/server/controllers/keyboard.py
```diff
@@ -256,6 +257,33 @@ async def set_input_source(self, layout: str, variant: str, **kwargs):
...
+        if new is not None:
+            layout, variant = new
+        cmd = [
+            "localectl",
+            "--no-ask-password",
+            "--no-convert",
+            "set-x11-keymap",
+            layout,
+            "",
+            variant,
```
ogayot, line 279: Please add `--` after the last option. Also, it's not immediately clear what the `""` means so I would recommend
```python
  ...
  "--no-convert",
  "--",
  "set-x11-keymap",
  layout,
  "",  # model
  variant,
```

subiquity/server/controllers/tests/test_keyboard.py
```diff
@@ -64,11 +64,16 @@ def setUp(self):
...
+            ("us", "", "[('xkb','us')]", "us", ""),
+            ("fr", "latin9", "[('xkb','fr+latin9')]", "fr", "latin9"),
+            # non-latin layouts get expanded to a dual layout with a latin
+            # fallback, same as for the "main" set_keyboard flow.
+            ("ru", "", "[('xkb','ru')]", "us,ru", ","),
         ]
     )
-    async def test_input_source(self, layout, variant, expected_xkb):
+    async def test_input_source(
+        self, layout, variant, expected_xkb, expected_locale_layout, expected_variant
```
ogayot: I feel like "variant" and "expected_variant" being different comes out confusing. Would you rename the latter to make it more specific?

ogayot (commented):
Thanks Erich,

This looks promising. I am not in expert in desktop environments so I can't tell if that's going to have an impact on GNOME desktops too, and whether we should completely replace the gsettings config by something like what you did. @d-loose can maybe help confirm here?

Otherwise, I have a few suggestions from the code standpoint.

subiquity/server/controllers/keyboard.py
eeickmeyer: We could certainly do that, but it creates a caveat: the pre-existing `gsettings/setxkbmap/systemd-run --setenv=DISPLAY=...` path in `_run_gui_command` (already in the codebase, not something I added) is more X11-flavored in appearance (it sets DISPLAY), but gsettings set itself just writes to dconf and doesn't actually require a display connection, so it works under Wayland sessions too since GNOME Shell reads the same org.gnome.desktop.input-sources schema regardless of session type. The DISPLAY env var there is vestigial rather than required.

So both the existing and newly-added persistence paths should work under Wayland.

eeickmeyer:
Hi Olivier,

I went ahead with those changes. Let me know what you think!

PR #2431, opened 2026-09-10
Only preserve TPM/FDE partitions that have the same type/flag

ogayot:
For a reason I don't know, when we're doing a TPM/FDE installation, we try to preserve partitions that are at the same offset and size. However, before honoring `preserve: true` for partitions, curtin does additional checks, which include partition type / flag.

Therefore, if we try to reuse a partition but want to change the partition type, we run into a curtin error. Instead we should delete and recreate the partition so that the new one has the right type/flag.

LP:#2166668

### DIFF
--- a/subiquity/server/controllers/storage.py
+++ b/subiquity/server/controllers/storage.py
@@ -1171,8 +1171,18 @@ class StorageController(SubiquityController, StorageManipulator):
                 }
 
                 for _struct, offset, size in on_volume.offsets_and_sizes():
-                    if (offset, size) in parts_by_offset_size:
-                        preserved_parts.add(parts_by_offset_size[(offset, size)])
+                    if (offset, size) not in parts_by_offset_size:
+                        continue
+                    part = parts_by_offset_size[(offset, size)]
+                    if on_volume.schema == "gpt":
+                        # Curtin explicitly checks if partitions that we
+                        # preserve have the expected flag.
+                        type_uuid = _struct.gpt_part_type_uuid()
+                        if type_uuid and part.flag != ptable_part_type_to_flag(
+                            type_uuid
+                        ):
+                            continue
+                    preserved_parts.add(part)
 
                 for part in list(disk.partitions()):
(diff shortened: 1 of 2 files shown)


antlassagne:
I see that you went with keeping the preservation of the partition for safety. Wouldn't "right after the LTS release" be a good moment for a supposedly less safe change?

(that's only my opinion ofc)

ogayot:
> I see that you went with keeping the preservation of the partition for safety. Wouldn't "right after the LTS release" be a good moment for a supposedly less safe change?

@antlassagne 

That's a valid point. I don't know if we want to keep this functionality going forward (i.e. for 27.04) since I don't know what was the original purpose. 

OTOH, that is the type of bug I want to see addressed in 26.04.2, and I want the safer approach there.

dbungert:
@mwhudson - would you elaborate on the intentions here?  It's not clear to us on why we're attempting to preserve almost correct partitions in the TPMFDE format disk case.

mwhudson:
Olivier already asked me about this, unfortunately I have no recollection of the problem I was trying to solve when I implemented this! I can see from chat conversations that I talked to both Olivier and yourself about it before landing it but not in enough detail to remember what the point is. I guess someone could take this re-use out and see if anything breaks?

dbungert:
@ogayot - counter-proposal - instead of attempting harder to preserve these, I argue we don't need to be doing this for the TPMFDE partitions on the reformat case.  WDYT?

ogayot:
> @ogayot - counter-proposal - instead of attempting harder to preserve these, I argue we don't need to be doing this for the TPMFDE partitions on the reformat case. WDYT?

That's what Antoine was suggesting as well. I'm okay to proceed with this "feature drop" but I think this is a safer approach for a point release, if we want it fixed.

https://github.com/canonical/subiquity/pull/2431#issuecomment-5661027838

I can do a follow up PR to remove the partition reuse if that's ok.

dbungert (approved):

PR #2432, opened 2026-09-11
storage: ignore variations that have missing snapd system

ogayot:
If a source variation declares a snapd-system-label but no matching system seed is found, Subiquity automatically assumed the variation corresponds to a classic variation.

This is a problem in setups with multiple variations because Subiquity won't know which is the real classic variation, and will use the first variation declared instead.

While this should not happen for ISOs that are correctly built, the correct thing to do is to ignore such a variation instead of assuming it's a classic variation.

LP:#2167127

### DIFF
--- a/subiquity/server/controllers/storage.py
+++ b/subiquity/server/controllers/storage.py
@@ -660,7 +660,22 @@ class StorageController(SubiquityController, StorageManipulator):
                     # though it might be called again concurrently.
                     raise
 
+                if system is None:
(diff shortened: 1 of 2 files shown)


dbungert:
https://bugs.launchpad.net/ubuntu/+source/livecd-rootfs/+bug/2167127

dbungert (approved):
Looks right, and passes practical tests, thanks!

PR #2433, opened 2026-09-11
storage: ignore variations that have missing snapd system

dbungert:
If a source variation declares a snapd-system-label but no matching system seed is found, Subiquity automatically assumed the variation corresponds to a classic variation.

This is a problem in setups with multiple variations because Subiquity won't know which is the real classic variation, and will use the first variation decalred instead.

While this should not happen for ISOs that are correctly built, the correct thing to do is to ignore such a variation instead of assuming it's a classic variation.

LP: #2167127

(cherry picked from commit 54fa2ab46a9607b4d1166cbe3f55cbab19007c11)

### DIFF
--- a/subiquity/server/controllers/filesystem.py
+++ b/subiquity/server/controllers/filesystem.py
@@ -461,7 +461,22 @@ class FilesystemController(SubiquityController, FilesystemManipulator):
                     # though it might be called again concurrently.
                     raise
 
+                if system is None:
+                    # The snapd system referenced by this variation's label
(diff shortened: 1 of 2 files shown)


ogayot (approved):
Thank you!

dbungert:
Some modest test changes were needed for this cherry-pick, see https://github.com/canonical/subiquity/pull/2433/changes/e4c47839c8d08b61303415894025e630e61c7f9a

ogayot (approved):
Thanks for fixing the tests!

PR #2434, opened 2026-09-14
[stonking] backport - storage - ignore variations that have missing snapd system

ogayot:
This is a backport of #2432 to stonking.


dbungert (approved):
Thanks for this, should also do Resolute.

PR #2436, opened 2026-09-15
workflows: run TICS on PR modifying TICS workflow

dbungert:
The recent dependabot run was odd in that it didn't actually run the updated TICS workflow.  Adding this will mean that future modifications to .github/workflows/tics.yaml will trigger the TICS workflow to run.


dbungert:
Will resubmit using branch directly on canonical/subiquity

PR #2438, opened 2026-09-16
Expose storage requirements in the API

ogayot:
This PR introduces the notion of storage requirement into the storage/v2 API:
```json
{
  "status": "DONE",
  "error_report": null,
  "disks": [
     ...
  ],
  "install_minimum_size": 5657067520,
  "requirements": [
    {
      "kind": "MOUNT_ROOT",
      "satisfied": false
    },
    {
      "kind": "MOUNT_LOCAL_BOOT",
      "satisfied": true
    },
    {
      "kind": "SELECT_BOOT_DISK",
      "satisfied": false
    },
    {
      "kind": "USE_EXT4_BOOT",
      "satisfied": true
    },
    {
      "kind": "BOOT_ON_SIMPLE_SETUP",
      "satisfied": true
    }
  ]
}
```

Only requirements that apply to the [platform / Ubuntu version] are listed. This means, for instance, that `"BOOT_ON_SIMPLE_SETUP"` will not be present on s390x.

This should allow the desktop installer to:
* block the "Next" button when at least one requirement is not met
* restrict the UI. E.g., only offering "ext4" for "/boot" if `"USE_EXT4_BOOT"` is present
* display information messages when relevant.

From Subiquity's standpoint, a few adjustments were needed:
* we split the `requirement.applies_to` into two distinct functions:
    * `requirement.platform_applies_to` - which is what the API uses to determine if the requirement should be listed. This predicate should only include conditions that are not affected by the user's choices - and is therefore expected to be constant at runtime. Examples:
        * "does this requirement apply if we're installing Ubuntu 26.04?"
        * "does this requirement apply if we're installing on s390x?"
        * "does this requirement apply if our platform supports booting using NVMe/TCP booting?"
    * `requirement.layout_applies_to` - which covers everything else, and can change at runtime. Examples:
        * "does this requirement apply if the rootfs is not yet mounted?"
        * "does this requirement apply if we have a separate /boot partition?"
* we now consistently wait for the firmware probing operation to be finished before responding to the v2/storage API
LP:#2165727

cc: @d-loose


d-loose:
Thanks, the additions look good from the consumer's PoV. We talked about removing `need_boot` and `need_root` from the storage response to keep the API small, do you also want to do that here, or is there a reason for keeping them around? At the very least they should be marked as deprecated.

ogayot:
> We talked about removing `need_boot` and `need_root` from the storage response to keep the API small, do you also want to do that here, or is there a reason for keeping them around? At the very least they should be marked as deprecated.

Thanks Dennis! Yes, I'm planning to get rid of `need_boot` + `need_root` but mayybe they need to stay in 26.10 since we're close to beta. I can mark them deprecated here and get rid of them in a separate PR. Would that be okay for you? I also wanted to keep them around for the sake of testing; to make sure we're not getting inconsistent results i.e.,
* `"need_root": true` staying consistent with `{"kind": "MOUNT_ROOT", "satisfied": true}`
* `"need_boot": true` staying consistent with `{"kind": "SELECT_BOOT_DISK", "satisfied": true}`

d-loose:
That makes sense, fine with me!

subiquity/server/controllers/storage.py
```diff
@@ -1050,13 +1050,14 @@ async def guided(
...
     async def _probe_response(self, wait, resp_cls):
-        if not self._probe_task.done():
-            if wait:
-                await self._start_task
-                await self._probe_task.wait()
-                await self._probe_firmware_task.wait()
-            else:
-                return resp_cls(status=ProbeStatus.PROBING)
+        # Wait for both probing tasks to be finished.
+        for task in self._probe_task, self._probe_firmware_task:
```
dbungert, line 1054: the loop over tasks like this makes me nervous but I can't see that it's wrong.  I suppose it get's weird if the tasks interact.  No change requested here, just recording that thought.

dbungert (approved):

subiquity/common/types/storage.py
```diff
@@ -396,16 +396,45 @@ class StorageResponse:
...
+    USE_EXT4_BOOT = _("Use the ext4 filesystem for /boot")
+    BOOT_ON_SIMPLE_SETUP = _("Place /boot on a partition of a disk (or RAID 1 disk)")
+
+
+@attr.s(auto_attribs=True)
+class StorageRequirementStatus:
+    """The state of a single storage requirement."""
+
+    kind: GuidanceMessageKind
+    satisfied: bool
```
Dh2Cool, line 421: not sure if this is in scope, but should severity be part of this entry? the spec says only blocking requirements should disable "Next", if a WARNING requirement ever gets added, the desktop has no way to tell them apart. Everything else is good!

Dh2Cool (approved):
Thank you Olivier,

subiquity/common/types/storage.py
ogayot: Thanks Dhruv, good point. My initial use-case for a "warning" requirement (would recommendation be a better term?) is for requirements that don't apply today, but are likely to prevent a future upgrade otherwise. 

For instance, it's fine for users to install 26.04 with `/boot` on `zfs`. Ubuntu 26.04 supports it. But the upgrade path to 26.10, or 28.04 might be difficult, or even unsupported.

Since I didn't introduce any of such "recommendation" yet, I think it's fine to omit it in the API. But that would be more future proof, I suppose.

PR #2440, opened 2026-09-18
storage: add a guided btrfs layout

Dh2Cool:
Add an unencrypted guided btrfs layout alongside guided_direct and
guided_zfs: a separate ext4 /boot partition, then a single btrfs filesystem
carrying two subvolume actions — @ mounted at / and @home at /home. The
mounts reference the format action with subvol= options, which curtin
writes verbatim to fstab; the subvolume actions exist only to create the
subvolumes.

Verified end-to-end in a QEMU autoinstall (layout: {name: btrfs}): @ at /
and @home at /home on the installed system, matching fstab entries, and a
clean snapshot round-trip.

Note: the desktop installer's subiquity_client does not know the BTRFS
capability yet and fails to deserialize the guided response, so this should
land together with a subiquity_client update.


Dh2Cool:
Closing this to reopen via the web UI so reviewers can be requested directly.

subiquity/common/storage/spec.py
```diff
@@ -53,6 +53,7 @@ class RaidSpec(TypedDict):
         "mount": str | None,
         "wipe": str | None,  # NOTE: no wipe is different from wipe=None
         "use_swap": bool,
+        "options": str | None,  # mount options, e.g. subvol=@ for btrfs
```
ogayot: Please mention "comma separated"

subiquity/models/storage.py
```diff
@@ -1378,6 +1378,9 @@ class Filesystem:
     extra_options: Optional[List[str]] = None
 
     _mount: Optional["Mount"] = attributes.backlink()
+    _subvolumes: List["BtrfsSubvolume"] = attributes.backlink(
```
ogayot: This feels out of place. Maybe call that `_btrfs_subvolumes` for now? If we need something more generic, we can change the name later.

subiquity/common/storage/manipulator.py
```diff
@@ -104,7 +104,12 @@ def create_filesystem(self, volume, spec: FileSystemSpec):
     def delete_filesystem(self, fs):
         if fs is None:
             return
-        self.delete_mount(fs.mount())
+        for sv in list(fs._subvolumes):
```
ogayot: do we need the `list` construct here?

subiquity/models/storage.py
```diff
@@ -2493,16 +2511,21 @@ def remove_filesystem(self, fs):
...
-    def add_mount(self, fs: Filesystem, path):
-        if fs._mount is not None:
+    def add_mount(self, fs: Filesystem, path, options=None):
+        # multiple mounts are legitimate for a btrfs fs with subvolumes
+        if fs._mount is not None and not fs._subvolumes:
             raise Exception(f"{fs} is already mounted")
-        options = None
-        if fs.volume.on_remote_storage():
+        if options is None and fs.volume.on_remote_storage():
             options = "defaults,_netdev"
```
ogayot: This means we don't get the `_netdev` option for free anymore if we're specifying options. I suggest we preserve this behavior. That said, maybe it's best to move this logic to curtin, since it already adds `_netdev` automatically for ISCSI.

https://github.com/canonical/curtin/blob/e2fc2bb9e38c7336c181567864f6b963e5c3835b/curtin/commands/block_meta.py#L1280-L1282

If we add something similar for NVMe/TCP, it should be fine to drop these lines.

subiquity/common/storage/spec.py
Dh2Cool: done.

subiquity/common/storage/manipulator.py
Dh2Cool: I think we do. the _remove(sv) on the next line takes the subvolume back out of fs._btrfs_subvolumes, so the list shrinks while we're still looping over it and we'd end up skipping every other one. The list() makes a copy so we loop over that instead.

subiquity/models/storage.py
Dh2Cool: Right, changed it to _btrfs_subvolumes. There is also another mention of _subvolumes related to raid functionality, so it is better to change it.

Dh2Cool:
Also renamed the config field to **name** here, matching the curtin side

subiquity/common/storage/manipulator.py
ogayot: Thanks! That makes sense.

doc/.custom_wordlist.txt
```diff
@@ -77,6 +78,7 @@ debian
 dir
 el
 enablement
+filesystem
```
dbungert: how did we not have the word filesystem in our docs yet?

subiquity/common/types/storage.py
```diff
@@ -226,6 +226,7 @@ class GuidedCapability(enum.Enum):
     LVM_LUKS = enum.auto()
     ZFS = enum.auto()
     ZFS_LUKS_KEYSTORE = enum.auto()
+    BTRFS = enum.auto()
```
dbungert: in PR followup, please also implement BTRFS_LUKS

subiquity/models/storage.py
```diff
@@ -1580,6 +1563,21 @@ def iter_storage_chain(self) -> Iterator[Any]:
...
+@fsobj("btrfs_subvolume")
+class BtrfsSubvolume:
+    volume: Filesystem = attributes.ref(backlink="_btrfs_subvolumes")
+    name: str
+
+    _mount: Optional["Mount"] = attributes.backlink()
+
+    @property
+    def fstype(self):
+        return self.volume.fstype
```
dbungert: is this ever not btrfs?

dbungert (commented):
Generally OK, let's resolve curtin questions first

dbungert:
Also let's resolve the doc failure first so we're not merging on broken CI

